PR #83, opened 2015-06-10
ADDED filters to the Price and Quantity table headers.

richardtape:
This enables people to easily modify the content of these table headers and matches the pattern with the 'Available Tickets' header that already contains a filter.

Also, the event ID is passed to the filters. This enables, for example, a plugin/theme to be able to see if this event only allows people to select 1 ticket, in which case the radio button is displayed and in such a case the 'Qty' heading becomes superfluous. Same deal with 'Price' if the event is Free.

### DIFF
--- a/modules/ticket_selector/templates/ticket_selector_chart.template.php
+++ b/modules/ticket_selector/templates/ticket_selector_chart.template.php
@@ -13,9 +13,33 @@ $template_settings =  isset ( EE_Registry::instance()->CFG->template_settings->E
 					<span class="ee-icon ee-icon-tickets"></span><?php echo apply_filters( 'FHEE__ticket_selector_chart_template__table_header_available_tickets', __( 'Available Tickets', 'event_espresso' )); ?>
 				</th>
 				<?php if ( apply_filters( 'FHEE__ticket_selector_chart_template__display_ticket_price_details', TRUE )) { ?>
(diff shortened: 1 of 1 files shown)


richardtape:
Not entirely sure why this is saying there are merge conflicts... I can't see where they are. I cloned from master, branched and then pushed that branch.

Any ideas, I'll fix them up.

tn3rb:
@richardtape 
The reason it conflicted is most likely that you didn't pull before pushing.
Looks to me like we had been pushing changes to master in between the time you pulled, made your pull request, and then pushed. So your version of master was out of sync with our version of master.
In the future, just always do a pull, then merge any remote changes prior to pushing.

Anyways, I've created a FET-8343-add-filters-to-ticket-selector-table-header branch off of our master for integrating your changes, but have edited what you pushed. Would have pushed my changes to your forked repo, but GitHub isn't letting me.

As soon as our support staff test the changes, they'll be merged into our master and go out with that release. Can't say at this time when that will be.

Thanks for contributing.

richardtape:
Thanks @tn3rb 

Weird. Must have been seconds between me pulling most recent master and pushing that branch. Good to see stuff happening so frequently :)

Thanks for doing the dirty work for me, appreciate it.

Rich

richardtape:
Urgh. There's an error in this pull request, I'm so sorry. I'm going to submit an updated version, pulling your most recent changes from master first.

PR #85, opened 2015-06-10
ADDED filters for the Price and Qty. table headers in the ticket selector

richardtape:
to match the pre-existing FHEE__ticket_selector_chart_template__table_header_available_tickets

ADDED esc_html call to the pre-existing FHEE__ticket_selector_chart_template__table_header_available_tickets filter for consistency and security

The new filters receive the Event ID which allows plugins/themes to see if these headings need to exist at all (or changed them based on the available tickets).

This replaces the previous pull request https://github.com/eventespresso/event-espresso-core/pull/83 -- I'm so sorry I made a bit of a mess with that. Not only did I not have the most recent master (even though I simply forked the repo), I also has _e() in the filter rather than __() which meant the extra text was doubled up. Brilliant.

Noticed that in your cherry-picked commit that you'd also added esc_html__ to the pre-existing available tickets header, so I've added that in here too, so it should _hopefully_ just be a straight up replacement for thr  FET-8343-add-filters-to-ticket-selector-table-header branch that you'd created. Again, sorry about that mess. Won't happen again.

### DIFF
--- a/modules/ticket_selector/templates/ticket_selector_chart.template.php
+++ b/modules/ticket_selector/templates/ticket_selector_chart.template.php
@@ -10,12 +10,36 @@ $template_settings =  isset ( EE_Registry::instance()->CFG->template_settings->E
 		<thead>
 			<tr>
 				<th scope="col" class="ee-ticket-selector-ticket-details-th">
-					<?php echo apply_filters( 'FHEE__ticket_selector_chart_template__table_header_available_tickets', __( 'Available Tickets', 'event_espresso' )); ?> <span class="ee-icon ee-icon-tickets"></span>
+					<?php echo esc_html__( apply_filters( 'FHEE__ticket_selector_chart_template__table_header_available_tickets', __( 'Available Tickets', 'event_espresso' ) ), 'event_espresso' ); ?> <span class="ee-icon ee-icon-tickets"></span>
 				</th>
 				<?php if ( apply_filters( 'FHEE__ticket_selector_chart_template__display_ticket_price_details', TRUE )) { ?>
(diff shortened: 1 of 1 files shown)


tn3rb:
please note that there is a difference between: 

```
esc_html__('translatable string')
```
and 
```
esc_html('translatable string')
```

the former is basically short form for:
```
esc_html( __('translatable string') )
```
so it translates a string, THEN applies esc_html() to the results

doing:
```
esc_html__( __('translatable string') )
```
translates a string... then translates it AGAIN !!!... THEN applies esc_html() to the results

since you are wanting to filter the translated string, you need to use:
```
esc_html( apply_filters( __('translatable string') ) )
```
make sense?

richardtape:
:coffee: please.

Blimey. Sorry. Changes incoming. It's been a long day and it's not even lunch time ;)

richardtape:
Right. Again. Sorry. Pushed that change.

tn3rb:
no worries... 
thnx for following our filter naming schema by the way.

So what are you using EE4 for across the water there? (I'm in Parksville BC)

richardtape:
We're using it for the Centre for Teaching, Learning and Technology's 
events system at UBC. We are on a hacked version of EE3 at the moment - 
a couple years ago some co-op students went a bit off-plan and hacked 
the main plugin instead of asking/adding filters in certain places, but 
we're transitioning to EE4 over the next few weeks.

Vancouver is beautiful today, hope it is up in Parksville.

Brent Christensen wrote:

tn3rb:
this (or a version thereof) has been merged to master

PR #86, opened 2015-06-11
ADDED a filter to allow the complete removal of the maximum tickets purchased footnote

richardtape:
Currently the existing filter only allows you to change just the message shown in the footnote. (This also needs to be changed a little as it's always plural even if the max is 1, but that can be a separate PR).

If you simply return an empty string to the existing filter, you get a paragraph tag with just a '*' character. This is unnecessary. It's possible to hide this via CSS, but having a filter to show or hide this footnote is cleaner.


tn3rb:
this (or a version thereof) has been merged to master

PR #102, opened 2015-07-29
deserialize in prepare_for_get

robert-osborne:
A object->get('serialized_field') on a serialzed field containting an encoded array should return the array instead of the encoded string

### DIFF
--- a/core/db_models/fields/EE_Serialized_Text_Field.php
+++ b/core/db_models/fields/EE_Serialized_Text_Field.php
@@ -42,6 +42,22 @@ class EE_Serialized_Text_Field extends EE_Text_Field_Base{
 			return $data;
 		}
 	}
+	/**
+	 * Value provided should definetely be a serialized string. We should unserialize into an array
+	 * @param string $value_found_in_db_for_model_object
+	 * @return array
+	 */
+	function prepare_for_get($value_of_field_on_model_object) {
+		$data = maybe_unserialize($value_of_field_on_model_object);
+		//it's possible that this still has serialized data if its the session.  WP has a bug, http://core.trac.wordpress.org/ticket/26118 that doesnt' unserialize this automatically.
+		$token = 'C';
+		$data = is_string($data) ? trim($data) : $data;
+		if ( is_string($data) && strlen($data) > 1 && $data[0] == $token  && preg_match( "/^{$token}:[0-9]+:/s", $data ) ) {
+			return unserialize($data);
+		} else {
+			return $data;
+		}
+	}
 	
 	/**
 	 * Gets a string representation of the array


nerrad:
I'm going to ask @mnelson4 to chime in here as well.  However Robert, can you explain where you are seeing a problem that this fixes?  The `prepare_for_set_from_db` method takes care of unserializing serialized strings in the db when the `EE_Base_Class` object is instantiated so there's no need to repeat that in the `get` method.

robert-osborne:
Sure,  without this change, when I do a get on a serialized field, I get the encoded version:

Declaration:

'GRR_permissions' 	     => new EE_Serialized_Text_Field( 'GRR_permissions', __( "Group Permissions", "event_espresso" ), FALSE, __( "register_self", "event_espresso" ) ),

Usage:

$perms = $this->get('GRR_permissions’);
EEH_Debug_Tools::printr($perms, "Permissions array", __FILE__, __LINE__);

Before change:

Permissions Array

string 'a:2:{i:0;s:13:"register_self";i:1;s:15:"register_others";}' (length=58)

After the change:
Permissions Array

array (size=2)
  0 => string 'register_self' (length=13)
  1 => string 'register_others' (length=15)

nerrad:
Are you using `EE_Serialized_Text_Field` within a custom model?  This is a class intended only for `EE_Model_Field` declarations. Not for general usage.

robert-osborne:
Ah!  Yes I am.   So how does a plugin writer serialize a list in an approved way?   I was just following the lead of classes like EEM_Payment...

mnelson4:
hey @robert-osborne thanks for the pull request. 
I suspect you're actually serializing the value before passing it into the model object, which is why you're getting back a model object. Can you confirm this (or otherwise please share code, and preferably a unit test, showing the current problem and how your commit resolves it)
I just added a unit test to verify things are working as expected: https://github.com/eventespresso/event-espresso-core/commit/405505d2f62e666187b2fb8cc7773356df60ea85

Notice that if I pass in an array to the `set` method, when I call `get` (either on the model object directly created in PHP, or fetched from the database) I get back an array. But if I serialized that array BEFORE passing it into `set`, then I get back a serialized array when I call `get`. I think this current behaviour makes sense, right?

robert-osborne:
I checked that first.  The raw column data in the database is only encoded once:

a:2:{i:0;s:13:"register_self";i:1;s:15:"register_others”;}

I’ll double check.

Rob.

mnelson4:
Hmmm well @robert-osborne if you verified the data isn't being serialized before passing it to the model's `set` method, then something weird's going on. Once data corresponding to a serialized field is set on the model, it should always be in its raw form (eg an array or an object). And so unserializing it when you're `get`int it shouldn't ever be necessary. If you're finding it is necessary then something else is wrong. Can you please share a code snippet that reproduces the issue, that includes how you create the model object and set the `GRR_permissions` field?

robert-osborne:
I'm really not doing anything exciting.   Creating the object is initially via the EE_Base::instantiate_class_from_array_or_object and it's definitely an array at that point and gets into the database correctly.   I'm instantiating via get_one_by_ID.

Stepping through the get code reveals that the get goes directly to EE_Text_Field_Base::prepare_for_get which just stripslashes from the string and returns the raw "a:2:{i:0;s:13:"register_self";i:1;s:15:"register_others";}"

My "test code" is literally:

$testInstance = EEM_Ypogroup_Role::instance();
$testRole = $testInstance->get_one_by_ID($ID);
$testPerms = $testRole->get('GRR_permissions');

nerrad:
Robert is this code:

```
$testInstance = EEM_Ypogroup_Role::instance();
$testRole = $testInstance->get_one_by_ID($ID);
```

being done on the same request as when you create the class via `instantiate_class_from_array_or_object`?

robert-osborne:
The code above is in a MUCH later request ... days at this point :-)

mnelson4:
ya that's weird because when I use basically the same chunk of code there is no problem.
```
//		$log = EE_Change_Log::new_instance( array( 'LOG_message' => serialize( array( 'key1' => 'var1', 'key2' => 'var2' ) ) ) );
//		$log->save();
		$log = EEM_Change_Log::instance()->get_one_by_ID( 2 );
		var_dump( $log->get( 'LOG_message' ) );
```
I just put that inside `EE_System::initialize()` and uncommented the first 2 lines on one request, and then uncommented the other 2 (with the ID of the newly generated log entry) on the next. What version of WP, EE (I had assumed this master branch, but just checking), and PHP are you running?

robert-osborne:
Awww, bloody hell!   Tracked it down. The code I inherited has a custom constructor which peels off the "from db" flag so the set from db never gets called.  Had to step through both examples line by line to find this.

I'm REALLY sorry for wasting everybody's time, thanks for the help.

mnelson4:
oh good! glad we're not going crazy. Although we may have discovered some inconsistent behaviour as a result of this: 
```
$log = EE_Change_Log::new_instance( array( 'LOG_message' => serialize( array( ) ) );
var_dump($log->get('LOG_message' );//outputs a string
$log_from_db = EEM_Change_Log::reset()->get_one_by_ID( $log->ID() );
var_dump( $log_from_db->get('LOG_message' );//outputs an array
```

Which we're working on resolving atm

robert-osborne:
I'm glad some good came out of this :-)   Thanks for the heads up,  I'll watch for this.

Thanks for all your help @mnelson4!

mnelson4:
hey @robert-osborne here's the related work we did, on branch **BUG-8589-make-EE_Base_Class-get-for-serialized-fields-consistent**

If you're able, it would be nice if you could test that branch and let us know if there are issues. (It won't fix the exact issue you were having). We just need someone else to verify this branch doesn't break normal usage of serialized fields (for example, you could also visit the payment method's logs page, and view a payment log. If it appears like on the master branch, the changes look good).

If you do get around to testing that branch please let us know!

robert-osborne:
@mnelson4 I just did a quick check of my usage and it still works fine.  My log test works as well.   Unfortunately I don't have any payment data.  I'm testing registrations etc. with my new plugin and I'll stay on this branch for all of that and try to get some payments in there today.   I'll let you know if I see anything.

mnelson4:
great thanks!

PR #104, opened 2015-08-06
Update get_most_recent_registration_for_event to use new query params format

robert-osborne:
Basically just copied get_most_recent_registration but restricted the query by EVT_ID.


mnelson4:
thanks @robert-osborne! yeah looks good. Sorry we missed refactoring that

Issue #137, opened 2015-12-14
Translation missing in Checkout Steps


DavidSteinbauer:
Please add translation as the following:
![2015-12-14 at 12 47](https://cloud.githubusercontent.com/assets/3392092/11780248/cc70eb24-a260-11e5-8249-cbcc70d46963.png)

nerrad:
I closed the pull request, you can see my explanation there.

nerrad:
In case you're wondering, I'm fairly certain the value for `$reg_step->name()` is already translated.  That value is retrieved from the `$_name` property on the `EE_SPCP_Reg_Step` child classes.  For example, if you look at `EE_SPCO_Reg_Step_Attendee_Information` you'll see this defined in it's constructor:

```
$this->_name = __('Attendee Information', 'event_espresso');
```

PR #138, opened 2015-12-14
Translation #137

DavidSteinbauer:


nerrad:
Hi David,

Thanks for your pull request, however wrapping variables or function/method outputs in internationalization methods is a no go.  Gnu-text is not a php parser.

PR #166, opened 2016-08-18
Updated date filter to include WP timezone offset

jonathanjulian:


nerrad:
Hi jonathanjulian,

Thanks for your pull request.  Can you please explain what problem this pull request solves?

lorenzocaum:
@nerrad This looks like the support post: https://eventespresso.com/topic/core-plugin-not-using-wordpress-timezone-settings-in-filtering/

nerrad:
jonathanjulian, I've read through the support forum report and the change you are requesting here doesn't actually affect the displayed transactions on _initial_ page load of the transaction list table.

The change you are requesting, merely ensures that the default date listed for the filters matches the current date for the site (which is a change we _can_ make).

As far as I can tell, the actual transactions displayed in the list table on initial page load *should* include those for the current date (regardless of the initial filter display).  Is something different happening?

nerrad:
Although I merged this pull request into master, I must emphasize that it has no affect on the transactions returned for the list table on initial page load.  If that is a problem,  a new issue will need created to investigate.

PR #178, opened 2016-12-01
Update template_tags.php w/ time classes

dwlorimer:
Problem:
The dates and times listed in the UL, LI structure are plain text, without even a span or class to identify them.

For example, see the screenshot from this site:
http://www.skypeug.com/events/baltimore-fall-2016/

image

Solution:
I am simply proposing to wrap those dates in a span with a class, so they can be accessed properly.  We had to made this edit manually on our site because we need to access those dates in javascript.


dwlorimer:
Please let me know an update when you know the status.
I don't know the relationship between event-espresso-core, and event-espresso-decaf (I don't see a repo for decaf on here), but this change needs to be made in decaf also.

Thanks!

tn3rb:
merged to master branch
thnx for your contribution

PR #181, opened 2016-12-09
Rework count_checkins_not_checkedout() for clarity

KinnaT:
Whenever you check someone into or out of a datetime, it adds a row to `esp_checkin`. Misclicks can cause an attendee to be checked in, which are then checked out to be corrected. If the attendee is checked in again later, their check-in count is now 2 according to `count_checkins_not_checkedout()`.

This changes `count_checkins_not_checkedout()` to return a count of the number of datetimes that this registration is currently checked into to allow for accurate attendance reporting and ticket use display.
`count_checkins_only()` contains the function that `count_checkins_not_checkedout()` had, with a more appropriate name.
This is less specific than `verify_can_checkin_against_TKT_uses` as it isn't tied to a datetime and can be used with just a `REG_ID`.
Using this change, checking if `count_checkins_not_checkedout()` equals `$ticket->uses()` can accurately return a boolean of fully attended or not, allowing for extended functionality like certificates of attendance.

### DIFF
--- a/core/db_classes/EE_Registration.class.php
+++ b/core/db_classes/EE_Registration.class.php
@@ -951,12 +951,23 @@ class EE_Registration extends EE_Soft_Delete_Base_Class implements EEI_Registrat
 
 
 	/**
-	 * Returns the number of current Check-ins this registration is checked into for any of the datetimes the registration is for.  Note, this is ONLY checked in (does not include checkedout)
+	 * Returns the number of current Check-ins this registration is checked into for any of the datetimes the registration is for.
 	 * @return int
 	 */
 	public function count_checkins_not_checkedout() {
-		return $this->get_model()->count_related( $this, 'Checkin', array( array( 'CHK_in' => 1 ) ) );
-	}
+		$check_ins = $this->get_model()->count_related( $this, 'Checkin', array( array( 'CHK_in' => 1 ) ) );
+		$check_outs = $this->get_model()->count_related( $this, 'Checkin', array( array( 'CHK_in' => 0 ) ) );
+		
+		return $check_ins - $check_outs;
+    	}
+	
+	/**
+	 * Returns the number of times this registration has had CHK_in set to 1
+	 * @return int
+    	 */
+	public function count_checkins_only() {
+        	return $this->get_model()->count_related( $this, 'Checkin', array( array( 'CHK_in' => 1 ) ) );
+    	}
 
 
 
--- a/core/db_classes/EE_Registration.class.php
+++ b/core/db_classes/EE_Registration.class.php
@@ -948,26 +948,24 @@ class EE_Registration extends EE_Soft_Delete_Base_Class implements EEI_Registrat
 		return $this->get_model()->count_related( $this, 'Checkin' );
 	}
 
-
+	 /**
+	 * Returns the number of current Check-ins this registration is checked into (CHK_in = 1) that do not have a balancing checkout (CHK_out = 0)
+	 * @return int
+	 */
+	 public function count_active_checkins_only() {
+	 	$check_ins = $this->get_model()->count_related( $this, 'Checkin', array( array( 'CHK_in' => 1 ) ) );
+	 	$check_outs = $this->get_model()->count_related( $this, 'Checkin', array( array( 'CHK_in' => 0 ) ) );
+	 	return $check_ins - $check_outs;
+	 }
 
 	/**
-	 * Returns the number of current Check-ins this registration is checked into for any of the datetimes the registration is for.
+	 * Returns the number of current Check-ins this registration is checked into for any of the datetimes the registration is for.  Note, this is ONLY checked in (does not include checkedout)
 	 * @return int
 	 */
 	public function count_checkins_not_checkedout() {
-		$check_ins = $this->get_model()->count_related( $this, 'Checkin', array( array( 'CHK_in' => 1 ) ) );
-		$check_outs = $this->get_model()->count_related( $this, 'Checkin', array( array( 'CHK_in' => 0 ) ) );
-		
-		return $check_ins - $check_outs;
-    	}
-	
-	/**
-	 * Returns the number of times this registration has had CHK_in set to 1
-	 * @return int
-    	 */
-	public function count_checkins_only() {
-        	return $this->get_model()->count_related( $this, 'Checkin', array( array( 'CHK_in' => 1 ) ) );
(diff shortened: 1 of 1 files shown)


mnelson4:
> Whenever you check someone into or out of a datetime

Out of curiosity, are you using the wp-admin check-in page? Or the iOS or Android mobile app? That would help give a little bit of context.

> Misclicks can cause an attendee to be checked in, which are then checked out to be corrected.

So you're saying an event manager might accidentally check someone IN, and then in order to undo that checkin, they check them out, right? If so, that's not actually the intended way to negate check-ins. They should actually DELETE the check-in (from the [wp-admin check-in page, you'd click on "view"](https://drive.google.com/file/d/0B2KCao4zFjaxdkJyUzdGUzZpUFU/view?usp=drivesdk) and then select the [check-in record and delete it](https://drive.google.com/file/d/0B2KCao4zFjaxdHlaN1JEQXZYZUE/view?usp=drivesdk); but from the mobile apps this isn't possible because we wanted to keep the UI simpler).

So if you do that, I think this issue can be avoided.

@nerrad do you have feedback on this pull request? I think if you want to see how many times this ticket has been properly used, you instead need to check whether the registration was most recently check-in or checked-out of each datetime (theoretically it's possible someone could be checked into an event multiple times with no check-out records, especially if someone deleted the check-out records from the admin page I pointed to above; and of course the opposite is also possible). I think just counting the registration's total check-ins vs total check-outs isn't adequate. Right?

nerrad:
Hi KinnaT thanks for your participation and contribution.

I think I understand where you are coming from in that you want to ensure  you're only getting a count of checkins minus the count of checkouts. However, I agree with all of @mnelson4 's assessments here.  In particular this statement you made here is not really correct:

> This changes count_checkins_not_checkedout() to return a count of the number of datetimes that this registration is currently checked into to allow for accurate attendance reporting and ticket use display.

That's only true if each checkin and checkout is for a single datetime that the registration has access to. Keep in mind the `esp_checkin` table is joined with both registration AND datetime.  So in reality the changes you made are more accurately just "a count of checkins minus a count of checkouts for that registration".

At this point I won't be merging this pull request in for a three reasons:

1. It doesn't really do what you claim it does. It may appear to work in your specific circumstance, but we have to think of all the potential users who would be affected.
2. Related to the above we'd like more context about the environment where current behaviour is a problem to help evaluate a change like this.  Mike asked some good questions for this.
3. We won't changing behaviour of public methods unless causes a severe critical bug in the wild (related to the point I made with number 1).  If an existing method isn't doing what you like, we'd be more receptive to introducing a new method if the need warranted it instead of changing behaviour of an existing method.

Again, thanks for your contribution but we won't be accepting this pull request as is :)

thebeline:
Personally, I can see a semantic issue here, where `count_checkins_not_checkedout` expands naturally to say "Count Check-Ins that are not Checked-Out".  

A Checkin is an object identified by an incremental ID that defines an edge between Registrations and DateTimes with a Checkin Status and Checkin Time.  Because of this, a check-in can be considered checked-out when it has a balancing checkout edge defined.  Therefore, "Count Check-Ins not Checked-Out" would lead one, semantically, to understand that the method returns the number of Check-Ins that do not have a balancing Check-Out.

For example:

CHK_ID | REG_ID | DTT_ID | CHK_in | CHK_timestamp
----|----|----|----|----
25 | 27 | 3 | 1 | 2016-12-21 03:52:33
26 | 27 | 4 | 1 | 2016-12-21 03:52:33
27 | 27 | 3 | 0 | 2016-12-21 03:52:38

This ticket (Registration) has one un-balanced Check-In.  This is what I would expect this method to return.  I have no experience with this method, so I will defer to you: Would this method return `1` or `2`.  Looking at it, it looks like it would return `2` (a count of the related objects that have a `CHK_in` status of `1`, which is `2` objects).

Again, not testing it, but it looks, generally, like this fella's PR would return `1`.  No?

One concern I would have:  Via clever (or lazy) programming, it would be trivial to:

```
for ($i = 1; $i < 100; $i++) {
    EE_Checkin::new_instance( array(
	'REG_ID' => $REG_ID,
	'DTT_ID' => $DTT_ID,
	'CHK_in' => 1 // WOOPS
    ) )->save();
}
```

Which would bugger the whole thing up, no?

Also, through lazy (or unknowing) programming, one may simply modify a Checkin directly:

```
// Oh my, I have my hands on a real, live, EE_Checkin object.
// I need to check it out, so I may as well do it directly...

$checkin->set('CHK_in', 0);
$checkin->save();
```

There is no good solution.  But I would say that the method name is a little misleading...

nerrad:
Regardless of how the method name is interpreted the most important point I've already made that i want to highlight is this:

> We won't change behaviour of public methods unless it fixes a severe critical bug in the wild 

Changing _existing_ behaviour on a method simply because someone has indicated its a poorly named method (whether or not that is true) means that other developers who have built things using this method _based on its existing_ behaviour will see things broken when they update to the codebase containing the change in behaviour.

So I appreciate the time you took @thebeline to express your opinion on the name of the method but I've already given an answer expressing a path forward to hopefully alleviate the concern the original poster brought up.

thebeline:
That's fair.

KinnaT:
Finally had some time to sit down and rework this. Thanks for your feedback, everyone!

Because it _appears_ as if event admins can negate a check-in by checking the registration out again, and the delete option doesn't appear in the app at all, their expectations are not unwarranted.

Use case: I need the number of check-ins for a registration to match the number of ticket uses in order to conditionally award a certificate of completion (so...they actually have to fully complete the course to get one). I need to create functionality that's as foolproof as possible. The hope is all they'll have to do is make sure the check-in page's/app's check marks for the datetimes reflect the actual attendees correctly. That way, an event admin that never touches wp-admin and only uses the app will be just as effective as an admin that uses wp-admin without needing the admin to to double-check the app user's actions.

I completely understand your reasoning for not changing the behaviour of existing methods. Based on your comments, I moved the proposed changes into a new method `count_active_checkins_only()` so there will be no change to `count_checkins_not_checkedout()`. This would satisfy my use case without changing any existing methods.

nerrad:
Hi @KinnaT 

Sorry for the delay in attending to this pull request. The conflicts involved in bringing your branch up to date are too significant to resolve cleanly.  Can you create a new branch off of master and apply your changes directly to that branch please?  Then you can either force push to this pull or create a new pull.

Thanks.

PR #186, opened 2016-12-20
Added a few classes *I* needed for customization, should help others.

thebeline:
This just makes it a little easier for me, at least, to do CSS customization.  Should help others as well.  Shouldn't hurt, anyway.

### DIFF
--- a/core/templates/global_assets/css/espresso_default.css
+++ b/core/templates/global_assets/css/espresso_default.css
@@ -261,6 +261,17 @@ ul.ee-upcoming-events-widget-ul li{
 	list-style-type:none !important;
 }
 
+.ee-break-before:before {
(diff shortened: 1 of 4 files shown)


thebeline:
Just noticed someone is trying the same thing in [178](https://github.com/eventespresso/event-espresso-core/pull/178).  I may refactor my usage to behave similarly.  I do have a few additional classes and some fixes for the fall-out of hiding one or more of the elements.  So I will keep the PR.

thebeline:
I am going back on my previous statement, I think moving this way is likely better, as I wrap the whole sections, as opposed to just the text, which allows identifying all of the contents and formatting accordingly.

That said, I am going to close this PR, as the related changes are in a newer one.

PR #191, opened 2016-12-21
Basic WP_CLI updating.

### LINKED ISSUE #187: IMP: Hooking directly into the WP Upgrade System
I was interested in what it would take to hook directly into the WP upgrade system, and it looks like the filter `site_transient_update_plugins` would give you a list of plugins, which you could then modify the array of each's  `update` and `update_version` and `package`. 

I am having a hard time determining how you currently handle upgrades, but I am suggesting this method (or, hooks, rather) because this will seamlessly integrate with the WP Cli plugin, and, depending on how it is currently done (again, I have write permissions locked down, so I can't perform upgrades from with web anyway) it will also integrate with the core WP Upgrader.

Relevant WordPress Classes:
- https://core.trac.wordpress.org/browser/tags/4.5.3/src/wp-includes/option.php#L1538
- https://core.trac.wordpress.org/browser/tags/4.7/src/wp-admin/includes/class-plugin-upgrader.php#L148
- https://core.trac.wordpress.org/browser/tags/4.7/src/wp-admin/includes/class-wp-upgrader.php#L640

Hope this helps...  If the current update tests and hooks were a little clearer, I would make the mods myself, but I can not seem to decipher how add-ons are upgraded, so I will leave it to the pros.  ;-)

thebeline:
Tentatively resolves #187

### DIFF
--- a/core/EE_System.core.php
+++ b/core/EE_System.core.php
@@ -795,7 +795,8 @@ final class EE_System
         add_action('wp_enqueue_scripts', array($this, 'wp_enqueue_scripts'), 25);
         add_action('admin_enqueue_scripts', array($this, 'wp_enqueue_scripts'), 25);
         add_action('admin_bar_menu', array($this, 'espresso_toolbar_items'), 100);
-        if (is_admin() && apply_filters('FHEE__EE_System__brew_espresso__load_pue', true)) {
+        // NOTE - WP_CLI is not officially supported, the following is only to allow basic updating and comes with no warranty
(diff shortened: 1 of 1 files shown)


nerrad:
Internal use: https://events.codebasehq.com/projects/event-espresso/tickets/10624

Haven't forgot about this :)

thebeline:
It seems like this is resolved (no more header warnings at the CLI).  Closing.  If I notice anything, I will address it later.

PR #206, opened 2017-02-11
add action for removing filters

landor:
EED_Event_Single::use_filterable_display_order calls an action allowing modifying the filter order, but doesn't call an action to allow cleaning up afterwards.
This adds the action to clean up.

EED_Events_Archive::use_filterable_display_order already calls an action allowing cleanup.


tn3rb:
merged to master... 
thnx for your contribution

PR #208, opened 2017-02-18
Add Filter For DropDown Detail

JRyven:
Problem - Company uses manual event checkin. And company has many events with the same name. Trial and error is not the ideal way to identify the correct event. 

The proposed filter, FHEE__Event_Registrations__get_event_filter_toggle_html, allows us to modify the text output of the selector drop down to include other distinguishing information. 

For example, we can add the event start date after the event name, making events distinguishable:

	function jv_ee_add_date_to_text( $evts ) {

	foreach ($evts as $key => $e){

		global $wpdb; 
		$start_date = $wpdb->get_var( $wpdb->prepare(' SELECT `DTT_EVT_start` FROM `wp_2_esp_datetime` WHERE `EVT_ID` = %d ', $e['id'] ) );
		$start_date = date("m/d/Y", strtotime($start_date));

		$evts[$key]['text'] = $e['text'].' - '.$start_date;

	}

	return $evts;

	}
	add_filter( 'FHEE__Event_Registrations__get_event_filter_toggle_html', 'jv_ee_add_date_to_text', 10, 3 );

**Thanks for your pull request!**

Please answer the following questions in order to expedite its acceptance.

####What problem does this work solve, or what benefit does it have?

####Why do you think these changes are needed in Event Espresso core instead of being put in an add-on? 
> Additions to EE core should benefit 80% of its users, otherwise the work is probably best put in an add-on

####Why do you think this is the best implementation?

####Does this include unit tests? Or how can it be tested?
> Automated tests not only demonstrate the code works, but also help in ongoing maintenance. So pull requests with automated tests are preferred.

##Please indicate

### DIFF
--- a/caffeinated/admin/extend/registrations/EE_Event_Registrations_List_Table.class.php
+++ b/caffeinated/admin/extend/registrations/EE_Event_Registrations_List_Table.class.php
@@ -202,7 +202,7 @@ class EE_Event_Registrations_List_Table extends EE_Admin_List_Table {
 				}
                                 $evts[] = array(
 					'id'    => $evt->ID(),
-					'text'  => apply_filters('FHEE__EE_Event_Registrations___get_table_filters__event_name', $evt->get( 'EVT_name' ), $event),
+					'text'  => apply_filters('FHEE__EE_Event_Registrations___get_table_filters__event_name', $evt->get( 'EVT_name' ), $evt),
 					'class' => $evt->is_expired() ? 'ee-expired-event' : '',
 				);
 				if ( $evt->ID() === $current_EVT_ID && $evt->is_expired() ) {
--- a/caffeinated/admin/extend/registrations/EE_Event_Registrations_List_Table.class.php
+++ b/caffeinated/admin/extend/registrations/EE_Event_Registrations_List_Table.class.php
@@ -209,6 +209,11 @@ class EE_Event_Registrations_List_Table extends EE_Admin_List_Table {
 					$checked = '';
 				}
 			}
+			$evts = apply_filters( 
+				'FHEE__Event_Registrations__get_event_filter_toggle_html', 
+				$evts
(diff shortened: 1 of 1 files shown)


mnelson4:
thanks for the pull request @JRyven,
The filter is named slightly incorrectly, it should include the method's name in the filter name. eg `FHEE__EE_Event_Registrations___get_table_filters__events`.
Also, if you want to just filter how the event names are displayed in the list table, you could also just add a filter onto that. eg add a filter on line 205 so it becomes 

```
                                $evts[] = array(
					'id'    => $evt->ID(),
					'text'  => apply_filters('FHEE__EE_Event_Registrations___get_table_filters__event_name', $evt->get( 'EVT_name' ), $event),
					'class' => $evt->is_expired() ? 'ee-expired-event' : '',
				);
```
Then your callback would just be something like 

```
function add_date_onto_event_name( $event_name, $event){
        return $event_name . ' ' . $event->primary_datetime()->get_pretty('DTT_EVT_start');
}
add_filter( 'FHEE__EE_Event_Registrations___get_table_filters__event_name', 'add_date_onto_event_name', 10, 2);

What do you think?

JRyven:
@mnelson4 - I think your corrections and suggestions are wonderful. 

I had the filter after $evts in case someone wanted to modify something else about the dropdown options, but I can't think of a scenario when there would be a benefit to accessing the id. I suppose someone might want to add a class to certain events... but probably not. 

Your placement and implementation is far more elegant.

mnelson4:
great @JRyven, although I noticed a typo (which was my original bad): the variable we pass into the filter should be `$evt`, not `$event`. 
If you get a chance to amend that, then we can accept this as-is. If you're busy I might get a chance to do it later today

mnelson4:
ok merged. Thanks for the PR!

PR #214, opened 2017-03-12
Remove last invalid param from `wp_register_style` call ...

oldskool73:
...that breaks JQueryUI css inclusion.

Ref : https://eventespresso.com/topic/jquery-ui-css-not-included-correctly-when-using-datepicker/

**Thanks for your pull request!**

Please answer the following questions in order to expedite its acceptance.

####What problem does this work solve, or what benefit does it have?

See support post above. JQuery UI CSS media type is set to '1' (invalid) when datepicker is used in registration form.

####Why do you think these changes are needed in Event Espresso core instead of being put in an add-on? 

Fix the broken styles.

####Why do you think this is the best implementation?

Fixes the bug.

####Does this include unit tests? Or how can it be tested?

Not got time to fix your code AND write your tests for a paid product, sorry.

##Please indicate
* [NA] User input is adequately validated and sanitized
* [NA] all publicly displayed strings are internationalized (usually using `esc_html__()`, see https://codex.wordpress.org/I18n_for_WordPress_Developers)

### DIFF
--- a/core/libraries/form_sections/inputs/EE_Datepicker_Input.php
+++ b/core/libraries/form_sections/inputs/EE_Datepicker_Input.php
@@ -33,7 +33,7 @@ class EE_Datepicker_Input extends EE_Form_Input_Base{
 	 */
 	public static function enqueue_styles_and_scripts() {
 		// load css
-		wp_register_style( 'espresso-ui-theme', EE_GLOBAL_ASSETS_URL . 'css/espresso-ui-theme/jquery-ui-1.10.3.custom.min.css', array(), EVENT_ESPRESSO_VERSION, TRUE );
+		wp_register_style( 'espresso-ui-theme', EE_GLOBAL_ASSETS_URL . 'css/espresso-ui-theme/jquery-ui-1.10.3.custom.min.css', array(), EVENT_ESPRESSO_VERSION );
 		wp_enqueue_style( 'espresso-ui-theme');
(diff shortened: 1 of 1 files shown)


nerrad:
Hi there, thanks for catching this obvious error, _really_ appreciated.  We'll hopefully get this merged in to master sometime today.

I just want to comment on this statement:

> Not got time to fix your code AND write your tests for a paid product, sorry.

EE core is not a paid product.  We charge for a _support_ license which allows non technical user's to get access to professional support and automatic updates.  EE core is actually an open source plugin (as you can see) offered on github and we've actually found that many developers use this directly to provide solutions to their clients and these developers _are_ getting paid by clients to do so.  We don't have a problem with that, but we DO appreciate when the same developers offer their time to contribute to this project because it not only helps us, but it helps _them_. That's the nature of open source, and its a wonderful thing.

As maintainers of this project, we encourage (and in some cases _require_) unit tests for changes because its a way to ensure the quality of the code being added/modified is maintained and contributing to improving stability for _all_ users of Event Espresso.  However, there are cases where obviously this won't be as important of a requirement (such as in your case) because the nature of the code change is so minor.

Thanks again for your contribution!

mnelson4:
thanks for your contribution @oldskool73. Ya if unit tests don't really apply it's nice to just give us an idea of how to manually test.

PR #286, opened 2018-03-29
add event-select component

nerrad:
This introduces an `<EventSelect />` component that can be reused in various Gutenberg (and general react contexts).

Done as a part of the work for https://events.codebasehq.com/projects/event-espresso/tickets/11400

I haven't tested this with any block yet so that will have to be done once we have blocks to test with.

Also, before merging into the base branch I'd like to:

- load moment via external (see https://events.codebasehq.com/projects/event-espresso/tickets/11473)
- create and utilize a data store for the `events` so we can wrap with the wp.data.withData HOC for the default export rather than `withAPIData`.  This is needed for performance reasons (see https://events.codebasehq.com/projects/event-espresso/tickets/11472)

Optional that probably would be good after merge:

- the `mapOrderBy` and `whereConditions` functions might work well as utility functions that we can reuse in multiple contexts in our js.
- we might want to consider creating a dynamic `propTypes` schema for each of our models from our REST API and its cached as a property on our global `eejs` object.  That way we can easily pull those schema anytime we define propTypes for a component/block.
- we might also want to consider creating some sort of PropType generator for other various model query parameters that help with prepping for requests using data.


assets/src/components/selection/event-select/index.js
```diff
@@ -0,0 +1,165 @@
...
+import PropTypes from 'prop-types';
+
+/**
+ * WP dependencies
+ */
+const { Component } = wp.element;
+const { Placeholder, SelectControl, withAPIData } = wp.components;
+const { __ } = wp.i18n;
+
+const nowDateAndTime = moment();
```
tn3rb: Should this maybe be moved up with the External dependencies?
Its current location almost makes it appear as though it is a WP dependency.

assets/src/components/selection/event-select/index.js
```diff
@@ -0,0 +1,165 @@
...
+  };
+  return isUndefined(orderByMap[orderBy]) ? orderBy : orderByMap;
+};
+
+const whereConditions = ({showExpired, categorySlug, month}) => {
+  let where = [];
+  const GREATER_AND_EQUAL = encodeURIComponent('>=');
+  const LESS_AND_EQUAL = encodeURIComponent('<=');
+  //using .replace('Z', '') on the end of our moment strings because EE currently has a bug with parsing default
+  //moment formatted strings (@see https://events.codebasehq.com/projects/event-espresso/tickets/11368)
```
tn3rb: Did you forget to do this replacement?
I don't see this applied anywhere in the code below!?!

assets/src/components/selection/event-select/index.js
nerrad: ya good point, however the extra space I think delineates its not part of the block.

assets/src/components/selection/event-select/index.js
nerrad: hmm nope, copy/pasta from code I did elsewhere (the eea-gutenberg add-on), the issue got fixed upstream in EE core and thus its no longer needed.  I'll just remove this comment.

assets/src/components/selection/event-select/index.js
tn3rb: ya that's fine and I knew you would say that, but initially I was just looking for some kind of comment to add !

PR #287, opened 2018-04-01
Introduce Asset Caching by Filename

nerrad:
This pull request introduces the following things:

## Hash based filenames for assets.

All assets will now include a content hash in the filename.  This hash will change whenever the content of the file changes.  This is a more [modern best-practice for cache-busting assets](https://www.alainschlesser.com/bust-cache-content-hash/).  With content hash file names, the build process will also generate a `build-manifest.json` file that contains a map of asset "chunk-name" to final file name.   Example:

```json
{
    "reactVendor": {
        "js": "ee-reactVendor.83c902271dfaf7c14e74.dist.js"
    },
}
```

For ease of registering built-scripts, `EventEspresso\core\services\assets\Registry` has a new method called `getAssetUrl($namespace, $chunk_name, $asset_type)`.

`$namespace` should be a unique slug for referencing a specific manifest.`$chunk_name` should correspond to the name given the built-file.  For our example above, the "chunk name" would be `reactVendor`.  The third argument `$asset_type`, is for indicating whether the url requested is for `EventEspresso\core\services\assets\Registry::ASSET_TYPE_JS` or `EventEspresso\core\services\assets\Registry::ASSET_TYPE_CSS` (eventually we may do something similar for svg's or other core images as well, but that's not needed yet so left alone for now).

## Splitting out "vendor" libraries and webpack boilerplate from builds.

Since vendor libraries will likely not change too much over time, when doing this style of cache-busting, it's preferable to split out vendor libraries from the built files.  That allows browsers to cache these files (advantageous because they are usually bigger in size) far longer and app files that change more often can have their hash generated separately.  Another advantage of separate vendor builds is that we can group common library components together into one file and re-use that across different app.js builds.  It also means if WordPress comes _with_ those libraries installed eventually, it's much easier to just switch to the WordPress implementation.

Besides the vendor libraries.  Another thing that needed to be split out from our builds is the webpack `manifest.js` file.  This contains Webpack boilerplate code that also shouldn't change as frequently as app.js so is [recommended](https://webpack.js.org/guides/caching/#extracting-boilerplate) to split it out.

## Retooling `eejs` global to be built as our own internal library.

`eejs` was loaded using the old system.  I wanted to move it into the new build process both so its getting the hash file format on builds and also because `eejs-core` can now serve as the dependency for ALL our registered js scripts in the new system which will automatically ensure the `ee-manifest` (webpack boilerplate) is set as a dependency for our built files.  `eejs` is wired up so that you can import it the same as other modules eg:

```js
```

It's also wired up so that if you use the `EventEspresso\core\services\assets\Registry::addData` (and related methods) api to register data, that will be exposed to any files referencing it in your source.

### Finish implementation of `@wordpress/browserslist` and `autoprefixer` plugin.

[browserslist](https://github.com/browserslist/browserslist) is a npm package that provides a sematical way of indicating what browsers your app supports.  WordPress has extended that with their [own package](https://github.com/WordPress/packages/tree/master/packages/browserslist-config) that is used to define what [browsers WordPress supports](https://github.com/WordPress/packages/blob/master/packages/browserslist-config/index.js) and I figured it be fitting for us to use that as a dependency in our package (see entry in `package.json`).  Various other modules will use a defined browserslist configuration in the services they provide.

For example, I also implemented the `autoprefixer` plugin (via `postcss-loader`) which uses browserslist.  This plugin will add _needed_ vendor prefixes for selectors that are different across different browsers and different browser versions.  For example `border-radius` would automatically get `-moz-` and `-webkit-` prefixed **if** the browserlists configuration indicated support for a browser required it.  So as developers, we can just focus on writing css without worrying about vendor prefixes and the build process will automatically add them _if necessary_.  

##  Still to do:

Assuming this passes code review, this can be merged in as is (into master) and we can start using it for all our builds.  Once its merged into master I'll need to:

* [ ] https://events.codebasehq.com/projects/event-espresso/tickets/10039 will need updated for the new builds
* [ ] https://events.codebasehq.com/projects/event-espresso/tickets/6645 (which is dependent on 10039) will need updated.
* [ ] Any gutenberg work based branches will need updated.
* [x] Add documentation to our docs folder for the build process especially dealing with how to register scripts for this (using `Registry::getAssetsUrl`).  

Also, Right now, I have the hashing happening on dev and production builds.   For now its not really affecting the build time significantly to do on both.  If/when this becomes an issue we can look into only doing the chunk hashing on production builds and not doing that on dev builds.  I don't want to take the time to do that yet though because we may find its not necessary for a while (and there may be additional tooling that comes out later that could make it easier).

### DIFF
--- a/core/services/assets/Registry.php
+++ b/core/services/assets/Registry.php
@@ -138,7 +138,7 @@ class Registry
         );
         wp_register_script(
             'ee-vendor-react',
-            $this->getAssetUrl(self::ASSET_NAMESPACE, 'vendorReact', self::ASSET_TYPE_JS),
+            $this->getAssetUrl(self::ASSET_NAMESPACE, 'reactVendor', self::ASSET_TYPE_JS),
             array('eejs-core'),
             null,
             true


assets/src/eejs/index.js
```diff
@@ -0,0 +1,17 @@
+/**
+ * Exported to the `eejs` global.
+ */
+
+/**
+ * This will hold arbitrary data assigned by the Assets Registry.
+ * @type {{}}
+ */
+export let data = eejsdata.data || {};
```
nerrad: `eejsdata` is provided by `EventEspresso\core\services\assets\Registry` as a part of the data api.  Thus on page load, `eejs.data` will automatically be populated by whatever data was registered.   Things are done this way because `eejs.data` itself is protected and cannot be manipulated outside of the global (which I _think_ is probably a good thing?)

core/services/assets/Registry.php
```diff
@@ -256,6 +283,57 @@ public function getData($key)
...
+    {
+        if (! empty($this->cached_manifests)) {
+            //already registered get out.  This usually means that a $chunk_name (argument for getAssetUrl call) either
+            //doesn't exist in any registered manifest files.  This could be because the $chunk_name is an actual file
+            //url and thus not registered in the manifest, or a plugin registered its manifest files too late.
+            return;
+        }
+        $registered_manifest_files = apply_filters(
+            'FHEE__EventEspresso_core_services_assets_Registry__registerManifests__registered_manifest_files',
+            $this->domain->distributionAssetsPath() . 'build-manifest.json'
```
nerrad: I added this filter because this will allow our add-ons to register their manifests to automatically be cached as well (once add-ons start using the same build process).  The only catch to this is add-ons will have to make sure they register their manifests early enough in the loading process so we may want to add an official Plugin registration API interface for this (and in turn add it to the `EE_Register_Addon` api?

core/services/assets/Registry.php
```diff
@@ -256,6 +283,57 @@ public function getData($key)
...
+            if (! file_exists($file)) {
+                throw new InvalidFilePathException($file);
+            }
+            /** recommended to avoid array_merge in loops */
+            $this->cached_manifests[] = json_decode(file_get_contents($file), true);
+        }
+        //PHP below 5.6
+        $this->cached_manifests = call_user_func_array('array_merge', $this->cached_manifests);
+        //We can use this once we support PHP5.6+
+        //$this->cached_manifests = array_merge(...$this->cached_manifests);
```
nerrad: I soooooo wish we could just use the spread syntax now 😖!  It would be so much more concise (and clearer imo).

package.json
```diff
@@ -10,15 +10,22 @@
   },
   "author": "Event Espresso",
   "license": "GPL-2.0",
+  "browserslist": [
+    "extends @wordpress/browserslist-config"
+  ],
```
nerrad: This is super nifty because there's a number of packages that support the `browserslist` configuration.  I think one of the `babel` plugins we are using also follows it.

webpack.common.js
```diff
@@ -1,29 +1,71 @@
...
+            reactVendor: reactVendorPackages,
             //example
             // 'ee-shortcode-blocks': [
             //     assets + 'blocks/index.js'
             // ],
         },
         plugins: [
             new CleanWebpackPlugin(['assets/dist']),
-            new ExtractTextPlugin('ee-[name].dist.css')
+            new ExtractTextPlugin('ee-[name].[contenthash].dist.css'),
```
nerrad: This requires the use of `[contenthash]` instead of `[chunkhash]` because if I use `[chunkhash]` that will just inherit whatever was generated for the base js containing the css that is extracted.  `[contenthash]` is a command-reference specific to the `ExtractTextPlugin`.

webpack.common.js
```diff
@@ -36,13 +78,31 @@ config = [
...
                             query: {
                                 modules: true,
                                 localIdentName: '[local]'
+                            },
+                            //can't use minimize because cssnano (the dependency)
+                            //doesn't parser the browserlist extension in packag.json
+                            //correctly, there's a pending update for it but
+                            //css-loader doesn't have the latest yet.
+                            // options: {
+                            //   minimize: true
```
nerrad: I spent a bit of time on this because I really wanted to minimize our css on production builds.  However it can't be done yet because of some dependency conflicts between the `cssnano` hasn't been updated yet to use the latest `browserslist` version which supports the `extends` syntax within `package.json`.  So just left this comment so when `cssnano` is updated to version 4 (currently its a release candidate) we'll be able to use it.  

It's possible at that point I may need to update the webpack config to webpack4 as well (which was just released recently).  But that still has a few quirks with not all packages supporting webpack 4 yet so I'm holding off for the time being.

core/services/assets/Registry.php
```diff
@@ -81,20 +93,35 @@ public function __construct(
...
+            $this->getAssetUrl('manifest', self::ASSET_TYPE_JS),
             array(),
-            EVENT_ESPRESSO_VERSION,
+            null,
+            true
+        );
+        wp_register_script(
+            'eejs-core',
+            $this->getAssetUrl('eejs', self::ASSET_TYPE_JS),
+            array('ee-manifest'),
```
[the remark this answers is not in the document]
tn3rb: 👍

package.json
tn3rb: ya this was a great idea to add this.
also looking forward to not having to add vendor prefixes to css due to autoprefixer.

webpack.common.js
```diff
@@ -36,13 +77,31 @@ config = [
...
-                        combineLoaders([{
+                        combineLoaders([
+                          {
                             loader: 'css-loader',
                             query: {
                                 modules: true,
                                 localIdentName: '[local]'
+                            },
+                            //can't use minimize because cssnano (the dependency)
+                            //doesn't parser the browserlist extension in packag.json
```
tn3rb: `packag.json` typo?

assets/src/index.js
```diff
@@ -0,0 +1,7 @@
+import React from 'react';
+import './style.css';
+
+const js = 'something';
+console.log(js);
+console.log('something_else');
```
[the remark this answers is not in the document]
tn3rb: ignore this... just saw the later commit where this was removed

tn3rb:
@nerrad 
I commented on some of my comments after thinking some more about things, so you'll need to click on the "Show Outdated" links to see everything I wrote (one set of comments is indeed outdated)

assets/src/eejs/index.js
nerrad: Two things:

- all bundles are scoped to the bundle (unless exposed on a global).  This not only is scoped to a bundle but it is also declared with `let` so the scope will be limited to the object its attached to.

core/services/assets/Registry.php
nerrad: ya agreed, however I wonder if we should still keep the filter for plugin authors who want to do things the "wp" way vs going through our plugin api if they have a really simple extension.  Or would you rather we enforce registering via our plugin api?

webpack.common.js
nerrad: ergh yup will fix.

core/services/assets/Registry.php
```diff
@@ -256,6 +293,73 @@ public function getData($key)
...
+    /**
+     * Used to register a js/css manifest file with the registered_manifest_files property.
+     *
+     * @param string $manifest_file  The absolute path to the manifest file.
+     * @throws InvalidFilePathException
+     * @since $VID:$
+     */
+    public function registerManifestFile($manifest_file)
+    {
+        if (! file_exists($manifest_file)) {
```
nerrad: this file exists check is repeated in registerManifests because if we keep the filter we still have to do a file_exists check.  So options are:

1. Keep things as is here.
2. Ditch the file_exists check in this method and just leave the one in `registerManifests`
3. Ditch the filter and file_exists check in `registerManifests`.  

Which would you prefer?  I'm on the fence.  I'm leaning a bit towards explicitly requiring manifest files to be registered using `registerManifestFile` myself.

tn3rb: ya I would have `registerManifests()` simply call `registerManifestFile()` when looping over the returned array, then ALL of the validation logic and/or any other hooks/triggers can all reside in one place.

so 3 basically.

nerrad: Hmm... if you're suggesting ditch the filter, then I don't understand why `registerManifestFile` would get called from `registerManifests`?  If the filter is ditched then `registerManifests` will simply loop over the `registered_manifest_files` property which would have been already populated by calls to `registerManifestFiles` earlier in the request right?

tn3rb: my apologies... i read the code wrong... I was getting confused between the `$registered_manifest_files` and `$cached_manifest` properties.

I think this could be simplified.

We don't need two arrays for tracking these files, one is sufficient imho.

Here's what I recommend:

 * get rid of the `$cached_manifest` property array as there is no need to shuffle data back n forth like we are doing now (wait for the rest of the explanation before coming to any conclusions)
 * have `registerManifestFile()` as a public method for adding files and have it perform ALL of the necessary validation and decoding of the file content. All valid decoded file content gets added to  `$registered_manifest_files` array (which could even just be renamed to `$manifest_files` for extra simplicity)
 * `registerManifests()` doesn't need to loop over any existing array of data, it can simply supply an empty array as its filtered property for add-ons to add their manifest files to. If the current approach was to allow the filter to deregister files, then that's fine, although this can easily be achieved by also adding the `$manifest_files` array property as a parameter on the filter. This would allow callbacks to edit the existing data if need be and probably provide better access to that data rather than having no option but to completely overwrite an existing manifest file (although I'll admit that I don't know what that content will look like so maybe this is a moo point).
 * so any code with access to the assets Registry class (like the Register Addon API one day) can call `registerManifestFile()` and directly register a file, or code can hook into the filter and add their file that way, which will also end up calling  `registerManifestFile()`.
 * then when `getAssetUrl()` is called, it will still call `registerManifests()`, which adds any *extra* manifest files to the `$manifest_files` array which `getAssetUrl()` uses to return its result.

does that make sense?

does the same thing, but is a little more simple and straightforward.

nerrad:
tests are failing for this because there's no `build-manifest.json` in the `dist` directory yet.  I'll have to run a build and commit that in order for tests to pass.  I'll do that before the merge.

assets/src/eejs/index.js
nerrad: Just making note that Brent and I talked this out in slack and we're leaving as is.

PR #288, opened 2018-04-04
Fet/gutenberg/11400/block manager with new config

nerrad:
This brings in the changes to config recently merged into master.  Other things worth noting:

- looks like this "base" branch still had a number of references to classes not in this branch (yet) so I modified that.
- I've added `EventEspresso\core\services\assets\Registry` as a dependency on the various block management api and updated the corresponding dependency map where needed.

Once this passes review it can be merged back into the base branch for https://events.codebasehq.com/projects/event-espresso/tickets/11400.


nerrad:
Unfortunately there's so many changes in this pull request because it was quite out of date from master.  So along with the configuration changes, there's a number of other changes from the merge from master.

nerrad:
@tn3rb this might be a cleaner list of changes to review: https://github.com/eventespresso/event-espresso-core/pull/288/files/ece259a8d71158705082a3c9725187ae2e7a5e4b..072f2b1c84b96bd689db57ebc8e7040151994333

It doesn't contain all the changes from master (re the config changes) but those are things that you've already seen anyways.  The commits mostly pertain to additional edits needed to the base branch for integrating with the changes (plus some cleanup).

core/services/editor/EditorBlockRegistrationManager.php
```diff
@@ -84,12 +85,13 @@ protected function populateEditorBlockCollection()
...
-                        'EventEspresso\core\domain\entities\editor\blocks\shortcodes',
-                        'EventEspresso\core\domain\entities\editor\blocks\widgets',
-                    )
+                    array()
+                    // array(
+                    //     'EventEspresso\core\domain\entities\editor\blocks\common',
+                    //     'EventEspresso\core\domain\entities\editor\blocks\editor',
+                    //     'EventEspresso\core\domain\entities\editor\blocks\shortcodes',
+                    //     'EventEspresso\core\domain\entities\editor\blocks\widgets',
+                    // )
```
tn3rb: why this change?
this means we are no longer loading any blocks?

core/services/editor/EditorBlockRegistrationManager.php
```diff
@@ -173,14 +175,16 @@ public function registerScripts()
     {
         wp_register_script(
             'ee-core-blocks',
-            $this->domain->distributionAssetsUrl() . 'ee-core-blocks.dist.js',
+            $this->assets_registry->getAssetUrl(Registry::ASSET_NAMESPACE, 'core-blocks', Registry::ASSET_TYPE_JS),
```
tn3rb: this is great, but I still think we should add a couple of  helper methods to the assets Registry like:

 * getJsUrl( $namespace, $chunk);
 * getCssUrl( $namespace, $chunk);

which would both hit `getAssetUrl()` with the appropriate `ASSET_TYPE` constant

core/services/editor/EditorBlockRegistrationManager.php
nerrad: those blocks don't exist in the branch this branch is based off of.  I think you forgot to move the config for those blocks into whatever branch you created for them.  I discovered this when I loaded this branch and there was an exception thrown.

core/services/editor/EditorBlockRegistrationManager.php
nerrad: That's an awesome suggestion!  Too bad you didn't make it in [this pull request](https://github.com/eventespresso/event-espresso-core/pull/287) where `getAssetUrl` was added :)

I think that's something we should do in a ticket off master (since its a helper that could get used elsewhere), thoughts?

tn3rb: I did make that comment in that pull request, but then I added an additional comment afterwards and GH collapsed the original comment and marked it outdated. 

So then I added an additional comment to the entire thread stating:

>  I commented on some of my comments after thinking some more about things, so you'll need to click on the "Show Outdated" links to see everything I wrote (one set of comments is indeed outdated)

so you must have missed that somehow!

core/services/editor/EditorBlockRegistrationManager.php
tn3rb: those aren't blocks, they're folders and I think phpStorm, or maybe Git, is deleting them because hey are empty...

might have to keep your changes for the time being until something gets added to each folder

nerrad: You can't commit empty folders in git. They aren't getting deleted. You may have them existing locally on your machine but they aren't being tracked.  Usually what I do when I want to create an "empty" folder structure to be tracked I'll add an _empty_`.gitignore` file to the folder.  

I realize these are paths, I was just carrying off your statement "this means we are not loading any blocks?", and I was simply pointing out the "blocks" I thought you were referring to don't exist on this branch and then I went on to say: _I think you forgot to move the config for those blocks into whatever branch you created for them_ because I was assuming at one point there _were_ blocks in these paths and you simply removed them and added them to a different branch (but didn't remove the config and restore the config - or in this case _path references_ - in that same branch).

So what do you want to do?  Keep these changes or add `.gitignore` so the folders are tracked?   imo I think we should only setup paths when there's actually something in them to process but I'll leave it up to you.

core/services/editor/EditorBlockRegistrationManager.php
nerrad: Ya I missed the comment, definitely would have added it in the branch had I seen it. doh!

PR #292, opened 2018-04-05
Fet/11481/extract custom post types

tn3rb:
In order to convert our CPT editors to use Gutenberg we will initially need to tap into various admin routes (such as when the editor content is saved) in order to be able to run the necessary logic on the appropriate routes.
Currently our CPTs, as well as our taxonomies and terms, are all defined and registered within EE_Register_CPTs.
This logic needs to be extracted into separate classes so that the CPT definitions can be loaded on their own without having to go through the entire registration process.

plz also see: https://events.codebasehq.com/projects/event-espresso/tickets/11481

This work splits up the logic in `EE_Register_CPTs` into the following classes:

- CustomPostTypeDefinitions
- CustomTaxonomyDefinitions
- CustomTaxonomyTerm
- RegisterCustomPostTypes
- RegisterCustomTaxonomies
- RegisterCustomTaxonomyTerms
- RewriteRules

This makes things much more SOLID and allows other classes to inject `CustomPostTypeDefinitions` in order to get an  accurate list of EE custom post types at any time without having to wait for them to be registered.


core/EE_System.core.php
```diff
@@ -1105,17 +1112,24 @@ public function perform_activations_upgrades_and_migrations()
...
+            'EventEspresso\core\domain\services\custom_post_types\RegisterCustomTaxonomies'
+        );
+        $register_custom_taxonomies->registerCustomTaxonomies();
+        /** @var EventEspresso\core\domain\services\custom_post_types\RegisterCustomPostTypes $register_custom_post_types */
+        $register_custom_post_types = $this->loader->getShared(
+            'EventEspresso\core\domain\services\custom_post_types\RegisterCustomPostTypes'
+        );
+        $register_custom_post_types->registerCustomPostTypes();
+        // register Custom Post Types and Taxonomies
         $this->loader->getShared('EE_Register_CPTs');
```
nerrad: why are you still calling the deprecated class here?

core/EE_System.core.php
```diff
@@ -1207,6 +1221,11 @@ public function initialize()
     public function initialize_last()
     {
         do_action('AHEE__EE_System__initialize_last');
+        /** @var EventEspresso\core\domain\services\custom_post_types\RewriteRules $rewrite_rules */
+        $rewrite_rules = $this->loader->getShared(
+            'EventEspresso\core\domain\services\custom_post_types\RewriteRules'
+        );
+        $rewrite_rules->flushRewriteRules();
```
nerrad: why are you flushing the rewrite rules here?  This means rules are flushed on _every_ request and that will cause significant problems (not to mention being a very expensive thing to do on every request.

core/EE_System.core.php
tn3rb: we still have lots of code referencing that class in both core and add-ons and there may also be third party code referencing it, so it needs to be loaded. It no longer performs any CPT registration though.

core/EE_System.core.php
tn3rb: No it only looks like that because of the name of the method. 
The actual code checks a WP option and if it is set to `true`, THEN it will flush the rewrite rules.
This kinda follows the "Tell Don't Ask" concept where the client code simply issues a command but the called service or entity actually decides whether to follow through with it or not.

nerrad: Understood.  But how come this is _added_ here?  There wasn't anything similar existing in this location before - is this replacing similar functionality that was elsewhere?

tn3rb: this replaces the following code that used to be in the `EE_Register_CPTs` constructor

```
add_action('AHEE__EE_System__initialize_last', array(__CLASS__, 'maybe_flush_rewrite_rules'), 10);
```

nerrad: AHH, k thanks.  I should've realized that.

tn3rb:
###branch FET/11481/extract-custom-post-types
###testing

 - [ ] confirm that all EE CPTs load and save correctly and possess appropriate post support features (title, featured image, etc)
 - [ ] confirm that EE CPT taxonomies (ie: event categories) still operate as expected
 - [ ] confirm that EE CPT custom terms (ie: event tags) still operate as expected
 - go to WP Admin > EE Events > Templates, change the Event Slug, and click save 
 - check that WP Admin > Settings > Permalink Settings are set to "Post name"
 - [ ] confirm that the new event slug you entered above works and you are able to visit www.yoursite.test/your-new-event-slug/
 - feel free to experiment with CPTs in this branch and attempt to find situations where CPT related functionality has changed

sethshoultes:
Merged in the latest from master branch

sethshoultes:
Getting fatal errors with this branch activated:
` call_user_func_array:{/Users/sethshoultes/Documents/Websites/espresso-4-9.dev.cc/wp-includes/class-wp-hook.php:286} ( )../class-wp-hook.php:286 60.410462243784EE_Admin_Page_CPT->load_page_dependencies( )../class-wp-hook.php:286 70.410462244280EE_Admin_Page_CPT->_load_page_dependencies( )../EE_Admin_Page_CPT.core.php:308 80.471763004824EE_Admin_Page->_route_admin_request( )../EE_Admin_Page_CPT.core.php:406 90.473263011504call_user_func_array:{/Users/sethshoultes/Documents/Websites/espresso-4-9.dev.cc/wp-content/plugins/event-espresso-core-github/core/admin/EE_Admin_Page.core.php:1033} ( )../EE_Admin_Page.core.php:1033 100.473263011808Events_Admin_Page->_create_new_cpt_item( )../EE_Admin_Page.core.php:1033 110.474863012240EE_Admin_Page_CPT->_create_new_cpt_item( )../Events_Admin_Page.core.php:783 120.492463099056EE_Admin_Page_CPT->loadEditorTemplate( )../EE_Admin_Page_CPT.core.php:1390 130.494163291136include_once( '/Users/sethshoultes/Documents/Websites/espresso-4-9.dev.cc/wp-admin/edit-form-advanced.php' )../EE_Admin_Page_CPT.core.php:1421 140.614264733528do_meta_boxes( )../edit-form-advanced.php:707 150.614864735696call_user_func:{/Users/sethshoultes/Documents/Websites/espresso-4-9.dev.cc/wp-admin/includes/template.php:1063} ( )../template.php:1063 160.614864737504post_submit_meta_box( )../template.php:1063 170.632864753976get_delete_post_link( )../meta-boxes.php:282 180.633364754848apply_filters( )../link-template.php:1385 190.633364755656WP_Hook->apply_filters( )../plugin.php:203 200.633364757800call_user_func_array:{/Users/sethshoultes/Documents/Websites/espresso-4-9.dev.cc/wp-includes/class-wp-hook.php:286} ( )../class-wp-hook.php:286 210.633364758640EE_Admin_Page_CPT->modify_delete_post_link( )../class-wp-hook.php:286 220.633364758992EE_Admin_Page_CPT->_set_model_object( )../EE_Admin_Page_CPT.core.php:1217 230.633464759136EE_Register_CPTs::get_cpt_model_names( )../EE_Admin_Page_CPT.core.php:708 240.633564759296EventEspresso\core\domain\entities\custom_post_types\CustomPostTypeDefinitions->getCustomPostTypeModelNames( )../EE_Register_CPTs.core.php:158 `

`Catchable fatal error: Argument 2 passed to EventEspresso\core\domain\entities\custom_post_types\CustomPostTypeDefinitions::getCustomPostTypeModelName() must be of the type array, null given, called in /Users/sethshoultes/Documents/Websites/espresso-4-9.dev.cc/wp-content/plugins/event-espresso-core-github/core/domain/entities/custom_post_types/CustomPostTypeDefinitions.php on line 251 and defined in /Users/sethshoultes/Documents/Websites/espresso-4-9.dev.cc/wp-content/plugins/event-espresso-core-github/core/domain/entities/custom_post_types/CustomPostTypeDefinitions.php on line 265`

core/admin/EE_Admin_Page_CPT.core.php
```diff
@@ -703,10 +714,15 @@ protected function _set_model_object($id = null, $ignore_route_check = false, $r
             //get out cuz we either don't have a model name OR the object has already been set and it has the same id as what has been sent.
             return;
         }
-        //if ignore_route_check is true, then get the model name via EE_Register_CPTs
+        // if ignore_route_check is true, then get the model name via CustomPostTypeDefinitions
         if ($ignore_route_check) {
-            $model_names = EE_Register_CPTs::get_cpt_model_names();
             $post_type   = get_post_type($id);
+            \EEH_Debug_Tools::printr($post_type, '$post_type', __FILE__, __LINE__);
```
nerrad: should remove this debug code

core/domain/entities/custom_post_types/CustomPostTypeDefinitions.php
```diff
@@ -229,16 +229,13 @@ public function getCustomPostTypeModelNames($post_type_slug = '')
...
             if (
-                ! isset($cpts[ $post_type_slug ])
-                || (
-                    isset($cpts[ $post_type_slug ])
-                    && (
-                        empty($cpts[ $post_type_slug ]['class_name'])
-                        && empty($cpts[ $post_type_slug ]['model_name'])
-                    )
-                )
+                empty($cpts[ $post_type_slug ]['class_name'])
```
nerrad: I'm puzzled why you split out this condition into multiple conditions?  Was it for readability? If anything, it could have been simplified to:

```php
if (
    ! isset($cpts[ $post_type_slug ])
    || (empty($cpts[ $post_type_slug ]['class_name'])
         && empty($cpts[ $post_type_slug ]['model_name'])
     )
)
```

tn3rb: ya that would work too.

I split it up because it is more readable imo, and also because it means that sometimes only the first conditional will get tested, whereas using an OR means that both parts of that conditional will always get tested. So i prefer it as is now.

nerrad: I'm okay if you want to leave it as is, but just want to address this:

> whereas using an OR means that both parts of that conditional will always get tested.

That's not really accurate.   If the first part of the conditional is a match, other conditional parts will not be read, that's one of the reasons why in statements like this we always put the least expensive conditional checks first.

tn3rb:
@sethshoultes 
ready for further testing

PR #293, opened 2018-04-10
Implement WordPress code style standards for js (via es-lint)

nerrad:
This has a bunch of configuration modification for implementing eslint which will help with ensuring adherence to the same coding standards as WordPress.  

(original: https://events.codebasehq.com/projects/event-espresso/tickets/11491)

Also take note of https://github.com/eventespresso/event-espresso-core/pull/293/commits/8949773e3f8f294da446c81cee0bf69bba1a8f52.  With this commit, we no longer have to add a `import React from 'react'` statement to any of our js files using it because webpack is now configured to automatically use it on demand.  I _only_ did this with React for the time being because its usage could change if/when React is bundeled with core.

I've also tested existng javascript to make sure there is no breakage with existing features using our new build process in both development and production builds.


.eslintrc.js
```diff
@@ -0,0 +1,28 @@
...
+	globals: {
+		wpApiSettings: true,
+		eejsdata: true
+	},
+	plugins: [
+		'jest',
+	],
+	overrides: [
+		{
+			files: [ 'test/e2e/**/*.js' ],
```
nerrad: I left this in ... eventually I plan on adding e2e (end to end) tests for our javascript work.  The folder structure may change then and I'll modify this if/when that happens.

.eslintignore
```diff
@@ -0,0 +1,16 @@
...
+docs
+languages
+modules
+payment_methods
+public
+shortcodes
+tests
+ui
+widgets
+wp-assets
```
nerrad: A lot of folders excluded :)  The intent here is to only include .js files within the root folder and `assets/src` for now.  Otherwise we're going to be bombarded by js linting errors everytime we open legacy js :)

assets/src/exit-modal-survey/index.js
```diff
@@ -1,7 +1,6 @@
 /**
  * External dependencies
  */
-import React from 'react';
```
nerrad: Take note that we no longer have to import React for react related work.  This is mostly applicable where jsx is being used in the file.  However I think if we're using components of React (i.e. `Component`) it'd probably still be good to import it just to make it clear.

assets/src/exit-modal-survey/index.js
```diff
@@ -10,78 +9,76 @@ import './style.css';
...
-            window.location.href = link;
-        },
-        {once: true}
-    );
+const handleDeactivationClick = ( e, el ) => {
+	e.preventDefault();
+	const link = $( el ).attr( 'href' );
+	const appContainer = document.getElementById( 'ee-exit-survey-modal' );
+	ReactModal.setAppElement( appContainer );
+	const modalContainer = ReactDOM.render( <ExitModal { ...modalProps } />, //eslint-disable-line react/no-render-return-value
```
nerrad: take note of the way you can disable linting for a specific line.  In this case, I disabled the linting for this line because although it's not recommended to use the return value of `ReactDOM.render`, I'm using it in a way that is supposed to be safe (not the way which gave rise to the warning).  So the linting doesn't apply in this case.

eslint/config.js
```diff
@@ -0,0 +1,171 @@
+module.exports = {
+	parser: 'babel-eslint',
+	extends: [
+		'wordpress',
+		'plugin:wordpress/esnext',
+		'plugin:react/recommended',
+		'plugin:jsx-a11y/recommended',
+	],
+	env: {
+		browser: false,
```
nerrad: I think this is false because this is covering non-built linting, in other words what is being linted is NOT what gets served to the browser (we only serve _built_ files to the browser).

eslint/config.js
```diff
@@ -0,0 +1,171 @@
...
+		document: true,
+	},
+	plugins: [
+		'wordpress',
+		'react',
+		'jsx-a11y',
+	],
+	settings: {
+		react: {
+			pragma: 'wp',
```
nerrad: this should cover usages of `wp.createElement` etc for react components in Gutenberg.

eslint/config.js
```diff
@@ -0,0 +1,171 @@
...
+					object: 'Object',
+					String: 'string',
+					Void: 'void',
+				},
+				requireParamDescription: false,
+				requireReturn: false,
+			},
+		],
+		'valid-typeof': 'error',
+		yoda: 'off',
```
nerrad: and here's where `yoda` conditional checking should be turned off.

info.json
```diff
@@ -1 +1,61 @@
...
+    "src\/ISSUE_TEMPLATE.md",
+    "src\/PULL_REQUEST_TEMPLATE.md",
+    "src\/assets\/src\/**",
+    "src\/eslint\/**",
+    "src\/.babelrc",
+    "src\/.editorconfig",
+    "src\/.eslintignore",
+    "src\/.eslintrc.js",
+    "src\/.nvmrc",
+    "src\/.postcss.config.js"
```
nerrad: Notice I added a bunch more files/folders to the folder removal configuration (these are things that get removed on release builds)

package.json
```diff
@@ -6,23 +6,37 @@
...
   },
   "author": "Event Espresso",
   "license": "GPL-2.0",
   "browserslist": [
     "extends @wordpress/browserslist-config"
   ],
+  "engines": {
+    "node": ">=8.0.0",
+    "npm": ">=5.0.0"
+  },
```
nerrad: missed mentioning this in commit notes, I _think_ this will ensure the right version of node/npm is being used.  I've already updated our `ee.devbox.bots` server so we should be okay there.

package.json
```diff
@@ -6,23 +6,37 @@
...
+    "babel-plugin-transform-react-jsx": "^6.24.1",
     "babel-polyfill": "^6.26.0",
     "clean-webpack-plugin": "^0.1.18",
     "css-loader": "^0.28.11",
+    "eslint": "^4.19.1",
+    "eslint-config-wordpress": "^2.0.0",
+    "eslint-plugin-jest": "^21.15.0",
+    "eslint-plugin-jsx-a11y": "^6.0.3",
+    "eslint-plugin-react": "^7.7.0",
+    "eslint-plugin-wordpress": "git://github.com/WordPress-Coding-Standards/eslint-plugin-wordpress.git#1774343f6226052a46b081e01db3fca8793cc9f1",
```
nerrad: Changes in these files DO mean you'll need to run npm install when they get propagated to upstream branches.

readme.txt
```diff
@@ -3,6 +3,7 @@ Contributors: eventespresso, sethshoultes, garthkoyle, pebblo, nerrad, mnelson4,
 Donate link: https://eventespresso.com/pricing/?ee_ver=ee4&utm_source=wordpress_org&utm_medium=link&utm_campaign=ee4_decaf_plugin_donate_link&utm_content=Donate+link
 Tags: events, tickets, registration, event management, event manager, event booking, class booking, booking, class registration, conference registration, events calendar, events planner, event ticketing, online registration, registration, sports, training,  wordpress events, online event registration
 Requires at least: 4.1
+Requires PHP Version: 5.4
```
nerrad: heh, missed this in the notes as well.  WordPress.org now displays this tag with plugin listings so I thought it worthwhile adding.

assets/src/exit-modal-survey/index.js
```diff
@@ -10,78 +9,76 @@ import './style.css';
...
+			top: '80px',
+			left: '50%',
+			bottom: 'auto',
+			right: 'auto',
+			WebkitTransform: 'translateX(-50%)',
+			transform: 'translateX(-50%)',
+		},
+		typeFormStyle: {
+			width: '600px',
+			height: '400px',
```
tn3rb: any  plans to accommodate mobile for this? 
or is this beyond the scope of what I'm  supposed to be reviewing here?

docs/AA--Javascript-in-EE/javscript-coding-standards.md
```diff
@@ -0,0 +1,32 @@
...
+
+and
+
+```bash
+npm run lint:fix
+```
+
+The former will go through all js files in the `assets` folder and root path and report on any eslint rule violations.  The latter will automatically fix any rule violations for things that can be automatically fixed.
+
+## Setting up in PHPStorm
```
tn3rb: the official name is "PhpStorm"
see: https://www.jetbrains.com/phpstorm/specials/phpstorm/phpstorm.html?gclid=EAIaIQobChMIzYDs29ey2gIVlLXACh2IpwTrEAAYASAAEgIRgvD_BwE&gclsrc=aw.ds.ds&dclid=CPy6iN3XstoCFYpXfgod-6wDrw

docs/AA--Javascript-in-EE/javscript-coding-standards.md
```diff
@@ -0,0 +1,32 @@
...
+
+If you use PHPStorm the following steps will assist you with setting up your environment for using our javascript styles:
+
+- Go to `Languages & Frameworks > Javascript > Code Quality Tools > ESLint` and make sure the checkbox labelled `Enable` is checked.  The default options _should_ work unless you've already modified it.
+- Enable `.editorconfig` support if you haven't already.  You can do this by installing the `EditorConfig` phpstorm plugin and then go to `Editor > Code Style` and click the `EditorConfig` checkbox.  This will make sure line indentation and spacing is setup correctly.
```
tn3rb: let's also change "phpstorm" to "PhpStorm"

eslint/config.js
```diff
@@ -0,0 +1,171 @@
...
+	],
+	settings: {
+		react: {
+			pragma: 'wp',
+		},
+	},
+	rules: {
+		'array-bracket-spacing': [ 'error', 'always' ],
+		'brace-style': [ 'error', '1tbs' ],
+		camelcase: [ 'error', { properties: 'never' } ],
```
tn3rb: is it ok that some of these are not in brackets? ie: camelcase, eqeqeq, indent

info.json
tn3rb: wait... why are we removing "readme.txt" ???
and should we be keeping ISSUE_TEMPLATE.md and PULL_REQUEST_TEMPLATE.md ?

nerrad: Eventually we should be moving the github specific md files to a `.github` folder.  That's a convention github provides to keep that sort of things organized better.  Then I just have to do the pattern against `.github`

tn3rb: kk makes sense thnx

eslint/config.js
tn3rb: right kk didn't notice that

PR #294, opened 2018-04-11
Convert Exception to EE_Error for sites messing with `plugin_dir_url`

nerrad:
See https://events.codebasehq.com/projects/event-espresso/tickets/11493 for context.

When `WP_DEBUG` is true, this will still throw an exception.  When not true, an `EE_Error` notice will be created instead (and only display in the admin).

Here's what the notice appears like:

New notice

## To Test

You can add the following code to some plugin (I just added it temporarily to the beginning fo the `espresso.php` file) to cause the conditions for the notice appearing:

```
add_filter('plugins_url', function($url) {
    return preg_replace_callback(
        '!(https?://[^/|"]+)([^"]+)?!',
        function($matches) {
            if (isset($matches[0]) && $matches[0] === site_url()) {
                return "/";
            } elseif (isset($matches[0]) && strpos($matches[0], site_url()) !== false) {
                return $matches[2];
            } else {
                return $matches[0];
            }
        },
        $url
    );
});
```

* [x] Using the above code, ensure that the error notice appears when `WP_DEBUG` is not defined or set to `false` and an exception is thrown when `WP_DEBUG` is defined and `true`.

Also please provide feedback (reviewer and tester) for actual content of notice.

### DIFF
--- a/core/services/assets/Registry.php
+++ b/core/services/assets/Registry.php
@@ -383,24 +383,21 @@ class Registry
             );
         }
         if (filter_var($url_base, FILTER_VALIDATE_URL) === false) {
-            $message = printf(
-                esc_html__(
-                    'The url given for %1$s assets is invalid.  The url provided was: "%2$s". This usually happens when another plugin or theme on a site is using the "%3$s" filter or has an invalid url set for the "%4$s" constant',
-                    'event_espresso'
-                ),
-                'Event Espresso',
-                $url_base,
-                'plugins_url',
-                'WP_PLUGIN_URL'
-            );
-            if ( WP_DEBUG ) {
-                throw new InvalidArgumentException(
-                    $message
-                );
-            }
             if (is_admin()) {
                 EE_Error::add_error(
-                    $message
+                    sprintf(
+                        esc_html__(
+                            'The url given for %1$s assets is invalid.  The url provided was: "%2$s". This usually happens when another plugin or theme on a site is using the "%3$s" filter or has an invalid url set for the "%4$s" constant',
+                            'event_espresso'
+                        ),
+                        'Event Espresso',
+                        $url_base,
+                        'plugins_url',
+                        'WP_PLUGIN_URL'
+                    ),
+                    __FILE__,
+                    __FUNCTION__,
+                    __LINE__
                 );
             }
             return;


core/services/assets/Registry.php
```diff
@@ -382,16 +383,36 @@ public function registerManifestFile($namespace, $url_base, $manifest_file)
...
+                        '$url_base',
+                        $url_base
+                    )
+                );
+            }
+            if (is_admin()) {
+                EE_Error::add_error(
+                    sprintf(
+                        esc_html__(
+                            'The url given for %1$s assets is invalid.  The url provided was: "%2$s". This usually happens when another plugin or theme on a site is using the "%3$s" filter or has an invalid url set for the "%4$s" constant',
```
tn3rb: why not use this same message in the exception? I think it would be helpful for developers to know the possible cause.

nerrad: Ya I could I suppose.  The original message was based on the (incorrect) assumption that `plugins_dir_url` (via `EventEspresso\core\domain\Domain::distributionAssetsUrl`) would always return a valid url.  In the case where `registerManifestUrl` is being called by third party (or add-on) code to register a manifest, it'd be pretty clear that the provided url is incorrect.

So... the extra added here is mostly for non developers encountering this error in the wild in which case it's more _likely_ the cause is as described.

joshfeck:
I think a fatal error should be avoided in any circumstance. If WP_DEBUG is true then it should just display the same notice.

joshfeck:
The message is displayed on the front end of the site (not seeing a notice in the admin).

screen shot 2018-04-12 at 12 15 09 pm

joshfeck:
This is the code snippet I'm using to replicate the issue:

`function roots_root_relative_url($input) {
  $output = preg_replace_callback(
    '!(https?://[^/|"]+)([^"]+)?!',
    create_function(
      '$matches',
      // if full URL is site_url, return a slash for relative root
      'if (isset($matches[0]) && $matches[0] === site_url()) { return "/";' .
      // if domain is equal to site_url, then make URL relative
      '} elseif (isset($matches[0]) && strpos($matches[0], site_url()) !== false) { return $matches[2];' .
      // if domain is not equal to site_url, do not make external link relative
      '} else { return $matches[0]; };'
    ),
    $input
  );
  return $output;
}
// workaround to remove the duplicate subfolder in the src of JS/CSS tags
// example: /subfolder/subfolder/css/style.css
function roots_fix_duplicate_subfolder_urls($input) {
  $output = roots_root_relative_url($input);
  preg_match_all('!([^/]+)/([^/]+)!', $output, $matches);
  if (isset($matches[1]) && isset($matches[2])) {
    if ($matches[1][0] === $matches[2][0]) {
      $output = substr($output, strlen($matches[1][0]) + 1);
    }
  }
  return $output;
}
if ( !in_array($GLOBALS['pagenow'], array('wp-login.php', 'wp-register.php'))) {
  add_filter('bloginfo_url', 'roots_root_relative_url');
  add_filter('theme_root_uri', 'roots_root_relative_url');
  add_filter('stylesheet_directory_uri', 'roots_root_relative_url');
  add_filter('template_directory_uri', 'roots_root_relative_url');
  add_filter('script_loader_src', 'roots_fix_duplicate_subfolder_urls');
  add_filter('style_loader_src', 'roots_fix_duplicate_subfolder_urls');
  add_filter('plugins_url', 'roots_root_relative_url');
  add_filter('the_permalink', 'roots_root_relative_url');
  add_filter('wp_list_pages', 'roots_root_relative_url');
  add_filter('wp_list_categories', 'roots_root_relative_url');
  add_filter('wp_nav_menu', 'roots_root_relative_url');
  add_filter('the_content_more_link', 'roots_root_relative_url');
  add_filter('the_tags', 'roots_root_relative_url');
  add_filter('get_pagenum_link', 'roots_root_relative_url');
  add_filter('get_comment_link', 'roots_root_relative_url');
  add_filter('month_link', 'roots_root_relative_url');
  add_filter('day_link', 'roots_root_relative_url');
  add_filter('year_link', 'roots_root_relative_url');
  add_filter('tag_link', 'roots_root_relative_url');
  add_filter('the_author_posts_link', 'roots_root_relative_url');
}
// remove root relative URLs on any attachments in the feed
function roots_root_relative_attachment_urls() {
  $roots_options = roots_get_theme_options();
  if (!is_feed() && $roots_options['root_relative_urls']) {
    add_filter('wp_get_attachment_url', 'roots_root_relative_url');
    add_filter('wp_get_attachment_link', 'roots_root_relative_url');
  }
}
add_action('pre_get_posts', 'roots_root_relative_attachment_urls');`

nerrad:
> The message is displayed on the front end of the site (not seeing a notice in the admin).

Hey Josh, are you logged in to the site?  I'm wondering if the notice is getting set because it's getting generated within the context of the admin_menu bar (where briefly `is_admin` might return true).

nerrad:
Actually, nvm I just realized that the code has `printf` instead of `sprintf` which is why it was printing :/

PR #298, opened 2018-04-13
Move Github Meta files and tweak templates

nerrad:
Just some housekeeping here.  All Github templates can be moved to a `.github` folder and they’ll still get picked up by github.  This simplifies somewhat our build process because we can just remove this folder from builds.

I'm also going to use this to tweak the actual templates using comment blocks so users's don't have to do a lot of editing within the templates.


nerrad:
@sethshoultes I just did some housekeeping here on improving our github templates and moving them into a `.github` folder.  Thought I should run the improvements by you before merging.

nerrad:
Actually I'm just going to merge this in.  It can be iterated on if changes need made.

PR #300, opened 2018-04-16
Fix broken boolean indicator for whether the pricing metabox is being loaded for an event being created or not. (fixes #297)

### LINKED ISSUE #297: Filtering the defaults on creating a new event documentation
### Is this a bug report?

* [ ] yes
* [ ] no
* [x] kind of

### What branch did this occur in?
master branch

### What version of WordPress is on the site this issue happened?
4.9.5

It seems that this document may need to be updated:
https://github.com/eventespresso/event-espresso-core/blob/master/docs/T--Tutorials/modifying-defaults-for-events.md#modifying-the-defaults-on-creating-a-new-event
The reason I created this issue is because I ran through the example code from there, and each time I'd go to create a new event, `$creating_event` is always empty, so none of the code there executes.

nerrad:
See #297 for original report.  Turns out the problem isn't actually with the documentation but due to a change in behaviour for how the `_is_creating_event` property is being set in `espresso_events_Pricing_Hooks::pricing_metabox`.  I'm not sure exactly the root cause but it looks like at some point in either the evolution of WP or the evolution of EE, a post (and thus an event) ALWAYS has an initial ID on page load of the editor regardless of context (I'm guessing at some point the post_id started getting generated a lot earlier).  This in turn meant that `$this->_is_ creating_event` would always be `false`.

The fix for this simply involves checking for `post` in the request and if its empty, we're creating a event.  This will be backward compatible with older versions of WP (if older versions of WP we support were not affected by this bug).

## How has this been tested

Using the plugin [here](https://github.com/eventespresso/event-espresso-core/blob/master/docs/T--Tutorials/modifying-defaults-for-events.md#modifying-the-defaults-on-creating-a-new-event), I:

* [x] verified creating an event was filtered.
* [x] verified editing an event was left untouched.

## Checklist

* [x] My code is tested.

### DIFF
--- a/caffeinated/admin/new/pricing/espresso_events_Pricing_Hooks.class.php
+++ b/caffeinated/admin/new/pricing/espresso_events_Pricing_Hooks.class.php
@@ -996,7 +996,7 @@ class espresso_events_Pricing_Hooks extends EE_Admin_Hooks
         $event                 = $this->_adminpage_obj->get_cpt_model_obj();
         //set is_creating_event property.
         $EVT_ID                   = $event->ID();
-        $this->_is_creating_event = absint($EVT_ID) === 0;
+        $this->_is_creating_event = empty($this->_req_data['post']);
         //default main template args
         $main_template_args = array(
             'event_datetime_help_link' => EEH_Template::get_help_tab_link(


nerrad:
Hey @joshfeck this one is ready for testing:

## How has this been tested



Using the plugin [here](https://github.com/eventespresso/event-espresso-core/blob/master/docs/T--Tutorials/modifying-defaults-for-events.md#modifying-the-defaults-on-creating-a-new-event):

* [ ] verified creating an event was filtered.
* [ ] verified editing an event was left untouched.

joshfeck:
Verified this works a-OK. Ready for merging.

PR #302, opened 2018-04-17
Fet/11498/allow disabling form inputs

mnelson4:
Adds support for "disabling" form inputs, which will make the HTML tags they generate have the `disabled` attribute, and then have the request input be ignored server-side.
Corresponds to codebase ticket https://events.codebasehq.com/projects/event-espresso/tickets/11498


sethshoultes:
## What changed
Added a `disabled` property onto `EE_Form_Input_Base`, with setter `disable()` and getter `disabled()`. Disabling a form input will mean that its request data will be ignored server-side, and the HTML tag will have a "disabled" attribute added on it. 
(I tentatively decided to not add the extra hidden input because, while implementing #11262, I thought the server-side code will need to be aware when it's expected that inputs could be disabled anyway, and so the server-side code could just as easily mark the input as disabled based on other data. Besides, we're going to ignore the input's request data anyway)

## Review
Created pull request https://github.com/eventespresso/event-espresso-core/pull/302

## Testing
Use branch **FET/11498/allow-disabling-form-inputs**. This should just add a new feature which is currently unused in EE core. So we should just verify nothing was broken here. (Verifying the new feature works should be don eon #11262)

* [ ] Register using SPCO. 
* [ ] Update a payment method.

PR #305, opened 2018-04-17
Switch to wp_add_inline_script from wp_localize_script for eejs.data api

nerrad:
See #310 for context.  Also note, that this bumps the **required** minimum version of WordPress to **4.5** because `wp_add_inline_script` was added in that verison of WP.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -75,11 +75,11 @@ jobs:
       env: WP_MULTISITE=1
     - php: 5.6
       env: WP_MULTISITE=1
-    # wp 4.2 builds
+    # wp 4.5 builds
     - php: 5.6
-      env: WP_VERSION=4.2
+      env: WP_VERSION=4.5
     - php: 5.6
-      env: WP_VERSION=4.2 WP_MULTISITE=1
+      env: WP_VERSION=4.5 WP_MULTISITE=1
     - php: nightly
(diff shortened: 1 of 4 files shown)


nerrad:
So the failing unit test reveals an issue with the usage of `rest_validate_value_from_schema` in `EED_Core_Rest_Api`.  This function was added in WordPress 4.7.0.  The problem here is that our phpunit tests were excluding any REST API related tests if the version of WP in the test environment was less than 4.4.0 and outside of this branch our `.travis.yml` is configured to only run tests against WordPress 4.2 and the latest WP version, which is why this didn't surface until this branch.

After discussing this in slack (see [here](https://eventespresso.slack.com/archives/C02SY781M/p1523997411000198) ) we decided that I'll do this:

* [ ] Exclude REST API tests for WP versions < 4.7.0
* [ ] When this work is released (which includes bumping the minimum WP version) we should update the required text on the requirements page of EventEspresso.com to state we support WordPress 4.5+ _unless_ the REST API is needed in which case WP 4.7+ is needed.

nerrad:
The only testing that was needed for this is code using the eejs.data api which currently is the new exitmodal on deactivating ee core.  I verified that still works as expected.  The `addData` method is also covered by phpunit tests.

PR #314, opened 2018-04-18
Bug/11502/failing uni tests with wp 45

mnelson4:
Changes described on https://events.codebasehq.com/projects/event-espresso/tickets/11502


modules/core_rest_api/EED_Core_Rest_Api.module.php
```diff
@@ -894,11 +894,15 @@ protected function _get_write_params(
...
-                array(
+            if(function_exists('rest_validate_value_from_schema')){
+                $sanitize_callback = array(
                     'EED_Core_Rest_Api',
                     'default_sanitize_callback',
                 );
+            } else {
+                $sanitize_callback = null;
+            }
+            $arg_info['sanitize_callback'] = $sanitize_callback;
```
nerrad: So, if sanitize callback is null, that means that our built-in model sanitization will take over?

mnelson4: Yes, mostly. Well, if `sanitize_callback` is defined, it's used to validate the data and inform the user if there are any validation issues.
Regardless of whether that's defined, the models always sanitize the data sent into them so it matches the field's type.
So model sanitization is always there; it's just this extra "sanitization" (well, really it's validation because it outright rejects the input or accepts it) that is missing in WP 4.5 and 4.6

Issue #316, opened 2018-04-19
Convert `[ESPRESSO_EVENT_ATTENDEES]` shortcode to gutenberg block


tn3rb:
# branch Gutenberg/event-attendees-block
plz also see https://events.codebasehq.com/projects/event-espresso/tickets/11426 for further commit history

nerrad:
@tn3rb I committed a number of fixes to the Branch attached to this ticket (`Gutenberg/event-attendees-block`) along with some fixes in #466 (that should get merged to Gutenberg/master once approved).  Fixes/changes include:

- fixing some of the import schema's in the js (use the package name instead of the `wp` global, webpack takes care of building correctly)
- fix jed "domain not found" error because blocks script was importing from `@wordpress/i18n` instead of `@eventespresso/i18n`
- move files around so they are more granular/modular
- move EventSelect component to InspectorControls (you can't have inputs in the main edit area because they will be disabled - all inputs have to be in controls)
- change name of block to `EventAttendees` (react standard has components capitalized).  Also shortened the name because the source file path is a namespace so we don't have to be verbose with our component names.
- I wrapped the block in a `withSelect` hoc that was wired up to `eventespresso/lists` for storing/retrieving the list of attendees by eventId.  
- Added a **very simple** output of a list of attendee names to demonstrate things are working when event id is selected in the control.

So the block is far from finished but its at a state now where I think there will be enough for you to learn/evaluate and build off of.  I thought it'd be helpful to get it to this point mostly because it helped me iron out some bugs in some of the upstream code (stores, event-select component).

Not an exhaustive list, but things that need done:

- ensure selected event id is persisted (this is a configuration issue, if you get stuck let me know)
- style the attendee list output in the editor to be what it will be on the frontend
- consider if you want to add additional controls for controlling the attendee list output (to match existing shortcode options and maybe add additional options?)
- Work on how this will get compiled for display in the frontend content.  You _shouldn't_ load the same editor script for the frontend (block does not need registered javascript side on frontend).  Two options are to either use server rendering via block registration, or the `save` property value in the block can output a skeleton container that provides enough data for a frontend react script to pull in the attendee list.  If you go with the latter option, you'll probably want to put the javascript output of attendee list  in its own react component so it (along with any css) can be loaded either in the editor via the editor script or in the frontend via the frontend script.  If what I'm suggesting as options here is not clear, let me know 👍

nerrad:
[...]

Issue #315, opened 2018-04-19
WIP Update Editor Blocks and Editor Block Manager to Register Assets Utilizing New Build Process


tn3rb:
Editor block asset registration needs to work with new JS build process that Darren has been working on.

PR #317, opened 2018-04-19
Pull Request for 315

tn3rb:
Not sure what you have and haven't seen in this branch, and for whatever reason GH is showing a bunch of commits for CB ticket 11400.
You can likely skip those, but will want to look at those for CB ticket 11426 as well as those for #315 
A couple of the commits here are also for #316


nerrad:
@tn3rb It's showing the commits from the other branch because you based this pull off of master.  I'll redo the pull and copy your comments into a new pull request based off of the branch it gets merged back into.

nerrad:
Actually I was able to edit this pull and change the base that way.  So should be good now (aside from the obvious conflicts.  Just means it has to get a merge in from the base branch before you can merge back via github.

nerrad:
connect with #316

assets/src/blocks/widgets/event-attendees.js
```diff
@@ -0,0 +1,46 @@
+const {__}                = wp.i18n; // Import __() from wp.i18n
```
nerrad: This should be `import { __ } from '@wordpress/i18n'`  because we import from the package now.

assets/src/blocks/widgets/event-attendees.js
```diff
@@ -0,0 +1,46 @@
...
+      save: function(props) {
+        return (
+            <div className={props.className}>
+              <p>Next Upcoming Event Datetime:</p>
+            </div>
+        );
+      },
+    },
+);
+
```
nerrad: Please run `npm run lint` in this branch and/or set up your PhpStorm via [this doc](https://github.com/eventespresso/event-espresso-core/blob/master/docs/AA--Javascript-in-EE/javscript-coding-standards.md) for the javascript styles we're following.

core/domain/entities/editor/blocks/widgets/EventAttendees.php
```diff
@@ -0,0 +1,39 @@
...
+     * including setting the block type and supported post types
+     *
+     * @return void
+     */
+    public function initialize()
+    {
+        $this->setEditorBlockType(EventAttendees::BLOCK_TYPE);
+        $this->setSupportedPostTypes(array('espresso_events', 'post', 'page'));
+        $this->setAttributes(array());
+        $this->setAssetRegisterFqcn('EventEspresso\core\domain\entities\editor\blocks\CoreBlocksAssetRegister');
```
nerrad: Might be unneeded if the block is instantiated with an instance of `AssetRegister`?

core/domain/services/admin/ExitModal.php
```diff
@@ -2,6 +2,7 @@
 namespace EventEspresso\core\domain\services\admin;
 
 use EventEspresso\core\services\assets\Registry;
+use const FILTER_VALIDATE_BOOLEAN;
```
nerrad: This will cause fatals  for PHP versions < 5.6 ... it's been removed multiple times, I wonder how it crept back in :(

core/domain/values/session/SessionLifespan.php
```diff
@@ -104,11 +104,11 @@ public function inHoursMinutesSeconds($separator = ':')
      *
      * @param bool $utc If true, displays expiration in UTC
      *                  If false, displays expiration in local time
-     * @return int
+     * @return float
      */
     public function expiration($utc = true)
     {
-        return (int) current_time('timestamp', $utc) - $this->lifespan;
+        return current_time('timestamp', $utc) - $this->lifespan;
```
nerrad: did you merge from 11400 or directly from master?  This looks like a revert of changes made on master.  As mentioned yesterday I already took care of a bunch of conflicts that happened when merging master into 11400 so if you merged from 11400 then I might have resolved this conflict wrong. Yet, I'm pretty sure I resolved it to use `int` (from master) so the fact that I noticed this  here is kind of odd.

core/services/assets/Registry.php
```diff
@@ -634,8 +636,10 @@ private function removeAlreadyRegisteredDataForScriptHandle($script_handle)
         if (isset($this->script_handles_with_data[$script_handle])) {
             global $wp_scripts;
             if ($wp_scripts->get_data($script_handle, 'data')) {
-                unset($wp_scripts->registered[$script_handle]->extra['data']);
-                unset($this->script_handles_with_data[$script_handle]);
+                unset(
+                    $wp_scripts->registered[$script_handle]->extra['data'],
+                    $this->script_handles_with_data[$script_handle]
+                );
```
nerrad: k this is a bit alarming because it seems to indicate that the branch this is in is out of date. The actual code in this method _currently_ within master and `FET/11400/block-manager` is:

```
private function removeAlreadyRegisteredDataForScriptHandle($script_handle)
    {
        if (isset($this->script_handles_with_data[$script_handle])) {
            global $wp_scripts;
            $unset_handle = false;
            if ($wp_scripts->get_data($script_handle, 'data')) {
                unset($wp_scripts->registered[$script_handle]->extra['data']);
                $unset_handle = true;
            }
            //deal with inline_scripts
            if ($wp_scripts->get_data($script_handle, 'before')) {
                unset($wp_scripts->registered[$script_handle]->extra['before']);
                $unset_handle = true;
            }
            if ($wp_scripts->get_data($script_handle, 'after')) {
                unset($wp_scripts->registered[$script_handle]->extra['after']);
            }
            if ($unset_handle) {
                unset($this->script_handles_with_data[$script_handle]);
            }
        }
    }
```

So we've got a problem if its not this in the branch you are editing.  There will be some nasty conflicts to try to resolve because of the additional changes you've made in this class in this branch (and possible reverts of work could happen too)

core/services/collections/Collection.php
```diff
@@ -4,6 +4,7 @@
 use  EventEspresso\core\exceptions\InvalidEntityException;
 use  EventEspresso\core\exceptions\InvalidInterfaceException;
 use LimitIterator;
+use SplObjectStorage;
```
nerrad: since gutenberg branches are going to live outside of master for a bit.  I think it'd be wiser if you make these kind of updates in separate branches off master do a pull request, and merge them in.  Otherwise you increase the risk of conflicts when/if the same kind of cleanup happens in other branches.

core/services/collections/Collection.php
```diff
@@ -38,6 +47,37 @@ class Collection extends \SplObjectStorage implements CollectionInterface {
...
+
+
+	 /**
+      * creates a very readable unique 9 character identifier like:  CF2-532-DAC
+      * and appends it to the non-qualified class name, ex: ThingCollection-CF2-532-DAC
+      *
+	  * @return void
+	  */
+	 protected function setCollectionIdentifier() {
+         // hash a few collection details
```
nerrad: I'm curious to know the reasoning for this beyond just creating a human friendly identifier.  In other words, **why** are you creating a human friendly identifier?

Also, again, this is something that I think should be done in a separate branch as an atomic commit rather than lumped in with a potentially long living branch.

src/assets/src/blocks/query-panel/index.js
```diff
@@ -0,0 +1,127 @@
+/**
```
nerrad: all these files within `/src/assets/src` are in the wrong path.  I also don't think we should bring these into these EE core branches from `eea-gutenberg` until we actually implement them.

This is also is a possible indicator that you've merged from a different branch or have NOT merged from the `FET/Gutenberg/11400/block-manager` branch that I took care of updating from master yesterday because in that branch I deleted all these files (or maybe you saw they were deleted and added them back?)

tn3rb:
> It's showing the commits from the other branch because you based this pull off of master.

No I did not base this pull off of master. I tried initiating the pull request from my Git client (GitKraken ), which has to be done from the specific branch, so in other words, this was initiated from the `FET/Gutenberg/11426/event-attendees-block` branch. It looks to me like it just dumped in ALL of the commits since the branch had been cut. If my Git client had set the base to master, then my apologies, as that was not my intent. To avoid any future issues, I will not initiate any more PRs from GitKraken since it appears to do bad things.

> So when this work is merged, will it also complete #315 and #316?

No. This PR is **only** connected to #315. In order to get things working with the new build process changes you introduced, I needed to make some changes to the underlying block managers as well as the parent block.

> Since this branch is based off of FET/Gutenberg/11400/block-manager there's something awry and maybe its just because it hasn't been updated from that branch. Regardless, there will be a lot of conflicts and potential unnoticed reverts of work if this branch is merged back into that branch. Something has gone south here and needs fixed.
> This may be related to the branches not being in sync, but I'm concerned about the number of commits/changes in non gutenberg specific files that are showing up in this pull. These kinds of changes should be done in their own branches in more atomic changesets to reduce changes of unexpected regressions and or conflicts.

This branch just needed to be updated again from 11400. I've done that now and it's fine. You were looking at old cruft. Sry for forgetting to update before pushing. I did this later in the evening last night when I was tired (granted, I should have updated the branch prior to working on it)

> The AssetRegister stuff looks good (aside from the comments I made in a few places). But imo it would have been better had you implemented in it's OWN branch off of FET/Gutenberg/11400/block-manager and done a pull with just those changes and the files they change. The event-attendees-block branch should just be for doing an event-attendees-block. The purpose for the base branch was so that any Framework stuff is always kept up to date in that branch and the block branches are just based off of that branch and not making changes to framework stuff in their branches. Otherwise we will have conflict hell. We can talk more about this in slack if I'm not clear.



Before I could build any blocks, I needed to build the framework for loading blocks and their corresponding assets, etc.
I did that in 11400 and then started to work on 11426 to build the event attendees block that will replace the shortcode of the same name.
Then the new build process for assets was introduced and the above work stopped functioning. Which is fine.
But in order to build blocks such as the event attendees block, I need to fix the system for loading blocks and their corresponding assets, etc.,
and to do that, I need at least one working block so that I can determine if the system for loading blocks is fixed or not.

You may have noticed that the actual JS for the event attendees block does not incorporate any of the changes regarding the new build process.
That's because I haven't touched it yet because I needed to get the underlying system working first. Once the framework stuff is working, I can merge this branch into 11400, and then either start a new branch solely in GH for finishing the event attendees block, or just continue working on this one (I'll most likely cut a new branch). 

I understand your point about "not making changes to framework stuff" in other branches, but if framework changes are required to get Feature-X working, don't you think it might be difficult to do so in a branch that doesn't have Feature-X in it? How are you going to know when it is actually working or not? Like imagine trying to build a new framework for loading EE admin pages in a new branch that didn't have any actual admin page classes in it!!! Kinda tricky eh? So  I could have created a new branch from 11400 for doing this work, but I still would have had to include the event attendees block in order to actually know it was working, so... what's the difference? Either way you end up with framework changes **and a block** in your branch. It was easier and faster to just use the existing branch.

Anyways, I should have been more explicit in clarifying that this pull request was for #315 and that it needs to be completed and merged into 11400 before any further work on individual blocks can progress. This PR **is** connected to issue #315 but that's not really that obvious until you get to the very bottom of the screen and see where that detail is documented.

> Also with regards to branch naming. I've been following the schema of FET/Gutenberg/11400/{issue/ticketnum}/description so that its clear from the branch name that its based off of the 11400 branch. I think we need to try to have a way to be clear about branch base.

well how about we create a new Gutenberg base branch that's just named `FET/Gutenberg` or something similar? I certainly don't see the need of keeping 11400 in the name since that is a reference to the old Codebase ticket. We also don't want to have multiple base branches from which we are creating new branches, ie: we don't want to have `FET/Gutenberg/11426/{issue}/description` or `FET/Gutenberg/11500/{issue}/description` branches right? So maybe let's just drop 11400 from the branch name. 

Since we now have the option to create Projects and Epics, maybe we can use those as part of the naming schema. For example, I create a Project named "Gutenberg Integration" as a test just to see what was involved with that. So maybe we could move all of the Gutenberg related issues into that project, and then instead of using `FET` or `BUG` for the base of our branch names, we use the Project name (or an abbreviation of it) like `Gutenberg/{issue}/description` or `GTB/{issue}/description` ? So for example, instead of `FET/Gutenberg/11426/event-attendees-block` that branch would be named `Gutenberg/316/event-attendees-block` or `GTB/316/event-attendees-block`.

So branches starting with  `FET` or `BUG` indicate that they are cut from master, and any branch name starting with something else indicates a larger feature Project that will eventually get merged into master

Thoughts?

nerrad:
> That's because I haven't touched it yet because I needed to get the underlying system working first. Once the framework stuff is working, I can merge this branch into 11400, and then either start a new branch solely in GH for finishing the event attendees block, or just continue working on this one (I'll most likely cut a new branch).

> I understand your point about "not making changes to framework stuff" in other branches, but if framework changes are required to get Feature-X working, don't you think it might be difficult to do so in a branch that doesn't have Feature-X in it? How are you going to know when it is actually working or not? Like imagine trying to build a new framework for loading EE admin pages in a new branch that didn't have any actual admin page classes in it!!! Kinda tricky eh? So I could have created a new branch from 11400 for doing this work, but I still would have had to include the event attendees block in order to actually know it was working, so... what's the difference? Either way you end up with framework changes and a block in your branch. It was easier and faster to just use the existing branch.

I felt I had to make the point I did because it wasn't clear that's what you were already planning on doing because of all the stuff that still in the branch and the name of the branch etc.  So without that clarity I felt I had to make that point.  I totally dig what you're saying and agree sometimes you need a bare minimum of implementation to test functionality.  I'm satisfied that you plan on merging back into the base branch as soon as you know it will work.

> well how about we create a new Gutenberg base branch that's just named FET/Gutenberg or something similar?

Yup me like.

> Since we now have the option to create Projects and Epics, maybe we can use those as part of the naming schema. For example, I create a Project named "Gutenberg Integration" as a test just to see what was involved with that. So maybe we could move all of the Gutenberg related issues into that project, and then instead of using FET or BUG for the base of our branch names, we use the Project name (or an abbreviation of it) like Gutenberg/{issue}/description or GTB/{issue}/description ? So for example, instead of FET/Gutenberg/11426/event-attendees-block that branch would be named Gutenberg/316/event-attendees-block or GTB/316/event-attendees-block.

Ya I'm chill with doing that.  However, I imagine that there will be some cases where we simply create a pull request devoid of any issue number.  Because of how well pull requests work I don't think issue num has to be a requirement in the branch name.  So our pattern for branch names could be something like this:

```
{PROJECT_NAME/EPIC_NAME}|BUG|FET/[{issue-num}/]{concise-description}
```

_Base_  branches for a project should simply be `{PROJECT_NAME|EPIC_NAME}/base`

...With all the benefits  you mentioned.

Thoughts?

assets/src/blocks/widgets/event-attendees.js
```diff
@@ -0,0 +1,46 @@
+const {__}                = wp.i18n; // Import __() from wp.i18n
+const {registerBlockType} = wp.blocks; // Import registerBlockType() from wp.blocks
+
+registerBlockType(
+    'event-espresso/widgets-event-attendees',
+    {
+      title:       __('Event Attendees'),
```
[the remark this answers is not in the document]
tn3rb: will do in a PR for #316

assets/src/blocks/widgets/event-attendees.js
```diff
@@ -0,0 +1,46 @@
...
+      description: 'Displays a list of people that have registered for the specified event',
```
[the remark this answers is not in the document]
tn3rb: will do in a PR for #316

assets/src/blocks/widgets/event-attendees.js
```diff
@@ -0,0 +1,46 @@
...
+          source: 'children',
+          selector:
+                  'p',
+          type:
+                  'string',
+        },
+      },
+      // The "edit" property must be a valid function.
+      edit:        function(props) {
+        console.log(props);
```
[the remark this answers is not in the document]
tn3rb: will fix in a PR for #316

assets/src/blocks/widgets/event-attendees.js
```diff
@@ -0,0 +1,46 @@
...
+        return (
+            <div className={props.className}>
+              <p>Event Editor</p>
```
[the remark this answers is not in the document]
tn3rb: ya just ignore plz

assets/src/blocks/widgets/event-attendees.js
```diff
@@ -0,0 +1,46 @@
...
+            </div>
+        );
+      },
+
+      // The "save" property must be specified and must be a valid function.
+      //   <p>props.attributes.id</p>
+      save: function(props) {
+        return (
+            <div className={props.className}>
+              <p>Next Upcoming Event Datetime:</p>
```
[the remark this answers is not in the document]
tn3rb: ya just ignore plz

.github/label-sync.yml
```diff
@@ -30,6 +30,20 @@
...
+    - name: "status:next"
+      color: efefef
+    - name: "status:needs-testing"
+      color: efefef
+    - name: "status:merge"
+      color: efefef
+    - name: "status:review"
+      color: efefef
+    - name: "status:feedback-needed"
+      color: efefef
```
nerrad: it's things like this that make me think there was a merge to this branch directly from master instead of to 11400 and then this branch.  Currently in 11400 I'm NOT seeing these changes from master.

nerrad: Brent and i chatted about this in slack and got it resolved, 11400 just hadn't been pushed yet.

core/services/editor/EditorBlockAssetRegister.php
```diff
@@ -0,0 +1,218 @@
...
+
+/**
+ * Class EditorBlockAssetRegister
+ * Abstract parent for classes that handle asset registration for one or more editor blocks that share the same assets
+ *
+ * @package EventEspresso\core\services\editor
+ * @author  Brent Christensen
+ * @since   $VID:$
+ */
+abstract class EditorBlockAssetRegister implements AssetRegisterInterface
```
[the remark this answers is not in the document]
tn3rb: ya I initially didn't like the idea of **just** using "block" to reference these things because it just seems way too generic, but that said, we are using namespacing which pretty much negates that issue, and "block" is what WP uses, so ya... I'll change all usages of "EditorBlock" to just "Block"

core/services/editor/EditorBlockAssetRegister.php
```diff
@@ -0,0 +1,218 @@
...
+        DomainInterface $domain,
+        Registry $registry
+    ) {
+        $this->script_handle       = $script_handle;
+        $this->script_dependencies = $script_dependencies;
+        $this->style_handle        = $style_handle;
+        $this->style_dependencies  = $style_dependencies;
+        $this->domain              = $domain;
+        $this->registry            = $registry;
+    }
```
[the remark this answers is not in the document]
tn3rb: **Absolutely No**!!!

That is a bad idea that should be avoided like the plague, so please don't do this kind of thing anywhere, and I'll explain why.

A quick example that hopefully makes it instantly clear is "unit testing". 

The longer answer is that a class should directly declares its dependencies by type hinting for them in it's constructor, which allows other classes to know what it requires in order to operate. Obtaining dependencies via other objects, hides a class's dependencies and is a form of "service location", which is discouraged for many reasons (you can read more about that if you want to by Googling that term). This also binds both of those classes to each other and prevents them from being able to change if/when their responsibilities change. 

For example: 

 - class A requires both class B and C in order to function and is therefore dependent on both of those classes
 - class B is also dependent on class C and already has that dependency being injected into it
 - class A indirectly obtains class C from class B (so class A is **only** injecting class B) instead of requiring it directly
 - a new business requirement comes along (like GDPR) which forces changes to existing domain logic. 
 - what if the logic in class B that requires class C needs to be extracted into a new class (class D)? Class B would no longer have an instance of class C available to pass along to class A. Now class A needs to be refactored in order for it to obtain an instance of class C which it can no longer retrieve from class B.
 - or what if it is determined that class C needs to be split into multiple subclasses (like D & E) that extend class C, and class B's requirement changes but class A's does not? So class B now requires class D (subclass of C), and class A now requires class E (also a subclass of C). This issue might be harder to spot because Class A can still obtain an object from class B that appears to be an instance of class C, but is actually now an instance of class D, however the data and/or functionality that class A requires has now been moved into class E. 

So what this means is that class B is either forced to maintain dependencies that it no longer requires (bloat), or class A is forced to change solely because of changes required in class B (the domino effect).

Going back to the simple example of "unit testing" as a reason for not hiding dependencies, and we can see that service location obviously interferes with the ability to inject mocks into a class for the purpose of testing it. If a class is obtaining its dependencies from another class, then any injected mock now has to also inject a mock of the second class. This of course is not impossible but should immediately highlight how these classes have become tangled up together. It's even worse if that service location or injection is happening via static method calls because now the class providing the dependency can not even be substituted for a mock. This is a common problem in EE right now because of our past usage of singletons.

So bottom line: Injecting a class's dependencies directly allows them to be controlled externally at runtime, which can then be done via a DI container, or a unit testcase setup method, etc. and gives classes the freedom to change when required and also to not have changes forced upon them when not required. This greatly enhances the solidity of code because it better adheres to the Single Responsibility Principle.

core/domain/entities/editor/blocks/widgets/EventAttendees.php
tn3rb: ya that's probably a better way to do this...
I was trying to be clever and do everything via the parent class instead of having to add constructors and declare dependencies for all of the individual blocks, but in light of my big comment above regarding dependencies, this doesn't look so clever does it?

tn3rb:
Ya I think:

```
{PROJECT_NAME/EPIC_NAME}|BUG|FET/[{issue-num}/]{concise-description}
```

is a better / more concise schema for what I was suggesting, although I have a couple of suggestions/clarifications:

A) I don't know if we want to ever have `{issue-num}/` as optional for anything other than the absolute base branch for a project or epic. 

ie: we **never** want branch names like:

 - FET/some-stuff-i-done
 - BUG/fix

and even the base branches for Projects or Epics can just be the name (or abbreviation) for the Project or Epic:

```
{PROJECT_NAME|EPIC_NAME}
```

B) Do we want `PROJECT_NAME` **and `EPIC_NAME` separated by `/` or do we want that to be an `OR` ? ie: are we **always** going to use Epics within Projects? And how does that affect master branch? Can we use Epics on master without a corresponding Project?

nerrad:
> A) I don't know if we want to ever have {issue-num}/ as optional for anything other than the absolute base branch for a project or epic.

That would be true if we weren't using github.  But if we enforce the _always_ use pull request mantra in github, issue num is not as important because you have information about the branch in the pull request.   Another reason why I suggest its optional is because if one does a pull request only (i.e. there's no corresponding issue, which _will_ happen) we have a chicken/egg situation where you won't know the number of the pull request until its created.  So the switch to github makes this less of a "hard" rule as it had to be in the past.  However, I do agree if there IS an accompanying **issue** for the work, its useful to have the issue number in the branch name, I just don't think it _can_ be required.

>  Do we want PROJECT_NAME **and EPIC_NAME separated by / or do we want that to be an OR ? ie: are we always going to use Epics within Projects? And how does that affect master branch? Can we use Epics on master without a corresponding Project?

Oh, I think you misunderstood I was just using those as a OR condition mainly with the mindset that not every "project" might also be an "epic" (for instance we don't have a "Gutenberg" epic yet, but we do have a Gutenberg project.  I'm not sure we should have a hard rule on this as I imagine in _most_ cases we'll either have a project or an epic, but not both.  In the cases where we do, we could probably just do `PROJECT/EPIC` like you suggest.  But as I said, in _most_ cases I think it will end up being `PROJECT` _or_ `EPIC`.

tn3rb:
> ut if we enforce the always use pull request mantra in github, issue num is not as important because you have information about the branch in the pull request. Another reason why I suggest its optional is because if one does a pull request only (i.e. there's no corresponding issue, which will happen) we have a chicken/egg situation where you won't know the number of the pull request until its created. 

**GREAT** points. Had not considered that but totally agree.

> Oh, I think you misunderstood I was just using those as a OR condition mainly with the mindset that not every "project" might also be an "epic" 

oh ya I see now that you also suggested that as an OR, I had looked at your full schema suggestion first:
```
{PROJECT_NAME/EPIC_NAME}|BUG|FET/[{issue-num}/]{concise-description}
```
and saw the `/` instead of `|` and then didn't notice the change when you wrote:
> Base branches for a project should simply be {PROJECT_NAME|EPIC_NAME}/base

> I'm not sure we should have a hard rule on this as I imagine in most cases we'll either have a project or an epic, but not both. In the cases where we do, we could probably just do PROJECT/EPIC like you suggest. But as I said, in most cases I think it will end up being PROJECT or EPIC.

totally agree

KK so I'm going to go ahead and do the following then:

 - rename that Project I started to simply be "Gutenberg" 
 - add all of the related Gutenberg issues and PRs to the Gutenberg Project
 - rename the `FET/Gutenberg/11400/block-manager` branch to `Gutenberg/master` (using master instead of base to keep with git standards)
 - create issues for any existing Gutenberg  related tickets in codebase
 - create branches  `Gutenberg/{issue}/{description}` branches where applicable

Final thoughts?
We should also get @mnelson4's opinion on this since I like to do things as democratically as possible

nerrad:
@tn3rb there's been some moving around of things in this pull and based on your last comment I'm not sure exactly what I'm reviewing?  Am I reviewing the current state of code in _this_ branch?

core/domain/services/admin/ExitModal.php
tn3rb: removed

core/services/collections/Collection.php
nerrad: > yaaa... I could go through the trouble of creating another issue and branch for every little thing added to core, but that just adds extra stuff on to everyone's plates, and since this is not a high priority need, and this class rarely changes, I don't mind it being in this branch.

Just to be clear that's not what I'm saying here (that we have to create another issue and branch for every little thing added to core).  Put another way, what I mean is if the change we are making is something that is not directly related to this branch and touches a part of the codebase that might get changed elsewhere in another branch, it's probably _safer_ to do it as an atomic commit in a branch off of master and get it merged _back_ to master right away.  I don't think that adds extra load to us and actually potentially _helps_ prevent future issues because it reduces the chance of merge conflicts.  Also, the more atomic the commit, the quicker it gets through code review and the easier it is to verify via a quick test.

If you don't think this particular file will get modified in any other branch or that you could use your change elsewhere, then sure just leave it here.

core/domain/entities/editor/EditorBlockInterface.php
```diff
@@ -0,0 +1,93 @@
...
+     * @param string $post_type
+     * @return boolean
+     */
+    public function appliesToPostType($post_type);
+
+
+    /**
+     * @return array
+     */
+    public function getEditorContainer();
```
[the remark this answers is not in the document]
tn3rb: This was something I added based on code you wrote for the initial Gutenberg Integration Experiment plugin. 
I actually don't know what this code is for or what it is supposed to do because it wasn't documented in the original version for obvious reasons.
I think it pertains more to integrating our CPT editors with Gutenberg which we haven't gotten into fully of course, so this is very likely to change in the future.

core/domain/entities/editor/EditorBlock.php
```diff
@@ -0,0 +1,280 @@
...
+    }
+
+
+    /**
+     * @return array
+     */
+    public function getEditorContainer()
+    {
+        return array(
+            $this->namespacedEditorBlockType(),
```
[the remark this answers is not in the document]
tn3rb: see next comment

core/domain/entities/editor/EditorBlock.php
```diff
@@ -0,0 +1,280 @@
...
+
+    /**
+     * Registers the Editor Block with WP core;
+     * Returns the registered block type on success, or false on failure.
+     *
+     * @return WP_Block_Type|false
+     */
+    public function registerBlock()
+    {
+        $context ='core';
```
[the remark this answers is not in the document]
tn3rb: yup updated to get script handles from new asset register class

core/domain/DomainInterface.php
```diff
@@ -50,6 +50,18 @@ public function pluginUrl();
...
+    /**
+     * @return string
+     */
+    public function sourceAssetsPath();
+
+
+    /**
+     * @return string
+     */
+    public function sourceAssetsUrl();
```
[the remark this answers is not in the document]
tn3rb: removed

core/domain/DomainBase.php
```diff
@@ -125,6 +132,24 @@ public function versionValueObject()
     }
 
 
+    /**
+     * @return string
+     */
+    public function sourceAssetsPath()
```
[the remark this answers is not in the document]
tn3rb: removed

core/domain/values/session/SessionLifespan.php
tn3rb: branch just needed updating. is correct now

core/services/assets/Registry.php
tn3rb: branch just needed updating. is correct now

src/assets/src/blocks/query-panel/index.js
tn3rb: branch just needed updating. is correct now

tn3rb:
> Am I reviewing the current state of code in this branch?
yes plz. 
Everything except `assets/src/blocks/widgets/event-attendees.js` which will be fixed as part of #316 

> there's been some moving around of things in this pull
The classname changes were at your request as well as many of the other changes
I *think* I've addressed all of your comments. 
My apologies again for this PR being a bit of a mess.

nerrad:
> I pushed some other changes as well but touched issue #315 instead o f this PR, so plz also see the extra commits at the bottom of : #315

It looks like the commits touching #315 are actually in the `Gutenberg/event-attendees-block` branch and not the branch this pull request is for.  FYI, a _pull-request_ is directly attached to the branch it is referencing.  You don't have to touch issues for any commits added to a branch the pull request is for. 

I think it might be better at this point if we close this pull request and open a new one for the new branch and you can reference this pull request after creation.  Then I can just do the review on the new branch.  Right now I think its just too confusing to have code across two branches.

tn3rb:
replaced by https://github.com/eventespresso/event-espresso-core/pull/340

PR #331, opened 2018-04-24
Gutenberg Integration Main Feature Branch

nerrad:
This pull request is for the initial Gutenberg integration work we are doing.  

## Purpose

The work in this branch has the following purposes:

- Expose an api for registering blocks for the Gutenberg Editor (and controlling what post-types they are active for)
- Convert EE core shortcodes to Gutenberg Blocks.

Once we get to a point where we have working EE core shortcodes for Gutenberg we can prepare testing steps and prep for merge into master.  

Refactoring of the EE CPT route editors (events, venues, attendee) will happen _after_ this branch has been merged to master.

## Workflow

- Gutenberg specific changes are done in a branch off of this one and then merged back into this branch when its "stable".  So anyone checking out this pull request/branch should have a functional copy of EE core.
- Changes to js/css that assist with the Gutenberg work but are _not_ specific to it (i.e. not dependent on Gutenberg), should be done in pull requests/branches based of of master and not this branch.  This might include things like changes to build-processes or importing js libraries/npm packages that could be used elsewhere in EE js eventually.
- As much as possible, any work that is branched off of this branch should be merged back in as a squash merge.  This keeps a clean trail of "features" merged into this branch for final release.
- Do not rebase master onto this branch/pull request.  Since its serving as the base for multiple feature branches rebasing would complicate (and even potentially) break things in the various feature branches.

> For background/history, see: https://events.codebasehq.com/projects/event-espresso/tickets/11400


nerrad:
Currently there's some problems with the Block API:

* [ ] Block assets are being _enqueued_ on every route.  They should only be _registered_ on every route.
* [ ] `CoreAssetManager::JS_HANDLE_EE_JS_CORE` script handle has `setRequiresTranslation` method executed on it incorrectly (this is actually  a problem in master).  This file doesn't (and never _should_) have i18n strings associated with it.  I realize there are strings in the `translation-map.json` but those are actually coming from jed which is now bundled in `eejs` because we wrap` wp.i18n`.  They are not translation strings, just caught by the webpack mapper plugin.  The reason why we can't have `setRequiresTranslation` set on this handle is because it creates a race condition where `eejs.i18n` is called BEFORE `eejs` gets loaded (that's how the translations work).  I'll create a pull off of master for this.
* [ ] We should namespace `core-blocks` a bit better in `CoreBlocksAssetManager::JS_HANDLE_CORE_BLOCKS` and `CoreBlocksAssetManager::CSS_HANDLE_CORE_BLOCKS`.  It's very likely those handles will clash with some other script/style getting registered because of it being so generic.  I'm even concerned that `ee-` prefix might not be sufficient for the handles.  I would suggest changing to `eventespresso-core-blocks`
* [ ] There's a flaw with `i18nRegistry::queueI18n` in that it _could_ get called multiple times in a request but we need to make sure its only registering the inline script for those ONCE (I'm seeing behaviour where core blocks is its translations dumped twice in the dom).  I'll debug this and fix (probably will do a unit test as well for it - have ideas for that).  This will have to be fixed off of the master branch
* [ ] `BlockAssetManager::addScript` and `BlockAssetManager::addEditorScript` are not calling `setRequiresTranslation()` on the chained `addJavascript` call which causes errors for the ee.i18n function calls (such as `__` etc) in those scripts (you'll get a jed error saying something like 'event_espresso' domain not set).

I'll create pulls for some of these and fix them off of the appropriate branch.  I'll update this comment with the links to the pulls when they are published.

PR #332, opened 2018-04-24
Create wp.data store for retrieving/tracking events

nerrad:
> Note: for reference, original ticket is https://events.codebasehq.com/projects/event-espresso/tickets/11472

While working on creating an `EventSelector` component, I realized that although the `withApiData` HOC WP provides is useful, if that selector is used on multiple places (blocks etc), then every time the selector is loaded there's an api request to get the list of events. That could be problematic (especially when something like this lands on EventSmart). The `wp.data` feature in Gutenberg (which will eventually be a standalone package) will be useful for this because I can just register a custom store for this type of query and have it memoize (using resolvers I think) by some sort of hash based on the query. That way the same query for the events in the selector will just return the existing events object in the store.

So as I think about this, there are some things I want to do:

1. Keep a top level store that keeps tracks of entities already retrieved and being worked with.  For example `ee-core` store might have a top level `events` entities state that serves as the "authoritative" state for those event entities.
2. Need to have a way to record whether the state for an entity is dirty (i.e. `ee-core.events.1.object` would have the event entity itself and `ee-core-events.1.dirty` would be a boolean indicating whether its "dirty" (not in sync with the db = true) or not.
3. By default, any reducers in any store dealing with ee entities would need to go through some sort of middleware or otherwise so that depending on the purpose of that reducer (for example a reducer that is not intended to mutate the state of entities in the `ee-core` store), it will optionally mutate/not mutate entities if they exist (but still add new ones if they don't exist).  So for example, for this ticket I'm creating an `ee-list` store that is mostly for just retrieving lists of model entities from our api and populating the `ee-core` store from what is retrieved.  However `ee-list` resolvers and reducers will by default not mutate the entities in the `ee-core` store because they are just intended for querying for things like select dropdowns, checkboxes etc.  Things where if elsewhere the entity details are changed (i.e. Event Name) it may be desired to update that automatically within any objects referencing that store.
4. We may still want a way to optionally flush the dirty state of entities within the `ee-core` store (undo changes, or at least make it allowable to "restore" the state from some other source (db, local storage etc).

Also noting. For the resolvers, I'm using a variation of the async generator pattern that WP core is using for [their wp-core data resolvers](https://github.com/WordPress/gutenberg/blob/master/core-data/resolvers.js).  For reference here's some documentation I've been reading following for things related to them:

- [Try Selector Side Effects and withData HoC](https://github.com/WordPress/gutenberg/pull/5219)
- [Asynchronous Iterators for JavaScript](https://github.com/tc39/proposal-async-iteration)
- [Framework: Fetch categories by common terms resolver](https://github.com/WordPress/gutenberg/pull/5826)


webpack.common.js
```diff
@@ -6,6 +6,12 @@ const autoprefixer = require( 'autoprefixer' );
 const externals = {
 	jquery: 'jQuery',
 	'@eventespresso/eejs': 'eejs',
+	'@wordpress/api-request': {
+		this: [ 'wp', 'apiRequest' ],
+	},
+	'@wordpress/data': {
+		this: [ 'wp', 'data' ],
+	},
```
nerrad: doing this allows us to import as aliased package names in our source files and then if/when WrodPress releases these things as actual packages, we just have to update the webpack config and all our source files can keep the existing imports.

assets/src/components/selection/event-select/index.js
```diff
@@ -3,77 +3,79 @@
...
  * @todo some of these proptypes are likely reusable in various place so we may
  * want to consider extracting them into a separate file/object that can be
  * included as needed.
- * @type {{events: *, onEventSelect, selectLabel: *, selectedEventId: *,
- *   isLoading: *, attributes: {limit: *, orderBy: *, order: *, showExpired: *,
- *   categorySlug: *, month: *}}}
  */
 EventSelect.propTypes = {
-	events: PropTypes.shape( {
+	events: PropTypes.arrayOf( PropTypes.shape( {
```
nerrad: The original implementation of this was incorrect, so this fixes that (covered by tests as well).

assets/src/components/selection/event-select/index.js
```diff
@@ -3,77 +3,79 @@
...
 		EVT_name: PropTypes.string.required,
 		EVT_ID: PropTypes.number.required,
-	} ),
-	onEventSelect: PropTypes.func.required,
+	} ) ),
+	onEventSelect: PropTypes.func,
```
nerrad: made this not required because there's validation upstream within the `SelectControl` component.

assets/src/components/selection/event-select/index.js
```diff
@@ -101,6 +103,9 @@ EventSelect.defaultProps = {
 		order: 'desc',
 		showExpired: false,
 	},
+	selectLabel: __( 'Select Event', 'event_espresso' ),
+	isLoading: true,
+	events: [],
```
nerrad: I moved the defaults here instead of in the constructor for the component (since I was already defining defaults here).

assets/src/components/selection/event-select/index.js
```diff
@@ -160,9 +187,8 @@ export default withAPIData( ( props ) => {
 	if ( where ) {
 		queryString += '&' + where;
 	}
-
 	return {
-		events: `/ee/v4.8.36/events?${ queryString }`,
-		isLoading: false,
+		events: getEvents( queryString ),
+		isLoading: isRequestingEvents( queryString ),
```
nerrad: This is implementing the new HOC that was done in this branch.

assets/src/components/selection/event-select/test/__snapshots__/index.js.snap
```diff
@@ -0,0 +1,40 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`EventSelect with 2 events should render and match snapshot 1`] = `
```
nerrad: This is a snapshot created by Jest and the new tests I added.  This should _never_ get modified directly.

assets/src/data/ee-lists/index.js
```diff
@@ -0,0 +1,21 @@
...
+
+/**
+ * Internal dependencies
+ */
+import reducers from './reducers';
+import * as selectors from './selectors';
+import * as actions from './actions';
+import * as resolvers from './resolvers';
+
+const store = registerStore( 'eventespresso/lists', {
```
nerrad: The intended purpose for the `eventespresso/lists` store is for any components that need to retrieve a list of EE entities for given queries for use in elements of that component.  The intent being that the retrieved entities are cached by that query so if there are multiple instances of that query requested in the same view, we avoid extra hits to the REST API endpoint for that query.

core/services/assets/Registry.php
```diff
@@ -176,7 +177,10 @@ public function scripts()
                 true
             );
             $this->jsdata['eejs_api_nonce'] = wp_create_nonce('wp_rest');
-            $this->jsdata['paths'] = array('rest_route' => rest_url('ee/v4.8.36/'));
+            $this->jsdata['paths'] = array(
+                'rest_route' => rest_url('ee/v4.8.36/'),
+                'collection_endpoints' => EED_Core_Rest_Api::getCollectionRoutesIndexedByModelName()
+            );
```
nerrad: This is provided so there's an easy way for any of our React work to receive a list of collection endpoints for setting up EE REST API queries.

package.json
```diff
@@ -37,21 +37,26 @@
...
 		"eslint": "^4.19.1",
 		"eslint-config-wordpress": "^2.0.0",
 		"eslint-plugin-jest": "^21.15.1",
 		"eslint-plugin-jsx-a11y": "^6.0.3",
 		"eslint-plugin-react": "^7.7.0",
 		"eslint-plugin-wordpress": "git://github.com/WordPress-Coding-Standards/eslint-plugin-wordpress.git#1774343f6226052a46b081e01db3fca8793cc9f1",
 		"extract-text-webpack-plugin": "^3.0.2",
+		"gutenberg": "Wordpress/gutenberg#master",
 		"pegjs": "^0.10.0",
+		"pegjs-loader": "^0.5.4",
```
nerrad: This is all related to tests implemented for gutenberg type testing.  I anticipate that we'll only be doing this type of testing mostly within the gutenberg branch and thus can just get merged with the shortcode to block conversion but if it turns out this is needed in any future work on master I'll have to cherrypick changes to master and work out the conflicts when merging upstream.

tests/javascript-config/unit/jest.config.json
```diff
@@ -6,15 +6,23 @@
...
 	],
 	"preset": "@wordpress/jest-preset-default",
 	"testPathIgnorePatterns": [
 		"/node_modules/",
 		"/test/e2e"
+	],
+	"transformIgnorePatterns": [
+		"node_modules/(?!(gutenberg)/)"
 	]
 }
```
nerrad: I wrote a blog post about all the changes in this file here: http://unfoldingneurons.com/2018/testing-javascript-for-your-gutenberg-integrated-plugin

core/services/assets/Registry.php
nerrad: Hmm I don't think it's a super expensive request.  The unfortunate thing is I don't think your suggestion would work because this is actually getting generated on a _normal_ browser request *not* an api request.  It's to provide info to the js for _making_ api requests.

assets/src/data/ee-lists/reducers.js
```diff
@@ -0,0 +1,53 @@
...
+				...state,
+				[ modelName ]: {
+					[ queryString ]: items,
+				},
+			};
+	}
+	return state;
+}
+
+export default listItems;
```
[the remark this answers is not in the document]
nerrad: Ya it takes some reading :)  What I did was read up on redux and flux patterns where many of the concepts are outlined.  Then I took a hard look at what the Guteberg project is doing (see the README.md and code here: https://github.com/WordPress/gutenberg/tree/master/data)

assets/src/data/ee-lists/resolvers.js
```diff
@@ -0,0 +1,49 @@
...
+ * Resolver for event entities.
+ *
+ * @param {Object} state Data in state.
+ * @param {string} queryString Additional query string parameters passed on to
+ *   the REST request.
+ * @return {IterableIterator<*>} A async iterable.
+ */
+export function getEvents( state, queryString ) {
+	return getItems( state, 'event', queryString );
+}
```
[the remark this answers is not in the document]
nerrad: > should this last method go in the same file with the generic getItems() method?

Well it IS in the same file as the `getItems` async genrator and that's what it is calling :).  No the resolvers have to have the same function names as the selectors and they get "automagically" assembled together as a part of the store object when registering them with `wp.data.registerStore`.  In other words, a resolver that matches a selector will query the api ONCE to populate the store on the first call to that selector.  Then further calls to the selectors will just retrieve data from the store.

> and are we going to populate this same file with similar methods for other entity types? 

possibly.  I haven't decided yet.  I want to start using this in actual implementations before I decide on an appropriate pattern.  The tricky part here is that as I pointed out above there must be a matching resolver for any selector that needs populated via the REST api on the initial query.

Another spoke to this (which isn't found in this branch yet) is adding actions are _dispatched_ to the store for updating data.  I don't want to start adding those actions until we start seeing how this will work in full implementations.  

So I've focused on having the generic functions and a single concrete (event) to begin using and then we'll add more as we need them.  Once we start seeing patterns emerge we can work on implementing more DRY architecture.

assets/src/data/ee-lists/selectors.js
```diff
@@ -0,0 +1,51 @@
...
+/**
+ * Selector specific to events for checking if requesting events.
+ *
+ * @param {Object} state Data state.
+ * @param {string} queryString The query string for the request
+ * @return {boolean} Whether items are being requested or not.
+ */
+export function isRequestingEvents( state, queryString ) {
+	return isRequestingItems( state, 'event', queryString );
+}
```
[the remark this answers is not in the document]
nerrad: Much like the other comments I've made so far, ya we may end up with that, but I think its too early to scope that out yet.  I'd like us to watch for patterns to emerge first.

core/services/assets/Registry.php
tn3rb: ahhh kk  then nvmd

assets/src/data/ee-lists/resolvers.js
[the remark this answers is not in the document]
tn3rb: > `should this last method go in the same file with the generic getItems() method?`
> 
> Well it IS in the same file as the getItems async genrator and that's what it is calling :).

sigh... I realize that there are a **lot** of aspects to this new JS code that I don't understand and I am more than willing to admit when there is something I'm not grokking... but I **am** however fully capable of recognizing when two functions are in the same file, **especially** because I almost always checkout the branch and look at the actual code when dealing with a complicated diff or code that is more challenging. In this case it's also obvious that those two functions are in the same file because if they were not, then one of them would have to be called via an imported object (ie: `object.getItems()`) right? 

I think it would benefit you greatly in the future to give people the benefit of the doubt, and take a few seconds every now and then to consider if **maybe** the person you're about to respond to actually **does** understand the situation, and if **maybe** there is some other way that you could interpret what they have written in a way that doesn't presume them to be completely void of intelligence. 

I fully understood that those two functions were in the same file because I was looking at the actual code in my IDE. So what I was actually asking was **SHOULD** that last method go in the same file with the generic getItems() method... as in... does it belong there or should it go in its own file? In other words, I wanted to know if the plan was to have **ALL** of the resolvers in one file, or to separate specific entity types into their own files, ie: a generic parent `resolvers.js` plus `event-resolvers.js`, `datetime-resolvers.js`, `attendee-resolvers.js`, `registration-resolvers.js`, `transaction-resolvers.js`, etc, etc, etc.

The last sentence you wrote above is sufficient in answering my question and there is no need to respond to this.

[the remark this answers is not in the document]
nerrad: > I think it would benefit you greatly in the future to give people the benefit of the doubt, and take a few seconds every now and then to consider if maybe the person you're about to respond to actually does understand the situation, and if maybe there is some other way that you could interpret what they have written in a way that doesn't presume them to be completely void of intelligence.

And maybe it would benefit you greatly in the future if you don't immediately jump to the conclusion that someone is criticizing your intelligence.  My statement was intended clarify what thought you meant by "same file" since in one paragraph it seemed to refer to different things. I'm left asking "What "same file are you referring to'?",   "Is he asking if this should be in the same file as `selectors.js` (which also has a `getEvents` method already)?".   In the light of vagueness, I made a statement that was a segue into how the methods functioned which hopefully would have clarified why they were named as is.  I honestly did not know _what_ you were referring to or the context of your question so thought my reply might help expedite the conversation and clarify what's happening in these files.  It definitely was _not_ intended in the way you seem to have taken it by your reply.

I hope there could be a reconsideration of the default way my comments seem to be perceived.  My goal is to help advance the project not put others down.

Issue #337, opened 2018-04-24
Add ReservedInstanceInterface and tag classes that need to operate like a "singleton" then prevent Loader from creating more than one instance


tn3rb:
Fixes some issues with Caching and interface aliases with the Loader and EE_Registry.

for backstory plz see https://events.codebasehq.com/projects/event-espresso/tickets/11450

PR #339, opened 2018-04-24
PR for #337 

tn3rb:
# branch Fet/337/reserved instance interface


core/EE_Registry.core.php
```diff
@@ -164,53 +176,81 @@ class EE_Registry implements ResettableInterface
...
+        EE_Dependency_Map $dependency_map = null,
+        Mirror $mirror = null,
+        ClassInterfaceCache $class_cache = null,
+        ObjectIdentifier $object_identifier = null
+    ) {
         // check if class object is instantiated
-        if (! self::$_instance instanceof EE_Registry) {
-            self::$_instance = new self($dependency_map);
+        if (
+            ! self::$_instance instanceof EE_Registry
```
nerrad: PSR has the requirement that the first condition is on the same line as the bracket.

core/EE_Registry.core.php
```diff
@@ -164,53 +176,81 @@ class EE_Registry implements ResettableInterface
...
+     * @param Mirror              $mirror
+     * @param ClassInterfaceCache $class_cache
+     * @param ObjectIdentifier    $object_identifier
+     */
+    protected function __construct(
+        EE_Dependency_Map $dependency_map,
+        Mirror $mirror,
+        ClassInterfaceCache $class_cache,
+        ObjectIdentifier $object_identifier
+    ) {
```
nerrad: Is EE_Registry _only_ instantiated in core?  If so, this is fine as is.  But if not, I wonder if the new args should default to null and we use LoaderFactory to instantiate if they are null?  I prefer to _not_ have to do that if possible but just pointing it out in case.

core/EE_Registry.core.php
```diff
@@ -693,48 +755,52 @@ public function create(
...
-            ? 'addon'
-            : '';
+        $addon = $addon ? 'addon' : '';
         // $this->_cache_on is toggled during the recursive loading that can occur with dependency injection
         // $cache is controlled by individual calls to separate Registry loader methods like load_class()
         // $load_only is also controlled by individual calls to separate Registry loader methods like load_file()
         if ($this->_cache_on && $cache && ! $load_only) {
             // return object if it's already cached
-            $cached_class = $this->_get_cached_class($class_name, $addon);
+            $cached_class = $this->_get_cached_class($class_name, $addon, $arguments);
```
nerrad: ooo... so I'm guessing this was the source of some of the bugs we saw?

core/EE_Registry.core.php
```diff
@@ -1517,20 +1583,25 @@ public function reset_model($model_name)
...
+        $instance            = self::instance();
         $instance->_cache_on = true;
         // reset some "special" classes
         EEH_Activation::reset();
-        $hard = apply_filters( 'FHEE__EE_Registry__reset__hard', $hard);
-        $instance->CFG = EE_Config::reset($hard, $reinstantiate);
-        $instance->CART = null;
-        $instance->MRM = null;
-        $instance->AssetsRegistry = $instance->create('EventEspresso\core\services\assets\Registry');
+        $hard                     = apply_filters('FHEE__EE_Registry__reset__hard', $hard);
```
nerrad: I wonder if it'd be better to label this variable `$do_hard_reset` so that it's clearer what it represents?

core/services/loaders/LoaderFactory.php
```diff
@@ -85,27 +86,37 @@ class LoaderFactory
...
                 $core_loader,
                 new CachingLoader(
                     $core_loader,
-                    new LooseCollection('')
-                )
+                    new LooseCollection(''),
+                    $object_identifier
+                ),
+                $class_cache
             );
```
nerrad: Hmm, so it appears there's a change in logic here.  Prior to the changes if a generator wasn't provided and `$loader` was not an instance of `LoaderInterface` then the `$generator` would get set to `EE_Registry` and used for the new loader.  However, **now** it looks like _any_ of the _new_ incoming arguments to `getLoader` are null, there will be no loader instantiated (so if its already null, it will return null).  

It seems this has the potential to break things... or is that unlikely because core should always have instantiated a loader (even before any add-ons need it)?

core/EE_Dependency_Map.core.php
```diff
@@ -435,19 +440,13 @@ public function has_alias($class_name = '', $for_class = '')
...
     {
-        if (! $this->has_alias($class_name, $for_class)) {
-            return $class_name;
-        }
-        if ($for_class !== '' && isset($this->_aliases[ $for_class ][ $class_name ])) {
-            return $this->get_alias($this->_aliases[$for_class][$class_name], $for_class);
-        }
-        return $this->get_alias($this->_aliases[$class_name]);
+        return (string) $this->class_cache->getFqnForAlias($alias, $for_class);
     }
```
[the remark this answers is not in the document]
tn3rb: See previous comment about "good idea... will do". I'll do the same here.

core/EE_Registry.core.php
tn3rb: Really?
Where?
I can't find that anywhere at all. 
The closest rule I've found (which is what I am following) is:

```
Argument lists MAY be split across multiple lines, where each subsequent line
is indented once. When doing so, the first item in the list MUST be on the
next line, and there MUST be only one argument per line.
```

If you can find something more relevant than that, then I'm more than happy to change.

core/EE_Registry.core.php
tn3rb: only in core.
ideally, your DI container (which is kinda sorta what EE_Registry is behaving like) is constructed before anything else and is the source for all other objects.

core/EE_Registry.core.php
tn3rb: unfortunately I think there was more than one source. a bunch of little things that all added up to the error you saw

core/EE_Registry.core.php
nerrad: Ya that paragraph you pointed to relates specifically to arguments in methods (not control structures).  I guess there is not hard rule in the actual standards.  I just know my linter screams at me if I have a bracket followed by a new line in the control structure.  So I've just kind of got used to that style.

No worries, since there's no hard rule I'll just try to remember to ignore this going forward 👍

core/EE_Registry.core.php
tn3rb: huh... that must have gotten reformatted after I removed some debug code... didn't mean to adjust that line at all

core/services/loaders/LoaderFactory.php
tn3rb: ya this happens during the very first step we take during bootstrapping at `plugins_loaded` priority 0.
Please note that it's also extremely detrimental to have more than one Loader floating around since that could result in duplicate objects that should be shared instances,
so having the construction of this object a bit complex is not a problem if that hinders the ability to create a second Loader.

core/EE_Dependency_Map.core.php
```diff
@@ -388,44 +390,47 @@ public function class_loaders()
...
-    public function has_alias($class_name = '', $for_class = '')
+    public function has_alias($fqn = '', $for_class = '')
     {
-        return isset($this->_aliases[$for_class][$class_name])
-               || (
-                   isset($this->_aliases[$class_name])
-                   && ! is_array($this->_aliases[$class_name])
-               );
+        return $this->class_cache->isAlias($fqn, $for_class);
     }
```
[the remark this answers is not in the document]
tn3rb: good idea... will do

core/EE_Registry.core.php
tn3rb: which linter are you using?

tn3rb:
> I wonder if now's a good opportunity to soft deprecate any non-PSR style method names in EE_Registry and move the logic to new methods following PSR style while soft deprecating the old method names that just wrap the new ones for now?

I'd rather put time into replacing uses of EE_Registry with injected dependencies and new Factory classes.
Other good changes would be extracting any other publically used logic into its own classes (like storing all EE_Addons in a shared Collection instead of on an EE_Registry property) and then replace that everywhere so that we can get away from EE_Registry altogether.

> I may have missed it, but do you have test coverage for classes implementing the new ReservedInstanceInterface?

 added a data set to the dataProvider for ``\EventEspresso\tests\testcases\core\services\loaders\ObjectIdentifierTest::getIdentifierProvider()` that ensures classes implementing `ReservedInstanceInterface` don't have their arguments utilized when generating a caching identifier (meaning the raw FQCN will always be used).

tn3rb:
The changes in this branch affect some low-level logic pertaining to dependency injection and singleton resetting.

# may affect blog switching on EventSmart

### testing

- activate EE
- maybe perform a migration
- add/edit an event
- register for a paid event and make a payment
- **IF** you have access to a multisite install and have administrator privileges, try switching blogs and performing some admin duties then switching back
- [ ] confirm that all activities operate as expected
- feel free to experiment with anything else in this branch and attempt to find errors that do not occur elsewhere

sethshoultes:
Testing notes part two:

- activated Event Espresso 4.6
- added an event and registered
- activated this branch
- added an event
- edited an existing event
- registered for a paid event
- registered for a free event

No issues.

sethshoultes:
@tn3rb this is working as expected for me. However, I haven't tested on a multisite install. I think we should merge this, then test on demoee.org. Thoughts?

tn3rb:
ya I agree, thnx

tn3rb:
KK seth, this has been merged to master, which should have resulted in it being pushed to demoee.org
can you plz test some of the multisite admin features there, primarily seeing if switching blogs causes any noticable effects.

sethshoultes:
@tn3rb I tested on demoee.org  and didn't notice any issues. I was able to switch to several different sites, view their admin dashboard, and Event Espresso events.

tn3rb:
KK thnx @sethshoultes , I'm going to assign to @nerrad now in case he wants/needs to do anything with it for ES

PR #340, opened 2018-04-24
Gutenberg/event attendees block

tn3rb:
replaces #317 which got messed up

PR #354, opened 2018-04-25
Build GutenbergBlockManager and supporting classes for locating and loading Gutenberg blocks

tn3rb:
replaces #317 AND #340 which both got messed up for various reasons

Codebase ticket: https://events.codebasehq.com/projects/event-espresso/tickets/11400

already need something like this in place to start converting REM over to work with Gutenberg (otherwise the replacement code I write now for REM will just have to be converted AGAIN later on)

MVP will be really simple and straightforward as we don't yet know what functionality the various interfaces will require

The pseudo code would be something like:

EE_System (or similar class) loads GutenbergBlockManager
GutenbergBlockManager scans a core directory (maybe: core/services/gutenberg/blocks/) to return a list of folder names that correspond to EE block loaders and then wraps the resulting array in a filter so that addons can inject paths
GutenbergBlockManager then (or at a later hookpoint) loops through those folder paths and grabs files matching a particular naming schema (like: *GutenbergBlockLoader.php), instantiates each class, and adds each service to a collection which restricts what kind of objects can be added
GutenbergBlockManager then (or at a later hookpoint) loops through that collection, maybe calls some sort of init() method which would allow blocks the opportunity to set hooks for preventing other blocks from loading (ie: allows addons to override blocks), and then tells each loader to load its corresponding block.
each individual GutenbergBlockLoader class would determine whether it should proceed to load its corresponding block or just do nothing because it has been told not to, or because its corresponding block is not valid for the current request.


core/domain/DomainBase.php
```diff
@@ -128,35 +135,42 @@ public function versionValueObject()
...
+     * @return void
      */
-    public function distributionAssetsUrl()
+    private function setAssetNamespace()
     {
-        return $this->pluginUrl() . 'assets/dist/';
+        $asset_namespace  = explode('/', $this->plugin_basename);
+        $asset_namespace = reset($asset_namespace);
+        $asset_namespace = $asset_namespace === 'event-espresso-core'
+            ? Registry::ASSET_NAMESPACE_CORE
```
nerrad: Would this maybe be better as:

```php
$asset_namespace = $asset_namespace === Registry::ASSET_NAMESPACE_CORE
    ? Registry::ASSET_NAMESPACE_CORE
    : $asset_namespace
```

core/domain/entities/editor/Block.php
```diff
@@ -104,21 +84,33 @@ public function editorBlockType()
     /**
      * @return string
      */
-    public function namespacedEditorBlockType()
+    public function namespacedBlockType()
     {
-        return EditorBlock::NS . $this->editor_block_type;
+        return Block::NS . $this->editor_block_type;
```
nerrad: Should we be avoiding abbreviations like `NS`?

core/domain/entities/editor/Block.php
```diff
@@ -181,21 +173,19 @@ public function setDynamic($dynamic = true)
...
         $args          = array(
             'attributes'      => $this->attributes(),
-            'editor_script'   => "ee-{$context}-blocks",
-            'editor_style'    => "ee-{$context}-blocks",
-            'script'          => "ee-{$context}-blocks",
-            'style'           => "ee-{$context}-blocks",
+            'editor_script'   => $this->asset_register->scriptHandle(),
+            'editor_style'    => $this->asset_register->styleHandle(),
+            'script'          => $this->asset_register->scriptHandle(),
+            'style'           => $this->asset_register->styleHandle(),
```
nerrad: Hmm In not sure about this.   `editor_script` and `script` _could_ be two different things because `editor_script` is intended for scripts only needed in the editor context and `script` is intended for scripts only needed in the frontend when the block is parsed for display (and the same with styles).  Now its entirely feasible that in most cases these scripts and styles will be the same in both contexts however  we should still allow for separation where needed.

core/services/assets/AssetRegisterInterface.php
```diff
@@ -0,0 +1,43 @@
...
+
+    /**
+     * @return string
+     */
+    public function scriptHandle();
+
+    /**
+     * @return string
+     */
+    public function styleHandle();
```
nerrad: So based on my previous comment, I wonder if we should either have specific methods for `editorscript` and `editorstyle` to differentiate from just `script` and `style` or just use a flag in these methods to indicate context for retrieval?

core/services/editor/BlockAssetRegister.php
```diff
@@ -0,0 +1,219 @@
...
+     * @param array           $style_dependencies
+     * @param DomainInterface $domain
+     * @param Registry        $registry
+     * @throws InvalidDataTypeException
+     */
+    public function __construct(
+        $script_handle,
+        array $script_dependencies,
+        $style_handle,
+        array $style_dependencies,
```
nerrad: based on my previous comments about the difference between `editor_script` and `script` (and etc with styles).  I wonder if this interface needs tweaking.  Maybe a VO or something to allow for multiple script handles and their dependencies to be set on this class?

core/services/editor/BlockAssetRegister.php
```diff
@@ -0,0 +1,219 @@
...
+     * @param array $script_dependencies
+     */
+    private function setScriptDependencies(array $script_dependencies)
+    {
+        $this->script_dependencies = $script_dependencies + array(
+                'eejs-core',
+                'wp-blocks',    // Provides useful functions and components for extending the editor
+                'wp-i18n',      // Provides localization functions
+                'wp-element',   // Provides React.Component
+                'wp-components' // Provides many prebuilt components and controls
```
nerrad: hmm.  I don't know if we should automatically assume all these dependencies would be registered with every script.  Certainly in the editor context it will usually be true, but in the frontend context it may not be.

core/domain/DomainBase.php
tn3rb: no, we are actually changing 'event-espresso-core' into 'core' because

```php
const ASSET_NAMESPACE_CORE     = 'core';
```

because 'event-espresso-core' is coming from the plugin basename...
oooooo... i just realized this is not going to work because if someone renames their EE folder, 
then the first part of their plugin basename will NOT be 'event-espresso-core' !!!
dang it.. going to hit you up in chat to discuss this...

core/domain/entities/editor/Block.php
tn3rb: why? What else could it be misinterpreted as:

- Nova Scotia
- Natural Science
- Non Smoker
- Nice Shot
- Nederlandse Spoorwegen
- Naughty Simian

errr.. ok... I'll change it...

core/domain/entities/editor/Block.php
tn3rb: kk agreed... will add specific getters and setters

core/services/assets/AssetRegisterInterface.php
tn3rb: I think I'll go with the former, since it's just more explicit imho

core/domain/entities/editor/Block.php
sethshoultes: I like Non Smoker 😆

tn3rb:
ok this has been reset to a point prior to the messed up asset-manager branch getting merged in

tn3rb:
Needed to back port some changes after working on the Event Attendees block.
Considering the number of commits here and all of the back n forth in the code, you may want to just open up the code itself and look at that files directly.
There's not actually that many classes at play, but it is a little bit intricate

nerrad:
I think it will be helpful if I hold off reviewing this until #371 has been merged into master and then into the `Gutenber/master` branch (which this pull is based off of).  That way the only changes that will show in this pull request for the diff will be the changes _specific_ to this pull request and it will thus make it MUCH easier to review.  I realize that slows things down a bit, but this can't really be merged into Gutenberg/master anyways until #371 has been merged upstream because 371 is _based_ off of the *master* branch _not_ Gutenberg/master.   So I'll make this _dependent_ on #371 and give it a "on-hold" status for now.

core/domain/entities/editor/Block.php
```diff
@@ -20,78 +17,60 @@
...
     }
 
 
     /**
      * @return string
      */
-    public function editorBlockType()
+    public function blockType()
     {
         return $this->editor_block_type;
```
nerrad: should this property name change?

core/domain/entities/editor/Block.php
```diff
@@ -100,21 +79,32 @@ public function editorBlockType()
     /**
      * @return string
      */
-    public function namespacedEditorBlockType()
+    public function namespacedBlockType()
     {
-        return EditorBlock::NS . $this->editor_block_type;
+        return self::NAME_SPACE . $this->editor_block_type;
```
nerrad: should `editor_block_type` name change.

core/domain/entities/editor/Block.php
```diff
@@ -100,21 +79,32 @@ public function editorBlockType()
...
     }
 
 
     /**
      * @param string $editor_block_type
      */
-    protected function setEditorBlockType($editor_block_type)
+    protected function setBlockType($editor_block_type)
     {
         $this->editor_block_type = $editor_block_type;
```
nerrad: should this property name change (and the corresponding argument name as well)?

core/domain/entities/editor/BlockInterface.php
```diff
@@ -18,9 +19,11 @@
  * @author  Brent Christensen
  * @since   $VID:$
  */
-interface EditorBlockInterface
+interface BlockInterface
 {
 
+    const NAME_SPACE = 'event-espresso/';
```
nerrad: I'd prefer to see `eventespresso` used as the namespace.  I realize its two words but hyphens could be a problem in certain contexts in the future and we're already using `@eventespresso` for package names etc.   I definitely will _not_ push hard on this though so if you want to leave this we can do that.

core/domain/entities/editor/BlockInterface.php
```diff
@@ -70,18 +81,6 @@ public function appliesToPostType($post_type);
     public function getEditorContainer();
```
nerrad: Offhand I'm not sure what the purpose of this method is for (its not documented), but also other elements in this interface have been genericized, would that apply to this as well?

core/domain/entities/editor/blocks/CoreBlocksAssetManager.php
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * Class CoreBlocksAssetManager
+ * Manages Editor Block Assets for Event Espresso Core
+ *
+ * @package EventEspresso\core\domain\entities\editor\blocks
+ * @author  Brent Christensen
+ * @since   $VID:$
+ */
+class CoreBlocksAssetManager extends BlockAssetManager
```
nerrad: It may have been explained elsewhere already and if so my apologies, I've forgotten.  What exactly are `Core Blocks` again?

core/services/assets/BlockAssetManager.php
```diff
@@ -0,0 +1,316 @@
...
+    {
+        $assets = array(
+            $this->getEditorScript(),
+            $this->getEditorStyle(),
+            $this->getScript(),
+            $this->getStyle(),
+        );
+        foreach ($assets as $asset) {
+            if ($asset instanceof BrowserAsset && $asset->isRegistered()) {
+                $asset->enqueueAsset();
```
nerrad: This may be already addressed, but just want to make sure that if blocks have been registered for only specific cpts (i.e. posts or pages and NOT EE Cpts), are their assets still only going to be _enqueued_ for those pages?  This is a bit related to my question about the purpose for `core-blocks` other words, is `core-blocks` going to be for blocks that are available on any editor?

webpack.common.js
```diff
@@ -6,6 +6,7 @@ const autoprefixer = require( 'autoprefixer' );
 const externals = {
 	jquery: 'jQuery',
 	'@eventespresso/eejs': 'eejs',
+	'@eventespresso/i18n': 'eejs.i18n',
```
nerrad: there's a problem with this entry because there are now two entries for `@eventespresso/i18n`.  This might be a merge artifact but there should only be one.  If you remove this one, it will most closely match what I see in `Gutenberg/master`

nerrad:
for now we can ignore the js tests until I figure out what I'm going to do to workaround [this issue](https://github.com/WordPress/gutenberg/issues/6718)

core/domain/entities/editor/Block.php
tn3rb: ya good catch... I missed that. will change

core/domain/entities/editor/Block.php
tn3rb: yup

core/domain/entities/editor/Block.php
tn3rb: yup

core/domain/entities/editor/BlockInterface.php
tn3rb: no that's fine with me. will chagne

core/domain/entities/editor/BlockInterface.php
tn3rb: This was based on code that you had added in the initial Gutenberg integration add-on. 
It wasn't documented there and I couldn't quite grok what it did, but assumed it was necessary.
I can remove until we determine if it is even needed still.

core/domain/entities/editor/blocks/CoreBlocksAssetManager.php
tn3rb: Core Blocks are Blocks for Core of course ! :laughing:
But seriously this is the `BlockAssetManager` for the EE core JS,
as opposed to something like `RecurringEventsBlocksAssetManager`.
This could maybe be broken up into more granular managers like `AdminBlockAssetManager` for admin-only assets, and `PublicBlockAssetManager` for frontend-only assets, but I *think* everything in here is shared between the frontend and admin.

webpack.common.js
tn3rb: ya they are identical, will fix

core/services/assets/BlockAssetManager.php
tn3rb: `CoreBlocksAssetManager` is basically all of the asset configuration that used to be directly within `core\services\assets\Registry`
so it would have run regardless of the CPT type. 
The `Asset::isRegistered()` method is just to prevent assets from being registered multiple times and has no correlation to CPT type or whether a script gets added for a particular post type. Some of that logic may still need to be figured out.

nerrad: My comment was less related to the `$asset->isRegistered()` bit and more related to the `$asset->enqueueAsset()` bit here.   Registration can happen on any route, but enqueueing should only be done where the script is actually needed.  In the case of blocks this is important because although there is a server side php registration, if a script is enqueued, then the block is also registered.  Which comes back to the latter part of my question which is, "is `core-blocks` going to be for blocks that are available in any editor?

nerrad: The other part of this is that the WP Gutenberg block registration php side _should_ automatically take _care_ of _enqueueing_ assets for the block.  Since that's the case, why is this even needed here?

PR #355, opened 2018-04-26
Gdpr/privacy policy hook

mnelson4:
This adds our privacy policy (right now its just placeholder text) to WP's new privacy policy page (if the needed function in WP exists).
Please note I made sure this works with an add-on by creating a related pull request in the infusionsoft add-on


core/services/privacy_policy/Manager.php
```diff
@@ -0,0 +1,99 @@
...
+
+/**
+ * Class PrivacyPolicyManager
+ * Manages setting up the hooks to add the EE core and add-ons' privacy policies
+ *
+ * @package        Event Espresso
+ * @author         Mike Nelson
+ * @since          $VID:$
+ */
+class Manager
```
tn3rb: I know that `privacy_policy` is in the namespace, but can we plz name this `PrivacyPolicyManager` so that it is always explicitly clear, so we can avoid conversations like this:

> Mr A : If you look in the Manager class you'll see such n such...
> Mr B : Which Manager class?
> Mr A : `\core\services\privacy_policy\Manager`
> Mr B : ahhh kk... thnx... uhhh...  I don't see "such n such" anywhere !?!?
> Mr A : Oh ya... that wasn't supposed to represent any actual code or text and was only added as a literary device to move this hypothetical conversation along in order to get to its real purpose.
> Mr B : Hypothetical conversation? Is... is none of this real???
> Mr A : Ya I'm sorry to inform you Mr B, but you are nothing more than an imaginary construct!
> Mr B : <existential crisis>

core/services/privacy_policy/Manager.php
```diff
@@ -0,0 +1,99 @@
...
+        add_action( 'edit_form_after_title', array( $this, 'possiblyAddPrivacyPolicy' ), 9 );
+    }
+
+
+
+    /**
+     * For all the registered `PrivacyPolicyGeneratorInterface`s, add their privacy policy content
+     * @param WP_Post $post
+     */
+    public function possiblyAddPrivacyPolicy($post)
```
tn3rb: another nitpicky issue... can we plz just call this `addPrivacyPolicy()` ?
I understand that it is not guaranteed to always add the privacy policy content to the supplied post, but the existence of some conditionals to bypass that process should not warrant prepending the method name with "possibly". If that were the case, then a huge number of methods within our codebase, especially any with conditionals that result in an early exit, could be similarly prefixed with "possibly", but that would be silly to see everywhere wouldn't it?

core/domain/services/admin/privacy_policy/PrivacyPolicy.php
```diff
@@ -17,7 +17,7 @@
  * @author         Mike Nelson
  * @since          $VID:$
  */
-class PolicyGenerator implements PolicyGeneratorInterface
+class PrivacyPolicy implements PrivacyPolicyInterface
```
tn3rb: please be sure to update the phpdocs for this class as well which I think are currently left at:

```php
 * Class Core
 * Description
```

core/services/privacy_policy/PrivacyPolicyInterface.php
```diff
@@ -7,18 +7,19 @@
 
 
 /**
- * Class PolicyGeneratorInterface
+ * Class PrivacyPolicyInterface
  * Description
```
tn3rb: just remove this if you are not going to write something for a description

core/services/privacy_policy/Manager.php
mnelson4: agreed, let's not go down that road 🙃

core/domain/services/admin/privacy_policy/PrivacyPolicy.php
mnelson4: ya those aren't super descriptive, are they?

core/services/privacy_policy/PrivacyPolicyInterface.php
mnelson4: oups! I'll descrive it

core/services/privacy_policy/Manager.php
mnelson4: hmmm ya, I suppose there are a lot of functions that don't necessarily perform their described action based on some conditions. But I think those usually don't perform their action because of some error, whereas this method expects to USUALLY not perform this action because usually users aren't on this specific page.
I'd like to just go along with your suggestion, but it gets less clear if I, someday, decide to put all the policy-adding code into a new function (which ideally would also be named `addPrivacyPolicy`), and call it from this one (which you're suggesting should have that same name).
Would the prefix "maybe" work better? (I see it's been used a bit.) I'd be inclined to only add it when it's anticipated that the function won't perform its action in non-error conditions.

core/domain/services/admin/privacy_policy/PrivacyPolicy.php
```diff
@@ -0,0 +1,51 @@
...
+{
+
+    /**
+     * Returns the name of the plugin and will be shown in the privacy policy's postbox header
+     * @return string
+     */
+    public function getName()
+    {
+        return esc_html__('Event Espresso', 'event_espresso');
```
tn3rb: ooo actually... I missed this on my first review, but I don't think we localize 'Event Espresso' since it is a proper name. You could verify this by looking at other places it is used in the plugin and/or confirm with Seth/Garth

core/libraries/plugin_api/EE_Register_Privacy_Policy.lib.php
```diff
@@ -0,0 +1,75 @@
...
+    protected static $privacy_policies = array();
+
+
+
+    /*
+     * @param string $plugin_id
+     * @param array $FQNSs can be the fully qualified namespaces each containing only privacy policies,
+     *              OR fully qualified class names of privacy policies
+     */
+    public static function register($plugin_id = null, $FQCNs = null)
```
tn3rb: sry missed this too, but the parameter defaults should match their data type, so strings should default to `''`, arrays should default to `array()`, and `null` should be used as a default for optional objects.

As well, we should type hint for array as much as possible, so in other words, this method signature should change to:

```public static function register($plugin_id = '', array $FQCNs = array())```

one lasts thing... do we even want to have either of these as optional?

File: core/libraries/plugin_api/EE_Register_Privacy_Policy.lib.php [hunk 7]
```diff
@@ -0,0 +1,75 @@
...
+            array('EE_Register_Privacy_Policy', 'addPrivacyPolicies')
+        );
+    }
+
+
+
+    /**
+     * @param null $ID
+     */
+    public static function deregister($ID = null)
```
tn3rb: same idea here... if $ID is a string, then it should default to `''` or even not be optional at all, considering the purpose of this method

core/libraries/plugin_api/EE_Register_Privacy_Policy.lib.php
```diff
@@ -0,0 +1,75 @@
...
+
+
+
+    /**
+     * Adds our privacy policiy generators registered by add-ons
+     *
+     * @param string[] $privacy_policies
+     * @return string[]
+     */
+    public static function addPrivacyPolicies($privacy_policies)
```
tn3rb: and `$privacy_policies` should type hint for an array

core/services/privacy_policy/PrivacyPolicyManager.php
```diff
@@ -0,0 +1,101 @@
...
+ * @author         Mike Nelson
+ * @since          $VID:$
+ */
+class PrivacyPolicyManager
+{
+
+    public function __construct()
+    {
+        if (function_exists('wp_add_privacy_policy_content')) {
+            add_action('edit_form_after_title', array($this, 'maybeAddPrivacyPolicy'), 9);
```
tn3rb: need to fix callback method name

core/services/privacy_policy/PrivacyPolicyManager.php
```diff
@@ -0,0 +1,101 @@
...
+    {
+        $loader = new CollectionLoader(
+            new CollectionDetails(
+            // collection name
+                'privacy_policies',
+                // collection interface
+                'EventEspresso\core\services\privacy_policy\PrivacyPolicyInterface',
+                // FQCNs for classes to add (all classes within that namespace will be loaded)
+                apply_filters(
+                    'FHEE__EventEspresso_core_domain_services_admin_privacy_policy_Manager__privacy_policies',
```
tn3rb: should update filter named after classname change

core/domain/services/admin/privacy_policy/PrivacyPolicy.php
mnelson4: I JUST had this exact conversation with WordPress core: https://core.trac.wordpress.org/ticket/43620#comment:41. But someone pointed out that translations that use other languages usually prefer to not show romantic language characters (eg Hindi, Cantonese, etc). Also, I see "Event Espresso" in tons of translated strings. But I'll bring this up with Seth

core/libraries/plugin_api/EE_Register_Privacy_Policy.lib.php
mnelson4: kk, ya `$FQCNs` should type-hint for array. 
The interface dictates they need to be optional though

core/libraries/plugin_api/EE_Register_Privacy_Policy.lib.php
mnelson4: I blame it on the interface again

core/services/privacy_policy/PrivacyPolicyManager.php
mnelson4: ah good call

core/services/privacy_policy/PrivacyPolicyManager.php
mnelson4: good catch

core/domain/services/admin/privacy_policy/PrivacyPolicy.php
joshfeck: We do localize 'Event Espresso' because of Eventsmart.

core/libraries/plugin_api/EE_Register_Privacy_Policy.lib.php
mnelson4: oh, actually the interface dictates it can't be type-hinted for, actually. So I just changed it to default to an array

core/libraries/plugin_api/EE_Register_Privacy_Policy.lib.php
mnelson4: yes sir!

core/domain/services/admin/privacy/policy/PrivacyPolicy.php
```diff
@@ -0,0 +1,51 @@
+<?php
+
+namespace EventEspresso\core\domain\services\admin\privacy\policy;
+
+use EEH_Template;
+use EventEspresso\core\services\privacy\policy\PrivacyPolicyInterface;
+
+defined('EVENT_ESPRESSO_VERSION') || exit('No direct script access allowed');
```
tn3rb: you'll need to run CodeSniffer on this file because the above line (as well as some of the spaces below it) are no longer kosher and need to be removed.

core/libraries/plugin_api/EE_Register_Privacy_Policy.lib.php
```diff
@@ -0,0 +1,75 @@
+<?php
+defined('EVENT_ESPRESSO_VERSION') || exit('No direct script access allowed');
```
tn3rb: same here, we have to remove these from now on from files that contain classes.
They are ok to add to files that do not contain symbols (ie: files with code that executes upon loading)

core/libraries/plugin_api/EE_Register_Privacy_Policy.lib.php
```diff
@@ -0,0 +1,75 @@
...
+     * @param string $plugin_id
+     * @param array $FQNSs can be the fully qualified namespaces each containing only privacy policies,
+     *              OR fully qualified class names of privacy policies
+     */
+    public static function register($plugin_id = null, $FQCNs = array())
+    {
+        self::$privacy_policies[$plugin_id] = $FQCNs;
+        // add to list of modules to be registered
+        add_filter(
+            'FHEE__EventEspresso_core_domain_services_admin_privacy_policy_Manager__privacy_policies',
```
tn3rb: I think this hook name changed to:
`FHEE__EventEspresso_core_services_privacy_policy_PrivacyPolicyManager__privacy_policies`

core/services/privacy/policy/PrivacyPolicyManager.php
```diff
@@ -0,0 +1,101 @@
...
+use EventEspresso\core\exceptions\InvalidEntityException;
+use EventEspresso\core\exceptions\InvalidFilePathException;
+use EventEspresso\core\exceptions\InvalidIdentifierException;
+use EventEspresso\core\exceptions\InvalidInterfaceException;
+use EventEspresso\core\services\collections\CollectionDetails;
+use EventEspresso\core\services\collections\CollectionInterface;
+use EventEspresso\core\services\collections\CollectionLoader;
+use WP_Post;
+
+defined('EVENT_ESPRESSO_VERSION') || exit('No direct script access allowed');
```
tn3rb: needs to be removed

core/services/privacy/policy/PrivacyPolicyInterface.php
```diff
@@ -0,0 +1,42 @@
+<?php
+
+namespace EventEspresso\core\services\privacy\policy;
+
+defined('EVENT_ESPRESSO_VERSION') || exit('No direct script access allowed');
```
tn3rb: needs to be removed

core/domain/services/admin/privacy/policy/PrivacyPolicy.php
mnelson4: kk!

core/libraries/plugin_api/EE_Register_Privacy_Policy.lib.php
mnelson4: 👍

core/libraries/plugin_api/EE_Register_Privacy_Policy.lib.php
mnelson4: good catch

core/services/privacy/policy/PrivacyPolicyManager.php
mnelson4: 👍

core/services/privacy/policy/PrivacyPolicyInterface.php
mnelson4: 👍

mnelson4:
This will be tested in downstream branches

mnelson4:
a downstream branch was merged into master, so this doesn't need to be merged anymore

PR #359, opened 2018-04-26
Improve tests to allow passes when times are within one minute (fixes #341)

### LINKED ISSUE #341: Fix phpunit test that fails from being one second off.
Full fail message:

```
1) EEM_Base_Test::test_current_time_for_query__timestamp
Failed asserting that two DateTime objects are equal.
--- Expected
+++ Actual
@@ @@
-2018-04-24T20:17:42.000000+0000
+2018-04-24T20:17:43.000000+0000
/home/travis/build/eventespresso/event-espresso-core/tests/testcases/core/db_models/EEM_Base_Test.php:487
```

See https://travis-ci.org/eventespresso/event-espresso-core/jobs/370753306 for an example (unless the test got restarted).

This should be able to use `EE_UnitTestCase::assertDateWithinOneMinute` method for the test to avoid these kind of fails.

nerrad:
Precision is tricky in unit tests because things can affect the second precision when tests are run all in one go.  So changes for test artifacts to be off by one second are greater.  I think it’s sufficient to have a time tolerance of one minute difference within the unit tests for any comparisons between date and time strings.

### DIFF
--- a/tests/testcases/core/db_models/EEM_Base_Test.php
+++ b/tests/testcases/core/db_models/EEM_Base_Test.php
@@ -479,18 +479,17 @@ class EEM_Base_Test extends EE_UnitTestCase
         $datetime->setTimestamp(
             EEM_Datetime::instance()->current_time_for_query('DTT_EVT_start', true)
         );
-        $this->assertEquals(
-            Datetime::createFromFormat(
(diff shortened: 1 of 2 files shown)


nerrad:
Basically just need to make sure no unit tests fail and then can be merged to master.

PR #371, opened 2018-04-27
Add Asset Value Objects plus Collection and Manager classes and update Registry to utilize them

tn3rb:
Because of the new asset build system, it is becoming more important for assets to be loaded via the Registry in a consistent manner. 
The Registry should handle Asset registration and localization of variable data as well as localization, but should **NOT** handle the actual configuration of every script within Event Espresso. This would be a serious violation of the Single Responsibility Principle.

This branch adds a new set of Asset value objects plus a Collection and abstract Manager class which work together to make it (somewhat) simple for any class in EE core or an add-on, to add assets and have them loaded properly via the new build system. 

This is how it works:

- AssetManager classes inject a shared instance of the AssetCollection and can add assets (JS & CSS) to the Collection
- Manifest files can also be added but only one per Domain (if we need more than that, then let me know)
- The AssetCollection is also injected into the asset Registry class which has had all asset configuration logic extracted (it's now in an AssetManager)
- The Registry then simply has to retrieve all assets from the AssetCollection and register them with WordPress
- The Registry also handles any data localization and translation required by scripts when registering them

Plz note: the first three commits should not have been part of this branch and were later removed (the ones pertaining to Editor Blocks) so plz ignore them.


core/domain/services/assets/CoreAssetManager.php
```diff
@@ -0,0 +1,357 @@
...
+     * registers assets for cleaning your ears
+     *
+     * @param JavascriptAsset $script
+     */
+    public function loadQtipJs(JavascriptAsset $script)
+    {
+        // qtip is turned OFF by default, but prior to the wp_enqueue_scripts hook,
+        // can be turned back on again via: add_filter('FHEE_load_qtip', '__return_true' );
+        if (
+            $script->handle() === CoreAssetManager::JS_HANDLE_EE_WP_PLUGINS_PAGE
```
nerrad: This is a new condition that I don't recognize from before the changes.  Prior to your changes there was only the `apply_filters`.  The qtip js is something the _entire_ ee admin references, not just the page the script with the handle `self::JS_HANDLE_EE_WP_PLUGINS_PAGE` loads on.

This may have been something you did by mistake, but I wonder if in the future when you make logic changes like this in refactoring code if it would be helpful if you document for the reviewer why the change was made?

core/domain/services/assets/CoreAssetManager.php
```diff
@@ -0,0 +1,357 @@
...
+    private function loadCoreJs()
+    {
+        $this->addJavascript(
+            CoreAssetManager::JS_HANDLE_EE_MANIFEST,
+            $this->registry->getJsUrl($this->domain->assetNamespace(), 'manifest')
+        );
+
+        $this->addJavascript(
+            CoreAssetManager::JS_HANDLE_EE_JS_CORE,
+            $this->registry->getJsUrl($this->domain->assetNamespace(), 'eejs'),
```
nerrad: This comment applies to everywhere  you invoke `$this->registry->getJsUrl`:  the second argument is the script handle that the url is being retrieved for, so you can use the constants you created for that.  So for instance, in this specific case  you'd end up with something like (assuming you use `self` like I mentioned in another comment in my review):

```js
$this->addJavascript(
    self::JS_HANDLE_EE_JS_CORE,
    $this->registry->getJsUrl($this->domain->assetNamespace(), self::JS_HANDLE_EE_JS_CORE),
    array(self::JS_HANDLE_EE_MANIFEST)
)-> setRequiresTranslation()
    -> setHasLocalizedData()
```

core/domain/values/assets/BrowserAsset.php
```diff
@@ -0,0 +1,139 @@
...
+
+    /**
+     * @return string
+     * @throws InvalidDataTypeException
+     */
+    public function version()
+    {
+        // if version is NOT set and this asset was NOT built for distribution,
+        // then set the version equal to the EE core plugin version
+        if ($this->version === null && substr($this->source, -8) !== '.dist.js') {
```
nerrad: I wonder if `.dist.js` should be a constant.  It's doubtful we'll ever change this suffix for distributed assets but if we do it'd be nicer to just have to change the constant.

Also, this is only picking up `.dist.js` from the filename.  What about css assets?

core/domain/values/assets/StylesheetAsset.php
```diff
@@ -0,0 +1,66 @@
...
+     * CssFile constructor.
+     *
+     * @param                 $handle
+     * @param string          $source
+     * @param array           $dependencies
+     * @param                 $media
+     * @param DomainInterface $domain
+     * @throws InvalidDataTypeException
+     */
+    public function __construct($handle, $source, array $dependencies, $media, DomainInterface $domain)
```
nerrad: should there be a default for the $media argument?

core/services/assets/Registry.php
```diff
@@ -381,7 +379,25 @@ public function getJsUrl($namespace, $chunk_name)
...
+     * @throws InvalidFilePathException
+     */
+    public function registerManifestFiles()
+    {
+        $manifest_files = $this->assets->getManifestFiles();
+        foreach ($manifest_files as $manifest_file) {
+            $this->registerManifestFile(
+                $manifest_file->assetNamespace(),
+                $manifest_file->urlBase(),
+                $manifest_file->filepath() . Registry::FILE_NAME_BUILD_MANIFEST
```
nerrad: I'm not sure this is correct here.  Is manifest name going to have the same name?

core/services/assets/Registry.php
```diff
@@ -135,62 +96,60 @@ public function getI18nRegistry()
...
+                $script->source(),
+                $script->dependencies(),
+                $script->version(),
+                $script->loadInFooter()
+            );
+            if ($script->requiresTranslation()) {
+                $this->registerTranslation($script->handle());
+            }
+            if ($script->hasLocalizedData()) {
+                $this->addRegisteredScriptHandlesWithData($script->handle());
```
nerrad: This is a change of logic from the previous behaviour.  The `addRegisteredScriptHandlesWithData` prior to your changes was always done AFTER the call to `wp_localize_script` or `wp_add_inline_script`.  Here you have it happening before right?  Have you tested to make sure this doesn't change behaviour (notably you'll see duplicates of the js variable printed to the html, or missing data in the final printed data object).

core/domain/services/assets/CoreAssetManager.php
```diff
@@ -0,0 +1,357 @@
...
+    // WordPress core / Third party JS asset handles
+    const JS_HANDLE_JQUERY                = 'jquery';
+
+    const JS_HANDLE_JQUERY_VALIDATE       = 'jquery-validate';
+
+    const JS_HANDLE_JQUERY_VALIDATE_EXTRA = 'jquery-validate-extra-methods';
+
+    const JS_HANDLE_UNDERSCORE            = 'underscore';
+
+    const JS_HANDLE_ACCOUNTING_CORE       = 'ee-accounting-core';
```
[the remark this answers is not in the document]
tn3rb: I prefixed constants with "EE_" if they were scripts that contained code written by EE. 
This particular handle is for the `accounting.js` core library, but I _think_ you had namespaced it to avoid conflicts with other plugins that may be loading a different version (or using a different set of base configurations). So I kept the EE namespacing for the actual handle but set the constant name to reflect that is is not actually Event Espresso code.

core/EE_Dependency_Map.core.php
```diff
@@ -744,6 +742,32 @@ protected function _register_core_dependencies()
...
+                'EventEspresso\core\services\assets\BlockAssetManagerCollection' => EE_Dependency_Map::load_from_cache,
+                'EventEspresso\core\domain\entities\editor\BlockCollection'      => EE_Dependency_Map::load_from_cache,
+                'EventEspresso\core\services\request\Request'                    => EE_Dependency_Map::load_from_cache,
+                'EventEspresso\core\domain\Domain'                               => EE_Dependency_Map::load_from_cache,
+                'EventEspresso\core\services\assets\Registry'                    => EE_Dependency_Map::load_from_cache
+            ),
+            'EventEspresso\core\domain\entities\editor\blocks\widgets\EventAttendees' => array(
+                'EventEspresso\core\domain\entities\editor\blocks\CoreBlocksAssetManager' => EE_Dependency_Map::load_from_cache
+            ),
+            'EventEspresso\core\domain\entities\editor\blocks\CoreBlocksAssetManager' => array(
```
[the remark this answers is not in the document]
tn3rb: ya missed that.
I was also working on the Block asset system that this was replacing (to make sure everything worked together), and had moved that code into a stash that I can apply to another branch, but obviously missed this.

core/domain/services/assets/CoreAssetManager.php
```diff
@@ -0,0 +1,357 @@
...
+     * @since $VID:$
+     * @throws DuplicateCollectionIdentifierException
+     * @throws InvalidArgumentException
+     * @throws InvalidDataTypeException
+     * @throws InvalidEntityException
+     */
+    private function loadCoreJs()
+    {
+        $this->addJavascript(
+            CoreAssetManager::JS_HANDLE_EE_MANIFEST,
```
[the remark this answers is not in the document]
tn3rb: I've always preferred using the class name for the following reasons:

 * it is explicit. You *always* know what class the constant (or property, or whatever) belongs to whereas with `self` you *need* to also know the class that the referenced code belongs to in order to know what `self` actually means. There have been times where people have posted code in chat (or forums, or whatever) that has included `self::something` but no mention of the class, which makes it impossible to know what that code is actually referring to. Your comment that "using self makes it clear the constant is from this class" has no value to me because why does it matter if the constant is from this class or another? How would anything change if the constant in the above code *was* from another class? IMO it doesn't change anything whatsoever. How does it change if we know that it *IS* from this class? Again, it doesn't. Knowing it is from *this* class has no benefit. Class constants work the same as static properties and are scoped globally, so it doesn't matter what class they come from as long as you know the actual class, but with `self` you can't tell that from looking at the constant alone.  I don't see how switching to using the explicit class name versus `self` would have any affect on how you would look at the code, because (using the code snippet above as an example) regardless of which is used, you would still need to look at the header for that code snippet (where it lists the file path `core/domain/services/assets/CoreAssetManager.php`) to see what class the code came from. In other words:
    - If the above code snippet had used `self::JS_HANDLE_EE_MANIFEST` then you would need to look at the snippet header to see the actual class that `self` referred to, because what is the point of knowing if a constant refers to *this* class if you don't know what *this* class is?
    - whereas when the above code snippet uses `CoreAssetManager::JS_HANDLE_EE_MANIFEST` then you only need to look at the snippet header to see that it belongs to *this* class. 
So either way you have to look at the snippet header to know where the code comes from. But again, knowing that a constant (or whatever) belongs to *this* class does not provide any extra benefit whatsoever, at least not in any way that I can see, but maybe you have some insight that I am completely overlooking.
- using the explicit class name instead of `self` makes it easier to copy n paste that constant (or whatever) into other classes that might need to reference it without having to do anything else but import the class. This will absolutely be the case if we switch over to using class constants for asset handles because those assets will be getting enqueued from some other class. So the other class will have no other option but to use  `CoreAssetManager::JS_HANDLE_EE_MANIFEST` so why not just use that everywhere? Although I don't think this is currently the case with any of our code, there is the potential to have a constant (or property, or whatever) with the same name in more than one class (imagine a static $ID property on multiple classes). In this (currently hypothetical) scenario, it should be obvious how using `self::$ID` could be disastrous or at least confusing, especially if you needed to reference that property from another class. Using the explicit class name avoids this issue completely.

core/domain/services/assets/CoreAssetManager.php
tn3rb: sure I can explain this change in a phpdoc comment.
This may not be necessary at all, but the reason for it was to ensure that the qTip assets loaded at the same time as they did in the old code.

Previously the code was like this:

```php
        $this->loadAccountingJs(); 
        $this->loadQtipJs(); 
        $this->registerAdminAssets(); 
```

but the qTip code could not be converted to use the new asset manager in this branch because that `loadQtipJs()` method was just doing this:

```php
    private function loadQtipJs() 
    { 
        // qtip is turned OFF by default, but prior to the wp_enqueue_scripts hook, 
        // can be turned back on again via: add_filter('FHEE_load_qtip', '__return_true' ); 
        if (apply_filters('FHEE_load_qtip', false)) { 
            EEH_Qtip_Loader::instance()->register_and_enqueue(); 
        } 
    } 
```

So without getting into completely converting the logic in `EEH_Qtip_Loader` (which is fairly complex) we needed to run that same logic before the logic that was previously in `registerAdminAssets()`. That logic was simply registering assets under the 'ee-wp-plugins-page' handle. I did this using the following:

```php
        $this->loadAccountingJs();
        add_action(
            'AHEE__EventEspresso_core_services_assets_Registry__registerScripts__before_script',
            array($this, 'loadQtipJs')
        );
        $this->registerAdminAssets();
```

which hooks the `loadQtipJs()` method into the block of code in `Registry::registerScriptsAndStyles()` (need to update the above hook name):

```php
            do_action(
                'AHEE__EventEspresso_core_services_assets_Registry__registerScriptsAndStyles__before_script',
                $script
            );
            wp_register_script(
                $script->handle(),
                $script->source(),
                $script->dependencies(),
                $script->version(),
                $script->loadInFooter()
            );
```

That hook passes the current script object along as a parameter, so if we wanted to do something immediately before registering the script using the 'ee-wp-plugins-page' handle, then we need to hook into 'AHEE__EventEspresso_core_services_assets_Registry__registerScripts__before_script' with a cllback that checks for that script handle:


```php
    public function callback(JavascriptAsset $script)
    {
        if (
            $script->handle() === CoreAssetManager::JS_HANDLE_EE_WP_PLUGINS_PAGE
        ) {
            // do something
        }
    }
```

IF it doesn't matter that the qTip assets are registered before the 'ee-wp-plugins-page' script, then I can maybe add another hookpoint to  `Registry::registerScriptsAndStyles()` after all of the other scripts run, and use that to trigger the `loadQtipJs()` method, but then that would mean the qTip stuff is happening after all other scripts load. The way I did it now ensures it happens at the exact same time it used to.

core/domain/services/assets/CoreAssetManager.php
tn3rb: > the second argument is the script handle that the url is being retrieved for

No the second argument is the "chunk" name. 
Here's the actual method signature:

```php
public function getJsUrl($namespace, $chunk_name)
```

Here is the original code that the above snippet replaces:

```php
wp_register_script( 
    'eejs-core', 
    $this->getJsUrl(self::ASSET_NAMESPACE, 'eejs'), 
    array('ee-manifest'), 
    null, 
    true 
); 
```

As you can see, the asset handle is  `'eejs-core'` but the second parameter passed to `getJsUrl()` is `'eejs'`.

That said, if you want to refactor how the manifest files are built so that the chunk name is the same as the asset handle, then that would certainly make things easier and less error prone since we would only need to reference that one identifier. 

I would enthusiastically support your changing the manifest files to work this way

core/domain/values/assets/BrowserAsset.php
tn3rb: > I wonder if .dist.js should be a constant.

ya I agree that is a good idea and will implement

>  this is only picking up .dist.js from the filename. What about css assets?

DOH. good catch. Will fix.

core/domain/values/assets/StylesheetAsset.php
tn3rb: a default is declared in `\EventEspresso\core\services\assets\AssetManager::addStylesheet()` which would be the preferred way to declare assets since that method also takes care of adding them to the `AssetCollection` class.
I hadn't previously realized that the PSR-2 guidelines state:

> Method arguments with default values MUST go at the end of the argument list.

which throws a bit of a wrench into how we are using our DI container and auto-injected dependencies, which I have been suggesting to always place as the last constructor arguments because otherwise you would have to manually supply them when overriding a default argument.

For example, if we wanted the `StylesheetAsset` class to declare a default for `$media` **AND** not violate PSR-2, then its signature would need to be:

```php
public function __construct($handle, $source, array $dependencies, DomainInterface $domain, $media='all')
```

which would then mean that any client code would then be responsible for providing an instance of `DomainInterface`. 

The following code:

```php
$default = $this->loader->getNew(
    'EventEspresso\core\domain\values\assets\StylesheetAsset',
    array(
        'default-css',
        EE_GLOBAL_ASSETS_URL . 'css/espresso_default.css',
        array('dashicons'),
        null,
        'screen'
    )
);
```

would result in the following:

> Fatal error: Uncaught TypeError: Argument 4 passed to EventEspresso\core\domain\values\assets\StylesheetAsset::__construct() must implement interface EventEspresso\core\domain\DomainInterface, null given

The reason for this is because the DI container sees that you have supplied a fourth argument in the array of values passed to the Loader, despite that value being `null`. It can not differentiate between a value of `null` representing the desire to use the default value for an argument, and an actual value of `null`.
Of course, this can probably be easily overcome by introducing a constant to indicate that the default value for an argument should be used instead of null, 
ie: something like:

```php
$default = $this->loader->getNew(
    'EventEspresso\core\domain\values\assets\StylesheetAsset',
    array(
        'default-css',
        EE_GLOBAL_ASSETS_URL . 'css/espresso_default.css',
        array('dashicons'),
        Loader::USE_DEFAULT_VALUE,
        'screen'
    )
);
```

which is actually a really good idea, and something I should probably implement regardless.

As well, if we are going to add a default value for `$media` for this class, then we should likely also add an empty array as the default value for `$dependencies`

core/services/assets/Registry.php
tn3rb: That's your call I guess. 
Do you want **every** manifest file to have the same name?
They will have different file paths of course, but using the same file name would certainly make it easy to identify them.

You tell me how you want this done.

core/services/assets/Registry.php
tn3rb: I can move this down into `enqueueData()` which also means I can remove the conditional because the call to `AssetCollection::getJavascriptAssetsWithData()` already calls `hasLocalizedData()` on each script when determining which ones to return.

I looked at how the existing code worked and didn't think it would make any difference where this was called as long as it was set somewhere.

However, I did indeed test things by visiting admin pages that utilized localized data (such as the Countries tab in the General Admin) as well as by performing some registrations on the frontend, which also utilize localized vars. I didn't notice any duplicate data, which in the past (iirc) has always resulted in the vars not working at all, but I didn't check all of the localized vars to make sure that one wasn't a duplicate. I can also do that after moving that line of code.

core/services/assets/Registry.php
nerrad: Well that's my point, the third argument is:

```php
$manifest_file->filepath() . Registry::FILE_NAME_BUILD_MANIFEST
```

Unless I'm misunderstanding something, the only unique thing for this argument as you loop through the `$manifest_files` collection is the `filepath()`.  So I was asking you if you were intending for every file manifest file to have the same name because you currently have this using `Registry::FILE_NAME_BUILD_MANIFEST` for every filename.  I think it'd be preferable to just have the third argument be whatever `filepath` returns?

tn3rb:
> I'm a bit concerned about what impact this may have on resource usage with all the extra objects etc being created for assets. It's something we may want to benchmark as a before/after picture just to ensure we're not seeing a significant impact from the changes.

I can do some benchmarking.
The benefit of these changes is it allows a consistent way for both EE core AND all of our add-ons to specify assets that can then be registered and managed by the Registry. It is not advisable (nor is it even doable) to put ALL of that logic into one class. So something like this needed to be done regardless.

> I'm also wondering if you tested this before turning over for review. It appears that a number of the things I'm pointing out (in particular the new Asset VO's) have problems that would surface in some testing. 

Ya of course I tested it. I visited as many admin pages I could think of that utilized javascript and especially those that use localized vars, so for example I tested:

 - the event editor and the tickets and datetimes section which heavily relies on JS
 - the registration form admin page, which relies on JS for adding options to select, checkbox and radio button question types
 - the Countries admin tab that is almost completely powered by JS and localized vars
 - I performed some registrations on the frontend which also utilize JS and localized vars

> - Test loading EE (with console showing): there's a console error related to the asset stuff (Failed to load resource: net::ERR_NAME_NOT_RESOLVED)

I'm not seeing **any** errors in the console so maybe something got messed up when you changed branches? Maybe you need to rebuild the assets?

> - Test the deactivation exit-modal (it's broken and not working) (that's currently the only system using the new data api stuff) - lots of console errors for that one.

I don't see anything happen when I deactivate EE, no modal whatsoever, so maybe something is broken there. I am however not seeing **ANY** console errors. Here's a screenshot showing that the `exitModalInfo` data is indeed part of the `eejsdata` var on the plugins page:
exitmodal screenshot_20180430112922

> - a console error appears to be on most EE routes that I loaded related to the stuff you did

Nope, not seeing any at all

> - None of the qtips appear to be working anywhere.

You mean like these: qtip screenshot_20180430113215

> - I wasn't able to test the new i18n for javascript stuff because of the broken exit modal but that would be where you'd verify that is working as well.

AHA... just realized that `warnings` was turned off in my console, so ya the Exit modal scripts were not loading, so that was not working, however everything else **was** working and (at least on my system) was not getting interrupted by those warnings.

Will fix the Exist Modal stuff

core/services/assets/Registry.php
tn3rb: > the only unique thing for this argument as you loop through the $manifest_files collection is the filepath()

that's right, and since it appears to me that we only require one manifest file per domain, then even though the filenames would be the same, we would end up with something like:

```
\event-espresso-core\assets\dist\build-manifest.json
\eea-automated-upcoming-event\assets\dist\build-manifest.json
\eea-wait-lists\assets\dist\build-manifest.json
```

which still points to uniques files.
 
So maybe a better question is: 

Is there any reason not to have **all** manifest files named the same like that?

ie: is there any benefit to having something like:

```
\event-espresso-core\assets\dist\build-manifest.json
\eea-automated-upcoming-event\assets\dist\auen.json
\eea-wait-lists\assets\dist\what-exactly-is-this-file.json
```

Is having random file names like that going to make it more difficult to work with them?
I'm guessing it won't make a difference since we just register the full file path and webpack takes care of the rest, but I just want to be absolutely certain.

I personally see no reason to not have all of our manifest files named the same because doing otherwise, means that we now have to have add-ons manually specify the file name they are using, which then also means more boilerplate in this new code I've been adding. Using the same filename also means that an example/empty manifest can be included in the EE add-on skeleton code for everyone to use as a base to start from.

core/domain/services/assets/CoreAssetManager.php
nerrad: > sure I can explain this change in a phpdoc comment.

Just to be clear,  I was not suggesting all changes need to be made in a phpdoc comment.  I was mainly thinking that you could go and do your own initial review of the code here in github and leave comments for the reviewer drawing attention to changes like this.   If that's not something you want to do that's okay.

core/services/assets/Registry.php
nerrad: Ya what you say makes sense.  It's likely we're just going to be copying the same config between core and add-ons so the filename can just stay the same for add-ons.

tn3rb:
### what changed

this branch extracts all asset configuration from `assets/Registry` into a set of new classes that can be used by any class in EE core or an add-on to add assets that can be included in the new asset build process. Or primary interest when testing is any Javascript that utilizes localized variables.

### testing

- edit an event in the admin and examine the Ticket and Datetimes section
- [ ] confirm that Ticket and Datetime advanced settings work as expected (the gear icon), as well as setting/editing datetimes, as well as linking Ticket with Datetimes
- mouse over the headings in the "Available Tickets" table
- [ ] also confirm that qTips work (text popups with descriptions in them)
- visit the Countries admin page in the General Settings
- [ ] confirm that everything works as expected
- visit the WP plugins.php page and deactivate Event Espresso
- [ ] confirm that the Exit Modal survey appears and functions as expected
- switch to the frontend, select two or more tickets for a paid event and register
- confirm that everything works as expected, including: 
    - success/error/validation notices 
    - copying attendee details between registrants
    - advancing to the next reg step
    - selecting a payment method and making a payment (like Stripe)
- feel free to experiment with functionality in this branch and attempt to find situations where issues with Javascript or CSS occur

tn3rb:
needed to make some changes for the Gutenberg block assets, can you plz review again?

tn3rb:
ok... reset this back to before the vendor folder was committed and then redid my changes.
In order to avoid any chance of further contamination, if you need/want to look at the code directly, can you **PLEASE** delete your local branch first and then checkout the one here? thnx

core/services/assets/Registry.php
```diff
@@ -135,61 +102,99 @@ public function getI18nRegistry()
...
+                $script
+            );
+            $registered = wp_register_script(
+                $script->handle(),
+                $script->source(),
+                $script->dependencies(),
+                $script->version(),
+                $script->loadInFooter()
+            );
+            if(WP_DEBUG && ! $registered) {
```
nerrad: I know in the past @joshfeck has expressed reluctance for us throwing exceptions even when `WP_DEBUG` is true.  I'm on the fence here but wonder if it should be run by him first?  I still would like to know what conditions would cause an asset to fail getting registered.  If its when there's already a script handle registered, then that is probably less likely to occur the way we're doing our registrations, which means this condition for throwing an exception is probably fine.

core/domain/values/assets/Asset.php
```diff
@@ -145,4 +150,20 @@ public function handle()
...
+        return $this->registered;
+    }
+
+    /**
+     * @param bool $registered
+     */
+    public function setRegistered($registered = true)
+    {
+        $this->registered = filter_var($registered, FILTER_VALIDATE_BOOLEAN);
+    }
```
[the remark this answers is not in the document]
tn3rb: to avoid running the same code repeatedly.

core/services/assets/AssetRegistrationException.php
```diff
@@ -0,0 +1,39 @@
...
+                    'The "%1$s" script could not be registered with WordPress core.',
+                    'The "script-handle" script could not be registered with WordPress core.',
+                    'event_espresso'
+                ),
+                $script_handle
+            );
+        }
+        parent::__construct($message, $code, $previous);
+    }
+}
```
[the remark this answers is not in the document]
tn3rb: there's a try catch in EE_System and this could be used in multiple places if we wanted

tn3rb:
sorry, but had some extra changes I had to make after working on the Event Attendees block some more, could you plz also have a look at the last few commits ?

core/services/assets/Registry.php
tn3rb: `$registered` might not be `true` if the asset is already registered or if some other issue occurs that prevents things from working (ex: there's a `doing_it_wrong()` notice that could be triggered but I can't remember if those bring the card house down or not) and you are correct that if "there's already a script handle registered, then that is probably less likely to occur the way we're doing our registrations".

core/services/assets/AssetManager.php
```diff
@@ -166,4 +166,36 @@ public function getStylesheetAssets()
...
+     * @return void
+     * @since $VID:$
+     */
+    public function enqueueJsAsset($handle)
+    {
+        if($this->assets->has($handle)){
+            /** @var JavascriptAsset $script */
+            $script = $this->assets->get($handle);
+            if ($script->isRegistered()){
+                wp_enqueue_script($handle);
```
nerrad: I'm puzzled, I thought the changes made in `JavaScriptAsset` and `StyleSheetAsset` meant you could just call something like the following:

```php
$script->enqeueueAsset()
```

tn3rb: ya I guess that would be better wouldn't it?
these two methods for enqueueing were added because I thought I was going to need to enqueue things from the manager class but that ended up not being the case.
I still think it will be good to be able to do so form the asset managers, but I like your idea better so will change it

core/services/assets/Registry.php
joshfeck: Could you check for `EE_DEBUG` instead?

core/services/assets/AssetManager.php
```diff
@@ -169,33 +163,19 @@ public function getStylesheetAssets()
...
      */
-    public function enqueueCssSAsset($handle)
+    public function enqueueAsset($handle)
     {
         if ($this->assets->has($handle)) {
-            /** @var StylesheetAsset $script */
-            $stylesheet = $this->assets->get($handle);
-            if ($stylesheet->isRegistered()) {
-                wp_enqueue_style($handle);
+            $script = $this->assets->get($handle);
```
nerrad: minor thing, but since this could be a script or a stylesheet I wonder if it'd be better to just name this variable `$asset`?

core/services/assets/Registry.php
tn3rb: sure, will do

core/services/assets/AssetManager.php
tn3rb: right... copy pasta... will fix

tn3rb:
### what changed

this branch extracts all asset configuration from `assets/Registry` into a set of new classes that can be used by any class in EE core or an add-on to add assets that can be included in the new asset build process. Or primary interest when testing is any Javascript that utilizes localized variables.

### testing

- edit an event in the admin and examine the Ticket and Datetimes section
- [ ] confirm that Ticket and Datetime advanced settings work as expected (the gear icon), as well as setting/editing datetimes, as well as linking Ticket with Datetimes
- mouse over the headings in the "Available Tickets" table
- [ ] also confirm that qTips work (text popups with descriptions in them)
- visit the Countries admin page in the General Settings
- [ ] confirm that everything works as expected
- visit the WP plugins.php page and deactivate Event Espresso
- [ ] confirm that the Exit Modal survey appears and functions as expected
- switch to the frontend, select two or more tickets for a paid event and register
- confirm that everything works as expected, including: 
    - success/error/validation notices 
    - copying attendee details between registrants
    - advancing to the next reg step
    - selecting a payment method and making a payment (like Stripe)
- feel free to experiment with functionality in this branch and attempt to find situations where issues with Javascript or CSS occur

sethshoultes:
Testing notes:

- Ticket and Datetime advanced settings work as expected
- hovering my mouse over the available tickets section heading displays the qtips as expected
- adding taxes to a ticket works as expected
- adjusted the event template settings and verified that the display order tool is working as expected
- viewed the front-end to make sure the display order settings took effect
- adjusted the settings in the countries admin and verified that states/provinces loaded when when the settings were changed
- deactivated EE4 and verified the exit survey modal appeared
- submitted the exit survey modal (works as expected)
- visited an event on the front-end and verified the show details works
- tried submitting a registration without details but validation prevented that (validation is working)
- advancing to the next reg step works as expected
- selecting a payment option works as expected
- onsite payment (authorize.net) works as expected
- testing with multiple attendees and copying data between registrations works as expected
- advanced copy link works as expected
- payment methods page works as expected
- admin help tours work as expected
- messages shortcode selector works as expected
- activating/deactivating message templates works as expected
- messages > email settings tab works as expected

PR #374, opened 2018-04-30
Remove empty index.js and add .gitignore

nerrad:
This commit still preserves folder structure via the use of `.gitignore` files.

I'm guessing the empty `index.js` files were used as placeholders to preserve the file structure.  It's better to use a more neutral placeholder file for cases like this.  So I just used empty `.gitignore` files.

## How has this been tested

No testing needed for this.


nerrad:
@tn3rb said this would be okay to merge once I fixed the failing test in slack but didn't mark it approved so I'm just going to go ahead and merge anyways.

Issue #327, opened 2018-04-23
Erase EE Registration data upon request for GDPR


mnelson4:
## Issue Overview
When users request to have their data anonymized, EE data should be part of that data.

https://core.trac.wordpress.org/ticket/43637 and https://core.trac.wordpress.org/ticket/43602 add a system for anonymizing WP data for users and commenters, we should just need to hook in to anonymize EE registraiton data

See our internal discussion: http://p2.eventespresso.com/2018/04/19/gdpr-for-ee-plan-not-eventsmart-or-ee-com/

mnelson4:
Here is a code snippet that does this. Pagination didn't seem necessary because it's all done in a query or two.


```
function ee_erase_answers( $email_address, $page = 1 ) {

	$page = (int) $page;
	EM_Answer::instance()->update(
		array(
			'ANS_value' => ''
		),
		array(
			array(
				'Registration.Attendee.ATT_email' => $email_address,
				'Question.QST_type' => array('NOT_IN', EEM_Question::instance()->question_types_in_category('multi-answer-enum'))
			)
		)
	);
	EEM_Answer::instance()->update(
		array(
			'ANS_value' => array()
		),
		array(
			array(
				'Registration.Attendee.ATT_email' => $email_address,
				'Question.QST_type' => array('IN', EEM_Question::instance()->question_types_in_category('multi-answer-enum'))
			)
		)
	);

	return array(
		'items_removed' => true,
		'items_retained' => false, // always false in this example
		'messages' => array(), // no messages in this example
		'done' => true,
	);
}

function register_my_plugin_eraser( $erasers ) {
	$erasers[] = array(
		'eraser_friendly_name' => __( 'Event Espresso Attendee Data' ),
		'callback'             => 'ee_erase_attendee',
	);
	$erasers[] = array(
		'eraser_friendly_name' => __( 'Event Espresso Registration Answers Data' ),
		'callback'             => 'ee_erase_answers',
	);
	return $erasers;
}

add_filter(
	'wp_privacy_personal_data_erasers',
	'register_my_plugin_eraser',
	10
);

```

mnelson4:
ok, so now when someone chooses to erase a user's personal data in WP, it also blanks out their attendee details and answers

PR #384, opened 2018-05-04
Gdpr/user data export

mnelson4:
For issue https://github.com/eventespresso/event-espresso-core/issues/326

Also adds the framework code for user data erasers, but doesn't implement any of them or hook them in (I was lazy and put that into this ticket becuase it seemed faster)

PR #385, opened 2018-05-04
Gdpr/erasure

mnelson4:
From ticket https://github.com/eventespresso/event-espresso-core/issues/327
Based on branch GDPR/user-data-export


core/domain/services/admin/privacy/erasure/EraseAnswers.php
```diff
@@ -0,0 +1,102 @@
+<?php
+
+namespace EventEspresso\core\domain\services\admin\privacy\erasure;
```
tn3rb: erasure?
like the 80s synth pop band?

core/domain/services/admin/privacy/erasure/EraseAnswers.php
```diff
@@ -0,0 +1,102 @@
...
+        $multi_value_questions_updated = $this->answer_model->update(
+            array(
+                'ANS_value' => array(),
+            ),
+            array(
+                array(
+                    'Registration.Attendee.ATT_email' => $email_address,
+                    'Question.QST_type'               => array(
+                        'IN',
+                        $this->question_model->question_types_in_category('multi-answer-enum'),
```
tn3rb: since this is called a couple of times in this method, and the method performs a bunch of array manipulations, can we plz capture the result of the first call and use it here instead of repeating that method call again?

core/domain/services/admin/privacy/export/ExportAttendeeBillingData.php
```diff
@@ -0,0 +1,133 @@
...
+        $export_items = array();
+        $found_something = false;
+        foreach ($attendees as $attendee) {
+            foreach ($payment_methods as $payment_method) {
+                try {
+                    $billing_info = $attendee->billing_info_for_payment_method($payment_method);
+                } catch (EE_Error $e) {
+                    $billing_info = null;
+                }
+                if (! $billing_info || ! $billing_info instanceof EE_Form_Section_Proper) {
```
tn3rb: only need the second part of this conditional

core/domain/services/admin/privacy/erasure/EraseAnswers.php
mnelson4: never heard of them before... listening... yes, this is named after that band, and their tambourine.

core/domain/services/admin/privacy/erasure/EraseAnswers.php
mnelson4: sure

core/domain/services/admin/privacy/export/ExportAttendeeBillingData.php
mnelson4: resolved that on https://github.com/eventespresso/event-espresso-core/pull/384

mnelson4:
Note: the "WIP" in a commit is causing one of the checks to not complete. I jus shouldn't add those to the commits from now on.

PR #386, opened 2018-05-04
fix add-on Domain registration

tn3rb:
allow add-on to supply an already constructed Domain object, or use supplied Domain FQCN to construct one when registering.

Fatal Errors during add-on registrations caused by improper Domain object creation and revealed by failing unit tests

## How has this been tested
unit tests pass, add-on can be activated without errors


tn3rb:
### testing

 - activate a few add-ons such as Wait List, Stripe, etc
- [ ] confirm that no fatal errors occur and activation proceeds as expected

PR #393, opened 2018-05-07
Bug/fix error in console with jed initialization (fixes #372)

### LINKED ISSUE #372: Fix error in console with exit-modal regarding jed initialization.
## Issue Overview

When loading the WordPress plugins list table page, the following error appears in the console:

```
ee-wp-plugins-page.2260f8293749cc2eb3fc.dist.js:1 Jed localization error: 

Error: Domain `event_espresso` was not found.
n @ ee-wp-plugins-page.2260f8293749cc2eb3fc.dist.js:1
(anonymous) @ ee-wp-plugins-page.2260f8293749cc2eb3fc.dist.js:1
n @ ee-wp-plugins-page.2260f8293749cc2eb3fc.dist.js:1
i @ ee-wp-plugins-page.2260f8293749cc2eb3fc.dist.js:1
(anonymous) @ ee-wp-plugins-page.2260f8293749cc2eb3fc.dist.js:1
IQQD @ ee-wp-plugins-page.2260f8293749cc2eb3fc.dist.js:1
n @ ee-manifest.f5984092809447a8fba8.dist.js:1
uhrh @ ee-wp-plugins-page.2260f8293749cc2eb3fc.dist.js:1
n @ ee-manifest.f5984092809447a8fba8.dist.js:1
1 @ ee-wp-plugins-page.2260f8293749cc2eb3fc.dist.js:1
n @ ee-manifest.f5984092809447a8fba8.dist.js:1
window.webpackJsonp @ ee-manifest.f5984092809447a8fba8.dist.js:1
(anonymous) @ ee-wp-plugins-page.2260f8293749cc2eb3fc.dist.js:1
```

This is related to the recent i18n work.  It's possible it is only an artifact of development builds and won't happen with production builds (because translation files will be present).  However at a minimum it should still be investigated.  It doesn't affect behaviour of the modal but it _would_ affect any eventual translations for the modal likely.

## Bug report or feature request?

* [x] Bug
* [ ] Feature
* [ ] Neither

## Environment Data:

Version of EE: master

Version of WordPress: 4.9.5

PHP Version:  7.1 

Browser used:  Chrome

nerrad:

There was an error that appeared in the console related to the loading of wp.i18n.  As mentioned in the original issue:

> So the issue here is that there is an assumption that the @wordpress/i18n package is exposed as a global everywhere, when that's not necessarily the case if gutenberg is not installed/active (and the wp global is not enqueued).

The solution is:

-  wrap the `@WordPress/i18n` package import so its exposed on the `eejs` global. 
- configure webpack so that `eejs.i18n` is aliased to `@eventespresso/i18n` (for package imports).
- modify our source so that instead of importing from `@wordpress/i18n` we import from `@eventespresso/i18n` everywhere we need i18n functions.
- tweak `EventEspresso/core/services/i18nRegistry::registerInlineScript` so its calling `setLocaleData` on the `eejs.i18n`global instance rather than (the likely non-existent) `wp.i18n` global instance.

## How has this been tested

* [x] Loaded plugins.php and made sure there were no longer any console errors/warnings.
* [x] Verify `eejs.i18n.__` works via the console (which means its working in loaded javascript as well).
* [x] Verify the current js is showing strings okay.

## Checklist

* [x] My code is tested.

## Additional

After this passes code review and is merged into master, some followup tasks are:

* [ ] Updating `Gutenberg/master` so that anywhere we're importing `@wordpress/i18n` is changed to `@eventespresso/i18n`
* [ ] Going through all add-ons that I recently added our webpack config to and updating them to use the same alias for `@eventespresso/i18n`


assets/src/exit-modal-survey/index.js
```diff
@@ -8,7 +8,7 @@ import ExitModal, { CLOSE_MODAL_EVENT } from '@eventespresso/react-exit-modal-ty
 import './style.css';
 import { data } from '@eventespresso/eejs';
 import { stringify } from 'querystringify';
-import { __ } from '@wordpress/i18n';
+import { __ } from '@eventespresso/i18n';
```
tn3rb: why would we do this if this functionality is now located at `eejs.i18n` ?

docs/AA--Javascript-in-EE/javascript-i18n.md
```diff
@@ -39,7 +39,7 @@ See: http://www.diveintojavascript.com/projects/javascript-sprintf
 Within the javascript files, make sure you import the wpi18n package via:
 
 ```js
-import { __ } from '@wordpress/i18n'
+import { __ } from '@eventespresso/i18n'
```
tn3rb: see question above re: `eejs.i18n`

assets/src/exit-modal-survey/index.js
nerrad: `@eventespresso/i18n` is an alias (via webpack) for `eejs.i18n`.  This gives us flexibility in case in the future we need to point `@eventespresso/i18n` to something else.  See [this change](https://github.com/eventespresso/event-espresso-core/pull/393/commits/d7ad7d98e23564604634cf156da34c36a951d026#diff-5c572bb8f24a00a84aa28f22d3a39ba5R9)

Issue #373, opened 2018-04-30
Create `ee-core` data store for top level EE data used in various contexts (using wp.data api)


nerrad:
## Issue Overview

This issue is broken out from #332 and relates to creating a "master" or "top-level" data-store of sorts that is subscribed to by any components utilizing ee data.  As mentioned in the pull request, this might involve:

- Keeping a top level store that keeps tracks of entities already retrieved and being worked with. For example `eventespresso/core` store might have a top level events entities state that serves as the "authoritative" state for those event entities.
- Exposing a way to record whether the state for an entity is dirty (e.g. `eventespresso/core.state.events.1.object` would have the event entity itself and `eventespresso/core.state.events.1.dirty` would be a boolean indicating whether its "dirty" (not in sync with the db = true) or not.
- By default, any reducers in any store dealing with ee entities would need to go through some sort of middleware or otherwise so that depending on the purpose of that reducer (for example a reducer that is not intended to mutate the state of entities in the `eventespresso/core` store), it will optionally mutate/not mutate entities if they exist (but still add new ones if they don't exist). So for example, in #332 I created an `eventespresso/lists` store that is mostly for just retrieving lists of model entities from our api and populating the `eventespresso/lists` store from what is retrieved (indexed by the query generating the list of entities). However, `eventespresso/lists` resolvers and reducers will by default not mutate the entities in the `eventespresso/core` store because they are just intended for querying for things like select drop-downs, checkboxes etc - things where if elsewhere the entity details are changed (i.e. Event Name) it may be desired to update that automatically within any objects referencing that store.
- We may still want a way to optionally flush the dirty state of entities within the `eventespresso/core` store (undo changes, or at least make it allowable to "restore" the state from some other source (db, local storage etc).

Once an appropriate approach is figured out, then the work that was done in #332 (likely merged) will need to be tweaked to integrate with the `eventespresso/core` store.

I've created this issue because at _some_ point I think we will need to implement something like this, but I don't want to implement it until we have something concrete to use it with.  For now the `event-espresso/lists` store will be sufficient for the immediate needs involving shortcodes conversion to blocks.   So for now this can linger in the backlog at the moment (while still being a part of the Gutenberg Epic).

## Bug report or feature request?

* [ ] Bug
* [x] Feature
* [ ] Neither

## Related Information:

- Gutenberg has some example middle-wares they use that can be examined.  I'm still not sure this is something that would involve middle-ware but I think I should at least investigate it.

PR #408, opened 2018-05-08
Add initial structure for ee-core data store

nerrad:

See #373 for details.  

This pull request will establish a `eventespresso/core` data store that will be used as the authoritative source for all ee entities worked with for a given view.  Other stores will interact with this for storage of entity data.

The pull accomplishes:

###  Structural changes

Data components folder layout has been restructured so they live within `assets/src/data/eventespresso`.  This follows the schema for store registration so we have `eventespresso/lists` and `eventespresso/core` (so far).

Also, `assets/src/data/model` is where all javascript related to model information and utilities will be found.

### State schema for `eventespresso/core` redux store.

`eventespresso/core` will serve as the "authority" for entities within the current view.  Any future work involving persisting data to the database will be done through this redux store.  To support this, the data model within the state follows a structure like:

```js
const state = {
  entities: {
    events: {
      10: { EVT_ID: 10, name: 'Event 10', /* other event entity data */ },
      20: { EVT_ID: 20, name: 'Event 20', /* other event entity data */ },
    },
    terms: {
      '10:20': { TERM_ID: 10, TAXONOMY_ID: 20, /* other term entity data */}
    }
  },
  entityIds: {
    events: [ '10', '20' ],
    terms: [ '10:20' ]
  },
  dirty: {
    events: [ 20 ],
    terms: []
  }
}
```

Notable elements:

* state is a collection of objects indexed by `entities`, `entityIds`, and `dirty`.
* `state.entities` is a collection of entities indexed by their modelName and then primary key.
* `state.entityIds` is collection of arrays indexed by modelName where each element is an array of entityIds.
* `state.dirty` is a collection of arrays indexed by modelName where each element is an array of entityIds for entities are "dirty".  An entityId here means that the entity is different from what was retrieved from the server (or is a newly created entity that doesn't exist on the server yet).
* In the case of models that have combined primary keys, the structure of the values for those primary keys are a concatenation the primary keys together with a `:`.

### Exposure of Primary Keys for all registered EE models via the eejs.data object.

`EED_Core_Rest_API` has a new method added to it: `getPrimaryKeyNamesIndexedByModelName` and this is used to expose an object containing all the primary keys for each model.  This is then exposed on the `eejs.data` global via `eejs.data.paths.primary_keys`.  This in turn is used by various new helpers (see `assets/src/data/model`).  

### Actions, Reducers and Selectors oh my!

Since `eventespresso/core` is a redux store (in the wp.data api pattern), there have been reducers, actions and selectors added (see `assets/data/eventespresso/core`).  Currently this allows for the following actions:

- Flushing the dirty state for entities in the store (either via a provided array of entity objects or an entityId).  This removes entity ids from the appropriate dirty record in the state.
- adding and/or updating new entity records in the core store.  For updates, this will automatically update the dirty record for any entity records added that differ from what is already in the state.

Along with the actions, selectors allow for:

- finding out what entities are dirty.
- retrieving all entities for a given model from the state.
- retrieving a specific entity for a given model from the state.

### You get tests, you get tests, and YOU get tests!

All of the new code has had unit tests added verifying expected behaviour (and is also a good way to learn what the expected behaviour is!).  I've also enhanced some of the existing tests for the `eventespresso/lists` store (which also involved fixing a couple of bugs in it).

### What about connecting `eventespresso/lists` and `eventespresso/core`?

Although I originally planned on doing that in this ticket and did some preliminary exploration of using middleware.  I decided its a bit premature to do that because the only time that will be needed is when we connect other stores to `eventespresso/core` so that any changes in `eventespresso/core` entities are reflected in any other stores (and thus the connected components to those stores).  For the most part, this won't happen until we start working on our editor integration with gutenberg so I'd like to be working that out _while_ we do the editor so we can make sure needs are met appropriately.

## How has this been tested

- Javascript unit tests
- implementation and use with EE Gutenberg blocks (will have to happen in later iterations)

## Checklist

* [x] My code is tested.


nerrad:
## Some planning

### Actions

Rough initial list of actions that will be needed

* [ ] RECEIVE_ENTITY_RECORDS:  used to update entities in the store
* [ ] SET_DIRTY : used to set if an entity is dirty or not.

### Reducers

* [ ] setEntityDirty(): manages the dirty state for an entity
* [ ] setEntitiesDirty(): manages the dirty state for a given group of entities.
* [ ] entityRecordsByKind(): manages entity records state for a given kind.  This ignores the dirty state of entities.

### Resolvers

I'm not sure we should have any resolvers for this store.  Since this is the top level store I think it should be populated by other stores such as the `@eventespresso/list` store.  It's possible at some point this may get hydrated on a page load (i.e. event editor populating the store) but that too could happen via middleware maybe.

### Selectors

* [ ] isEntityDirty() : whether a given entity is dirty or not.
* [ ] getEntity() : retrieve a requested entity from the store.
* [ ] getEntitiesByKind() :  retrieve all the entities currently in the store for the given Kind. 

### Middleware

> Note: I refer to middleware in here, but its possible I won't use full fledged middleware but instead just something similar to the `loadAndPersist` wrapper that Gutenberg uses for persisting state in `window.localStorage`.  Instead of persisting state, the wrapper would 'connect' with the provided store for enhanced reducers and actions.

Once I get the core data store setup, the next step will be to create middleware for connecting other stores with the core data. This will kind-of follow what Gutenberg did for its `persist` middleware (see [here](https://github.com/WordPress/gutenberg/blob/master/data/persist.js), [here](https://github.com/WordPress/gutenberg/blob/a81e0dc704fd922d778073f23fceee0bc0f81895/edit-post/store/middlewares.js) and [here](https://github.com/WordPress/gutenberg/blob/a81e0dc704fd922d778073f23fceee0bc0f81895/edit-post/store/index.js) ). Also, [this article](https://medium.com/@pragyan88/writing-middleware-composition-and-currying-elegance-in-javascript-8b15c98a541b) has some good pointers on creating middlewares in javascript. This middleware will likely involve:

#### Enhance reducer(s) and actions

This is a loose list of things the enhanced reducer/actions would be for.

* [ ] RECIEVE_NON_DIRTY_ENTITIES : This action will result in only non-dirty entities being updated in the core-data state.  This will enhance any existing entity related reducers in the connected store

#### Subscribes and Dispatches

* [ ] Subscribes the connected store to entity changes in core-data so anytime an entity changes, its related record anywhere in the connected store is updated via a dispatch to that connected store.

assets/src/data/index.js
```diff
@@ -0,0 +1,5 @@
+/**
+ * Register stores
+ */
+import './eventespresso/core';
+import './eventespresso/lists';
```
nerrad: This is necessary to ensure that the `default` export from each of these locations is executed by whatever script imports this file.  Currently nothing is doing so, I think I may introduce a new built file explicitly for eventespresso/data which will depend on the `eejs-core` script.  Then `eventespresso/data` (or whatever the script handle ends up being for this built file) will automatically be setup for any script that needs it.  I think this can be done in a separate pull request once we start implementing blocks.

assets/src/data/model/endpoints.js
```diff
@@ -1,32 +1,29 @@
...
-import { data, exception as GeneralException } from '@eventespresso/eejs';
-import { isUndefined } from 'lodash';
+import { data } from '@eventespresso/eejs';
 
 /**
- * WP Dependencies
+ * Internal imports
  */
-import { sprintf } from '@eventespresso/i18n';
+import { validateEntityHasKey } from './validators';
```
nerrad: This is another extraction since the logic wrapped by `validateEntityHasKey` was being used in multiple places.  Keeps things DRY.

assets/src/data/model/test/__mocks__/@eventespresso/eejs.js
```diff
@@ -0,0 +1,20 @@
...
+		primary_keys: {
+			events: 'EVT_ID',
+			tickets: 'TKT_ID',
+			venues: 'VNU_ID',
+			terms: [ 'TERM_ID', 'TAXONOMY_ID' ],
+		},
+	},
+};
+
+module.exports.Exception = Exception;
```
nerrad: This mock allows for some nice prepped data for various tests that will rely on the data passed from the server.  Note, this is simply passing through the `Exception` function so it can be relied on as is.

assets/src/eejs/index.js
```diff
@@ -21,9 +21,13 @@ export const i18n = wpI18n;
...
  * @param {string} msg
  */
-export const exception = function( msg ) {
+export const Exception = function( msg ) {
 	this.message = msg;
+	this.name = 'Exception';
+	this.stack = ( new Error() ).stack;
 };
+
+Exception.prototype = new Error;
```
nerrad: I tweaked the current generic EE exception error handler after doing some more reading up on it.  This makes it inherit more properties of the native `Error` class in javascript.  We can use this pattern for any custom Exceptions we want to create for client side error handling (such as comparables to what we have PHP side).  I think its valuable to keep all these exceptions exposed on the `eejs` global (so `eejs.Exception` for instance), so our add-ons can take advantage of them as well.  If start increasing the number of custom exception classes we can break them out into their own file and just export them from here.

package.json
```diff
@@ -37,46 +37,48 @@
...
 		"eslint": "^4.19.1",
 		"eslint-config-wordpress": "^2.0.0",
 		"eslint-plugin-jest": "^21.15.1",
 		"eslint-plugin-jsx-a11y": "^6.0.3",
-		"eslint-plugin-react": "^7.7.0",
+		"eslint-plugin-react": "^7.8.1",
 		"eslint-plugin-wordpress": "git://github.com/WordPress-Coding-Standards/eslint-plugin-wordpress.git#1774343f6226052a46b081e01db3fca8793cc9f1",
 		"extract-text-webpack-plugin": "^3.0.2",
-		"gutenberg": "Wordpress/gutenberg#master",
+		"gutenberg": "Wordpress/gutenberg#v2.8.0",
```
nerrad: This is a workaround for the [issue I posted here](https://github.com/WordPress/gutenberg/issues/6718).  By pinning gutenberg to a specific tag, this will prevent the errors from happening that killed our js test runs.  This will fix what's happening in the `Gutenberg/master` branch once this branch is merged back into it.

nerrad:
Ya I concur with your observations about the current state shape.  Based on that, and additional reading/consideration, what are your thoughts on this for the state shape (just using events as an example).

```js
const state = {
	events: {
		entitiesById: {
			10: { EVT_ID: 10, name: 'event 10' },
			20: { EVT_ID: 20, name: 'event 20' }
		},
		entities: [ 10, 20 ]
	},
	terms: {
		entitiesById: {
			'10:20': { TERM_ID: 10, TAXONOMY_ID: 20 }
		},
		entities: [ '10:20' ]
	},
	dirty: {
		events: [ 10 ],
		terms: [ '20:20' ]
	}
}
```

tn3rb:
Sorry but I'm totally not grokking what the benefit would be to having a schema like that?

If you used:

```js
const state = {
	events: {
		10: {
			EVT_ID: 10,
			name: 'Event 10',
			/* other event entity data */
		},
		20: {
			EVT_ID: 20,
			name: 'Event 20',
			/* other event entity data */
		}
	},
	terms: {
		'10.20': {
			TERM_ID: 10,
			TAXONOMY_ID: 20,
			/* other term entity data */
		}
	},
	dirty: {
		events: [10],
		terms: ['10.20'],
	},
};
```

and wanted to just get an array of Event IDs could you not just use `state.events.keys` ?

nerrad:
The example I gave was basically going by the example given in the redux doc you linked to (where they did something similar for the toDos).  However another reason why this _might_ work well is because to get the number of events in the state one could just go `state.entities.length`.  One _could_ go `state.entitiesById.keys.length` but that would create a new array every time its called.   I'm not sticky on this point and although it is in my next example I'm fine with just removing it.

So the next example I have considers preserving the shape consistently between elements in the state.  Since `dirty` is a top-level property and has the models as indexes beneath it, I'm wondering if the same shape should be followed for other top-level elements.  So we'd end up with something like this:

```js
const state = {
  entitiesById: {
    events: {
      10: {
        EVT_ID: 10,
        name: 'Event 10',
        /* other event entity data */
      },
      20: {
        EVT_ID: 20,
        name: 'Event 20',
        /* other event entity data */
      }
    },
    terms: {
      '10.20': {
        TERM_ID: 10,
        TAXONOMY_ID: 20,
        /* other term entity data */
      }
    },
  },
  entities: {
    events: [ 10, 20 ],
    terms: [ '10:20' ]
  },
  dirty: {
    events: [10],
    terms: ['10:20'],
  }
};
```

(entitiesById can just be dropped if you still don't see the need for it).

nerrad:
After discussing this with @tn3rb in slack, we've decided we're going to go with this shape:

```js
const state = {
  entities: {
    events: {
      10: {
        EVT_ID: 10,
        name: 'Event 10',
        /* other event entity data */
      },
      20: {
        EVT_ID: 20,
        name: 'Event 20',
        /* other event entity data */
      }
    },
    terms: {
      '10.20': {
        TERM_ID: 10,
        TAXONOMY_ID: 20,
        /* other term entity data */
      }
    },
  },
  entityIds: {
    events: [ 10, 20 ],
    terms: [ '10:20' ]
  },
  dirty: {
    events: [10],
    terms: ['10:20'],
  }
};
```

assets/src/data/eventespresso/core/selectors.js
```diff
@@ -0,0 +1,64 @@
...
+ * @param { Object } state
+ * @param { string } modelName
+ * @return {null|{}} If there are no records then null is returned.  Otherwise
+ * a collection object indexed by primary key values for the entity records is
+ * returned.
+ */
+export function getEntityRecordsForModel( state, modelName ) {
+	return state.entities[ modelName ] ?
+		state.entities[ modelName ] :
+		null;
```
tn3rb: what backwards style of formatting is this?

plz tell me this is a mistake and should be:

```
return state.entities[ modelName ]
	? state.entities[ modelName ]
	: null;
```

assets/src/data/eventespresso/core/test/reducers.js
```diff
@@ -0,0 +1,364 @@
...
+	it( 'returns original state when provided modelName is not in state',
+		() => {
+			const state = receiveEntityRecords(
+				originalState,
+				receiveRecordsAction(
+					'tickets',
+					[ { TKT_ID: 10, name: 'Ticket 10' } ],
+				),
+			);
+			expect( state ).toBe( originalState );
```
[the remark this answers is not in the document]
tn3rb: so explicit equality, meaning same instance AND exact same properties ?

assets/src/data/index.js
tn3rb: what we be the benefit of doing that over what is here now?

assets/src/data/model/primary-keys.js
```diff
@@ -0,0 +1,114 @@
...
+ * @type { memoized }
+ * @return { string } The string representation for the values.
+ * @throws { Exception }
+ */
+export const valuesForCombinedPrimaryKeys = memoize( ( keys, entity ) => {
+	if ( ! isArray( keys ) ) {
+		throw new Exception(
+			__( 'The provided value is not an array.', 'event_espresso' ),
+		);
+	}
```
tn3rb: `! isArray()` is used in at least a couple of places so I wonder if it would be a good candidate for another function in `validators.js`?

assets/src/eejs/index.js
tn3rb: agreed. sounds great

assets/src/data/eventespresso/core/selectors.js
nerrad: It's not a mistake.  That's the standard for javascript formatting with our current eslint rules.  I also think there's a technical reason _why_ it is this way (see [this stack overflow comment](https://stackoverflow.com/a/39068992/439720).

A quick perusal of various projects:

- jQuery style guide has it this way (http://contribute.jquery.org/style-guide/js/#multi-line-statements)
- airBnb has the way you prefer: https://github.com/airbnb/javascript#comparison--nested-ternaries (currently, there's an issue discussing this: https://github.com/airbnb/javascript/issues/1281
- while [Google code style](https://google.github.io/styleguide/jsguide.html#formatting-where-to-break) isn't explicit about ternary wrapping. Their standard does seem to indicate it would be preferable to have it as I have it since that would be breaking at the "higher syntatical level"
- npm has it following your method: https://www.npmjs.com/package/javascript-style-guide
- [Javascript Standard Style](https://standardjs.com/rules.html) has it using your method.

So there seems to be a mix of different methods being implemented everywhere.  I'll let you make the call.  I prefer the current way in javascript mostly because it is less maintenance overhead to maintain different rules from what we've adopted.  However, I'll be happy to find the rule to modify and update our linters if you'd prefer to adjust so `?` and `:` start the new line.

assets/src/data/index.js
nerrad: I'm sorry, even with your edit I'm not sure I understand exactly what you are asking.  Specifically I don't know what "what is here now?" is referring to.  Is it referring to the lack of this file in the Gutenberg/master branch?

Right now, the stores are not actually registered anywhere (although there is the code in place _to register_, that code is not actually _executed_ anywhere).  In order to regsiter the stores, `core/index.js` and `lists/index.js` would need to be invoked.  This file is prep work for that, all I'd do is add a configuration to the webpack file to build js file with the store registration in them, and expose it (likely on `eejs.stores`) and then ensure that file is loaded via a dependency on any built javascript requiring them.

Does that help or was there something else you were needing to know?

assets/src/data/model/primary-keys.js
nerrad: hmm good point. Will do.

assets/src/data/eventespresso/core/test/reducers.js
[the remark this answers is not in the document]
nerrad: yes.

assets/src/data/eventespresso/core/selectors.js
nerrad: I should add, if you want to do a switch in linting rules for this, then it should probably be done in a new branch off of master.  Then any javascript breaking the new rules will just have to be updated as we go.

tn3rb: I've stated that as far as I'm concerned, you're taking the lead on this JS stuff as you have the most experience with it, so I'll leave the final decision in your hands. 

If I were to make that decision, in situations like this where there is not a unanimous consensus amongst the community, I would consider all of the ways to choose an option that could best serves our needs. This could include:

- going with the option that has the greatest community support (like what we did with PSR) which also involves having to weigh the importance of each community (ie: Google vs jQuery vs AirBnB vs npm, etc) as well as speculate as to where those communities may stand in coming years (ie: is jQuery use waning?) or where those communities themselves may be headed (ie: are they currently entrenched in old school practices but leaning towards a renaissance?) 
- ignoring the greater community in favour of an internal vote for preference amongst our devs, which is easy, but not guaranteed to select the best option (although one can suppose that there are no guarantees for anything)
- making an executive decision by yourself and choosing an option that may not have the most community support, or be the most favoured option internally, but may offer significant benefits over others and/or appears to be an option that is gaining support in the community and could in fact become the defacto standard down the road, so best to adopt now. Choice is totally on your shoulders, which could make you look great if your speculations turn out to be correct, or make you look like a fool if things go some other direction, so good luck and hope for the best.

anyways, ya, your decision.

assets/src/data/model/validators.js
```diff
@@ -0,0 +1,66 @@
...
+				'event_espresso',
+			),
+			entity,
+			key,
+		) :
+		customMessage;
+	if ( ! entity.hasOwnProperty( key ) ) {
+		throw new Exception( message );
+	}
+};
```
tn3rb: except for the extra comma after translation namespace:

```
+				'event_espresso',
```

this all looks fine, but in general, I would suggest not swapping variables around so much as it just uses more more memory, both for the actual computers as well as for the devs that have to keep track of everything while reading code.

So instead of having a `customMessage` variable AND a `message` variable, why not just use `message`, and if an  incoming message is not supplied, then use a default value:

```
export const validateEntityHasKey = ( key, entity, message = '' ) => {
	if ( message === '' ) {
		message = sprintf(
			__(
				'The provided entity (%s) does not have the given property (%s)',
				'event_espresso'
			),
			entity,
			key,
		)
	}
	if ( !entity.hasOwnProperty( key ) ) {
		throw new Exception( message );
	}
};
```

bottom line:  there is no reason to introduce another variable when there is no need to preserve the original value, especially in a simple function like this.

nerrad: > except for the extra comma after translation namespace:

That's to satisfy the eslint rules.

I'm not too picky about the variables because the files are built and that gets taken care of in the built results anyways.  However, to keep things concise I will make the requested change.

assets/src/data/eventespresso/core/selectors.js
nerrad: Gonna ponder.  I don't want to invest time (yet) in possible eslint rule changes (might be easy to do but don't know yet).  I'm leaning towards consistency with our php psr rules just because of muscle memory, however that will require a rule modification.

PR #411, opened 2018-05-08
Bug/chicken is a required question

### LINKED ISSUE #407: Fix form validation messages aka the "Chicken is a required question" issue
From Codebase ticket: https://events.codebasehq.com/projects/event-espresso/tickets/11245

Apparently the registration forms default validation message, the one that displays in the pop up message, uses the first answer option as the question name. It's important that this gets fixed because if you set up a required question with radio buttons like below:
Choose your meal preference:*
Chicken
Beef
Fish
then if the registrant does not complete the field, the pop up error message says ‘Chicken is a required question’. 😂

https://slack-files.com/T02SY781D-F8EJQ8Y0N-c1e918bf7c

tn3rb:
Fixes required multi-choice questions using option names in default validation errors instead of the question name, resulting in notices like "Chicken is a required question"

would be overkill to create an add-on just for fixing an issue in core ;P

## How has this been tested
created a required multi-choice question and left the "Required Text" field empty to trigger the use of the default validation notice, then registered for an event using that question, but did not supply a value.

## Checklist

* [x] My code is tested.

fixes #407


core/libraries/form_sections/strategies/display/EE_Checkbox_Display_Strategy.strategy.php
```diff
@@ -19,16 +19,18 @@ class EE_Checkbox_Display_Strategy extends EE_Compound_Input_Display_Strategy
...
         $label_size_class = $input->get_label_size_class();
-        $html             = '';
+        $html = '';
         if (! is_array($input->raw_value()) && $input->raw_value() !== null) {
             EE_Error::doing_it_wrong(
                 'EE_Checkbox_Display_Strategy::display()',
                 sprintf(
-                    esc_html__(
+                    esc_html_x(
                         'Input values for checkboxes should be an array of values, but the value for input "%1$s" is "%2$s". Please verify that the input name is exactly "%3$s"',
```
mnelson4: Changing this to a multi-line string breaks the translations, right? So I think its better to leave it as a long string, right?

nerrad: The second argument isn't actually the translation, it's a comment for the translation - so it won't break translations.  However I do think that it'd be good to concatenate the second argument rather than adding line breaks and all the spacing (it'll might look ugly and really hard to follow in whatever client is used for the translation).

mnelson4: oh right, ya that's the explanation string. Sure good idea.

nerrad: > What's the benefit of concatenating the second argument?

You mean other than what I already wrote, which was:

> it might look ugly and really hard to follow in whatever client is used for the translation.

In case that isn't clear, what I mean by that, is the formatting of the translation explanation you did might result in odd line-breaks and spacing in the end users client.  I don't know for sure, I just thought that might be the case.  I suppose, however, argument could be made that the script generating the pot might not capture the argument correctly either because it doesn't actually parse php (and thus miss that its concatenated).

So, in the end, it might be better to leave translation explanations as long strings as well.

mnelson4: oh ok. I'm inclined to leave this as-is for now, then check out how this string appears on translate.eventespresso.com (the string should appear there as soon as its committed right?)

nerrad: No, it won't appear there as soon as its committed.  Someone on the team (I think usually @Pebblo ) triggers pot builds using grunt and then imports them into the glotpress instance.  _Then_, the strings show up there.

tn3rb: > what I mean by that, is the formatting of the translation explanation you did might result in odd line-breaks and spacing in the end users client. 

oooohhhhh  kk.. I thought you were saying that concatenation might look funny, ie: 

```
'Input values for checkboxes should be an array of values, '
 . 'but the value for input "form-input-id" is "form-input-value". '
 . 'Please verify that the input name is exactly "form_input_name[]"',
```

was uglier than 

```
'Input values for checkboxes should be an array of values, 
but the value for input "form-input-id" is "form-input-value". 
Please verify that the input name is exactly "form_input_name[]"',
```

but ya, you mean that what I did could look bonkers in a translation app. Makes sense, will change.

mnelson4: yeaaah just making the line really long is my preference, I'm not sure how translation apps will handle line breaks

sethshoultes:
Looks like this working as expected now: https://www.screencast.com/t/bRhmYVZZpPt

I also tested using checkboxes: https://www.screencast.com/t/8YDcdtIc

Issue #418, opened 2018-05-10
Fix Invalid field (count) passed to constructor of `EE_Term_Taxonomy`


joshfeck:
[TopOnline reports](https://eventespresso.com/topic/bug-two-small-bugs-fixes/):

in event-espresso-core-reg/core/db_models/EEM_CPT_Base.model.php on line 323, it reads:
`'count' => 1,`
This throws an `EE_Error` Exception:
```
Invalid field (count) passed to constructor of EE_Term_Taxonomy. Allowed fields are :term_taxonomy_id, term_id, taxonomy, description, parent, term_count
```
as the error indicates, it should be:
`'term_count' => 1,`
according to the model, in file event-espresso-core-reg/core/db_models/EEM_Term_Taxonomy.model.php
on line 66:
`'term_count' => new EE_Integer_Field(`

sethshoultes:
@mnelson4 want to fix this really quick?

mnelson4:
https://github.com/eventespresso/event-espresso-core/pull/419 implements josh's fix

sethshoultes:
@joshfeck can you please test?

joshfeck:
No, sorry @sethshoultes . Please see the discussion in the pull request ( #419 ) The fix cannot be tested because the method with the bug in it isn't used by EE core anywhere.

mnelson4:
ya, I should have noticed that. The associated pull request has been merged into master so this is resolved (it has unit tests to verify this works)

PR #419, opened 2018-05-10
Bug/invalid field count

mnelson4:
Addresses https://github.com/eventespresso/event-espresso-core/issues/418

### DIFF
--- a/tests/testcases/core/db_classes/EE_CPT_Base_Test.php
+++ b/tests/testcases/core/db_classes/EE_CPT_Base_Test.php
@@ -134,7 +134,7 @@ class EE_CPT_Base_Test extends EE_UnitTestCase{
             'Events look at important Rebel Space Ships'
         );
         $this->assertTrue($ee_term_taxonomy instanceof EE_Term_Taxonomy);
-        $this->assertNotEquals(0, $ee_term_taxonomy);
+        $this->assertNotEquals(0, $ee_term_taxonomy->get('term_count'));
     }
 }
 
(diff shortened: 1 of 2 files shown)


mnelson4:
## Testing
* [ ] verify this resolves https://github.com/eventespresso/event-espresso-core/issues/418

mnelson4:
> Hey Mike I'm going to need some steps to test this partly because a user reported the issue and I'm afraid I'm in the dark about what this is supposed to resolve.

Oh sorry, I thought you knew more about it. I looked into it some more and the method with the bug in it isn't used by EE core anywhere, so they must have a custom code that's using this. 

Seeing how the custom code isn't used anywhere, we could probably skip manual testing. Or I could maybe add a unit test... that might be good...

joshfeck:
A unit test might be good in case this happened to happen because of a merge conflict that went bad.

youri--:
As for the custom code... Issue #431 describes a custom code example (see "Reproduce" section) of how this bug came to light (I reported #418 on the EE forum). I.e. using PHP to programmatically add a event_espresso_category to an event.

mnelson4:
Ok I added a unit test that shows the issue was resolved. I think that will do in place of manual testing. @nerrad can you please re-review?

tests/testcases/core/db_classes/EE_CPT_Base_Test.php
```diff
@@ -118,6 +118,24 @@ public function testGetPrettyCurrentPostRemainingTheSame()
...
+        $e = $this->new_model_obj_with_dependencies('Event');
+        /**
+         * @var $e EE_Event
+         */
+        $ee_term_taxonomy = $e->add_event_category(
+            'Rebel Ships',
+            'Events look at important Rebel Space Ships'
+        );
+        $this->assertTrue($ee_term_taxonomy instanceof EE_Term_Taxonomy);
+        $this->assertNotEquals(0, $ee_term_taxonomy);
```
nerrad: Not sure both these assertions are necessary?  All that needs to be asserted is just the first right?

mnelson4: oh oups, the second one should be testing the count

Issue #417, opened 2018-05-10
Fix issue with constants related to `EE_Event_Category_Taxonomy`


joshfeck:
[TopOnline reports](https://eventespresso.com/topic/bug-two-small-bugs-fixes/):
in the file event-espresso-core-reg/core/db_models/EEM_CPT_Base.model.php
on line 2, it reads:
`define('EE_Event_Category_Taxonomy', 'espresso_event_category');`
and it should be, as used everywhere else in EE:
`define('EE_Event_Category_Taxonomy', 'espresso_event_categories');`
The above is problematic only in the same file, where it’s used in the add_event_category() and remove_event_category() methods.

mnelson4:
Pull request https://github.com/eventespresso/event-espresso-core/pull/420 impliments josh's fix

sethshoultes:
@joshfeck can you please test?

joshfeck:
No @sethshoultes , because this cannot be tested.

mnelson4:
My bad for missing this- the PR has unit tests that test it, I just didn't get around to merging the associated PR after it passed code review. But I just merged it now.

PR #420, opened 2018-05-10
its espresso_event_categories not espresso_event_category

mnelson4:
Addresses https://github.com/eventespresso/event-espresso-core/issues/417

### DIFF
--- a/tests/testcases/core/db_classes/EE_CPT_Base_Test.php
+++ b/tests/testcases/core/db_classes/EE_CPT_Base_Test.php
@@ -135,6 +135,7 @@ class EE_CPT_Base_Test extends EE_UnitTestCase{
         );
         $this->assertTrue($ee_term_taxonomy instanceof EE_Term_Taxonomy);
         $this->assertNotEquals(0, $ee_term_taxonomy);
+        $this->assertEquals('espresso_event_categories', $ee_term_taxonomy->taxonomy());
     }
 }
 
--- a/core/db_models/EEM_CPT_Base.model.php
+++ b/core/db_models/EEM_CPT_Base.model.php
@@ -16,7 +16,7 @@
 abstract class EEM_CPT_Base extends EEM_Soft_Delete_Base
 {
 
-    const EVENT_CATEGORY_TAXONOMY = 'espresso_event_category';
+    const EVENT_CATEGORY_TAXONOMY = 'espresso_event_categories';
 
     /**
      * @var string post_status_publish - the wp post status for published cpts


mnelson4:
## Testing
* [ ] verify this resolves https://github.com/eventespresso/event-espresso-core/issues/417, although I'm not sure where the affected code was actually in use in EE core (so this was probably an issue only for custom code that made use of these helper functions)

joshfeck:
If this was probably an issue only for custom code that made use of these helper functions, then there's not much that can be done with regards to testing, correct?

youri--:
As for the custom code... Issue #431 describes a custom code example of how this bug came to light (I reported it on the EE forum). I.e. using PHP to programmatically add a event_espresso_category to an event.

mnelson4:
Merged BUG/invalid-field-count into this because otherwise we couldn't use the method `EE_CPT_Base::add_event_category()`

mnelson4:
Passing for review again

tests/testcases/core/db_classes/EE_CPT_Base_Test.php
```diff
@@ -118,6 +118,25 @@ public function testGetPrettyCurrentPostRemainingTheSame()
...
+        /**
+         * @var $e EE_Event
+         */
+        $ee_term_taxonomy = $e->add_event_category(
+            'Rebel Ships',
+            'Events look at important Rebel Space Ships'
+        );
+        $this->assertTrue($ee_term_taxonomy instanceof EE_Term_Taxonomy);
+        $this->assertNotEquals(0, $ee_term_taxonomy);
+        $this->assertEquals('espresso_event_categories', $ee_term_taxonomy->taxonomy());
```
nerrad: as mentioned in my review in the other ticket, I don't think the second assertion is necessary.

mnelson4: ah good points, I meant to do the first one but got dizzied by the rest of this UI. The second one I hadn't thought of but makes a ton of sense

mnelson4:
this will be ready ot merge once the other passes review and I resolve the conflict

mnelson4:
huh, because this was merged into BUG/invalid-field-count AFTER it had already been merged into master, I don't think these changes are in master

mnelson4:
yep, this isn't yet in master. For example, `EEM_CPT_Base::EVENT_CATEGORY_TAXONOMY` is still `espresso_event_category` on master branch, but this branch changed that

PR #424, opened 2018-05-11
Add one more check to make sure espresso-notices loads where expected

joshfeck:
Sometimes espresso-notices end up at the very top of the document because some plugins and themes will fire loop_start very early in the request

screen shot 2018-05-11 at 2 28 25 pm

## How has this been tested

Checked to make sure notices display when and where expected.

## Checklist

* [x] My code is tested.


joshfeck:
Seth this could use a bit of testing before it gets merged, where we'd make sure the ee notices that display during checkout display as expected. Like when you try to go submit the ticket selector without selecting quantities, when there's a form validation error, or after a promotion code is input.

Issue #432, opened 2018-05-14
Assigning event category by slug, which also exists as regular post category, creates duplicates


youri--:
Using EE version 4.9.61.p.

**Description of bug**
When using custom code to assign an event category (using slug) to an event, EE looks up the term by slug/name to get an ID. It however fails to specify its own taxonomy in this query (ie. 'event_espresso_categories'), so if there is a regular post category with this same slug, it will fetch that. 

After that it will try to fetch the term_taxonomy record for the term with the previously retrieved ID and its own taxonomy ('espresso_event_categories'). Of course when the term initially gotten is a regular post category, there's no relationship with the espresso_event_categories taxonomy, so this returns `null`.

Subsequently, the code assumes the EE category therefore doesn't yet exist and creates a new category with the same slug.

The next time we try to add this same category to an event, the same thing will happen, and yet another category will be created.

**Expected behavior**
The lookup for an existing term should specify the taxonomy 'espresso_event_categories', so it stays within its own taxonomy.

**To Reproduce**
1- Use the WP admin to add a new, regular, post category with the slug "security".
2- Use the WP admin to add a new Event Espresso Category with the slug "security".
3- Use custom PHP code to get or create an EE_Event object and add an existing event category to it:

```
$new_event = EE_Event::new_instance( array(
	'EVT_name' =>  'Event Name',
	'EVT_desc' =>'Some text',
	'EVT_display_ticket_selector' => 0,
	'EVT_external_URL' => 'http://google.com/',
) );
$new_post_id = $new_event->save();

$new_event->add_event_category( 'security' );
```

This will call in turn call method `add_event_category()` in EEM_CPT_Base.model.php.

**Additional context**
in the file `event-espresso-core-reg/core/db_models/EEM_CPT_Base.model.php` starting on line 293:
```
        $term = EEM_Term::instance()->get_one(array(
            array(
                'OR' => array(
                    'name' => $category_name,
                    'slug' => $category_slug,
                ),
            ),
        ));
```
This doesn't specify the taxonomy, so it matches regular post categories as well. If that happens, and a term from the 'category' taxonomy (regular post category) is fetched, then when the term_taxonomy is looked up, it fails, because there's obviously no relationship between a regular category term and a espresso_event_categories term.
```
        $term_taxonomy = EEM_Term_Taxonomy::instance()->get_one(array(
            array(
                'term_id'  => $term->ID(),
                'taxonomy' => EE_Event_Category_Taxonomy,
            ),
[...]

PR #434, opened 2018-05-14
Bug/duplicate event categories

mnelson4:
Addresses https://github.com/eventespresso/event-espresso-core/issues/432 by adding a unit test, and verifies that EEM_CPT_Base::add_event_category creates a new term if none for the event category exists.


mnelson4:
merging this into its base branch. The affected code isn't actually used in EE core, so manual testing isn't possible, but that's why I added a unit test

mnelson4:
the base branch was previously merged into master so this is actually in master

PR #435, opened 2018-05-14
update labels

tn3rb:
makes dem labels look all purdy

### DIFF
--- a/.github/label-sync.yml
+++ b/.github/label-sync.yml
@@ -238,11 +238,6 @@
         - "type-invalid 🦄"
     - name: "type:impacts-eventsmart"
       color: 00f2ff
-    # other labels
-    - name: "Epic 🤘"
-      color: 5d498c
-      previousNames:
-        - Epic
   options:
     ignoreMilestones:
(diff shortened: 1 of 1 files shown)


.github/label-sync.yml
```diff
@@ -199,24 +213,36 @@
...
         - invalid
         - "type:invalid"
+        - "type-invalid 🦄"
     - name: "type:impacts-eventsmart"
       color: 00f2ff
+    # other labels
+    - name: "Epic 🤘"
+      color: 5d498c
+      previousNames:
+        - Epic
```
nerrad: hmm I think Epic is a label zenhub adds.  We should make sure this doesn't break how Zenhub does epics before making this change?

tn3rb: kk I'll try it

tn3rb: changed it and it appears to still work... but we can wait and see....
my guess is that GitHub uses a hidden internal identifier for the labels and the actual label text doesn't matter,
which would allow for localization, etc

Issue #431, opened 2018-05-14
SQL bug: update_term_taxonomy_counts() clears all term_counts when passed a term_taxonomy_id


youri--:
Using EE version 4.9.61.p.

**Description of bug**
When using custom code to assign an event category to an event, all term counts are reset to 0, except for the term_taxonomy that is the category assigned, which gets the count of all term counts combined. See "Additional context" for details and proposed change.

**Expected behavior**
Term count should be updated for the $term_taxonomy_id passed.

**To Reproduce**
Use an EE_Event object and add an existing event category to it.

```
$new_event = EE_Event::new_instance( array(
	'EVT_name' =>  'Event Name',
	'EVT_desc' =>'Some text',
	'EVT_display_ticket_selector' => 0,
	'EVT_external_URL' => 'http://google.com/',
) );
$new_post_id = $new_event->save();

$new_event->add_event_category( 'my-existing-event-category-slug' );
```

This will call in turn call method `add_event_category()` in EEM_CPT_Base.model.php and from there eventually we'll end up in method `update_term_taxonomy_counts()` where the SQL is problematic when a term_taxonomy_id was passed.

**Stack**
```
EEM_Term_Relationship.model.php:174, EEM_Term_Relationship->insert()
EE_HABTM_Relation.php:169, EE_HABTM_Relation->add_relation_to()
EEM_Base.model.php:2562, EEM_Event->add_relationship_to()
EEM_Event.model.php:710, EEM_Event->add_relationship_to()
EEM_CPT_Base.model.php:331, EEM_Event->add_event_category()
EE_CPT_Base.class.php:125, EE_Event->add_event_category()
taf.php:1405, taf::post_form_shortcode()
shortcodes.php:319, do_shortcode_tag()
shortcodes.php:197, preg_replace_callback()
shortcodes.php:197, do_shortcode()
class-wp-hook.php:286, WP_Hook->apply_filters()
plugin.php:203, apply_filters()
post-template.php:240, the_content()
page.php:169, include()
template-loader.php:74, require_once()
wp-blog-header.php:19, require()
index.php:17, {main}()
```

**Additional context**
in the file `event-espresso-core-reg/core/db_models/EEM_Term_Relationship.model.php` starting on line 129, we find the following method:
```
    /**
     * Makes sure all term-taxonomy counts are correct
     *
     * @param int   $term_taxonomy_id the id of the term taxonomy to update. If NULL, updates ALL
     * @global wpdb $wpdb
[...]

PR #436, opened 2018-05-15
Bug/update term taxonomy counts clears all term counts

mnelson4:
Addresses https://github.com/eventespresso/event-espresso-core/issues/431

### DIFF
--- a/tests/testcases/core/db_models/EEM_Term_Relationship_Test.php
+++ b/tests/testcases/core/db_models/EEM_Term_Relationship_Test.php
@@ -60,6 +60,74 @@ class EEM_Term_Relationship_Test extends EE_UnitTestCase{
 //		$this->assertEquals( 0, $tt_1->count() );
 //		$this->assertEquals(0,$tt_2->count() );
 	}
+
+
+    /**
+     * @group current
+     */
+	public function testUpdateTermTaxonomyCountsPassNull()
+    {
+        //create a few term taxonomies
+        $term1 = $this->factory()->term->create_and_get();
+        $term2 = $this->factory()->term->create_and_get();
+        $term3 = $this->factory()->term->create_and_get();
+        //add them to some posts
+        $post1 = $this->factory()->post->create_and_get();
+        $post2 = $this->factory()->post->create_and_get();
+        wp_set_post_terms($post1->ID, array($term1->term_id, $term2->term_id));
+        wp_set_post_terms($post2->ID,array($term2->term_id, $term3->term_id));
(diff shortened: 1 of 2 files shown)


tests/testcases/core/db_models/EEM_Term_Relationship_Test.php
```diff
@@ -60,6 +60,74 @@ public function test_insert_and_update_and_delete() {
...
+        //verify that when we update the counts, they aren't set to zero
+        EEM_Term_Relationship::instance()->update_term_taxonomy_counts($term2->term_id);
+        wp_cache_flush();
+        $term1 = get_term($term1->term_id);
+        $term2 = get_term($term2->term_id);
+        $term3 = get_term($term3->term_id);
+        $this->assertEquals(1, $term1->count);
+        $this->assertEquals(2, $term2->count);
+        $this->assertEquals(1, $term3->count);
+    }
```
nerrad: I don't get the purpose of having identical tests where the only difference is the name of the test.  Going by the name of the second test, I don't think its necessary to assert ints either because that's creating a test for WP behaviour right?  Either way, I don't think its necessary and maybe could just be removed?

mnelson4: the setup and assertsions for both tests are the same, but the first one does `EEM_Term_Relationship::instance()->update_term_taxonomy_counts();` and the second does `EEM_Term_Relationship::instance()->update_term_taxonomy_counts($term2->term_id);`.
Before this ticket's work, the first one passed but the second one failed; with this work they both pass (last I checked!)

mnelson4: does that explain why I did things this way? Do you still have recommendations?

mnelson4: I probably should add some more info to each test's PHPdocs to explain what they're doing and what's different

nerrad: hmm ya I missed that.  I know I usually follow the rule of three before making code more DRY (and recommend that) but in this case it might be better to extract the boilerplate setup stuff into its own method here.  I think that may make the actual assertion differences stand out better if the same boilerplate isn't in the same tests.

mnelson4: hmm ya good suggestion IMO

mnelson4:
putting into previous milestone because it was merged into master before that was released

PR #458, opened 2018-05-17
Fix i18n api loading strings twice per handle

nerrad:

While testing some gutenberg stuff, I noticed in the branch I was testing that there were duplicate calls to `eejs.i18n.setLocaleData` before a script.  In this pull request:

* I created a phpunit test that demonstrates the flaw.
* I refactored i18n to fix the flaw.

Essentially the core problem is that `i18nRegistry` hooks into `print_scripts_array` to work its magic.  However this hook can potentially get invoked multiple times in the request (at least twice, once for the header, and once for the footer).  If `I18nRegistry::registerScriptI18n` only gets called once in a request for a script handle, then there would be no issues.  But if there's a set of conditions where it gets called subsequently _after_ `print_scripts_array` has been invoked, then the existing logic attempting to prevent multiple calls to `wp_add_inline_script` for a script handle will fail.  So the fix in this pull addresses that.

## How has this been tested

* [ ]  Currently automated phpunit tests.
* [x] Verify that the systems using this mechanism (currently only the exit modal stuff) continue to work as expected and there are no js related errors in the browser console.

## Checklist

* [x] My code is tested.

### DIFF
--- a/core/services/assets/I18nRegistry.php
+++ b/core/services/assets/I18nRegistry.php
@@ -20,7 +20,7 @@ class I18nRegistry
     private $domain;
 
     /**
-     * Will hold all registered i18n scripts.
+     * Will hold all registered i18n scripts.  Prevents script handles from being registered more than once.
      *
      * @var array
      */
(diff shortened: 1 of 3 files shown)


nerrad:
So what I think is happening is we currently use `$this->registered_I18n` to track what things have been registered but there's no prevention of `re-registering` a handle after its already been queued up later in the request.  So I'll have to use another mechanism for tracking handles that have already been queued up in the request.

PR #459, opened 2018-05-17
update block asset registration

tn3rb:
the first, third, and fifth points in the bullet list here:  https://github.com/eventespresso/event-espresso-core/pull/331#issuecomment-389844213

### DIFF
--- a/core/services/assets/BlockAssetManager.php
+++ b/core/services/assets/BlockAssetManager.php
@@ -163,7 +163,7 @@ abstract class BlockAssetManager extends AssetManager implements BlockAssetManag
             ),
             $this->addDefaultBlockScriptDependencies($dependencies)
         )
-        ->setHasLocalizedData();
+        ->setRequiresTranslation();
     }
 
 
@@ -214,7 +214,7 @@ abstract class BlockAssetManager extends AssetManager implements BlockAssetManag
             ),
             $this->addDefaultBlockScriptDependencies($dependencies)
         )
-        ->setHasLocalizedData();
+        ->setRequiresTranslation();
     }
 
 
--- a/core/services/assets/BlockAssetManager.php
+++ b/core/services/assets/BlockAssetManager.php
@@ -2,6 +2,7 @@
 
 namespace EventEspresso\core\services\assets;
 
+use EventEspresso\core\domain\entities\editor\BlockInterface;
 use EventEspresso\core\domain\services\assets\CoreAssetManager;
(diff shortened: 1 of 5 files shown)


core/services/assets/BlockAssetManager.php
```diff
@@ -110,6 +120,9 @@ public function getStyleHandle()
      */
     public function setStyleHandle($style_handle)
     {
+        if (strpos($style_handle, BlockInterface::NAME_SPACE . '-') !== 0) {
+            $style_handle = BlockInterface::NAME_SPACE . '-' . $style_handle;
+        }
```
nerrad: Regarding using `BlockInterface::NAME_SPACE` as the prefix for all these methods, will EE add-ons be able to override this to add their own prefix?  So instead of `eventespresso-script-handle`, an addon might have `eea-automated-upcoming-events-script-handle`?

core/services/assets/BlockAssetManager.php
```diff
@@ -142,14 +155,16 @@ public function addEditorScript($handle, array $dependencies = array())
...
-        return parent::addJavascript(
+        $asset = parent::addJavascript(
             $handle,
             $this->registry->getJsUrl(
                 $this->domain->assetNamespace(),
                 $handle
             ),
             $this->addDefaultBlockScriptDependencies($dependencies)
         );
+        $asset->setHasLocalizedData();
```
nerrad: I assumed you'd just be able to chain this call (like what you did in `CoreAssetManager`) removing the need to set another variable (the methods are fluent right?)?  Also, are you calling the wrong method here?  Shouldn't it be `setRequiresTranslation`?

core/services/assets/BlockAssetManager.php
```diff
@@ -192,14 +207,16 @@ public function addScript($handle, array $dependencies = array())
...
+        return $asset;
```
nerrad: Same here as previous comment.

core/services/assets/BlockAssetManager.php
tn3rb: No the namespacing is just to ensure that our scripts don't conflict with another vendor. 
Add-ons will need to still supply a handle that is unique within the `eventespresso` namespace.
If an add-on wants to do that by prepending their own namespace, then that's fine with me, doesn't really matter.
So as long as the Automated Upcoming Events Notifications add-on (or any other) doesn't try to register a script with the handle `core-blocks` or any other handle that is already used, then everything will be fine.

PR #461, opened 2018-05-17
Always use the 'pretty' value for the TXN paid column (fixes #452)

### LINKED ISSUE #452: Fix transaction page paid column display back to 0.00
It seems that the paid column monetary value for nothing paid changed from `$0.00` to `0` for some reason:
 https://www.screencast.com/t/RDqeLTKF

Can we please change that back to `$0.00`?

Pebblo:
In version 4.9.45 the `$transaction_paid` var was cast to a string which meant the ternary was actually broken but worked in favour of displaying the 'pretty' price.

The condition was fixed in 4.9.46 which meant we actually got the 'correct' value.

This commit removes the condition to display the int and always uses the 'pretty' version of TXN paid.

This fixes #452


tn3rb:
Just a heads up that you'll need to merge master into this branch prior to merging.

Pebblo:
Sorry Brent, I missed this.

Free registrations assigned the transaction_paid var to an int, paid registrations used a float so this actually only affected free registrations.

Anyway, it just always uses the pretty version now and doesn't set the var to an int as there's no need to.

PR #465, opened 2018-05-18
Bug: Fix store and configuration

nerrad:

This pull contains a number of fixes related to testing in the event-attendees-block branch.

### DIFF
--- a/assets/src/data/eventespresso/lists/index.js
+++ b/assets/src/data/eventespresso/lists/index.js
@@ -12,7 +12,7 @@ import * as actions from './actions';
 import * as resolvers from './resolvers';
 
 export default registerStore( 'eventespresso/lists', {
-	reducers,
+	reducer: reducers,
 	actions,
 	selectors,
 	resolvers,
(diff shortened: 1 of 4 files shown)


assets/src/data/eventespresso/lists/index.js
```diff
@@ -12,7 +12,7 @@ import * as actions from './actions';
 import * as resolvers from './resolvers';
 
 export default registerStore( 'eventespresso/lists', {
-	reducers,
+	reducer: reducers,
 	actions,
 	selectors,
 	resolvers,
```
tn3rb: any idea if you need to do something similar for any of these other lines? ie:

```js
 	action: actions,
 	selector: selectors,
 	resolver: resolvers,
```

nerrad: no because the properties for those are actually `actions`, `selectors` and `resolvers` so this uses a feature of ES6 that automatically sets the properties to the same value.  I had to manually define the `reducer` property here because I imported the reducers and assigned them to a `reducers` named variable.  If I had of imported as `reducer` then I wouldn't have needed to do this (and I could still modify that but its fine this way too).

PR #466, opened 2018-05-20
Gutenberg/fix data stores and event select

nerrad:

While working on the `Gutenberg/event-attendees-block` branch for Brent (fixing bugs and working out issues), there were a number of things that were broken/needed enhanced for the new data stores and the `event-select` component.  All that work is contained in this pull request.  The `Gutenberg/event-attendees-block` will **require** this pull in order to work.  So it's beneficial to get it merged into `Gutenberg/master` asap and then it's available to pull into `Gutenberg/event-attendees-block`.

A short-list of some of the things this fixes/enhances:

- better indication of state resolution for the list resolvers using the enhanced `wordpress/core` resolver api.
- fix problems with the items reducer
- fix some bugs with the event-select component
- expose the event-select component via a new `@eventespresso/components` alias.  This allows you to use components by doing something like `import { EventSelect } from '@eventespresso/components'`.  Along with making things simpler to import, it also future proofs for when we eventually release our components as a package (we'd just have to change the webpack config).
- export validators from `assets/src/data/models` so that they are exposed anywhere models are imported.
- expose ee components as its own file and via `@eventespresso/components` aliased to `eejs.components`.  This enables ee add-ons to make use of any components.
- fix all tests related to changes and _improve_ tests to catch a bug I discovered while working in the `Gutenberg/event-attendees-block` branch where queryString entries in the store state for lists was not being retained through consecutive store state additions with additional `queryString` results.

## How has this been tested

* [x] Automated jest tests updated and passed (locally).
* [x] Changes were tested in a local branch comprised of the `Gutenberg/event-attendees-blocks` and this work to ensure that the basic functionality of the block worked as expected.

## Checklist

* [x] My code is tested.


nerrad:
This requires the changes in #471 to be merged in to master and then upstream to this pull before I can continue to adjust it so that the `components` library is exposed (currently living in a stash named "final-final-changes" in my local repo).

assets/src/components/index.js
```diff
@@ -0,0 +1 @@
+export { default as EventSelect } from './selection/event-select';
```
tn3rb: is there any chance that using "default" will cause name collisions?
or is it standard practice to export things named as "default" because the name gets changed at some other point?
what happens if one needed to utilize two different components that were both exported using the "default" name?

assets/src/components/selection/event-select/index.js
```diff
@@ -42,27 +43,35 @@ export const EventSelect = ( {
...
-			}
-		</Placeholder>;
+	if ( isEmpty( events ) ) {
+		return (
+			<Fragment>
+				<Placeholder
+					icon="calendar"
+					label={ __( 'EventSelect', 'event_espresso' ) }
+				>
+					{ ! isLoading && isEmpty( events ) ?
```
tn3rb: isn't this second call to `isEmpty()` redundant because we are already within the brackets for the first `isEmpty()` check?
ie: if `events` is NOT empty... then how do we get to this code in the first place?

assets/src/components/selection/event-select/index.js
```diff
@@ -42,27 +43,35 @@ export const EventSelect = ( {
...
+						__(
+							'There are no events to select from. You need to create an event first.',
+							'event_espresso',
+						) :
+						<Spinner />
```
tn3rb: I'm assuming `<Spinner />` is a self closing JSX tag and does not require any other attributes?
(and yes I know that it was imported from `@wordpress/components`)

assets/src/components/selection/event-select/index.js
```diff
@@ -172,8 +182,9 @@ const whereConditions = ( { showExpired, categorySlug, month } ) => {
  * eventespresso/lists store.
  */
 export default withSelect( ( select, ownProps ) => {
-	const { limit, order, orderBy } = ownProps.attributes;
-	const where = whereConditions( ownProps.attributes );
+	const { attributes = EventSelect.defaultProps.attributes } = ownProps;
```
tn3rb: is this kinda like double assigning the incoming `ownProps` variable to both `EventSelect.defaultProps.attributes` AND the `attributes` constant?

If so, then are there any actual *default* attributes saved to `EventSelect.defaultProps.attributes` that would be missed if you go on to only use the  `attributes` constant?

for example:
if `EventSelect.defaultProps.attributes` starts as `{ 'a' = 'a', 'b' = 'b', 'c' = 'c' }` 
but then the incoming `ownProps` is `{ 'a' = 'A', 'b' = 'B' }`,
then  the `attributes` constant will also be `{ 'a' = 'A', 'b' = 'B' }` and not have a value for `'c'` 

therefore should `ownProps` be merged with `EventSelect.defaultProps.attributes` before assigning to the `attributes` constant?

assets/src/components/index.js
nerrad: `default` is used here because I'm pulling in the default export from `./selection/event-select`  So this is essentially saying, export the default from the given file's default export and alias it to `EventSelect` for any files importing this.  

This is a standard way to pass through exports.

assets/src/components/selection/event-select/index.js
nerrad: ya it's redundant.  This went through a variety of conditional checks so I probably missed this when I switched the parent condition.  Good catch.

assets/src/components/selection/event-select/index.js
nerrad: correct.

assets/src/components/selection/event-select/index.js
nerrad: Well the tricky thing here is that attributes is actually something that should be defined by the parent block using this component.  If the parent block doesn't specify attributes then they will not be persisted with the block.  So in some ways its _good_ if this errors for any properties that aren't included.  The only reason why I set the default from the default properties of the EventSelect is to make it easier to initialize the EventSelect via `<EventSelect />` in the parent block.  However even then, at a minimum there would need to be passed through the onSelect callback etc.

assets/src/components/index.js
tn3rb: omg... somehow I read `as` incorrectly... can't even remember what I thought it was, 
but ya, what you said is painfully obvious now :facepalm:  sry about that

Issue #330, opened 2018-04-23
Facilitate having checkbox for consenting to site's privacy policy for GDPR


mnelson4:
When users register for an event, they should explicitly consent to the site's privacy policy, so we should have an input on registration forms for this.
Right now I think it should be optional in case someone adamantly refuses to have their regform modified

mnelson4:
The easiest way to achieve this would be:
-add a new system question, whose label is something like "Do you consent to having your data stored according to the Privacy Policy (link)?" and there would be a checkbox with the value "I consent to the Privacy Policy". 

This approach has a few drawbacks though:
1. Ideally this question will always appear at the very bottom of the reg form. We'd probably like to default to having this be part of the Personal info question group, but then if there are address questions or other systme questions, it will actually be towards the top of the form.
2. If we put it in a new system question group that has a very high order value so it always appears at the bottom, then site admins will need to manually include it on every event (and they probably want to use it on ALL or none)
3. If it's a regular checkbox, we'd probably want to allow admins to change the text, but certainly not provide more checkbox options, which by default the question would allow. Also, once the question is answered, they won't be able to change the text right next to the checkbox.
4. The value to this question will appear on the registration approved email, CSV reports, etc. That will probably be annoying. Ideally it just appears during registration, maybe even only during initial registration, and that's it.
5. We don't want this to appear for EACH registration on a group registration, just once on the form

So it's becoming clearer to me that this should NOT be new system question afterall.

I think there should be a checkbox in the admin saying "Require consent to Privacy Policy in Registraiton Form?", and a text field allowing the admin to choose what language is used for the field. Then, when that's checked, SPCO takes care of adding this extra checkbox right above the submit button, and preferably only when they first fill out the form (not on revisits)

Thoughts @sethshoultes ?

sethshoultes:
@mnelson4 i think that sounds like a good plan. Thanks for thinking this through.

mnelson4:
PR https://github.com/eventespresso/event-espresso-core/pull/470 was just merged into master

PR #470, opened 2018-05-21
Pull request of Gdpr/consent for ticket 330

mnelson4:
Addresses https://github.com/eventespresso/event-espresso-core/issues/330.

## What this adds

Adds a section to the registration form admin tab, see https://drive.google.com/a/eventespresso.com/file/d/1xGxnWSEBHSBDy9qi8h7DeRcoJi84SsWQ/view?usp=drivesdk. Specifically it adds
* with a dropdown for enabling a "privacy policy consent checkbox". 
* a textarea input (which is only shown when the checkbox is enabled) which will be the text of what the user consents to. This text defaults to having a link to the privacy policy page, if it's set.

When the checkbox is enabled, it appears in SPCO attendee info step (when it's not a revisit). Registrants are required to check it in order to register. Eg https://drive.google.com/a/eventespresso.com/file/d/1ciWAO4xZDrcym7JhMAwZ9D8sTF4ZyjxC/view?usp=drivesdk

## Current Drawbacks

The text for the checkbox is editable, and the default uses the company's name and the URL of the privacy policy page. But if either of those change, it doesn't get automatically updated. Admins need to manually update it. I thought maybe we should use shortcodes for those, and think it would be a very convenient UX, but want to avoid overcomplicating the UI and code (especially given this is on a timeline, technically for 4 days from now). Also, the page where this feature can be enabled/disabled is currently a caf-only page, so decaf users don't have access to it. So if their company's name or privacy policy changes, they're stuck with it. Using shortcodes for those values, or providing a reset button, or changing the admin page to be accessbible to non-admins are all options.

Right now, the new checkbox is defaulting to being ON. For new installs, I think that makes sense, especially given that lots of sites are defaulting to treating as if the GDPR applies to all visitors. I think most sites should have it. But some admins will probably insist on NOT adding a privacy policy page (or having one at all) and so will probably be annoyed that it appears all-of-a-sudden (mind you, WP currently doesn't give admins a choice for the new checkbox on comment fields... so we're still being more accomodating than WP core in that respect)


caffeinated/admin/extend/registration_form/Extend_Registration_Form_Admin_Page.core.php
```diff
@@ -1182,6 +1191,73 @@ protected function _email_validation_settings_form()
...
 
+    /**
+     * @throws EE_Error
+     * @throws InvalidArgumentException
+     * @throws InvalidDataTypeException
+     * @throws InvalidInterfaceException
+     */
+    public function privacyConsentFormContent()
+    {
+        echo $this->getPrivacyConsentForm()->get_html_and_js();
```
tn3rb: ideally I would prefer that you created a `PrivacyConsentForm` class that was its own class that extended `EE_Form_Section_Proper` and injected `EE_Registration_Config` into its constructor
OR
have a `PrivacyConsentFormHandler` class that extends `\EventEspresso\core\libraries\form_sections\form_handlers\FormHandler` which would negate the need for a lot of the boilerplate logic you currently have in `updatePrivacyConsentCheckboxSettingsForm()`,
and then instead of using the `getPrivacyConsentForm()` getter here, you would use the Loader to retrieve an instance of `PrivacyConsentFormHandler` 

for example:

if  `PrivacyConsentFormHandler`  extended `EventEspresso\core\libraries\form_sections\form_handlers\FormHandler`, then `privacyConsentFormContent()` could maybe change to something like:

```php
public function privacyConsentFormContent()
{
    /** @var \EventEspresso\core\libraries\form_sections\form_handlers\FormHandler $privacy_consent_form_handler*/
    $privacy_consent_form_handler= LoaderFactory::getLoader()->getShared('F/Q/C/N/PrivacyConsentFormHandler');
    echo $formHandler->display();
}
```

and `updatePrivacyConsentCheckboxSettingsForm()` could maybe change to something like:

```php
public function updatePrivacyConsentCheckboxSettingsForm(EE_Registration_Config $reg_config)
{
    /** @var \EventEspresso\core\libraries\form_sections\form_handlers\FormHandler $privacy_consent_form_handler */
    $privacy_consent_form_handler = LoaderFactory::getLoader()->getShared('F/Q/C/N/PrivacyConsentFormHandler');
    $privacy_consent_form_handler->process($reg_config);
    return $reg_config;
}
```

where the `PrivacyConsentFormHandler::generate()` method returned the actual `EE_Form_Section_Proper` object and `PrivacyConsentFormHandler::process()` method did something like:

```php
public function process(EE_Registration_Config $reg_config)
{
    try {
        $valid_data = (array) parent::process();
        $reg_config->setConsentCheckboxEnabled($valid_data['enable'] === 'enable-privacy-consent');
        $reg_config->setConsentCheckboxLabelText(
            $valid_data['enable-privacy-consent']['consent_assertion']
        );
    } catch (EE_Error $e) {
        // handle error
    }
    return $reg_config;
}
```

The benefit of this is that it keeps our classes simpler and focused solely on one responsibility.
It also makes it easier down the road to inject the `PrivacyConsentFormHandler` into this controller class when this route is called and not even have to bother with the Loader.

ie: one day it could be something as simple as:

```php
public function privacyConsentFormContent()
{
    echo $this->form_handler->display();
}

public function updatePrivacyConsentCheckboxSettingsForm(EE_Registration_Config $reg_config)
{
    return  $this->form_handler->process($reg_config);
}
```

core/libraries/form_sections/strategies/layout/EE_Template_Layout.strategy.php
```diff
@@ -112,6 +112,7 @@ public function layout_form_begin()
      */
     public function layout_input($input)
     {
+        echo "layotu input;";
```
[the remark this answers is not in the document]
mnelson4: OUPS

caffeinated/admin/extend/registration_form/Extend_Registration_Form_Admin_Page.core.php
mnelson4: yeah that sounds like a good suggestion. The refactoring should really just affect the one file so won't be too bad I think

tn3rb:
With regards to the **Current Drawbacks** addressed above:

I would think that any changes to a company's URL and/or name would require an extensive search through their content to catch any places that either of those had been used and personally don't think it would be fair for anyone to expect any software they use to manage that for them automatically (but of course some ppl will expect that and more). I know that for URL changes, it's best to use one of the many available scripts that finds and replaces content directly via the database, so that any URLs saved in serialized arrays have that data changed correctly, which prevents things from blowing up due to serialization errors (ie: unserialize the data, replace the URL, re-serialize the data, resave). We could implement a couple of simple shortcodes like `{COMPANY_NAME}` and `{SITE_URL}` and then perform a find a replace for those using the appropriate callbacks when rendering the final text. So they wouldn't be regular WordPress shortcodes, but would achieve the same thing in a pretty straightforward manner.

With regards to turning the option on or off by default, I agree that it's fine to have it turned on by default especially since that is inline with what WordPress is doing for comment forms. IF we wanted to get really fancy, then during activation (or upgrade, etc) we could pull the info for the site's country as set in the General Admin options and saved to the `EE_Organization_Config` object and then check the corresponding `CNT_is_EU` field from the database (assuming that info is still correct). THEN, we could set the privacy policy consent form default to "on" for countries that are part of the EU, and leave it off for all others. This of course requires a little bit of extra work, but not a whole lot.

mnelson4:
so @tn3rb the last remaining issue is that when you submit the form with invalid data, the form with the invalid data isn't shown to the user again- they're just shown the original form.

I traced the code to find out what's happening, and I see that on the POST request, `EE_Session` is constructed properly, has the form data added to it properly, and we attempt to save it, but saving the session fails.
I think this line is the culprit: https://github.com/eventespresso/event-espresso-core/blob/master/core/EE_Session.core.php#L888
Seeing how this is an admin request, there is no cart. And the method is getting called from https://github.com/eventespresso/event-espresso-core/blob/master/core/EE_Session.core.php#L847 with the default value for `$clear_session`. Is this a case that maybe wasn't considered when adding that? If so, I'm guessing that should be a separate ticket (and I assume any changes to that will require regression testing).

Alternatively, this admin code could NOT rely on the invalid form data being put into the session- we could do it the ol' fashioned way (where if the form is invalid on a POST request, we actually don't redirect and instead just show the invalid form immediately- as if it were a GET request. This is the way Django did it and the way the EE forms system originally did it, it's just not as slick a putting the invalid data into the session).

Thoughts?

core/EE_Config.core.php
```diff
@@ -2407,6 +2409,48 @@ public function set_default_max_ticket_on_EEM_Event()
...
+     * constructed because that happens before we can get the privacy policy page's permalink.
+     *
+     * @throws InvalidArgumentException
+     * @throws \EventEspresso\core\exceptions\InvalidDataTypeException
+     * @throws \EventEspresso\core\exceptions\InvalidInterfaceException
+     */
+    public function setDefaultCheckboxLabelText()
+    {
+        if ($this->getConsentCheckboxLabelText() === null
+            || $this->getConsentCheckboxLabelText() === '') {
```
tn3rb: was this not set during construction to "I Consent to Privacy Policy" ???

tn3rb: oh right... this was getting overwritten with nothing after loading the config from the db eh?

core/domain/services/admin/privacy/forms/PrivacySettingsFormHandler.php
```diff
@@ -0,0 +1,136 @@
...
+                        EEH_HTML::h2(esc_html__('Privacy Policy Consent Settings', 'event_espresso'))
+                    ),
+                    'enable'                   => new EE_Select_Reveal_Input(
+                        array(
+                            'enable-privacy-consent' => esc_html__('Enabled', 'event_espresso'),
+                            'disable'                => esc_html__('Disabled', 'event_espresso'),
+                        ),
+                        array(
+                            'default'         => $reg_config->isConsentCheckboxEnabled() ? 'enable-privacy-consent'
+                                : 'disable',
```
tn3rb: formatting looks funky here.. `? 'enable-privacy-consent'` should be on its own line

core/domain/services/admin/privacy/forms/PrivacySettingsFormHandler.php
```diff
@@ -0,0 +1,136 @@
...
+                            'html_label_text' => esc_html__('Privacy Consent Checkbox', 'event_espresso'),
+                            'html_help_text'  => esc_html__(
+                                'When enabled, a checkbox appears in the registration form requiring users to consent to your site\'s privacy policy.',
+                                'event_espresso'
+                            ),
+                        )
+                    ),
+                    'enable-privacy-consent'   => new EE_Form_Section_Proper(
```
tn3rb: why is this a whole new form section?
why not just add this as another input?

core/domain/services/admin/privacy/forms/PrivacySettingsFormHandler.php
```diff
@@ -0,0 +1,136 @@
...
+        try {
+            $valid_data = parent::process($submitted_form_data);
+            $reg_config = $this->config->registration;
+            $reg_config->setConsentCheckboxEnabled($valid_data['enable'] === 'enable-privacy-consent');
+            $reg_config->setConsentCheckboxLabelText(
+                $valid_data['enable-privacy-consent']['consent_assertion']
+            );
+            return $this->config->update_espresso_config(false, false);
+        } catch (InvalidFormSubmissionException $e) {
+            //the form was invalid, it show be re-displayed with errors
```
tn3rb: it **should** be ?

core/EE_Config.core.php
mnelson4: I think you're onto something, something was fishy. I think multiple iterations of this code got mixed up. 
We shouldn't set the default value during construction because this should be setting the default.

core/domain/services/admin/privacy/forms/PrivacySettingsFormHandler.php
mnelson4: d'oh! right

core/domain/services/admin/privacy/forms/PrivacySettingsFormHandler.php
mnelson4: currently, the select-reveal input works best when the section-to-be-revealed is a proper formsection, not just a form input. I tried setting it to an input, and it does successfully hide-and-reveal the requested input, but the input's label and help text etc doesn't get hidden. I'm ok with that restriction because we'll often want to hide groups of inputs anyways, and I'm not sure how much refactoring of the layout strategies would be required to change this.

core/domain/services/admin/privacy/forms/PrivacySettingsFormHandler.php
mnelson4: ya good catch

mnelson4:
kk I tried to implement your suggestions and fixes.
I'm not super excited that anytime we want to make use of the session anywhere outside the cart we need to remember to `add_filter('FHEE__EE_Session___save_session_to_db__abort_session_save', '__return_false');`, but I'm not sure what would be a better solution (and that's probably the best solution for this ticket). Maybe I'll have a suggestion on what might be better later.
Anyways, this is ready for another review

core/domain/services/admin/privacy/forms/PrivacySettingsFormHandler.php
tn3rb: OHHHHH kk... didn't connect the dots that it was done because of the select-reveal input. makes total sense now thnx

core/EE_Session.core.php
```diff
@@ -885,7 +890,10 @@ private function _save_session_to_db($clear_session = false)
     {
         // don't save sessions for crawlers
         // and unless we're deleting the session data, don't save anything if there isn't a cart
-        if ($this->request->isBot() || (! $clear_session && ! $this->cart() instanceof EE_Cart)) {
+        if (apply_filters(
+            'FHEE__EE_Session___save_session_to_db__abort_session_save',
+            $this->request->isBot() || (! $clear_session && ! $this->cart() instanceof EE_Cart)
```
tn3rb: I would prefer the filter to be added as an additional condition after the `EE_Cart` instance check so that we only check the filter if everything else evaluates to `true`, as any other result means the conditional will immediately pass and enter the return block..

The way it is in this commit means it will **always** check the filter which is unnecessary if any of the other conditions are true, ie:

 - if the current request is being made by a bot... we don't need to check the filter
 - if we are attempting to clear the session... we don't need to check the filter
 - if the session cart property contains a valid cart... we don't need to check the filter

we **only** need to check the filter if none of those conditions apply but we still want to save the session.

core/libraries/form_sections/base/EE_Form_Section_Proper.form.php
```diff
@@ -378,7 +378,7 @@ protected function get_submitted_form_data_from_session()
      * @throws InvalidInterfaceException
      * @throws InvalidDataTypeException
      */
-    protected function flush_submitted_form_data_from_session()
+    public static function flush_submitted_form_data_from_session()
```
tn3rb: why static?

does the action below not work correctly if it were:

```php
add_action(
    'shutdown', 
    array($this, 'flush_submitted_form_data_from_session')
);
```

core/libraries/form_sections/form_handlers/FormHandler.php
```diff
@@ -587,7 +587,7 @@ public function display()
         ) {
             $form_html .= $this->form()->form_open($this->formAction());
         }
-        $form_html .= $this->form(true)->get_html($this->form_has_errors);
+        $form_html .= $this->form(true)->get_html();
```
tn3rb: why are you making this change?
this could totally affect how other forms function?

core/libraries/form_sections/base/EE_Form_Section_Proper.form.php
mnelson4: good question. Because if there's a form with sub-forms, each one will set this as a callback from `EE_Form_Section_Proper::populate_from_session()` as that gets called on each sub-section. If this weren't static, then we'd call this method again for each sub-form which seems unnecessary. 
Granted, we could also achieve this by leaving it non-static, and making sure we only call it for formsections that have no parents. Would you prefer that?

core/libraries/form_sections/form_handlers/FormHandler.php
mnelson4: I tried to explain that in the commit message (https://github.com/eventespresso/event-espresso-core/pull/470/commits/f86313d72b5fc5a8a3134b9543e3b5ac1640bc27).
But basically, it seemed to me this logic was wrong: if the form was submitted on a previous request and found to be invalid, and then the invalid data was put into the session, then on the subsequent request, when this line gets ran as we're fetching the HTML to display the form, the instantiation of the form **does not yet have any errors**. 
And this logic is saying "well, if there are no errors in the form at the moment, ignore the session data, just display the form normally." So, even though there is stored invalid form data in the session, this code is saying to NOT check the session because we are unaware of any errors in the form.

Of course, if we were to only look in the session data, we'd see there is invalid form data in there. 

So I believe this new logic is right because just because the form hasn't been marked as "invalid" (ie, has validation errors) during THIS request, doesn't mean it wasn't marked as invalid during a previous request and that invalid data was stuffed into the session.

So this new logic basically says "regardless of whether we've found the form to be valid or invalid so far during this request, we still need to check in the session for invalid form data, because if there is we should display it."

Does that make sense?

And from what I understand, yes this will affect other forms, because the previous logic seemed like it was buggy (but you would know if somehow other code was relying on this logic.)

core/EE_Session.core.php
mnelson4: I prefer it this way as it's more flexible (other plugins can completely override the default logic in both to force either true or false, whereas your suggestion only allows them to make it false). And to my understanding, running a filter isn't onerous (we run thousands of them). iirc this is what I've been doing everywhere I want to allow code to override an `if` statement, so this seems consistent... yes, even if our current understanding dictates that, theoretically, we shouldn't need to override the `if` statement to force it to be true.

Is efficiency the main reason you're making this recommendation? Or is there something else I'm missing?

mnelson4:
awaiting feedback on my feedback of your feedback 🙃

core/EE_Session.core.php
tn3rb: The way I worded it sounds like my concern is efficiency but no, it's actually because we don't want those other conditions being overridden.

I guess the bot issue is debatable... but I don't know why someone would want to allow a bot to save session data, and the other two scenarios will just outright cause problems if some one erringly returns the wrong value and those events do not occur. Support already has to deal with enough issues with hosts/servers blocking the session, so I don't want to give them another potential way to cause that problem.

core/libraries/form_sections/base/EE_Form_Section_Proper.form.php
tn3rb: no that's a valid use of static imho. good call.

core/libraries/form_sections/form_handlers/FormHandler.php
tn3rb: > I tried to explain that in the commit message

hmmmm... ya that's stinky... GitHub doesn't always show all of the individual commit notes when reviewing because it kinda amalgamates all of the changes into one big final diff.

> Does that make sense?

roger Roger

core/EE_Session.core.php
mnelson4: kk I'll go with that then

mnelson4:
ok adjusted the filter. MOAR reviewzes

mnelson4:
ok switched the consent box to NOT appear by default, and so this technically requires a re-review.

We could implement brent's earlier suggestion of making it dynamic, based on whether, at the time of setting the default, the org is in an EU or non-EU country. That will work well for sites updating to this version of EE, but I don't think it will work as well for new installations, as we'd be setting this config option's default value BEFORE they have a chance to change their country. (And then we'd have to decide whether we'd automatically change this value when users change their country etc. It could work but it's starting to get unnecessarily complicated IMO. Just let users decide if they want to use it or not.)

## Testing

* [ ] Start with a site where there is NO privacy policy page setup yet, using WP 4.9.6 or higher. Switch to this branch and register for an event. It should work fine and nothing should have changed.
* [ ] Visit EE general settings page, then go to the privacy tab. Enable the consent box. The default text should be sensible to you. Save the form. It should stick. Try changing the default text, it should again stick.
* [ ] Register for an event again. The consent box should appear and be required.
* [ ] Revisit the attendee info step. The consent box should NOT appear this time (we don't ask for consent again)
* [ ] Go to WP's Privacy SEttings page and set a privacy page. Then go to the EE General settings then privacy settings page and remove all the privacy description and save. The updated text should have a link to the privacy policy page. Register for another event and you should see it, and when you click the link it should open the privacy policy page in a new tab.
* [ ] Try again using WP 4.9.5 or lower. There should be no fatal errors. If you reset the privacy policy text, there should again be no fatal error. However, you should still be able to enable the consent checkbox in EE and use it normally.

PR #472, opened 2018-05-22
Update license validation config

nerrad:

Just an update to our license validation config:

- remove whitelist items that pass okay now
- add CC-BY-3.0 to list of accepted licenses

In this pull request I also tweaked .travis.yml so that we aren't doing as many multisite jobs.  It's only really necessary to do one job in a multisite environment.

### DIFF
--- a/js-green-licenses.json
+++ b/js-green-licenses.json
@@ -10,24 +10,21 @@
 		"BSD",
 		"BSD-2-Clause",
 		"BSD-3-Clause",
-		"CC0-1.0"
+		"CC0-1.0",
+		"CC-BY-3.0"
 	],
 	"packageWhitelist": [
 		"@eventespresso/react-exit-modal-typeform",
-		"@wordpress/jest-preset-default",
 		"js-green-licenses",
 		"@typeform/embed",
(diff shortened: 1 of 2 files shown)


js-green-licenses.json
```diff
@@ -10,24 +10,21 @@
...
 	"packageWhitelist": [
 		"@eventespresso/react-exit-modal-typeform",
-		"@wordpress/jest-preset-default",
 		"js-green-licenses",
 		"@typeform/embed",
 		"scrollbar-width",
 		"webfontloader",
 		"validate-npm-package-license",
 		"spdx-correct",
-		"spdx-exceptions",
```
mnelson4: were these removed on purpose? I don't see why they're being removed

mnelson4: otherwise these changes seem fine

nerrad: the were removed because they no longer need to be whitelisted.  The license validation now correctly passes on these packages.

mnelson4: ahhhh- and they pass now because you added a new "greenLicense". Makes sense!

PR #474, opened 2018-05-23
Bug/duplicate event categories for 432

mnelson4:
Same pull request as https://github.com/eventespresso/event-espresso-core/pull/434, except that merged into a branch that was already merged into master, so that didn't get this branch into master. So this needs a new pull request (what I should have done is: change that pull request's base branch before merging)


mnelson4:
In creating this pull request, I see I made the same mistake on https://github.com/eventespresso/event-espresso-core/pull/420, which was merging it into its base branch AFTER I already merged that base branch into master. So 420's work is also not in master. So I think that's why that pull request's changes are in here, but I'm not sure why https://github.com/eventespresso/event-espresso-core/pull/419 's changes are appearing on this pull request. Its changes are already in master

nerrad:
> but I'm not sure why #419 's changes are appearing on this pull request. Its changes are already in master

If #419 was merged into master via a squash commit, that's why.  One way you can avoid this kind of thing with these kind of branch setups is to ALWAYS do rebase of the base branch onto tips.  That way when a base branch is merged into master you can rebase from master into the tip and it should just reflect the changes for that branch in the pull (plus have a cleaner commit history for reference).

PR #475, opened 2018-05-23
add type specific getters and... hasers ??

tn3rb:
Since the same handle may be shared by both a javascript and css asset, we need a way to differentiate between those assets stored in an AssetCollection and can not rely on the handle alone as an identifier.

## How has this been tested
added initially to the Gutenberg/event-attendees-block to see if it solved some problems there pertaining to this issue.

### DIFF
--- a/core/services/assets/AssetCollection.php
+++ b/core/services/assets/AssetCollection.php
@@ -64,7 +64,7 @@ class AssetCollection extends Collection
 
     /**
      * @param $type
-     * @return array
+     * @return JavascriptAsset[]|StylesheetAsset[]|ManifestFile[]
      * @since 4.9.62.p
      */
     protected function getAssetsOfType($type)
@@ -72,7 +72,7 @@ class AssetCollection extends Collection
         $files = array();
         $this->rewind();
         while ($this->valid()) {
-            /** @var \EventEspresso\core\domain\values\assets\Asset $asset */
+            /** @var Asset $asset */
             $asset = $this->current();
             if ($asset->type() === $type) {
                 $files[ $asset->handle() ] = $asset;
@@ -93,7 +93,7 @@ class AssetCollection extends Collection
         $files = array();
         $this->rewind();
         while ($this->valid()) {
-            /** @var \EventEspresso\core\domain\values\assets\JavascriptAsset $asset */
+            /** @var JavascriptAsset $asset */
             $asset = $this->current();
             if ($asset->type() === Asset::TYPE_JS && $asset->hasInlineData()) {
                 $files[ $asset->handle() ] = $asset;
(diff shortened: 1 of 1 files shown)


core/services/assets/AssetCollection.php
```diff
@@ -103,4 +103,119 @@ public function getJavascriptAssetsWithData()
...
+     * @access public
+     * @param  mixed $identifier
+     * @param string $type
+     * @return bool
+     */
+    public function hasAssetOfType($identifier, $type = Asset::TYPE_JS)
+    {
+        $this->rewind();
+        while ($this->valid()) {
+            if ($this->getInfo() === $identifier && $this->current() === $type) {
```
nerrad: Should this be `$this->current()->type()` === $type`?

core/services/assets/AssetCollection.php
```diff
@@ -103,4 +103,119 @@ public function getJavascriptAssetsWithData()
...
+     * @access public
+     * @param  mixed $identifier
+     * @param string $type
+     * @return JavascriptAsset|StylesheetAsset
+     */
+    public function getAssetOfType($identifier, $type = Asset::TYPE_JS)
+    {
+        $this->rewind();
+        while ($this->valid()) {
+            if ($this->getInfo() === $identifier && $this->current() === $type) {
```
nerrad: should this be `$this->current()->type() === $type`?

tn3rb:
FYI: last commit was just phpdoc changes so I used `[ci skip]` but apparently GitHub doesn't like that and is still asking for tests to pass.
here's the test results from the previous push:
https://travis-ci.org/eventespresso/event-espresso-core/builds/383275287

tn3rb:
can't test some of the changes in this branch because they would only affect the Gutenberg branches, but any issues would show up there.
we can however test the effects of https://github.com/eventespresso/event-espresso-core/pull/475/commits/30556be0a3a7e93b6b0860e53fc2321a4672ae79
since that will show up on master and fixes the issue that @joshfeck mentioned in chat: 

> Is this stylesheet supposed to be loading on every page of the front end?
> `/assets/dist/ee-wp-plugins-page.12612f6ab71570ab1af8.dist.css?ver=4.9.63.rc.009`

### testing

 * checkout core master branch
 * visit any non-EE page on the frontend of a site
 * examine the page's HTML header using your browser's inspection tool
 * [ ] confirm that a file named something like `/assets/dist/ee-wp-plugins-page.12612f6ab71570ab1af8.dist.css?ver=4.9.63.rc.009` is being loaded
 * now checkout the `add-type-specific-asset-methods` branch
 * refresh the page from above
 * [ ] confirm that `/assets/dist/ee-wp-plugins-page.12612f6ab71570ab1af8.dist.css?ver=4.9.63.rc.009` is no longer being loaded

PR #476, opened 2018-05-23
Gutenberg/use type specific asset methods

tn3rb:
see #475


core/services/assets/AssetCollection.php
```diff
@@ -205,9 +198,7 @@ public function getStylesheetAsset($identifier)
 
 
     /**
-     * has
-     * returns TRUE or FALSE
-     * depending on whether the Javascript Asset is within the Collection
+     * returns the Javascript Asset from the Collection
```
nerrad: Nitpick: `returns` should be sentence case (so `Returns`) and there should be a period at the end of the comment sentence for all of the above.  However, I'm ambivalent about whether you change it or not because I know none of us have been super consistent with it.

PR #483, opened 2018-05-28
QueryLimit: A React component for adding a WordPress Range Control input for setting a query's limit clause

tn3rb:
extracts logic that will very likely be highly repeated, into a component that will hopefully reduce a bit of the work required to build other components/blocks

because they will be used by core

## How has this been tested
browser tested only by import into a test block
## Checklist

* [x] My code is tested.


tn3rb:
> Darren Ethier [1:34 PM]
> based on the summary in here though @br3nt I think it’d be helpful to include in the pull request why you think a range control component is the best fit for this.

Given the current WordPress editor form components available:

 * autocomplete
 * base-control
 * button
 * button-group
 * checkbox-control
 * clipboard-button
 * code-editor
 * color-palette
 * dashicon
 * date-time
 * disabled
 * draggable
 * drop-zone
 * dropdown
 * dropdown-menu
 * external-link
 * focusable-iframe
 * font-size-picker
 * form-file-upload
 * form-toggle
 * form-token-field
 * higher-order
 * icon-button
 * keyboard-shortcuts
 * menu-group
 * menu-item
 * menu-items-choice
 * navigable-container
 * notice
 * panel
 * placeholder
 * popover
 * query-controls
 * radio-control
 * range-control
 * responsive-wrapper
 * sandbox
 * scroll-lock
 * select-control
 * server-side-render
 * slot-fill
 * spinner
 * tab-panel
 * text-control
 * textarea-control
 * toggle-control
 * toolbar
 * tooltip
 * tree-select

there does not appear to be a more applicable control type!

nerrad:
Let me clarify my question Brent.  Why are  you landing on using a `range` input method for implementing query _limits_ as opposed to using a simple number input?  I think I'm generally in agreement that a range-control is a good fit for this but I think its useful to articulate for our future selves why it's a good fit for indicating query limits.  That also can inform what choices you made for the code.

Also did you meant to commit the test block stuff in here?

PR #484, opened 2018-05-29
QueryLimit: A React component for adding a WordPress Range Control input for setting a query's limit clause

tn3rb:
adds a granular component for setting a query's limit value

## How has this been tested
added to the Event List component (not in this branch) and supported by Jest tests

## Checklist

* [x] My code is tested.


.eslintrc.js
```diff
@@ -13,9 +14,9 @@ module.exports = {
 		wpApiSettings: true,
 		eejsdata: true
 	},
-	plugins: [
-		'jest',
-	],
+	rules: {
+		"linebreak-style": 0
+	},
```
nerrad: I know your commit message said you added this linebreak-style for windows but I'm wondering _why_ you are adding this?  We've imported the `eslint-config-wordpress` ruleset which has this:

```
// Disallow mixed "LF" and "CRLF" as linebreaks
'linebreak-style': ['error', 'unix'],
```

In other words we're INTENTIONALLY avoiding CRLF in our code.   So I'd prefer you did not override this rule.  If you're getting linting fails locally, that's something you need to fix locally as the linter is performing as expected.  There's many options that you can use to do so, one of them is implement in phpStorm the reading of our  `.editorconfig` file.  Here's a github issue that may be useful to you as well: https://github.com/diegohaz/arc/issues/171

assets/src/components/query/limit/index.js
```diff
@@ -0,0 +1,76 @@
...
+ * 		  					limit={ limit }
+ * 		  					onLimitChange={
+ * 		  						( value ) => setAttributes( { limit: value } )
+ * 		  					}
+ * 		  				/>
+ * 		  			</PanelBody>
+ * 		  		</InspectorControls>
+ * 		  	);
+ * 		}),
+ */
```
nerrad: I prefer not to have documentation like this in the component.  Having the documentation in our docs folder (as you did) is sufficient or we could adopt the same pattern as GB where there's a README.md located in the same directory that has documentation for the things in that directory.  That might work better for js documentation as we remove the entire source directory for release builds.

assets/src/components/query/limit/index.js
```diff
@@ -0,0 +1,76 @@
...
+	return onLimitChange && (
+		<RangeControl
+			key={ 'query-limit' }
+			value={ limit }
+			label={ label }
+			min={ min }
+			max={ max }
+			onChange={ onLimitChange }
+		/>
+	);
```
nerrad: Should there be an error or something human friendly error message when `onLimitChange` is undefined?  I'm on the fence because it will already be clear something isn't working correctly likely.

assets/src/components/query/limit/test/index.js
```diff
@@ -0,0 +1,54 @@
...
+			expect( limit.props().max ).toBe( 50 );
+			// and label
+			const label = wrapper.find( 'label' );
+			expect( label.text() ).toEqual( 'No Limit' );
+			// test onChange
+			limit.simulate( 'change', { target: { value: '30' } } );
+			expect( onChange ).toHaveBeenCalledWith( 30 );
+		} );
+	} );
+} );
```
nerrad: A++ love it :)  the only change I'd like to see here is a test for when `onChange` is undefined.

docs/P--Gutenberg Blocks and Components/README.md
```diff
@@ -0,0 +1,13 @@
...
+The following is a list of Event Espresso Gutenberg Blocks that can be utilized in your site's content:
+
+ * [Event Attendees List](blocks/event-attendees.md) 
+ 
+### Components
+If you are building your own Gutenberg Block to use within Event Espresso, then the following is a list of additional React components can be utilized:
+ 
+ 
+ * [QueryLimit](components/query/limit.md) - A React component for adding a WordPress Range Control input for setting a query's limit clause
+
```
nerrad: Gutenberg is a code name and I'm not sure we should use it in our documentation.  In the case of the folder structure it means that if we remove the name down the road it will break any direct bookmarks people have made. 

So I have a couple suggestions:

- `P--Blocks-and-Components`  (please use dashes because they are easier to work with for urls).  
- or nest it in the existing `AA-Javascript-in-EE` folder so it would be `AA-Javascript-in-EE/Blocks-and-Components/`  I prefer this option because this stuff is _mostly_ js related.  However, I'm not going to be sticky on this point because I also see the value of it being a top-level folder.

webpack.common.js
```diff
@@ -97,6 +97,9 @@ const config = [
 			'data-stores': [
 				assets + 'data/index.js',
 			],
+			'eventespresso-core-blocks': [
+				assets + 'blocks/index.js',
+			],
```
nerrad: I don't think this should be in this branch since there is no `assets/src/blocks/index.js` in this branch?

nerrad: Ya so this is what is causing the fail on travis because the `npm build` step in the job fails with this entry.  So removing it will fix that.

.eslintrc.js
nerrad: I also know your rule just disables the linebreak style checks.  I'd still prefer if you left the rules as is.

nerrad: I really don't know what to say then, because what you are seeing does not jive with either what I see locally or what travis sees.

nerrad: as a short term fix for your needs locally, you can run just the js unit-tests without a lint check via `npm run test-unit`.   However, long-term you'll want to look into what is changing the line endings on your machine.

assets/src/components/query/limit/index.js
tn3rb: Is there an actual reason other than preference?

I added this here so that someone looking through our components folder that wanted to use `QueryLimit` would know how to do so without having to go searching elsewhere for documentation about it. So maybe having a README.md here would be good as well as the other documentation I provided (ie: just duplicate all block/component documentation so that there are local in place READMEs as well as the docs folder where everything is combined).

We may even want to follow this pattern for PHP and include READMEs within individual folders, so for example the `core/services/commands` folder could have it's own README.md which would be a duplicate (or link to) the existing documenation at `\docs\N--Core-Functionality\command-bus-hexagonal-architecture.md`.

thoughts?

docs/P--Gutenberg Blocks and Components/README.md
tn3rb: will rename and move

PR #490, opened 2018-05-30
Bug/privacy policy fatal error in maintenance mode fixes #489

### LINKED ISSUE #489: Fix Level 2 maintenance mode causing fatal error from PrivacyPolicy.php
**Describe the bug**
If you activate Level 2 maintenance mode by setting up a new site and activating an older version of EE and then switch to 4.9.62+ you'll get a fatal error:

```
Fatal error: Uncaught EE_Error: Event Espresso Level 2 Maintenance mode is active. That means EE can not run ANY database queries until the necessary migration scripts have run which will take EE out of maintenance mode level 2. Please inform support of this error. in D:\xampp\htdocs\ee4\wp-content\plugins\event-espresso-core-reg\core\db_models\EEM_Base.model.php:2378 Stack trace: 
#0 D:\xampp\htdocs\ee4\wp-content\plugins\event-espresso-core-reg\core\db_models\EEM_Base.model.php(1080): EEM_Base->_do_wpdb_query('get_results', Array) 
#1 D:\xampp\htdocs\ee4\wp-content\plugins\event-espresso-core-reg\core\db_models\EEM_Base.model.php(954): EEM_Base->_get_all_wpdb_results(Array, 'ARRAY_A', NULL) 
#2 D:\xampp\htdocs\ee4\wp-content\plugins\event-espresso-core-reg\core\db_models\EEM_Payment_Method.model.php(134): EEM_Base->get_all(Array) 
#3 D:\xampp\htdocs\ee4\wp-content\plugins\event-espresso-core-reg\core\domain\services\admin\privacy\policy\PrivacyPolicy.php(65): EEM_Payment_Method->get_all_active('CART') 
#4 D:\xampp\htdocs\ee4 in D:\xampp\htdocs\ee4\wp-content\plugins\event-espresso-core-reg\core\db_models\EEM_Base.model.php on line 2378
```

The above happens if you trigger any migrations, so if you had an older version of the calendar active previously (but no longer use it), then activate the calendar with EE4 core v4.9.62+ you'll get a fatal so marking as critical.

mnelson4:
Addresses https://github.com/eventespresso/event-espresso-core/issues/489 by only loading the privacy policy code when the site isn't in full maintenance mode. 
Also introduces `EE_Admin::initModelsReady()` which should be used for admin code that only runs when the site is in NOT in full maintenance mode.

### DIFF
--- a/core/EE_System.core.php
+++ b/core/EE_System.core.php
@@ -1199,17 +1199,9 @@ final class EE_System implements ResettableInterface
         );
         $rewrite_rules->flushRewriteRules();
         add_action('admin_bar_init', array($this, 'addEspressoToolbar'));
-        if ($this->request->isAdmin()) {
-            $this->loader->getShared('EventEspresso\core\services\privacy\policy\PrivacyPolicyManager');
+        if($this->request->isAjax() && $this->maintenance_mode->models_can_query()) {
             $this->loader->getShared('EventEspresso\core\services\privacy\export\PersonalDataExporterManager');
             $this->loader->getShared('EventEspresso\core\services\privacy\erasure\PersonalDataEraserManager');
-        }
-        if ($this->request->isAjax()) {
-            $this->loader->getShared('EventEspresso\core\services\privacy\export\PersonalDataExporterManager');
-            $this->loader->getShared('EventEspresso\core\services\privacy\erasure\PersonalDataEraserManager');
-        }
-        if ($this->request->isAdmin() && function_exists('wp_add_privacy_policy_content')) {
-            $this->loader->getShared('EventEspresso\core\services\privacy\policy\PrivacyPolicyManager');
         }
     }
 
--- a/core/EE_System.core.php
+++ b/core/EE_System.core.php
@@ -1199,7 +1199,8 @@ final class EE_System implements ResettableInterface
         );
         $rewrite_rules->flushRewriteRules();
         add_action('admin_bar_init', array($this, 'addEspressoToolbar'));
-        if ($this->request->isAjax() && $this->maintenance_mode->models_can_query()) {
+        if (($this->request->isAjax() || $this->request->isAdmin())
+            && $this->maintenance_mode->models_can_query()) {
             $this->loader->getShared('EventEspresso\core\services\privacy\export\PersonalDataExporterManager');
             $this->loader->getShared('EventEspresso\core\services\privacy\erasure\PersonalDataEraserManager');
         }
--- a/core/EE_System.core.php
+++ b/core/EE_System.core.php
@@ -1199,7 +1199,7 @@ final class EE_System implements ResettableInterface
         );
         $rewrite_rules->flushRewriteRules();
         add_action('admin_bar_init', array($this, 'addEspressoToolbar'));
-        if($this->request->isAjax() && $this->maintenance_mode->models_can_query()) {
+        if ($this->request->isAjax() && $this->maintenance_mode->models_can_query()) {
             $this->loader->getShared('EventEspresso\core\services\privacy\export\PersonalDataExporterManager');
             $this->loader->getShared('EventEspresso\core\services\privacy\erasure\PersonalDataEraserManager');
         }
--- a/core/admin/EE_Admin.core.php
+++ b/core/admin/EE_Admin.core.php
@@ -8,6 +8,7 @@ use EventEspresso\core\interfaces\InterminableInterface;
 use EventEspresso\core\services\container\exceptions\ServiceNotFoundException;
 use EventEspresso\core\services\loaders\LoaderFactory;
 use EventEspresso\core\services\notifications\PersistentAdminNoticeManager;
+use WPStaging\Utils\Loader;
 
 /**
  * EE_Admin
@@ -28,6 +29,11 @@ final class EE_Admin implements InterminableInterface
      * @var PersistentAdminNoticeManager $persistent_admin_notice_manager
      */
     private $persistent_admin_notice_manager;
+
+    /**
+     * @var Loader
+     */
+    protected $loader;
 
     /**
      * @singleton method used to instantiate class object
@@ -215,6 +221,22 @@ final class EE_Admin implements InterminableInterface
 
 
     /**
+     * Gets the loader (and if it wasn't previously set, sets it)
+     * @return \EventEspresso\core\services\loaders\LoaderInterface|Loader
+     * @throws InvalidArgumentException
+     * @throws InvalidDataTypeException
+     * @throws InvalidInterfaceException
+     */
+    protected function getLoader()
+    {
+        if (! $this->loader instanceof Loader){
+            $this->loader = LoaderFactory::getLoader();
+        }
+        return $this->loader;
+    }
+
+
+    /**
      * Method that's fired on admin requests (including admin ajax) but only when the models are usable
      * (ie, the site isn't in maintenance mode)
      * @since $VID:$
@@ -222,9 +244,8 @@ final class EE_Admin implements InterminableInterface
      */
     protected function initModelsReady()
     {
-        $loader = LoaderFactory::getLoader();
         // ok so we want to enable the entire admin
-        $this->persistent_admin_notice_manager = $loader->getShared(
+        $this->persistent_admin_notice_manager = $this->getLoader()->getShared(
             'EventEspresso\core\services\notifications\PersistentAdminNoticeManager'
         );
         $this->persistent_admin_notice_manager->setReturnUrl(
@@ -582,8 +603,6 @@ final class EE_Admin implements InterminableInterface
      */
     public function admin_init()
     {
-        $loader = LoaderFactory::getLoader();
-
         /**
          * our cpt models must be instantiated on WordPress post processing routes (wp-admin/post.php),
          * so any hooking into core WP routes is taken care of.  So in this next few lines of code:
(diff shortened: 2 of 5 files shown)


core/EE_System.core.php
```diff
@@ -1199,18 +1199,10 @@ public function initialize_last()
         );
         $rewrite_rules->flushRewriteRules();
         add_action('admin_bar_init', array($this, 'addEspressoToolbar'));
-        if ($this->request->isAdmin()) {
-            $this->loader->getShared('EventEspresso\core\services\privacy\policy\PrivacyPolicyManager');
+        if($this->request->isAjax() && $this->maintenance_mode->models_can_query()) {
```
mnelson4: FYI I left these methods in here because it turns out `EE_Admin` isn't loaded on admin ajax that doesn't have `ee_admin_ajax`

tn3rb: just need a space after the `if` in this line in order to pass the code sniffer

core/admin/EE_Admin.core.php
```diff
@@ -194,24 +194,7 @@ public function init()
...
-                    ),
-                    EE_ADMIN_URL
-                )
-            );
-            $this->maybeSetDatetimeWarningNotice();
-            // at a glance dashboard widget
-            add_filter('dashboard_glance_items', array($this, 'dashboard_glance_items'), 10);
-            // filter for get_edit_post_link used on comments for custom post types
-            add_filter('get_edit_post_link', array($this, 'modify_edit_post_link'), 10, 2);
+           $this->initModelsReady();
```
tn3rb: indenting needs fixing here to pass the code sniffer

mnelson4:
Resolved the phpcs issues, so this needs another review

## Testing
* [ ] make sure the Wp option "ee_maintenance_mode" has value 2 and you're on the latest release of WP. Visit the admin. There was a fatal error on master branch, but not on this branch
* [ ] Visit the maintenance page (it will automatically take the site out of maintenance mode when you do this). Then visit the settings -> privacy, then view the privacy policy guide. EE's section should still appear
* [ ] Create a request to export user data for an EE registration (ie, create a request to export data for an email used by an EE registration). Confirm the request (from the email that WP sends) and create the export file. It should still include EE's data
* [ ] Create a request to erase userdata for an email used by an EE registration. The EE attendee should again be renamed anonymous and have their answers removed
* [ ] Run EE with an older version of WP (eg 4.9.5 or earlier). It should still work fine

sethshoultes:
@mnelson4 please assign to someone on the support team once ready for testing

Pebblo:
Hey Mike, are these testing notes correct? The reason I ask is I'm setting my site to MM2 using:

`update_option('ee_maintenance_mode', 2);`

That's in a custom functions plugin, I refresh the page to set the option then comment out the code and refresh again just to be sure, but I'm not getting any fatal errors (I'm on master).

mnelson4:
> Hey Mike, are these testing notes correct? ...

Yes I was getting the fatal error on master. Are you on WP 4.9.6 or newer? Which admin page were you visiting @Pebblo ?

Pebblo:
Yep, definitely 4.9.6 and MM2 - http://take.ms/PXv10

I can view any admin page, this is the same site I found the problem on but using a different DB.

Something really strange, if I switch to the DB I used to find this originally, I get the error as soon as I set MM, but on this other DB I don't get any errors (I only have EE and the plugin I'm using to set MM active).

Want a copy of the 'working' DB?

mnelson4:
I saw in the code that this the privacy policy code doesn't execute when there's no privacy policy page set- I think that explains why one site was having the issue (it had a privacy policy page defined) and one site didn't (it didn't have a privacy policy page defined).
However, that was out-of-date logic because this content now appears on a different page, the privacy policy guide page. 
So I modified the code so the privacy policy page content is only generated on the privacy policy guide page.

core/services/privacy/policy/PrivacyPolicyManager.php
```diff
@@ -26,25 +26,27 @@ class PrivacyPolicyManager
...
-        $policy_page_id = (int) get_option('wp_page_for_privacy_policy');
-        if (! $policy_page_id) {
-            return;
-        }
-        // load all the privacy policy stuff
-        // add post policy text
-        foreach ($this->loadPrivacyPolicyCollection() as $privacy_policy) {
-            wp_add_privacy_policy_content($privacy_policy->getName(), $privacy_policy->getContent());
+        $screen = get_current_screen();
+        if ($screen instanceof \WP_Screen
```
tn3rb: the `do_action( 'current_screen', $current_screen );` hookpoint passes the global `$current_screen` object as a parameter, so instead of calling `get_current_screen()` you can simply type hint for `WP_Screen` in your callback.

As well, plz import all classes, even those from the root namespace

core/admin/EE_Admin.core.php
```diff
@@ -569,6 +582,7 @@ public function wp_loaded()
      */
     public function admin_init()
     {
+        $loader = LoaderFactory::getLoader();
```
tn3rb: you could instead add a `$loader` property to the class and change this line to:

```php
$this->loader = LoaderFactory::getLoader();
```

and then use `$this->loader` from within `adminInitModelsReady()` instead of having to use the `LoaderFactory` again.

This also has the benefit of making it easier in the future to inject the `Loader` and simply replace the one line of code where the `LoaderFactory` is being used.

core/services/privacy/policy/PrivacyPolicyManager.php
mnelson4: oh right- both good calls

core/admin/EE_Admin.core.php
mnelson4: ya good idea

core/admin/EE_Admin.core.php
```diff
@@ -8,6 +8,7 @@
 use EventEspresso\core\services\container\exceptions\ServiceNotFoundException;
 use EventEspresso\core\services\loaders\LoaderFactory;
 use EventEspresso\core\services\notifications\PersistentAdminNoticeManager;
+use WPStaging\Utils\Loader;
```
tn3rb: oops... think you grabbed the wrong Loader class  😆 

oh and plz always reference the interface everywhere except in the dependency map (or unit test setup)

core/admin/EE_Admin.core.php
```diff
@@ -29,6 +30,11 @@
      */
     private $persistent_admin_notice_manager;
 
+    /**
+     * @var Loader
```
tn3rb: plz always reference interface

core/admin/EE_Admin.core.php
```diff
@@ -214,6 +220,22 @@ public function init()
     }
 
 
+    /**
+     * Gets the loader (and if it wasn't previously set, sets it)
+     * @return \EventEspresso\core\services\loaders\LoaderInterface|Loader
```
tn3rb: plz import all FQCNs

core/admin/EE_Admin.core.php
mnelson4: 😖 good catch! And ya, I'll try to reference the loader interface instead

core/admin/EE_Admin.core.php
mnelson4: :+1:

core/admin/EE_Admin.core.php
mnelson4: :+1:

mnelson4:
ok @Pebblo this is ready for testing agian. I updated the first test to mention that you need to have a WP privacy policy page created in order to reproduce the error on EE master (see https://github.com/eventespresso/event-espresso-core/pull/490#issuecomment-393661770)

Pebblo:
> I think that explains why one site was having the issue (it had a privacy policy page defined) and one site didn't (it didn't have a privacy policy page defined).

Yeah, that'd make sense, I did some digging and the 'working' db was an update from WP4.9.5, the broken db was a clean install of 4.9.6.

The updated DB does not have a privacy page selected.

---

* [ ] Visit the maintenance page (it will automatically take the site out of maintenance mode when you do this). Then visit the settings -> privacy, then view the privacy policy guide. EE's section should still appear

The text for billing information is only displayed if you have an ON site payment method, I had only PayPal Express enabled and there's no text under 'Billing information', it just shows the next heading.

---

* [ ] Create a request to export user data for an EE registration (ie, create a request to export data for an email used by an EE registration). Confirm the request (from the email that WP sends) and create the export file. It should still include EE's data

This feature isn't working on this branch. It works on master, but not on this branch - http://take.ms/Jki0b

(Same request after switching to master and refreshing the page works fine).

---

* [ ] Create a request to erase userdata for an email used by an EE registration. The EE attendee should again be renamed anonymous and have their answers removed

This has the same issue as above, doesn't work on this branch but does on master.

mnelson4:
@Pebblo 
> The text for billing information is only displayed if you have an ON site payment method, I had only PayPal Express enabled and there's no text under 'Billing information', it just shows the next heading.

That's normal- iirc we don't store billing info for off-site payment methods, so that text is meant to be dynamic. I don't think that's any different than how it is on master, right? 

> This feature isn't working on this branch. It works on master, but not on this branch

Argh, ya I'm reproducing. I thought I had tested that but apparently made another change afterwards that borked it.

@tn3rb 
It turns out the exporter and erasure managers need to load on admin requests too, as they need to be loaded when showing the data export and erasure pages (they need to be sent to the page's javascript for the subsequent ajax requests to work properly). I had thought about only calling `add_filter` calls on the exact tools page where they're used (ie, inside `PersonalDataExporterManager::__construct()` having the request injected and checking if it's an ajax request OR an admin request to the personal data exporter page), but it seemed doing the logic of checking would probably be about the same speed as just adding the filter like it is now. So I figured I'd initially just leave it as-is.

Pebblo:
> That's normal- iirc we don't store billing info for off-site payment methods, so that text is meant to be dynamic. I don't think that's any different than how it is on master, right?

Yeah it's the same on master but I still thinks its wrong (it's checking for on-site payment methods in 2 conditionals, I think the second should be offsite, just a different var). I'll do another ticket for this as its not related to these changes.

---

The testing requirements above all work fine now.

The phone number isn't removed for the user when the contact is set to Anonymous and I think that should be included but will do another ticket for it.

So I think this can be merged @mnelson4

PR #497, opened 2018-06-01
Bug/registry check addons exist before using them

mnelson4:
This resolves the saas issue https://github.com/eventespresso/eventsmart.com-website/issues/36.
Apparently if you try to access a non-existent property on an `stdClass` it can cause an error.
I don't think there is any way in EE core to reproduce the issue (besides the unit tests that I just added).

### DIFF
--- a/core/EE_Registry.core.php
+++ b/core/EE_Registry.core.php
@@ -1454,7 +1454,11 @@ class EE_Registry implements ResettableInterface
     public function getAddon($class_name)
     {
         $class_name = str_replace('\\', '_', $class_name);
-        return $this->addons->{$class_name};
+        if (isset($this->addons->{$class_name})) {
+            return $this->addons->{$class_name};
+        } else {
+            return null;
+        }
     }
 
 
@@ -1467,7 +1471,9 @@ class EE_Registry implements ResettableInterface
     public function removeAddon($class_name)
     {
         $class_name = str_replace('\\', '_', $class_name);
-        unset($this->addons->{$class_name});
+        if (isset($this->addons->{$class_name})) {
+            unset($this->addons->{$class_name});
+        }
     }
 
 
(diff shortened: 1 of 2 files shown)


mnelson4:
Requesting a review from Darren in order to free up Brent (as far as I know that policy hasn't been rescinded), especially when this seems pretty straightforward

core/EE_Registry.core.php
```diff
@@ -1467,7 +1471,9 @@ public function getAddon($class_name)
     public function removeAddon($class_name)
     {
         $class_name = str_replace('\\', '_', $class_name);
-        unset($this->addons->{$class_name});
+        if (isset($this->addons->{$class_name})) {
+            unset($this->addons->{$class_name});
```
nerrad: isset checks are not needed before unsetting something. It's completely safe to just unset.

nerrad:
Also, @mnelson4 can you please add some correct labels to this pull?  Thanks!

core/EE_Registry.core.php
mnelson4: :+1:

mnelson4:
re-review plz!

mnelson4:
hmm, this needed master merged in, so I rebased this branch on master (ie, `git checkout {this-branch}; git rebase master;`) but that dismissed @nerrad 's review (or marked it as stale)... 
Anyways, this apparently needs another review. 

## Testing

* [ ] Activate an add-on (Eg calendar). It should behave normally. (If you notice a bug with it, double-check that bug doesn't occur with EE master; if so, that bug report should probably go into that add-on's repo.)

Otherwise, the unit tests and testing on ES should take care of testing this

mnelson4:
@sethshoultes would you like to do the manual test for this? Or reassign to someone else for testing?

PR #498, opened 2018-06-01
Fix CPT query filters being applied twice Fixes #496

### LINKED ISSUE #496: Fix WordPress database error: [Not unique table/alias: 'wp2_esp_datetime']
This is a new bug since 4.9.62.p

Steps to replicate:

1. Switch to the twentyseventeen theme
2. Unzip and upload the attached custom archive template to the twentyseventeen folder
3. View the /events/ archive on the front end of the site
[archive-espresso_events.php.zip](https://github.com/eventespresso/event-espresso-core/files/2063566/archive-espresso_events.php.zip)

tn3rb:
In #496 it was discovered that running an additional `WP_Query` for an EE CPT within a template configured to load on that same EE CPT Archives page resulted in an SQL error because the same table alias was being applied to the query multiple times. There are apparently other conditions that can cause this to occur, but that was the only scenario presented in that issue.

This was occuring because of changes introduced by #339 which provides better control over caching of objects. Specifically it prevents caching when attempting to create a new shared instance of an object using a new set of properties, for example, a `User` object with the name property set via constructor args to "Bob" could be shared and a request for a shared instance of `User` with the name "Doug" would result in a new instance being returned. Previous to 339, a request for a new shared instance of "Doug" would result in "Bob" being returned because objects were being cached solely on their classnames with no regard to the initial constructor arguments.

Prior to #339, this is what would happen if a theme employed a custom template for the `/events/` archive page and that template ran its own `WP_Query` to retrieve a list of `espresso_events` posts.

-  `EE_CPT_Strategy` loads in order to monitor any `WP_Query` being performed to determine if they are for an EE CPT 
- because the current request was for the `/events/` archive endpoint, the WP main query would be for the `EE_Event` CPT
- `EE_CPT_Strategy` would generate a new instance of `EventEspresso\core\CPTs\CptQueryModifier` for the current `WP_Query` 
- `CptQueryModifier` would determine the requested EE CPT and use `EE_Registry` to load an instance of `EE_CPT_Event_Strategy` (as well as other stuff)
- **IMPORTANT** - `EE_Registry` would determine whether to load a new or cached version of `EE_CPT_Event_Strategy` based solely on the class name, which would result in a **new** instance of  `EE_CPT_Event_Strategy` being generated since no instances exist
- upon construction,  `EE_CPT_Event_Strategy` would set some class properties based on the supplied constructor parameters (<= also important)
- `EE_CPT_Event_Strategy` would then apply a set of filters to the `WP_Query` that was about to be performed
- **IMPORTANT** - `EE_CPT_Event_Strategy` did **not** remove the filters being applied, meaning that **any** additional `WP_Query`s would have those filters applied
- the `WP_Query` would run and return its results which at this point were for the WP main query that is run **prior to template loading**
- eventually the custom event archives template would load which would then generate its own `WP_Query` for the `EE_Event` CPT
- `EE_CPT_Strategy` which is still monitoring any `WP_Query`s sees another query for the `EE_Event` CPT
- `EE_CPT_Strategy` would generate another new instance of `EventEspresso\core\CPTs\CptQueryModifier` for this new `WP_Query` 
- `CptQueryModifier` would determine the requested EE CPT and use `EE_Registry` to load an instance of `EE_CPT_Event_Strategy`
- **IMPORTANT** - `EE_Registry` would determine whether to load a new or cached version of `EE_CPT_Event_Strategy` based solely on the class name, which would result in the **previously cached** instance of  `EE_CPT_Event_Strategy` being returned. Because a previously cached instance of  `EE_CPT_Event_Strategy` was returned, the constructor would not run again, and therefore the same properties for the previous query would be used for any subsequent queries. This was really not good because Query B was being modified based on some of the properties of Query A.
- **SUPER IMPORTANT** - because `EE_CPT_Event_Strategy` never removed the filters it had applied to the previous query, those same filters would run again on the new query. For any dissimilar queries this would not have mattered because the filter callbacks have some conditionals in place to determine whether they should still be applied or not, and most of the conditionals pertained to the `WP_Query` supplied to the callback, however, some pertained to the class properties that had been set upon construction.  In other words, despite the filters never being removed, a subsequent `WP_Query` for anything other than an `EE_Event` CPT would almost certainly not get applied because the incoming query was different enough, but any subsequent `WP_Query`s for an `EE_Event` CPT would have filters applied based partially on the **first**  `WP_Query`s for an `EE_Event` CPT.

- **TL;DR: requests for shared `EE_CPT_Event_Strategy(A)` and `EE_CPT_Event_Strategy(B)` objects resulted in a shared instance of `EE_CPT_Event_Strategy(A)` being used to apply filters to both `WP_Query(A)` and `WP_Query(B)`**

So what changed after implementing #339?

- changes to object caching allowed for an object's constructor arguments to be factored into how it would get shared. So two objects of the same class with different properties could both be shared: ie `User("Bob")` and `User("Doug")`. With regards to queries this meant that two queries for the same EE CPT, but with different arguments, could both be shared, ie: `WP_Query(A)` and  `WP_Query(B)` would ultimately result in the creation of `EE_CPT_Event_Strategy(A)` **and** `EE_CPT_Event_Strategy(B)`
- however because `EE_CPT_Event_Strategy` never removed the filters it had applied to a query, the filters added by `EE_CPT_Event_Strategy(A)` to `WP_Query(A)` would also get applied to `WP_Query(B)` **as well as** the filters added by `EE_CPT_Event_Strategy(B)`
- this meant that some of the query clauses had aliases like `event_start_date` added more than once to the same query, resulting in the `Not unique table/alias` error

#### the fix

So `EE_CPT_Event_Strategy` already contained logic for removing all of the filters it added to a query but that logic was not being applied anywhere. It's possible that this was the case at some point in time, but it had been discovered that the filters were missing on subsequent queries and so the filter removal was itself removed because that would have been the simplest most straightforward fix, albeit a seriously imperfect one. The primary fix in this branch is simply removing the query filters after they have been applied. Additional changes include always using the new `Loader` class (and/or injecting dependencies) for obtaining class instances because that minimizes the chance of caching errors.

## How has this been tested
browser testing only using the template supplied by @joshfeck in #496 

## Checklist

* [x] My code is tested.

### DIFF
--- a/core/CPTs/CptQueryModifier.php
+++ b/core/CPTs/CptQueryModifier.php
@@ -4,7 +4,8 @@ namespace EventEspresso\core\CPTs;
 
 use EventEspresso\core\exceptions\InvalidDataTypeException;
 use EventEspresso\core\exceptions\InvalidInterfaceException;
-use EventEspresso\core\services\loaders\LoaderFactory;
+use EventEspresso\core\services\loaders\LoaderInterface;
+use EventEspresso\core\services\request\RequestInterface;
 use InvalidArgumentException;
 
 /**
@@ -55,15 +56,24 @@ class CptQueryModifier
     protected $model;
 
     /**
-     * @var \EE_Request_Handler $request
+     * @var \EE_Request_Handler $request_handler
      */
-    protected $request;
+    protected $request_handler;
 
     /**
      * @var \WP_Query $wp_query
      */
     protected $wp_query;
 
+    /**
+     * @var LoaderInterface $loader
+     */
+    protected $loader;
+
+    /**
+     * @var RequestInterface $request
+     */
+    protected $request;
 
     /**
      * CptQueryModifier constructor
@@ -71,15 +81,21 @@ class CptQueryModifier
      * @param string              $post_type
      * @param array               $cpt_details
      * @param \WP_Query           $WP_Query
-     * @param \EE_Request_Handler $request
+     * @param \EE_Request_Handler $request_handler
+     * @param RequestInterface    $request
+     * @param LoaderInterface     $loader
      */
     public function __construct(
         $post_type,
         array $cpt_details,
         \WP_Query $WP_Query,
-        \EE_Request_Handler $request
+        \EE_Request_Handler $request_handler,
+        RequestInterface $request,
+        LoaderInterface $loader
     ) {
-        $this->setRequest($request);
+        $this->loader = $loader;
+        $this->request = $request;
+        $this->request_handler = $request_handler;
         $this->setWpQuery($WP_Query);
         $this->setPostType($post_type);
         $this->setCptDetails($cpt_details);
@@ -199,21 +215,15 @@ class CptQueryModifier
 
 
     /**
+     * @deprecated $VID:$
      * @return \EE_Request_Handler
      */
     public function request()
     {
-        return $this->request;
+        return $this->request_handler;
     }
 
 
-    /**
-     * @param \EE_Request_Handler $request
-     */
-    protected function setRequest(\EE_Request_Handler $request)
-    {
-        $this->request = $request;
-    }
 
     // phpcs:disable PSR1.Methods.CamelCapsMethodName.NotCamelCaps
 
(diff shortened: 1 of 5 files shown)


joshfeck:
I've tested this with the Enfold theme (which has the same issue) and this PR's branch fixes the issue.

tn3rb:
I employed a much better way of getting class loaders for CPT strategies: https://github.com/eventespresso/event-espresso-core/pull/498/commits/31c63dc27e42338c24854b77a66e237b612a4c2c
which I'm pretty sure would be the only thing affecting `EE_CPT_Person_Strategy` as that class doesn't really do anything other than add some flags to the current `WP_Query` object.
So the duplicate filter issue doesn't really apply to that class since it appears that the filter is only used for loading the `EEH_People_View` helper

tn3rb:
@joshfeck 
you mentioned that there were others ways to reproduce this issue but did not provide any details, so I'm just going to pass this over to you for testing to go all freestyle on

joshfeck:
One of the other ways was install the Enfold theme and set its option to use the Grid template. I tested this branch and the issue reported here is resolved. I also checked the People add-on and found no regression.

This is ready for merging.

PR #502, opened 2018-06-05
Add transform class properties babel plugin

nerrad:

There's various reasons for adding this plugin:

- allows us to have field properties on js classes (useful for proptypes defined within the class itself)
- allows us to use class fields to bind callbacks for events.

For the second bullet point, this enables us to do something like this for event callbacks that will correctly bind `this` for usage in the callback.

```js
class SomeComponent extends Component {
    EventCallback = value  => this.props.setAttributes( { itemWithValue: value } );
    render() {
         return <SomeOtherComponent onChange={ this.EventCallback } />
    }
}
```

Without this babel plugin we can't do that because class fields (properties) is still not official spec.

## How has this been tested

- I merged this into the (not up to date) copy of the `EventAttendees` branch I have and implemented the callbacks using the new method.  It worked fine.

## Checklist

* [x] My code is tested.


nerrad:
Oh yea, I had to update the jest snapshot because of the intentional change in the component.

PR #504, opened 2018-06-05
Add DatetimeSelect component

tn3rb:
adds a select input component linked to the datetime state container

## How has this been tested
so far just in a branch


assets/src/components/selection/datetime-select/index.js
```diff
@@ -0,0 +1,231 @@
...
+ * @param {boolean} isLoading  Whether or not the selector should start in a
+ *                               loading state
+ * @return {Function}  A pure component function.
+ * @constructor
+ */
+class DatetimeSelect extends Component
+{
+	constructor() {
+		super( ...arguments );
+		this.placeHolder = this.placeHolder.bind( this );
```
nerrad: this is unnecessary because this isn't being passed down as a prop.

assets/src/components/selection/datetime-select/index.js
```diff
@@ -0,0 +1,231 @@
...
+		return (
+			<Fragment>
+				<Placeholder
+					icon="calendar"
+					label={ selectLabel }
+				>
+					{ ! isLoading ?
+						__(
+							'There are no datetimes to select from. You need' +
+							' to create a datetime first.',
```
nerrad: I know I've done this as well in the js and I _think_ its okay, however we'll have to make sure that the Webpack extraction plugin grabs these strings okay (I keep forgetting to double-check).

assets/src/components/selection/datetime-select/index.js
```diff
@@ -0,0 +1,231 @@
...
+							'event_espresso',
+						) :
+						<Spinner/>
```
nerrad: I _think_ there may be an .eslint error due to there being no space before the closing bracket.  In other words, this needs to be: `<Spinner />`

assets/src/components/selection/datetime-select/index.js
tn3rb: heh... i think my IDE did that on its own...

assets/src/components/selection/datetime-select/index.js
```diff
@@ -0,0 +1,228 @@
...
+ * @param {number} forEventId   ID for Event to retrieve datetimes from
+ * @param {boolean} isLoading  Whether or not the selector should start in a
+ *                               loading state
+ * @return {Function}  A pure component function.
+ * @constructor
+ */
+class DatetimeSelect extends Component {
+	constructor() {
+		super( ...arguments );
+		this.placeHolder = this.placeHolder.bind( this );
```
nerrad: This should still be removed please.

nerrad: Also when you remove this, there will be no need for the constructor.

tn3rb: that method uses `this` on its first line
```
	placeHolder() {
		const { isLoading, selectLabel } = this.props;
```

nerrad: I realize that.  But `placeHolder` is not passed through as a prop on an eventHandler so afaik that means it doesn't need `this` bound to that context.  You're simply calling it in the scope of the class its defined in.

nerrad: I also forgot to mention that I'd like to see some tests on this branch before merging.

tn3rb:
ya I kinda wanted to wait until after you made the changes to how these select components are going to operate because I really feel there is better use of my time than to write code that we know is going to be changed any day now

assets/src/components/selection/datetime-select/index.js
tn3rb: I'm not seeing any of the translations from this component getting picked up. Even the ones that are not concatenated.

Is there something I have to do other than `npm run build` to get them into `translation-map.json` ?

nerrad: hmm... just running `npm run build` _should_ be sufficient.   If they aren't getting picked up though then it's possible there's a yet undiscovered bug with the extractor I built.

PR #511, opened 2018-06-07
Change condition for offsite payment methods. (fixes issue #510)

### LINKED ISSUE #510: GDPR template - Billing information does not show any example text if you only use offsite payment methods.
To reproduce go to the WP privacy policy guide:

/wp-admin/tools.php?wp-privacy-policy-guide
(Dashboard -> Priviacy -> Checkout our guide)

Make sure you have the latest version of EE active and only an offsite payment method (offline payment methods don't matter so if you have invoice active leave it as is).

Look under billing information - http://take.ms/y1L9N

On that site I only have the PayPal Express payment method active.

If I also active PayPal Pro I get some info - http://take.ms/UuMUQ

Which make sense and covers the bases we need to, however with offsite payment methods I think it should be showing this - http://take.ms/RP3kP

Pebblo:
I think this fixes #510

### DIFF
--- a/core/domain/services/admin/privacy/policy/privacy_policy.template.php
+++ b/core/domain/services/admin/privacy/policy/privacy_policy.template.php
@@ -72,7 +72,7 @@
                 'event_espresso'
             ); ?></p>
     <?php } // IF OFFSITE PAYMENT METHOD ACTIVE
-    elseif (! empty($active_onsite_payment_methods)) { ?>
+    elseif (! empty($active_offsite_payment_methods)) { ?>
         <p><?php
             printf(
                 esc_html_x(


Pebblo:
To test go to: /wp-admin/tools.php?wp-privacy-policy-guide

In another tab, open up your payment methods and de-activate all online payment methods (both on site and off site).

Now refresh your privacy policy guide and look under EE's billing information section, it should not display anything.

Activate a couple of ONsite payment methods and refresh the policy page, the text should display something like:

> In order to process payments, we collect billing information on-site. Sensitive billing information is not stored on our server, but may be handled while in-transit to the payment processing server
>Please see the privacy policy of %1$s.

%1$s. should be a list of your active payment methods.

Now active some offsite payment methods (leave the onsites active).

CHeck those are listed with the **same** text above.

Now switch off the ONsite payment methods, leaving only off-site payment methods, now the text should be:

>Billing information is sent directly to the payment processor, and is not handled by our servers. Please see the privacy policy of %1$s.

%1$s. should be a list of your offsite payment methods.

joshfeck:
This is tested, works as expected, and can be merged.

PR #515, opened 2018-06-08
Implement react-select for our select input components

nerrad:

Right now in a few different places we are using `chosen` for a fancy select to help with auto-complete ux etc (Attendee Mover add-on, Barcode Scanner).  Although chosen is okayish, there are some challenges bringing it into react for usage (encountered while working on the Barcode Scanner).  Along with that, [chosen has a number of accessibility issues to address](https://github.com/harvesthq/chosen/issues/264).  In some research, I came across [react-select](https://deploy-preview-2289--react-select.netlify.com/home) and although version2 is in beta it introduces some great flexibility and while still not fully accessible (nothing is really yet), it is somewhat accessible and they are working on improving that.  Implementing this as a component for our select needs allows us flexibility in the various uses we may have for selects (single select, multi-select etc).  Brent and I chatted about this in slack and he was in agreement with this as well.

Along with the implementation of react-select, there's a number of other under the hood changes in this pull request.

### Abstracting out things specific to model related data interaction into the `data/model` folder.

Our existing selects had a lot of boilerplate code for building query's etc that will likely get used elsewhere beyond just selects.  It is also more fitting to have it connected with the `data` business logic for our modules.  So for the new selects, this has been abstracted more so that:

| file | purpose|
|----| ---------|
| data/model/base.js | This is a base function containing the logic for generating queryString from passed in arguments.
| data/model/model-names.js | this simply returns an array of modelNames from the data exposed in the view via eejs.
| model/event/index.js | It is expected that each model will have its own folder in this path containing functionality specific to the model.  In this particular file are a number of functions related to prepping things for a event REST query.  I fully expect this will be extracted even further into specific files eventually.

### Changing folder structure to more clearly reflect the component context.

Instead of `components/selection`, I've added the new selects in `components/form/select`.   This could likely even be further refined to be `components/form/select/modelName` for the individual models when necessary.

### Building a base `ModelSelect` component with an accompanying `withSelect` HOC wrapping it.

The `ModelSelect` component provides all the logic that exposes the `react-select` component for use.  Usage allows for indicating what model and modelentities are to be used for building the select as well as passing through the `react-select` component options to configure that component itself (via a `selectConfiguration` prop).

This then allows add-ons to utilize this as a base for their own composite components using this.

There's an example of a new `EventSelect` component that is composed of the `ModelSelect`.  It becomes much easier to construct any selects that utilize data from a model endpoint.

## What remains to do?

* [ ] Do some documentation of how to utilize `ModelSelect`.  I want to hold off doing this until its had a review (and merged into master).
* [ ] It's possible once we start implementing this we may want to play with the default styling of the select output a bit.  However, that's something we can do when we come to it.  `react-select` has a specific way it can be styled programatically so IF we decide to change the styles we'll want to investigate that.
* [ ] I think I may need to tweak how I'm using `derivedStateFromProps` in `ModelSelect` yet.  Basically I should only be changing state if there is a change detected for the incoming props with the current state ([see example here](https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html)).  I can probably utilize memoize a bit in the some of the helper functions as well.   This can be something done after this pull is merged and we are using it in blocks (so we can measure improvements).

## Important things to note for converting over older selects to this new component:

For the most part, the new `EventSelect` component can be used as a model (pun not intended) for how the conversion works.  However it's important to note that the signature for the event callbacks are specific to react-select so you need to keep that in mind when handling the event in the top-level component (i.e. `EventAttendees`).  Namely:

### `OnChange` callback

This differs from the WP `<SelectControl>` in that instead of the signature recieving back the id that was selected, there are two arguments passed into callback with the following schema:

| Argument | Possible types | Details |
|--------|---------------| ------ |
| selection | null \| undefined \|Array \<Object\> \| Object | The object is in the shape of `{ value: 10, label: 'Some label' }`
| action | Object <Enum> | An object with the shape of `{ action: <string> }` and the values are `select-option`, `deselect-option`, `remove-value`, `pop-value`, `set-value`, `clear`, `create-option` |

## How has this been tested

* [x] I've added new jest and enzyme tests (and modified any existing tests broken in the process) and ensured they passed.
* [x] I've tried this out on a branch based off of the EventAttendee block branch to ensure it works as expected.

## Checklist

* [x] My code is tested.


assets/src/components/form/select/model-selects/event.js
```diff
@@ -0,0 +1,62 @@
...
+export default class EventSelect extends Component {
+	state = {
+		modelName: 'event',
+	};
+
+	static defaultProps = {
+		selectConfiguration: {
+			loadingMessage: () => __( 'Retrieving Events.', 'event_espresso' ),
+			noOptionsMessage: () => __(
+				'There are no events available to select from.',
```
tn3rb: maybe "No Events" instead ? This is just too long for some use cases and should be simplified.

assets/src/data/model/datetime/index.js
```diff
@@ -0,0 +1,113 @@
...
+export const queryDataTypes = {
+	queryData: PropTypes.shape( {
+		limit: PropTypes.number,
+		orderBy: PropTypes.oneOf( [
+			'DTT_name',
+			'DTT_ID',
+			'start_date',
+			'end_date',
+		] ),
+		order: PropTypes.oneOf( [ 'asc', 'desc' ] ),
```
tn3rb: should we consider the values `'ASC'` and `'DESC'` as well 

and/or do one of the following:

- use a function to convert all incoming values for `order` to lowercase (or uppercase)
- use constants like `QUERY_ORDER_ASC` and `QUERY_ORDER_DESC` instead

assets/src/data/model/datetime/index.js
```diff
@@ -0,0 +1,113 @@
...
+
+/**
+ * Return a query string for use by a REST request given a set of queryData.
+ * @param { Object } queryData
+ * @return { string }  Returns the query string.
+ */
+export const getQueryString = ( queryData = {} ) => {
+	queryData = { ...defaultQueryData.queryData, ...queryData };
+	return baseGetQueryString( queryData, whereConditions, mapOrderBy );
+};
```
tn3rb: missing newline at end of file

assets/src/data/model/event/index.js
```diff
@@ -0,0 +1,115 @@
...
+		limit: PropTypes.number,
+		orderBy: PropTypes.oneOf( [
+			'EVT_name',
+			'EVT_ID',
+			'start_date',
+			'end_date',
+			'ticket_start',
+			'ticket_end',
+		] ),
+		order: PropTypes.oneOf( [ 'asc', 'desc' ] ),
```
tn3rb: same question as with the datetimes model... allow for caps, convert case, or use constants ?

assets/src/data/model/datetime/index.js
```diff
@@ -0,0 +1,113 @@
...
+		order: PropTypes.oneOf( [ 'asc', 'desc' ] ),
+		showExpired: PropTypes.bool,
+		month: PropTypes.month,
+	} ),
+};
+
+/**
+ * Default attributes for this model
+ * @type {{attributes: {limit: number, orderBy: string, order: string,
+ *   showExpired: boolean}}}
```
tn3rb: maybe this could be formatted better?

assets/src/data/model/event/index.js
```diff
@@ -0,0 +1,115 @@
...
+		showExpired: PropTypes.bool,
+		categorySlug: PropTypes.string,
+		month: PropTypes.month,
+	} ),
+};
+
+/**
+ * Default attributes for this model
+ * @type {{attributes: {limit: number, orderBy: string, order: string,
+ *   showExpired: boolean}}}
```
tn3rb: better formatting ?

assets/src/data/model/event/index.js
```diff
@@ -0,0 +1,115 @@
...
+		where.push( 'where[Datetime.DTT_EVT_start][]=' +
+			GREATER_AND_EQUAL +
+			'&where[Datetime.DTT_EVT_start][]=' +
+			moment().month( month ).startOf( 'month' ).local().format() );
+		where.push( 'where[Datetime.DTT_EVT_end][]=' +
+			LESS_AND_EQUAL +
+			'&where[Datetime.DTT_EVT_end][]=' +
+			moment().month( month ).endOf( 'month' ).local().format() );
+	}
+	return where.join( '&' );
```
tn3rb: If a component or block requires a where condition that is not here, then in your opinion should we:

- add the new condition and corresponding prop to this method and everywhere else it is required
- override this method (or another) and simply append the additional where condition to the query string

assets/src/components/form/select/build-options.js
```diff
@@ -0,0 +1,72 @@
...
+ * with label and value keys.
+ */
+const buildOptions = (
+	entities,
+	modelName,
+	addAllOptionLabel = '',
+	map = DEFAULT_MODEL_OPTIONS_MAP,
+) => {
+	const MAP_FOR_MODEL = map[ modelName ] ? map[ modelName ] : false;
+	const generatedOptions = entities && modelName && MAP_FOR_MODEL ?
```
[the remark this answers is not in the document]
nerrad: ya agreed, I'll fix that.

assets/src/components/form/select/build-options.js
```diff
@@ -0,0 +1,72 @@
...
+			}
+			return options;
+		}, [] ) :
+		[];
+	if ( addAllOptionLabel !== '' ) {
+		generatedOptions.unshift( {
+			label: addAllOptionLabel,
+			value: OPTION_SELECT_ALL,
+		} );
+	}
```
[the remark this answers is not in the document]
nerrad: agreed. fixing.

assets/src/components/form/select/default-select-configuration.js
```diff
@@ -0,0 +1,111 @@
...
+		singleValue: PropTypes.func,
+		valueContainer: PropTypes.func,
+	} ),
+	tabIndex: PropTypes.string,
+	tabSelectsValue: PropTypes.bool,
+	value: PropTypes.oneOfType( [
+		PropTypes.object,
+		PropTypes.array,
+	] ),
+};
```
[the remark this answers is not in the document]
nerrad: The properties this comment is attached to are all props used by the react-select component.  Our `<ModelSelect />` component just passes through any set props specifically for the react-select component via the `props.selectConfiguration` (which then gets added to the `state` in`ModelSelect`).  So, adding `data-*` attributes (if necessary) would need to be implemented via the `react-select` component.  If we encounter a place where we absolutely need them, we can likely use the [components prop](https://deploy-preview-2289--react-select.netlify.com/components) to override one of the react select components to add the data attributes.

assets/src/components/form/select/model-select.js
```diff
@@ -0,0 +1,170 @@
...
+			value: ModelSelect.getOptionObjectForValue(
+				selectConfiguration.defaultValue, options
+			),
+		};
+		return {
+			...REACT_SELECT_DEFAULTS,
+			...selectConfiguration,
+			...updated,
+		};
+	}
```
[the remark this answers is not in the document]
nerrad: I originally intended to link to the "You probably don't need derived state" in my original description for this pull but linked to a different (but still related) post by mistake.  For context, here's what I wrote in the description for the pull:

> I think I may need to tweak how I'm using derivedStateFromProps in ModelSelect yet. Basically I should only be changing state if there is a change detected for the incoming props with the current state (see example here). I can probably utilize memoize a bit in the some of the helper functions as well. This can be something done after this pull is merged and we are using it in blocks (so we can measure improvements).

We talked about this in slack but to summarize:

- I'm aware of the potential issues.
- I'm not convinced (yet) that we're doing things wrong (or in a way that is problematic). 
- I fully intend on investigating further to see if we can do what's needed without using `getDerivedStateFromProps` but think this is something that will be easier to investigate in actual implementations of this component.  So I'd prefer to get this merged in then create another pull to test with and try different potential solutions.

assets/src/components/form/select/model-selects/datetime.js
```diff
@@ -0,0 +1,64 @@
...
+export default class DatetimeSelect extends Component {
+	state = {
+		modelName: 'datetime',
+	};
+
+	static defaultProps = {
+		selectConfiguration: {
+			loadingMessage: () => __( 'Retrieving Datetimes.', 'event_espresso' ),
+			noOptionsMessage: () => __(
+				'There are no datetimes available to select from.',
```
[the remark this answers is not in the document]
nerrad: From my testing it doesn't expand the width of the input, but I like it more concise anyways so will fix.

assets/src/components/form/select/model-selects/event.js
nerrad: Am doing.

assets/src/data/model/datetime/index.js
nerrad: ya will do.

assets/src/data/model/datetime/index.js
nerrad: ya will do.  This was done automatically by PhpStorm but doesn't look like there is a way to fix the auto-formatting for this so manual fixes will be necessary.

assets/src/data/model/event/index.js
nerrad: fixed

assets/src/data/model/event/index.js
nerrad: One of the reasons I extracted these into their own files is so that they could be considered the "authority" for parsing where conditions for that model.  So for now, the option would be "add the new condition and corresponding prop to this method and everywhere else it is required".  

However, as we see patterns emerge, we may find that we can autogenerate some of these a bit (or at least start with a common base).

Issue #313, opened 2018-04-18
Add Support for PayPal Express Smart Payment Buttons


garthkoyle:
It is a new button that support alterative payment methods, like Venmo. It would be adding Java Script to the button code to make it dynamic.  It would detect if the customer has Venmo and it shows as an option during  checkout. PayPal is going to add more payment methods in the future too, like Apple Pay and Amazon Pay. 

Docs PayPal provided:
[Muse_and_SPB_Best_Practices.pptx](https://github.com/eventespresso/event-espresso-core/files/1924948/Muse_and_SPB_Best_Practices.pptx)
[Smart Payment Buttons Value Prop - 1.16.18.pptx](https://github.com/eventespresso/event-espresso-core/files/1924949/Smart.Payment.Buttons.Value.Prop.-.1.16.18.pptx)
[Solution Design Document Express Checkout using NVP.pdf](https://github.com/eventespresso/event-espresso-core/files/1924957/Solution.Design.Document.Express.Checkout.using.NVP.pdf)

mnelson4:
So there are really 3 different features mentioned in those documents.

1. PayPal Marketing Solutions (Muse). "...is a snippet of JS that a merchant has on their site that gathers insights and provides marketing messaging to consumers from PayPal." (Their Woocommerce plugin demo shows it pretty clearly: https://www.youtube.com/watch?v=tsSHL8NarAE&feature=youtu.be)
2. Smart Payment Button. Basically, the user needs to sign into PayPal first (preferably before we request registration information, so we can use their PayPal profile data to auto-populate some of the registration form), and then we can request PayPal generate a button image showing the payment options applicable to the current user. I think this is the feature that allows users to pay with Venmo etc
3. PayPal Express with checkout.js. Provides a payment experience basically identical to Stripe.


From the Express Checkout using NVP PDF
> You have to include the PayPal checkout.js on the page hosting the PayPal button. Make sure you do not
download the JS and host it from your server.

That conflicts with the WP.org plugin guidelines (we could break the WP plugin guidelines a little with this, or have this not be part of the EE core decaf, or try downloading the JS and hope to update it whenever PayPal updates the script)

garthkoyle:
We were contacted by PayPal for option 2.

shorrocks:
Hi Garth,

The goal is adding Smart Payment buttons because it is dynamic, this means that alternative payments will appear for the buyers during checkout.

This product pairs well with a few other things, which include:

[...]

PR #517, opened 2018-06-11
Add filters required for #313

mnelson4:
Adds paypal smart buttons. See https://github.com/eventespresso/event-espresso-core/issues/313


mnelson4:
actually, we're going to be making this an add-on. At very least initially, so it can be developed independently of the rest of EE and not be placed in front of ALL our users until it's had some more small-scale user testing. We'll re-evaluate whether this should go into core afterwards

mnelson4:
Seeing how PayPal Smart Buttons will be a separate add-on, I'm just going to use this for the filters I needed for that. (Yes, I may end up creating separate PRs for each, but I'm first making sure it will meet the add-on's needs)

mnelson4:
This is required for https://github.com/eventespresso/eea-paypal-smart-buttons/pull/1. 
## What changed
So, this just adds a few filters, and refactored EE_Radio_Button_Display_Strategy to use the less-old method `_attributes_string()` which makes the attributes more maliable, standardized, and filtered... although it turned out that didn't provide what I needed for the new gateway integration, it seemed like a good thing to leave anyway. What's more that code is unit tested so it helped protect from regression there.

## Testing
* [ ] Verify radio buttons work the same as ever. Eg, the payment method select radio-box input in SPCO should still work. 
* [ ] Custom questions using radio buttons should still work. Test having required and non-required radio button questions. Those should all work the same as before.

sethshoultes:
Testing notes check out.

@tn3rb can you please review this so I can get it merged to core?

PR #522, opened 2018-06-12
Documentation Updates for JS related work

nerrad:
## Problem this Pull Request solves

This pull adds additional documentation for latest work done in the GB Epic.  There are also a few minor tweaks to code related to issues spotted while documenting that will be added to this summary when submitted for review.

Docs added:

- `<AvatarImage />`
- `<ModelSelect />` and `<ModelEnhancedSelect />` components
- `<DatetimeSelect />`
- `<EventSelect />`
- various table of contents 

Code tweaked:

- The `<ModelSelect />` component was given a default function for the `getQueryString` callback.

PR #524, opened 2018-06-12
Fix event model query defaults

tn3rb:
same issue as described in #523 but with regards to the event model

## How has this been tested
unit tests


assets/src/data/model/event/index.js
```diff
@@ -87,29 +87,36 @@ export const mapOrderBy = ( orderBy ) => {
...
+} ) => {
 	const where = [];
 	const GREATER_AND_EQUAL = encodeURIComponent( '>=' );
 	const LESS_AND_EQUAL = encodeURIComponent( '<=' );
 
 	if ( ! showExpired ) {
-		where.push( 'where[Datetime.DTT_EVT_end**expired][]=>&where[Datetime.DTT_EVT_end**expired][]=' +
-			nowDateAndTime.local().format() );
+		where.push(
+			'where[Datetime.DTT_EVT_end**expired][]=>' +
```
nerrad: huh, I think I've missed this in multiple places (i.e. will affect our datetime and ticket implementation as well).  This line results in (with the other two):

`where[Datetime.DTT_EVT_end**expired][]=>&where[Datetime.DTT_EVT_end**expired[]='{some-date-string}'`

It doesn't makes sense?  I think the first part which is `where[Datetime.DTT_EVT_end**expired][]=>&` shouldn't even be present right?

nerrad: actually no, it's correct.  But we should be using `GREATER_AND_EQUAL` instead of `=>` in that line.

assets/src/data/model/event/test/index.js
```diff
@@ -24,7 +26,10 @@ describe( 'mapOrderBy()', () => {
 
 describe( 'whereConditions()', () => {
 	it( 'returns expected default for empty object passed in', () => {
-		expect( whereConditions( {} ) ).toEqual( '' );
+		expect( whereConditions( {} ) ).toEqual(
+			'where[Datetime.DTT_EVT_end**expired][]=>' +
```
nerrad: This will need fixed to be `GREATER_AND_EQUAL` instead of `=>` if the same fix is done in the method being tested.

assets/src/data/model/event/test/index.js
```diff
@@ -46,17 +51,12 @@ describe( 'whereConditions()', () => {
 		};
 		expect( whereConditions( testObject ) ).toEqual(
 			'where[Datetime.DTT_EVT_end**expired][]=>' +
```
nerrad: this will need changed from `=>` to `GREATER_AND_EQUAL` as well.

assets/src/data/model/event/test/index.js
```diff
@@ -66,8 +66,7 @@ describe( 'getQueryString', () => {
 		expect( getQueryString() ).toEqual(
 			'limit=100&order=DESC&order_by=Datetime.DTT_EVT_start' +
 			'&where[Datetime.DTT_EVT_end**expired][]=>' +
```
nerrad: This will need changed from `=>` to `GREATER_AND_EQUAL` as well.

assets/src/data/model/event/index.js
tn3rb: ya I never grokked that, but assumed it was something funny with the way the REST API worked

so... the correct line should be:

```
'where[Datetime.DTT_EVT_end**expired][]=' + GREATER_AND_EQUAL  +
```

or

```
'where[Datetime.DTT_EVT_end**expired][]' + GREATER_AND_EQUAL  +
```

I would think the former would be correct ( ie: replace `=>` with `=' + GREATER_AND_EQUAL` ),
but it sounds like you are saying the latter  ( ie: replace `=>` with GREATER_AND_EQUAL` ),
which makes no sense at all to me.

nerrad: OH, actually maybe its supposed to be `=` + `>` only.  In which case it'd probably be good to add `GREATER_THAN` and `LESS_THAN` constants that encode the equal and less than characters.  Since multiple models will likely be using these constants, it'd probably also be good to move them into the `data/model/base.js` module and they can just be included where needed?

tn3rb: kk no THAT makes sense and I totally agree.
will do that right now

tn3rb:
will add documentation after #522 is merged because of structural changes to docs

assets/src/data/model/datetime/index.js
```diff
@@ -89,22 +91,24 @@ export const whereConditions = ( {
...
+			'where[DTT_EVT_end**expired][]=' + GREATER_THAN +
+			'&where[DTT_EVT_end**expired][]=' +
+			nowDateAndTime.local().format()
+		);
 	}
 	if ( month && month !== 'none' ) {
-		where.push( 'where[DTT_EVT_start][]=' +
-			GREATER_AND_EQUAL +
+		where.push(
+			'where[DTT_EVT_start][]=' + GREATER_THAN +
```
nerrad: Why did you change this from `GREATER_AND_EQUAL` to `GREATER_THAN`?  I think it should remain as it was as we're wanting to _include_ the start of the month right?

assets/src/data/model/datetime/index.js
```diff
@@ -89,22 +91,24 @@ export const whereConditions = ( {
...
 			'&where[DTT_EVT_start][]=' +
-			moment().month( month ).startOf( 'month' ).local().format() );
-		where.push( 'where[DTT_EVT_end][]=' +
-			LESS_AND_EQUAL +
+			moment().month( month ).startOf( 'month' ).local().format()
+		);
+		where.push(
+			'where[DTT_EVT_end][]=' + LESS_THAN +
 			'&where[DTT_EVT_end][]=' +
```
nerrad: I think this should remain `LESS_AND_EQUAL` here because we want to include the last day of the month in the query right?

assets/src/data/model/datetime/test/index.js
```diff
@@ -40,19 +41,17 @@ describe( 'whereConditions()', () => {
...
-			'where[DTT_EVT_end**expired][]=>' +
+			'where[DTT_EVT_end**expired][]=' + GREATER_THAN +
 			'&where[DTT_EVT_end**expired][]=' + expectedNow +
-			'&where[DTT_EVT_start][]=' + GREATER_AND_EQUAL +
+			'&where[DTT_EVT_start][]=' + GREATER_THAN +
 			'&where[DTT_EVT_start][]=' + expectedStartofDate +
-			'&where[DTT_EVT_end][]=' + LESS_AND_EQUAL +
+			'&where[DTT_EVT_end][]=' + LESS_THAN +
 			'&where[DTT_EVT_end][]=' + expectedEndofDate +
 			'&where[Event.EVT_ID]=' + 20
```
nerrad: These tests will need modified if my other comments are acted on.

assets/src/data/model/event/index.js
```diff
@@ -87,31 +89,40 @@ export const mapOrderBy = ( orderBy ) => {
...
-		where.push( 'where[Term_Relationship.Term_Taxonomy.Term.slug]=' + categorySlug );
+		where.push(
+			'where[Term_Relationship.Term_Taxonomy.Term.slug]=' + categorySlug
+		);
 	}
 	if ( month && month !== 'none' ) {
-		where.push( 'where[Datetime.DTT_EVT_start][]=' +
-			GREATER_AND_EQUAL +
+		where.push(
+			'where[Datetime.DTT_EVT_start][]=' + GREATER_THAN +
```
nerrad: As noted elsewhere, this should remain as `GREATER_AND_EQUAL` and not be changed.

assets/src/data/model/event/index.js
```diff
@@ -87,31 +89,40 @@ export const mapOrderBy = ( orderBy ) => {
...
 			'&where[Datetime.DTT_EVT_start][]=' +
-			moment().month( month ).startOf( 'month' ).local().format() );
-		where.push( 'where[Datetime.DTT_EVT_end][]=' +
-			LESS_AND_EQUAL +
+			moment().month( month ).startOf( 'month' ).local().format()
+		);
+		where.push(
+			'where[Datetime.DTT_EVT_end][]=' + LESS_THAN +
```
nerrad: As noted elsewhere, this should remain `LESS_AND_EQUAL` and not be changed.

assets/src/data/model/event/test/index.js
```diff
@@ -37,26 +44,19 @@ describe( 'whereConditions()', () => {
...
-			'&where[Datetime.DTT_EVT_start][]=' +
-			expectedStartofDate +
-			'&where[Datetime.DTT_EVT_end][]=' +
-			LESS_AND_EQUAL +
-			'&where[Datetime.DTT_EVT_end][]=' +
-			expectedEndofDate,
+			'&where[Datetime.DTT_EVT_start][]=' + GREATER_THAN +
+			'&where[Datetime.DTT_EVT_start][]=' + expectedStartofDate +
+			'&where[Datetime.DTT_EVT_end][]=' + LESS_THAN +
+			'&where[Datetime.DTT_EVT_end][]=' + expectedEndofDate,
```
nerrad: These will need modified if my earlier comments are acted on.

assets/src/data/model/ticket/index.js
```diff
@@ -91,24 +93,21 @@ export const whereConditions = ( {
...
 		where.push(
-			'where[TKT_start_date][]=' + GREATER_AND_EQUAL +
+			'where[TKT_start_date][]=' + GREATER_THAN +
 			'&where[TKT_start_date][]=' +
 			moment().month( month ).startOf( 'month' ).local().format()
 		);
 		where.push(
-			'where[TKT_end_date][]=' + LESS_AND_EQUAL +
+			'where[TKT_end_date][]=' + LESS_THAN +
 			'&where[TKT_end_date][]=' +
```
nerrad: as noted earlier, you need to retain `GREATER_AND_EQUAL` and `LESS_AND_EQUAL` instead of changing those.  The only thing needing implemented is the `GREATER_THAN` to replace the `>` in the `showExpired` condition.  For months, we must include the starting date and the ending date in the query otherwise a query for the month of June would exclude events falling on June 1 or June 30.

assets/src/data/model/ticket/test/index.js
```diff
@@ -39,20 +40,18 @@ describe( 'whereConditions()', () => {
...
 		expect( whereConditions( testObject ) ).toEqual(
-			'where[TKT_end_date**expired][]=>' +
+			'where[TKT_end_date**expired][]=' + GREATER_THAN +
 			'&where[TKT_end_date**expired][]=' + expectedNow +
-			'&where[TKT_start_date][]=' + GREATER_AND_EQUAL +
+			'&where[TKT_start_date][]=' + GREATER_THAN +
 			'&where[TKT_start_date][]=' + expectedStartofDate +
-			'&where[TKT_end_date][]=' + LESS_AND_EQUAL +
+			'&where[TKT_end_date][]=' + LESS_THAN +
 			'&where[TKT_end_date][]=' + expectedEndofDate +
```
nerrad: this will need adjusted if my earlier comments are acted on.

Issue #410, opened 2018-05-08
Allow custom sortable columns to be used in list table order_by param?


Pebblo:
Reported here: https://eventespresso.com/topic/sorting-by-custom-column-in-events-registration-table/

In short if you add a custom column to the registrations list, then use this filter:

```$_sortable = apply_filters("FHEE_manage_{$this->screen->id}_sortable_columns", $_sortable, $this->_screen);```

To enable sorting on that column, although you can click to sort, and if you've done it correctly the URL will change to short something like `&orderby=REG_code&order=desc` the orderby parameter doesn't actually get used and defaults to REG_date

See https://github.com/eventespresso/event-espresso-core/blob/master/admin_pages/registrations/Registrations_Admin_Page.core.php#L1468

I switched this out so that `$orderby_field` was set to whatever was passed in `_req_data['orderby']` and then ran through the switch (setting it to whatever the value is if it doesn't match) and it worked as expected then but should this be done through a filter?

mnelson4:
IMO I think ideally, the default switch case would double-check they're requesting to filter by a valid column name, and if so, use it for ordering. We probably shouldn't order by just whatever is passed in, as the current user might not have permission to see that column they're requesting to filter by (so that would be a bit of a security issue)

Pebblo:
I think the work done in #525 covers the original request in this ticket, however, it doesn't filter the sortable columns based on what the user has access too as Mike suggested.

So @mnelson4 do you still think that is something that should be done or can this be closed?

mnelson4:
Hmmm, ya. That is technically still an issue. We shouldn't accept just any user input, we should have a filter that accepts the user's input, which plugin code can use (and validate the user's input), rather than just accepting any input user's submit.

PR #525, opened 2018-06-12
Filter the order_by array returned by _get_orderby_for_registrations_query()

Pebblo:
This pull request just adds a filter to:

https://github.com/eventespresso/event-espresso-core/blob/f94d697792594433aa83a74d22547b2a41430c4e/admin_pages/registrations/Registrations_Admin_Page.core.php#L1468

See #410


Pebblo:
@mnelson4 I've assigned this to you for review simply because you commented on the issue, if you think this should go to someone else please assign to them.

admin_pages/registrations/Registrations_Admin_Page.core.php
```diff
@@ -1506,7 +1506,13 @@ protected function _get_orderby_for_registrations_query()
         if (empty($orderby['REG_ID'])) {
             $orderby['REG_ID'] = $order;
         }
-        return array('order_by' => $orderby);
+        return apply_filters(
```
nerrad: If this is intended to change the order_by value only, it'd be better to just filter the `$orderby` value and not the entire value.

Pebblo:
I think I may have done this slightly different from your comments?

Something I did figure out is the filter is pretty much necessary as far as I can tell. If I add a couple of Registration question answers to the table, then wanted to order by either of them the models need 'Answer.ANS_value' or similar passed to them, you can't set the columns to both have that as the name so you have to filter for the column names and then set it.

Sorting by specific reg answers if you have more than one won't work but I think to accommodate for that we'd need further refactoring, right?

nerrad:
You still have some issues with linting that need fixed before this can be merged (and fixes will invalidate my approval so it will need reapproved after fixes).

nerrad:
@mnelson4 I just want to clarify why I reviewed this ticket because I'm concerned you "but whatever" statement indicates frustration.  @Pebblo pinged me in slack this morning asking if I could review it.  I just want to make sure you don't think I jumped in and reviewed this on my own initiative.

Pebblo:
To test, go to Event Espresso -> Registrations.

Sort the registrations by various columns ASC and DESC and confirm they sort as expected.

Then add this to your site: https://gist.github.com/Pebblo/88f2a0a9213c716e4886a249e7709245

That adds a sortable 'Custom Reg Question' column to your table, then if that's used to sort it sets the order by to 'Answer.ANS_value' rather than the column name (which wouldn't work). You'll need to change [line 29](https://gist.github.com/Pebblo/88f2a0a9213c716e4886a249e7709245#file-example-php-L29) to match an ID for a question on your site that has values in registrations.

Confirm that works as expected when sorting.

Pebblo:
Sorry @nerrad can you review this last change. It accommodates for using the TXN Date column sorting and I've then used that same value for the default if no order by is set.

mnelson4:
> @mnelson4 I just want to clarify why I reviewed this pull request because I'm concerned your "but whatever" statement indicates frustration. @Pebblo pinged me in slack this morning asking if I could review it. I just want to make sure you don't think I jumped in and reviewed this on my own initiative.

You usually do have the best insights on a topic, even when a ticket isn't assigned to you or we didn't ask your opinion (which sometimes is annoying 😛). But you should probably keep doing it anyway, because it's good stuff. So even if you did review this on your own initiative, that's fine.

But to be clear, I didn't think my review was necessary (you probably did a fine job), actually very helpful (I want to avoid "too many cooks spoil the broth") or the best use of my time.

But thanks for checking!

sethshoultes:
Testing notes:
- before adding the filter, I tested the filters in the registrations overview and they work fine
- added the filter and retested filtering, which seems to work fine
- testing filtering by the custom question and that works as well

PR #526, opened 2018-06-13
Add form StatusSelect component

tn3rb:
Adds a component for generating a select input for retrieving results from the EE status model

## How has this been tested
included unit tests

## Todo

* [x] documentation (will add after #522 is merged)


assets/src/components/form/select/model-selects/status.js
```diff
@@ -0,0 +1,60 @@
...
+
+	static defaultProps = {
+		selectConfiguration: {
+			loadingMessage: () => __( 'Retrieving Statuses.', 'event_espresso' ),
+			noOptionsMessage: () => __( 'No Statuses.', 'event_espresso' ),
+			placeholder: __( 'Select Status...', 'event_espresso' ),
+		},
+		...statusModel.defaultQueryData,
+		getQueryString: statusModel.getQueryString,
+		selectedType: 'registration',
```
nerrad: we can leave this until we see a pattern emerge if you want, but I wonder if `data/model/status/index.js` should include a set of constants for the different status "types" and then you just use the appropriate constant in here (i.e. `STATUS_TYPE_REGISTRATION`)?

assets/src/components/form/select/model-selects/status.js
```diff
@@ -0,0 +1,60 @@
...
+		selectLabel: __( 'Select Status', 'event_espresso' ),
+		addAllOptionLabel: __( 'All Statuses', 'event_espresso' ),
+	};
+
+	static propTypes = {
+		...statusModel.queryDataTypes,
+		selectedType: PropTypes.string.isRequired,
```
nerrad: Since this has a default set, the `isRequired` usage here is redundant.  If you want to use `isRequired` then remove the default (or leave the default and remove `isRequired`).

assets/src/data/eventespresso/lists/selectors.js
```diff
@@ -131,23 +131,23 @@ export function isRequestingTickets( state, queryString ) {
...
  *
  * @param {Object} state  Data state.
+ * @param {string} queryString The query string for the request
  * @return {Array} An array of event entities for the given model and query.
  */
-export function getRegistrationStatuses( state ) {
-	return getItems( state, 'status', 'where[STS_type]=registration' );
+export function getStatuses( state, queryString ) {
+	return getItems( state, 'status', queryString );
 }
```
nerrad: ooo yah I like this much better.

assets/src/components/form/select/model-selects/status.js
tn3rb: I like that idea. 
The only downside is that any new statuses added to the db need to be replicated in code, but that's kinda the case anyways, which was never my intent with statuses, although that's how they panned out regardless.

assets/src/components/form/select/model-selects/status.js
tn3rb: hmmm... think I'll remove the default then and leave the component more abstracted

assets/src/components/form/select/model-selects/status.js
```diff
@@ -27,14 +27,20 @@ export default class StatusSelect extends Component {
...
 		...statusModel.queryDataTypes,
-		selectedType: PropTypes.string.isRequired,
+		statusType: PropTypes.oneOf( [
+			statusModel.STATUS_TYPE_EMAIL,
+			statusModel.STATUS_TYPE_EVENT,
+			statusModel.STATUS_TYPE_MESSAGE,
+			statusModel.STATUS_TYPE_PAYMENT,
+			statusModel.STATUS_TYPE_REGISTRATION,
+			statusModel.STATUS_TYPE_TRANSACTION,
+		] ),
```
nerrad: You probably should also append `isRequired` to this as well UNLESS it's okay if its not included.

nerrad:
I also noticed that there are no tests for `data/model/status/index.js`, can you add them please?

assets/src/components/form/select/model-selects/status.js
tn3rb: ohh ya good catch... will do

tn3rb:
found some additional mistakes in the DatetimeSelect docs and also added docs for the TicketSelect component

PR #528, opened 2018-06-13
Fix duplicate asset registration when do_action('admin_enqueue_scripts') is called multiple times on the same admin page Fixes #501

### LINKED ISSUE #501: Avoid fatal errors related to "The namespace for this manifest file has already been registered"
This fatal error has been reported in 3 support topics:
https://eventespresso.com/topic/nav-menu-php-wont-load/
https://eventespresso.com/topic/white-screen-error-on-wp-admin-nav-menus-php/
https://eventespresso.com/topic/problems-when-people-try-to-register-an-event-on-there-mobile-phone/#post-270951

In each case the error happens on a specific admin page related to a specific plugin. So far these are the plugins known to be affected by this bug:

https://wordpress.org/plugins/megamenu/
https://wordpress.org/plugins/wp-megamenu/
https://wordpress.org/plugins/child-theme-configurator/

The first two are affected when viewing the WP > Menu Manager
The third is affected when using the Analyze tool on the Tool > Child Themes pag

tn3rb:
Fixes #501 
What's happening is that `do_action('admin_enqueue_scripts')` is triggered multiple times on some admin pages in various locations. 
This can result in our asset registry hooking into that action multiple times and attempting to run it's logic repeatedly.

I considered adding logic to the asset Registry for determining if `registerManifestFiles()` and/or other methods had already run, but was concerned that that may result in some assets not getting registered if their corresponding `AssetManager` failed to load in time.
So the reason for not proceeding with registration based on the `$hook_suffix` passed to the `admin_enqueue_scripts` action was because some of those actions are happening earlier than the typical `admin_enqueue_scripts` hookpoint you would expect for an admin page, and I was concerned that they would occur before all of the `AssetManager`s had a chance to load.
ie: if we only register assets once per request, and that happens during the `admin_enqueue_scripts` action where `$hook_suffix === 'nav-menus.php'`, but that occurs before an `AssetManager` is loaded, then those assets won't get registered.

## How has this been tested
confirmed fix using the Max Mega Menu plugin as described in 501 and via unit tests (no new ones added)

### DIFF
--- a/core/domain/DomainBase.php
+++ b/core/domain/DomainBase.php
@@ -160,6 +160,9 @@ abstract class DomainBase implements DomainInterface
      */
     private function setAssetNamespace()
     {
-        $this->asset_namespace = sanitize_key($this->plugin_basename);
+        $this->asset_namespace = sanitize_key(
+            // convert directory separators to dashes and remove file extension
+            str_replace(array('/', '.php'), array('-', ''), $this->plugin_basename)
+        );
     }
 }
--- a/core/services/assets/AssetManager.php
+++ b/core/services/assets/AssetManager.php
@@ -50,9 +50,25 @@ abstract class AssetManager implements AssetManagerInterface
         $this->assets = $assets;
         $this->registry = $registry;
         add_action('wp_enqueue_scripts', array($this, 'addManifestFile'), 0);
-        add_action('admin_enqueue_scripts', array($this, 'addManifestFile'), 0);
+        add_action('admin_enqueue_scripts', array($this, 'adminAddManifestFile'), 0);
         add_action('wp_enqueue_scripts', array($this, 'addAssets'), 2);
-        add_action('admin_enqueue_scripts', array($this, 'addAssets'), 2);
+        add_action('admin_enqueue_scripts', array($this, 'adminAddAssets'), 2);
+    }
+
+
+    /**
+     * @param string $hook_suffix
+     * @since $VID:$
+     * @throws DuplicateCollectionIdentifierException
+     * @throws InvalidDataTypeException
+     * @throws InvalidEntityException
+     */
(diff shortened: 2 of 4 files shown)


tn3rb:
> I wonder if it might be better to track when a manifest file is registered and ONLY allow it to be registered once per request?

that's essentially what is happening now:

```php
if (isset($this->manifest_data[ $namespace ])) {
    if(! $this->debug()){
        return;
    }
    throw new InvalidArgumentException(
```

except an exception gets thrown if `EE_DEBUG` is set to true so developers can be made aware if something they are doing is resulting in that happening

nerrad:
In that case, I don't think we need to have `dontRegisterForLocation` at all.  But if you want to leave it in, it's your call.  I'll mark this as approved for now.

tn3rb:
refactored to only bypass exceptions if debugging is not on

tn3rb:
### testing

- first checkout master branch
- download and activate the "Max Mega Menu" plugin
- visit the WP admin **Appearance** > **Menus** admin page
- [ ] confirm that the "The namespace for this manifest file has already been registered" exception is thrown
- now checkout this branch (BUG/duplicate-asset-registration) and refresh the page
- [ ] confirm that the exception is no longer thrown
- visit the WP admin **Plugins** admin page
- deactivate EE
- [ ] confirm that the **ExitModal** survey appears
- reactivate EE and visit the Event Editor for any event
- [ ] confirm that advanced settings for tickets and datetimes can be viewed by clicking on the gear icons
- feel free to experiment with pages that utilize JavaScript in this branch and attempt to find situations where JavaScript fails to load or function as expected

joshfeck:
Tested and verified the fix, and found no regressions with JS-related features of Event Espresso or its add-ons. This is ready for merging.

Issue #416, opened 2018-05-09
Fix UXIP option not saving without changing something else on the page


sethshoultes:
Per @Pebblo from Codebase:
https://events.codebasehq.com/projects/event-espresso/tickets/11503

Found here: https://eventespresso.com/topic/user-experience-improvement-program-uxip/

The user reported that the UXIP setting wouldn't save and always defaulted to Yes.

I tested this on one of my test sites and sure enough, when I changed the option to No and saved, it displayed Yes again on reload.

However, if I changed the Org name at the same time (just added a 1 to the name) and saved, it saved the correct option.

After doing the above, the option now saves without having to change anything else on the page.

I don't have any warnings, notices etc for the site from testing.

sethshoultes:
@nerrad can you please take a look?

nerrad:
There were also some [comments made in our slack](https://eventespresso.slack.com/archives/C080L8G68/p1528810769000423) about changing this to default to opt-in vs the current default of opt-out.  Did you want me to change that as well for this ticket?

joshfeck:
p2'ed:
http://p2.eventespresso.com/?p=5141

sethshoultes:
@nerrad yes, please update that to be opt-in, instead of opt-out by default.

joshfeck:
per the discussion in Slack, we're going to try to make this a single checkbox where you check the box to opt-in.

nerrad:
closing in favor of #530

PR #530, opened 2018-06-14
Update Your Organization Settings Page to new form system and UXIP option fixes and improvements

nerrad:

Ultimately this was started to fix the issue reported in #416 but as a part of that I also converted the `Your Organization Settings` admin page to use the new form system (as our process currently is to transition old code to new systems when we work on it).  As a summary, the following is accomplished in this pull:

- Convert the old form for Your Organization Settings page to new form system.
- ~Create and implement a new `EE_License_Key_Display_Strategy` which can be used with `EE_Text_Input` to display a license key valid indicator next to the text input.~ Turned out this display strategy was not needed.
- Change the default for UXIP to **opt-out** instead of **opt-in** and adjust all the text/language regarding that as necessary.

As a side effect of the work done on converting the form to the new form system the following bugs are fixed:

- uxip option set persists
- logo image upload is using the new media uploader and consistently works as expected (where before if `File Url` was not selected - its not the default - it would not insert the image as expected).

### Testing an existing site with set options on master and switching to this branch.

* [x] Verified that existing options on Your Organization Settings page were retained.

### Testing a fresh site

* [x] Verify UXIP defaults to opt-out.
* [x] Verify UXIP admin notice text is correct and that it automatically dismisses on saving the "Your Organization Settings" page.

### All contexts

* [x] Verify various states of the site license key indicator is correct (empty field, valid key, invalid key) when saved.
* [x] Verify all options on settings page persist when saved
* [x] Verify uploading an image for the logo works and persists across requests.

## Checklist

* [x] My code is tested.


core/libraries/form_sections/strategies/display/EE_License_Key_Display_Strategy.php
```diff
@@ -0,0 +1,33 @@
...
+    }
+
+    public function display()
+    {
+        $input = parent::display();
+        $verified_class = $this->verified ? 'ee-icon-color-ee-green' : 'ee-icon-color-ee-red';
+        $input .= '<span class="dashicons dashicons-admin-network ' . $verified_class . ' ee-icon-size-20"></span>';
+        return $input;
+    }
+}
```
tn3rb: I don't think this extra class is necessary simply because it has no potential reusability beyond this one use case

I think a better more reusable solution would be to add a filter in one of the following places (all of which would allow you to do what you need):

- `EE_Admin_Two_Column_Layout::layout_input()` - not so great because only applies to this form layout
- `EE_Text_Input_Display_Strategy::display()` - not so great because only applies to this input layout
- `EE_Form_Input_Base::get_html_for_input()` - great because it would apply to all inputs and layouts

and then simply add your dashicon before the input html.

Another possibly more robust solution, but far more complex, would be to add dashicons to the form system as part of the `EE_Form_Input_Base` class and then pass your configuration details along when constructing the form. This of course would be best added via a separate issue.

For now I would recommend adding a filter to `EE_Form_Input_Base::get_html_for_input()`

admin_pages/general_settings/OrganizationSettings.php
```diff
@@ -0,0 +1,487 @@
...
+    }
+
+
+    /**
+     * Return whether the site license key has been verified or not.
+     * @return bool
+     */
+    private function licenseKeyVerified()
+    {
+        if (empty($this->registry->NET_CFG->core->site_license_key)) {
```
tn3rb: same comment re: injecting `EE_Network_Core_Config` and all other uses in class

core/libraries/form_sections/strategies/display/EE_License_Key_Display_Strategy.php
nerrad: Your feedback prompted memory of something we already added to the form system that I had forgot about.  I can use the `VsprintfFilter` object for doing what's needed.

nerrad:
Ready for testing.  The testing steps are in the description for this pull.

PR #531, opened 2018-06-14
Ensure EE custom post type objects get added to WP_Post objects during AJAX requests Fixes #529

### LINKED ISSUE #529: EE_Event object not attached to post data on ajax requests.
Reported here: https://eventespresso.com/topic/ajax-request-return-data-doesnt-include-ee_event-information-in-post-data/

A quick way to reproduce is to use WP Debug bar console and add:

```
$posts = get_posts(array('post_type'=>'espresso_events','suppress_filters'=>false));

var_dump($posts);
```

With master, you'll just get WP_Post objects with no EE_Event objects.

The same code used in a template file returns the WP_Post and EE_event object.

If you switch to 4.9.59.rc.009 you'll get both objects on the same ajax request above, it looks like this was introduced with:

https://events.codebasehq.com/projects/event-espresso/tickets/11174

tn3rb:
Fixes #529
`EE_CPT_Strategy` was only loading for frontend HTML or frontend EE AJAX requests, but needed to also run for non-EE AJAX requests

## How has this been tested
confirmed via browser using Debug Bar example given in #529


joshfeck:
I took a quick spin with this branch and it looks like we're on our way to a fix, but there are a few issues.

With this branch, if I'm **on the front end of the site** and I open the WP Debug bar console and add:

`$posts = get_posts(array('post_type'=>'espresso_events','suppress_filters'=>false));`

`var_dump($posts);`

The EE_Event object **is** attached.

if I switch to the **admin side of the site**, open the WP Debug bar console and try the same code, the EE_Event object **is not** attached.

Also, the EE shortcodes do not parse if they're included in the calendar tooltip (which we've found to also be a regression from adding the bot detection middleware.

tn3rb:
@joshfeck 

> if I switch to the admin side of the site, open the WP Debug bar console and try the same code, the EE_Event object is not attached.

**Which page(s) in the admin are you going to?**

Because when I do the same thing from the regular post editor ( `http://local.wordpress.test/wp-admin/post.php?post=838&action=edit` ),
the `EE_Event` **is** attached, which makes sense considering the code I added in this branch:

```php
if (($this->request->isFrontend() || $this->request->isAjax())
    && apply_filters('FHEE__EE_Front_Controller__employ_CPT_Strategy', true)
    && ! $this->maintenance_mode->level()
) {
    // determine how to integrate WP_Query with the EE models
    $this->loader->getShared('EE_CPT_Strategy');
}
```

the above conditional is basically saying:

**IF**
- request is for the frontend OR **any** type of AJAX request
- AND this filter returns `true` (which defaults to `true`)
- AND we are not in any kind of Maintenance Mode
then load the `EE_CPT_Strategy` class

Since the Debug Bar console is performing AJAX requests (and assuming you are not toggling that filter or in M-Mode), 
the `EE_CPT_Strategy` class should be loading, which is responsible for injecting `EE_Event` objects into their corresponding `WP_Post` objects.

I've gone through and tried every top level WP admin page (Dashboard, Posts, Media, etc) as well as many secondary pages, as well as EE Admin pages and can not find anywhere the  `EE_Event` is not getting attached. The _only_ exception to this would be the WP Customizer, but that's only because the Debug Bar does not run on that page, therefore I can't run any AJAX requests from the Debug Bar console!

The `EE_CPT_Strategy` class is also getting loaded in `EE_Admin::get_request()` during admin requests, and since it appears we just want that happening anywhere, what I can do is remove that line of code from `EE_Admin::get_request()` and then also remove all conditionals surrounding the loading of `EE_CPT_Strategy` in `EE_System` so that it will simply be a part of every request.

I would still really like to know what admin page you were on where the above existing code was not working.

tn3rb:
> I can't shake the feeling there was a reason why we didn't load EE_CPT_Strategy everywhere 

oh but we were...  but the loading was split up...
frontend requests loaded it from`EE_Frontend_Controller` and **all** admin requests, including AJAX loaded it from `EE_Admin`
So the changes in this branch return to loading `EE_CPT_Strategy` everywhere.

tn3rb:
### testing

- activate the Debug Bar and Debug Bar Console plugins
- switch to the frontend, open the DBB console and enter the following:

```php
$posts = get_posts(array('post_type'=>'espresso_events','suppress_filters'=>false));
var_dump($posts);
```

- click `run`
- [ ] confirm that the displayed `WP_Post` object contains the corresponding `EE_Event` object - the `WP_Post` object is not that large and should fit within one screens length if you have a large monitor, whereas the `EE_Event` object will take up many many screens due to the inclusion of its related data (datetimes, tickets, etc) so it's pretty easy to determine if this is successful simply by the amount of data returned.

sethshoultes:
@Pebblo can you please test?

Pebblo:
So I tested this with the above notes and it seems to be working fine now, I get the EE_Event object returned on the posts whenever I run the above.

I navigated to multiple different pages both front and back end and tested the above in DBB.

Tested using Beaver builder, and Elementor simply because we've had issues with their ajax requests before so confirmed they still worked (just pulled some posts, the EE calendar, EE widget and so on)

However, Darren mentioned the scope for testing increased quite a bit and I'm not so sure I've done any extensive testing on anything specific here, which may be fine but just checking before settings this for merging.

@tn3rb Anything else I  should be looking at for this?

tn3rb:
No I think it should be ok... Darren (rightfully so) thought that the `EE_CPT_Strategy` class was not getting loaded everywhere because that process was split up and done from different places. The changes in this branch merely amalgamate that process and do all of the loading from `EE_System` now for basically all routes that generate standard `WP_Query` objects.

Issue #534, opened 2018-06-15
Fix typo in help tab for PayPal Express


lorenzocaum:
From the WP-admin (WordPress dashboard), go to Event Espresso --> Payment Methods. Then click on the Help tab in the top right and click on PayPal Express.

There is a typo: https://cl.ly/3Y0Y2n2N3C2h

sethshoultes:
@mnelson4 or @knazart can one of you please fix this really quick?

mnelson4:
created https://github.com/eventespresso/event-espresso-core/pull/535 to address this

PR #535, opened 2018-06-15
Bug/paypal pro typo for #534

mnelson4:
Addresses https://github.com/eventespresso/event-espresso-core/issues/534. Unfortunately, this string will need to be re-translated


mnelson4:
To test, just check the typo was fixed.
Once this is done, anyone can just click "sqaush and merge", then update the status label and move into the next release milestone

PR #537, opened 2018-06-15
Load shortcodes and template tags during AJAX requests Fixes #477

### LINKED ISSUE #477: Fix [ESPRESSO_EVENT_ATTENDEES] no longer parsing in the calendar content.
Reported here: https://eventespresso.com/topic/registered-members-short-code/

In ver 4.9.59+ the [ESPRESSO_EVENT_ATTENDEES] shortcode stopped parsing in the calendar content.

To reproduce just add the above shortcode to an event's description, then view that event on the calendar. The shortcode doesn't parse at all and just shows in the content -  http://take.ms/jKmrWi

Switching to EE core 4.5.58 the shortcode parses as expected.

tn3rb:
Fixes #477
Neither the ShortcodesManager nor the template_tags (which were being used by some shortcodes) were being loaded during AJAX requests.

## How has this been tested
confirmed that shortcodes load and are processed as expected in Calendar tooltip content


tn3rb:
### testing

- edit an existing event that has had some registrations for it already
- add the `[ESPRESSO_EVENT_ATTENDEES]` shortcode to the event's content and make sure there is no excerpt set (because the calendar will favour the excerpt)
- view the calendar, find the above event, and hover over the listing
- confirm that the attendee list appears in the content

sethshoultes:
@Pebblo can you please test?

Pebblo:
This is working as expected.

The shortcode is now parsing in the Ajax requests for the calendar. 
[ESPRESSO_EVENT_ATTENDEES]
[ESPRESSO_TICKET_SELECTOR event_id=9]
[restrict]This content is limited to logged in users.[/restrict] (Restrict content Pro)
[gallery]

All parse.

[ESPRESSO_EVENTS_TABLE_TEMPLATE] does not (although I wouldn't expect it to)

I think this is good to go

PR #543, opened 2018-06-20
Add formatting helpers and tweak model exports

nerrad:

This pull was spawned out of the need to have human friendly representations of dates in the default generated options for the datetime selector.  Accomplishing this resulted in a number of peripheral changes as well:

### Restructuring of how models are exposed.

The current models in `assets/src/data/models/{model}` are now exposed as named exports via `models.js` (which in turn is exported via `/data/models/index.js`.  This change means that any js within a model folder will be exposed on the named label for the model.

For example, instead of importing the datetime model via this:

```js
import dateTimeModel from '../../data/model/datetime'
```

You'd do:

```js
import { dateTimeModel } from '../../data/model';
```

I think this reduces slightly the need for the deeper nested relative paths and paves the way for eventually exposing models on `@eventespresso/models`.

### New `@eventespresso/helpers` import and corresponding work.

There's a new folder `data/helpers` that has the purpose of exposing various helpers related to data manipulation/usage.  This is exposed in the `ee-helpers.js` file via `eejs.helpers` and can be imported via `@eventespresso/helpers`.

Currently the helpers module includes date string formatters exposed on `helpers/datetime.js`.  This in turn uses some new server exposed data (see `helpers/site-data.js`) for formatting settings on the server.

### Base model formatter and concrete model wrappers.

The date time string helpers exposed via `helpers/datetime.js` are composed by the `data/model/base-date-formatter.js` into model specific formatters that handle automatically spitting out formatted date strings on entity properties from models.  This in turn is used to dynamically generate functions by the concrete models (see `/data/model/datetime/formatter.js`) which handle automatically passing the fields that get formatted.  So instead of client code having to do:

```js
import { datesToSiteFormatOnEntities } from '../data/model/base-date-formatter.js';
import { DATE_FIELDS } from '../data/model/datetime/formatter.js';

const exampleDateEntities = [
   /** an array of EE datetime entities */
];
const formattedEntities = datesToSiteFormatOnEntities(  exampleDateEntities, DATE_FIELDS );
```

Client code could do this instead:

```js
import { dateTimeModel } from '../data/model';

const exampleDateEntities = [
   /** an array of EE datetime entities */
];

const formattedEntities = dateTimeModel.formatter.datesToSiteFormatOnEntities( exampleDateEntities );
```

### `build-options.js` now implements `dateTimeModel.prettyDateFromDateTime` in the map for option labels.

buildOptions now accepts functions used with the map and the function receives the entity being mapped.  This allows for more complex label and/or value entries in the built option.

With the map now being able to use functions, the map for `datetime` entities now uses a function to build the label value for datetime options. This ensures that the default generated options are human friendly.  They will have the dates formatted according to the date format set for the WordPress site by default (and mysql format if that's not available).

## How has this been tested

* [x] All new stuff is covered by jest tests and passing locally.
* [x] I've tested changes with the work I'm doing on the barcode scanner app and it works great there.

## Checklist

* [x] My code is tested.


nerrad:
@tn3rb this is ready for review.  I'm pretty sure the code is ready to go and for review.  However, I will still have to tweak the tests so they are not susceptible to end-of-day type bugs (they currently are). So any failing tests are likely due to timezone issues in the travis environment.

I figured it'd still be good to give this over to you to start reviewing because it'll give you a preview of what's coming down the pipes.

assets/src/data/helpers/site-data.js
```diff
@@ -0,0 +1,30 @@
...
+	'YY-MM-DD';
+
+/**
+ * The time format used by the site or mysql time format if not set.
+ * @type { string }
+ */
+export const FORMAT_SITE_TIME = dateFormats.moment_split &&
+	dateFormats.moment_split.time ?
+	dateFormats.moment_split.time :
+	'HH:mm:ss';
```
tn3rb: same thing as above... what about:

 - moment default: `HH:mm:ss`
 - WP admin settings default:  `h:mm a`

assets/src/data/model/base-date-formatter.js
```diff
@@ -0,0 +1,220 @@
...
+ *
+ * @param { Array } entities  An array of entity objects
+ * @param { Array } entityDateFields  An array of field names that are date
+ *   fields.
+ * @param { boolean } local      Whether or not to convert the date field value
+ *   to the local timezone for the host.
+ * @return { Array }  Returns a new array of new entities with the date field
+ *   values formatted
+ */
+export const datesToMysqlFormatOnEntities = (
```
tn3rb: following the same pattern you have set already in `assets/src/data/helpers/datetime.js` (ie: `formatDateString()`, `formatMysqlDateString()`, `formatSiteDateString()`, etc), how about naming this `formatMysqlDatesOnEntities()` ?

or... 

if you are trying to be more descriptive, then rewrite the function name to be more inline with how people actually talk...
ie: you would never say "Hey, let's dates to MySQL format on the entities ok?"
but you might say something like "Hey, let's format the entities dates to MySQL ok?"
So that latter sentence might translate into a function name like: `formatEntitiesDatesToMysql()` which is far easier to understand the purpose of than `datesToMysqlFormatOnEntities()`

assets/src/data/model/base-date-formatter.js
```diff
@@ -0,0 +1,220 @@
...
+ * mutate original entity.
+ *
+ * @param { Object } entity  An array of entity objects
+ * @param { Array } entityDateFields  An array of field names that are date
+ *   fields.
+ * @param { boolean } local      Whether or not to convert the date field value
+ *   to the local timezone for the host.
+ * @return { Object }  Returns a new entity with the date field values formatted
+ */
+export const datesToMysqlFormatOnEntity = (
```
tn3rb: same as above, how about: 

- `formatMysqlDatesOnEntity()`
- or `formatEntityDatesToMysql()` ?

assets/src/data/model/base-date-formatter.js
```diff
@@ -0,0 +1,220 @@
...
+ *
+ * @param { Array } entities  An array of entity objects
+ * @param { Array } entityDateFields  An array of field names that are date
+ *   fields.
+ * @param { boolean } local      Whether or not to convert the date field value
+ *   to the local timezone for the host.
+ * @return { Array }  Returns a new array of new entities with the date field
+ *   values formatted
+ */
+export const datesToSiteFormatOnEntities = (
```
tn3rb: same as above, how about: 

- `formatSiteDatesOnEntities()`
- or `formatEntitiesDatesToSite()`
- or even  or `formatEntitiesDatesToSiteFormat()`

assets/src/data/model/base-date-formatter.js
```diff
@@ -0,0 +1,220 @@
...
+ * mutate original entity.
+ *
+ * @param { Object } entity  An array of entity objects
+ * @param { Array } entityDateFields  An array of field names that are date
+ *   fields.
+ * @param { boolean } local      Whether or not to convert the date field value
+ *   to the local timezone for the host.
+ * @return { Object }  Returns a new entity with the date field values formatted
+ */
+export const datesToSiteFormatOnEntity = (
```
tn3rb: same as above, how about: 

- `formatSiteDatesOnEntity()`
- or `formatEntityDatesToSite()`
- or `formatEntityDatesToSiteFormat()`

assets/src/data/model/base-date-formatter.js
```diff
@@ -0,0 +1,220 @@
...
+ * Converts date field values to moment objects for the provided entities.
+ * Does not mutate original entities.
+ *
+ * @param { Array } entities An array of entity objects
+ * @param { Array } entityDateFields An array of field names that are date
+ *   fields.
+ * @return { Array } Returns a new array of new entities with the date field
+ *   values converted to moment objects.
+ */
+export const datesToMomentOnEntities = (
```
tn3rb: how about: 

- `convertEntitiesDatesToMoment()`

assets/src/data/model/base-date-formatter.js
```diff
@@ -0,0 +1,220 @@
...
+ * Converts date field values to moment objects for the provided entity.
+ * Does not mutate original entity.
+ *
+ * @param { Object } entity An entity.
+ * @param { Array } entityDateFields An array of field names that are date
+ *   fields.
+ * @return { Object } Returns a new entity with the date field values converted
+ *   to moment objects.
+ */
+export const datesToMomentOnEntity = (
```
tn3rb: how about: 

- `convertEntityDatesToMoment()`

assets/src/data/model/datetime/formatter.js
```diff
@@ -0,0 +1,111 @@
...
+			);
+		}
+		if ( DateTimeEntity.DTT_EVT_end ) {
+			content += DateTimeEntity.DTT_EVT_end.format(
+				DATE_TIME_FORMAT_SITE
+			);
+		}
+	}
+
+	content += DateTimeEntity.DTT_name ? ')' : '';
```
tn3rb: `DTT_name` could be added here with only one ternary instead of before and after processing the dates, ie:

```js
content = DateTimeEntity.DTT_name ? DateTimeEntity.DTT_name + ' (' + content + ')' : content 
```

which would be more clear and less prone to mistakes if this needs refactoring or removal, etc

assets/src/data/helpers/datetime.js
```diff
@@ -0,0 +1,98 @@
...
+ *                                recommended the incoming dateString is in UTC
+ *   OR the format of the incoming string includes offset info.
+ * @return { string }  Returns a date string in the provided format.
+ */
+export const formatDateString = (
+	dateString = '',
+	format = DATE_TIME_FORMAT_ISO8601,
+	local = true,
+) => {
+	const date = dateString === '' ? moment() : moment( dateString );
```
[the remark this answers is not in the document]
nerrad: doh, obvious too isn't it :)

assets/src/data/helpers/datetime.js
```diff
@@ -0,0 +1,98 @@
...
+
+/**
+ * Receives an indefinite number of dateStrings as arguments and concatenates
+ * them together with the given separator.
+ * @param { string } separator
+ * @param { ...string } dateStrings
+ * @return { string }  Returns a string concatenating all the provided
+ *                       dateStrings together with the given separator.
+ */
+export const allDateTimesAsString = ( separator = ' - ', ...dateStrings ) => {
```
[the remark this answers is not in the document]
nerrad: sounds like premature refactoring but easy to add so I'll do.

assets/src/data/helpers/site-data.js
```diff
@@ -0,0 +1,30 @@
...
+export const { date_formats: dateFormats = {} } = data.site_formats;
+
+/**
+ * The date format used by the site or mysql date format if not set.
+ * @type { string }
+ */
+export const FORMAT_SITE_DATE = dateFormats.moment_split &&
+	dateFormats.moment_split.date ?
+	dateFormats.moment_split.date :
+	'YY-MM-DD';
```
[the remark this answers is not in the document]
nerrad: > I'm imagining a scenario where the site format is not getting set on a particular page for whatever reason

This won't happen because of the dependency set for the helpers (at least in a WP context).
The purpose of this is to have a human readable format.  Currently, the moment.default is ISO8601.  I figured that wasn't a really human friendly format and mysql would be more neutral.  I'm certainly open to changing the default to something else though.

assets/src/data/helpers/site-data.js
nerrad: moment default is not `HH:mm:ss`

I also don't want to tie this specifically to WP because it might not always be used in a WP environment.  I just figured mysql format is a good explicit default.

assets/src/data/model/base-date-formatter.js
```diff
@@ -105,7 +105,7 @@ export const datesToMysqlFormatOnEntities = (
  *   to the local timezone for the host.
  * @return { Object }  Returns a new entity with the date field values formatted
  */
-export const datesToMysqlFormatOnEntity = (
+export const formatEntityDateToMysql = (
```
tn3rb: sry for being nitpicky.. but shouldn't it be `formatEntityDatesToMysql()` because it may format more than one date?

nerrad: ya good catch

Issue #544, opened 2018-06-21
Payment logs: if we can't record the payment the log is for, at least record the transaction


sethshoultes:
**Codebase ticket:** 
https://events.codebasehq.com/projects/event-espresso/tickets/11367

**Testing notes:**
To test this use master and the iPay88 payment method (it doesn't need to be set up to function, just add M03222 as the merchant code and Tmnx7fmSu4 as the merchant key)

The reason for using that payment method is right now it creates logs that use the transaction object rather the the payment object.

Setup a paid ticket and register onto it using iPay88.

Check the PM logs and you'll find 2 of them do not link to a transaction - http://take.ms/9Q0SK

Now switch to BUG/11361/fix-log-calls-using-the-transaction-object-rather-than-the-payment-object

Repeat the same test and check the payment method logs, this time you'll see the logs link to the transaction, but show a payment method of unknown - http://take.ms/LpjnU

Confirm that the payment method in both the unknown logs and the 'normal' logs show correctly when you view the log itself:
Unknown - http://take.ms/aR8rW
Normal - http://take.ms/Ec0zV

There will be another ticket to improve this but for now its better to know the logs are for a specific transaction and work from there, than to know theres 'a' log entry for a payment method, but not know what its for.

The above payment will fail, now select a payment method to finalize the registration (complete it using another like Stripe if preferred).

Confirm the logs still show correctly for the iPay88 payments and the new one.

sethshoultes:
@Pebblo does this still need to be tested? I see a related Codebase ticket that still needs to be merged:
https://events.codebasehq.com/projects/event-espresso/tickets/11361

Pebblo:
Yes, this still needs to be tested, but I got multiple conflicts on when merging master so I've created a new branch with the changes.

WRT - https://events.codebasehq.com/projects/event-espresso/tickets/11361

That's for the iPay88 payment method and the reason I haven't merged it in yet is to help with testing this branch. The iPay88 payment method uses the transaction object for all log calls currently, so its really simple to test this change to confirm it works.

It's 'better' to use the payment object for the logs which is what that ticket changes in iPay88.

Pebblo:
OK, this was fixed with #545 so closing out.

PR #545, opened 2018-06-22
FET/544/fix log calls when using transactions object

Pebblo:
If a payment method passes an EE_Transaction object to the log call the log entry shows 'Could not be determined' because it expects a payment object.

This pull request allows it to use an EE_Transaction object so we know what the log entry relates to.

## How has this been tested
See issue https://github.com/eventespresso/event-espresso-core/issues/544

This was tested by me on an older branch, however, I'm getting conflicts due to formatting changes etc when merging master into that branch, so I created this branch and reapplied the commits to it.

I'll retest this on this branch before merging, then remove both branches.


admin_pages/payments/Payment_Log_Admin_List_Table.class.php
```diff
@@ -198,8 +198,10 @@ public function column_PMD_ID(EE_Change_Log $item)
             return $item->object()->admin_name();
         } elseif ($item->object() instanceof EE_Payment && $item->object()->payment_method()) {
             return $item->object()->payment_method()->admin_name();
+        } elseif ($item->object() instanceof EE_Transaction) {
+            return esc_html__('Unknown', 'event_espresso');
```
nerrad: Why using "Unknown" here?  Is there some other label that can be used that might provide some more helpful information?  Or would it be better to maybe exclude logs that are specifically a transaction object from showing in this list table instead?

Or maybe it'd be better to get the payment method from the transaction and then use _it's_ `admin_name()`?

admin_pages/payments/templates/payment_log_details.template.php
```diff
@@ -31,34 +31,40 @@
...
                     <label>
-                        <?php _e('Payment Method', 'event_espresso'); ?>
+                        <?php esc_html_e('Payment Method', 'event_espresso'); ?>
                     </label>
                 </th>
                 <td>
-                    <?php echo $payment_method
+                    <?php
+                    if ($payment_log->object() instanceof EE_Transaction) {
+                        esc_html_e('Unknown', 'event_espresso');
```
nerrad: I think it'd be better here to get the payment method from the transation and then use that payment method's `admin_name()`?

nerrad:
Hey @Pebblo I added some more labels for you, but would  you be okay with adding labels that are applicable to any pulls/issues you create going forward?  I _think_ that's something we're supposed to be doing but maybe not - so you can double-check with Seth or Brent too maybe.

nerrad:
Also @Pebblo you still need to add a changelog entry for this please

admin_pages/payments/Payment_Log_Admin_List_Table.class.php
Pebblo: Good question and the reason for unknown is we can't be certain that this log is for the transactions last payment method which makes it harder for support to follow whats happening and can easily lead us down the wrong path.

Lets say a user attempts a payment with payment method A and it shows as payment failed, so switches to payment method B and it's successful, both logs will then show as payment method B.

Mike initially did use the payment method name from the transaction and I changed it because of the above, see:
https://events.codebasehq.com/projects/event-espresso/tickets/11367#update-49090081

admin_pages/payments/templates/payment_log_details.template.php
Pebblo: This is happening for the same reason as above, if the Transaction object is passed to the log call, we can't be sure that the last payment method on a transaction is the 'correct' payment method.

CHANGELOG.md
```diff
@@ -27,4 +27,6 @@ Headings should be (only add headings as needed):
 - Fixed typo in paypal pro ([535](https://github.com/eventespresso/event-espresso-core/pull/535))
 - Fixed EE CPT objects not being added to WP_Post objects during Ajax Requests ([531](https://github.com/eventespresso/event-espresso-core/pull/531))
 
+### Changed
 
+- Link payment logs to the transaction if an EE_Transaction object is passed to the log call (545(https://github.com/eventespresso/event-espresso-core/pull/545))
```
nerrad: I think this should go in the fixed section, its a bug right?  Also please use declarative (or indicative) mood, you can see docs on our changelog entries [here](https://github.com/eventespresso/event-espresso-core/blob/master/docs/A--Best-Practices/change-log.md)

Finally, your formatting for the link to the pull is incorrect. You can get the correct format from the other changelog entries.

Thanks!

Pebblo:
Sorry, that better?

Pebblo:
The last 2 commits are from me missing a changed Mike made here: https://events.codebasehq.com/projects/event-espresso/repositories/32-core/commit/b4ebe83c2cbb13f085b50c5c55af6936c79ef9c4

We need to use the transaction_id variable rather than trying to pull it from the object again.

Assigning to Mike for review.

Pebblo:
OK, So I've tested this and confirmed that existing payment methods that pass the payment object continue to log as expected and also tested a payment method that passes the transaction object to the log.

Both worked as expected and display the log entries as expected so I'm merging this in.

PR #554, opened 2018-06-28
Add attendee model

tn3rb:
adds attendee model in `assets/src/data/model/attendee/`

## How has this been tested
via included unit tests


assets/src/data/model/attendee/index.js
```diff
@@ -0,0 +1,128 @@
...
+// import from constants file when that gets added
+const REGISTRATION_STATUS_IDS = [
+	'RAP',
+	'RCN',
+	'RDC',
+	'RIC',
+	'RNA',
+	'RPP',
+	'RWL',
+];
```
nerrad: I think it's be good if this was part of the registration model and then pulled in here from there.  I imagine other models might need this as well.  So what you could do is get the model pull you did (#553 - its approved) merged into GB/master and then into this pull and then move this constant into the model/registration/constants.js file?

assets/src/data/model/base.js
```diff
@@ -26,14 +25,21 @@ export const getQueryString = (
...
-		pickBy( queryArgs, value => ! isUndefined( value ) ),
-	);
+	const queryParams = [];
+	if ( ! isUndefined( limit ) ) {
+		queryParams.push( `limit=${ limit }` );
+	}
+	if ( ! isUndefined( mapOrderBy( orderBy ) ) ) {
+		if ( isArray( mapOrderBy( orderBy ) ) ) {
+			for ( const field of mapOrderBy( orderBy ) ) {
+				queryParams.push( `order_by[${ field }]=${ order }` );
```
nerrad: this looks good for now, however do you think there'll ever be a case where we want the two order_by fields to be ordered differently (i.e. one ASC, and one DESC)?  If so, I wonder if we should account for that pattern.

tn3rb: I thought about that as I was writing the code but couldn't think of any cases where we currently do that so was fine with letting that go as is for now.
It can always be improved upon later.

assets/src/data/model/attendee/index.js
nerrad: I know in the other pull the _individual status constants_ are in the status model and that's okay and I think it should stay that way.  However the specific code I'm referring to is this:
```
const REGISTRATION_STATUS_IDS = [
+	'RAP',
+	'RCN',
+	'RDC',
+	'RIC',
+	'RNA',
+	'RPP',
+	'RWL',
+];
```

This is different in that its used to provide an _array_ of registration statuses.   Since its informative about the `registration` model domain, I thought it would be useful if its actually kept with the registration model.  I can totally see other models like Transaction, or Ticket that might build queries relying on a `registration_status` prop in which case it would make sense to pull the specific allowed props enum from the registration model.

So to be clear, this array is acting more as an `enum` definition and thus I think it should reside in the model it relates to.

With that said, if you want to put all the specific status enum-link arrays like this in the status model I'm also fine with that, you could label the constants appropriately in it.

assets/src/data/model/attendee/index.js
```diff
@@ -0,0 +1,128 @@
...
+
+const REGISTRATION_STATUS_IDS = [
+	statusModel.STATUS_ID_REGISTRATION_APPROVED,
+	statusModel.STATUS_ID_REGISTRATION_CANCELLED,
+	statusModel.STATUS_ID_REGISTRATION_DECLINED,
+	statusModel.STATUS_ID_REGISTRATION_INCOMPLETE,
+	statusModel.STATUS_ID_REGISTRATION_NOT_APPROVED,
+	statusModel.STATUS_ID_REGISTRATION_PENDING_PAYMENT,
+	statusModel.STATUS_ID_REGISTRATION_WAIT_LIST,
+];
```
nerrad: I know in the other pull the _individual status constants_ are in the status model and that's okay and I think it should stay that way.  However the specific code I'm referring to is this:
```
const REGISTRATION_STATUS_IDS = [
+	'RAP',
+	'RCN',
+	'RDC',
+	'RIC',
+	'RNA',
+	'RPP',
+	'RWL',
+];
```

This is different in that its used to provide an _array_ of registration statuses.   Since its informative about the `registration` model domain, I thought it would be useful if its actually kept with the registration model.  I can totally see other models like Transaction, or Ticket that might build queries relying on a `registration_status` prop in which case it would make sense to pull the specific allowed props enum from the registration model.

So to be clear, this array is acting more as an `enum` definition and thus I think it should reside in the model it relates to.

With that said, if you want to put all the specific status enum-link arrays like this in the status model I'm also fine with that, you could label the constants appropriately in it.

tn3rb: ohhhhh ya ok... ya I misinterpreted what you said. ya I think that's a good idea. will fix

Issue #533, opened 2018-06-15
Mock Add-Ons Missing A Lot of Setup files


mnelson4:
Like:

* .github
* assets ?
* eslink.js ?

and all the other top-level files (like .babelrc, composer.json, etc) that EE core and most add-ons have now. I encountered this while working on https://github.com/eventespresso/eea-paypal-smart-buttons/pull/1

mnelson4:
related PR merged in

PR #555, opened 2018-06-28
Task/skeleton addon missing files for #533

mnelson4:
Addresses https://github.com/eventespresso/event-espresso-core/issues/533


mnelson4:
I originally did this as part of ICE time, but just double-checked with Seth and he's fine with me bringing someone else in to review this so it can get into core.
So, to my knowledge, this adds all the boilerplate files needed for nodejs and composer to the mock/skeleton add-ons. This should be helpful next time we go to create a new add-on

mnelson4:
...Although I assume I've missed some boilerplate code etc, so I'd be fine if you want to take a stab at adding anything I missed @nerrad

mnelson4:
yeah a WP CLI command for creating add-ons would probably be good whenever you get around to it

PR #556, opened 2018-06-28
Add higher order component for wrapping components with BaseControl

nerrad:

In the WordPress editor, Gutenberg has a component that takes care of adding useful styling surrounding inputs (typically in the inspector).  This is provided via the `<BaseControl />` component.  To ease the implementation of this for any time our selectors (or other input type components we create) are implemented in the admin I've created a [Higher Order Component](https://reactjs.org/docs/higher-order-components.html) that can be used to automatically compose BaseControl as the wrapper around the provided component for wrapping.

Implementation for this new component is fairly trivial, the syntax is:

```
// note if you are using this in a bundled component for EE core, please import directly not via the package.
import { withBaseControl } from '@eventespresso/higher-order-components';

//some component that gets wrapped
const MyComponent = props => <div { ...props } />

export default withBaseControl( 'div-my-component')( MyComponent );
```

More documentation is included with this pull.

Also as a part of this pull, I've introduced a `higher-order-components` folder for holding all hocs we build (located at `src/higher-order-components`).  This in turn is built to a `ee-hocComponents-{hash}.js` bundle that can be used as a dependency for client js.  Currently, this file is NOT a dependency on any default ee core files because I think its important that for the `ee-components` folder we do direct dependencies on hocs any of those components may be used (via importing the files directly in the source).  However, I still exported the HOCs as a separate bundles as its possible EE add-ons or other sources might only need the built components package and not the hoc package in certain contexts.  This decision may be revisited at some point, but for now I think it will work okay.

## How has this been tested

* [ ] Added new unit-tests for the new component


assets/src/components/form/select/model-selects/test/datetime.js
```diff
@@ -1,5 +1,7 @@
-import DatetimeSelect from '../datetime';
-import { shallow } from 'enzyme';
+import DatetimeSelect, { EditorDatetimeSelect } from '../datetime';
```
tn3rb: is this line correct ? looks wonky...

tn3rb: just looked at other tests below and you've done the same thing there... so I guess it is correct

 ` \_( O.o )_/ `

assets/src/higher-order-components/base-control.js
```diff
@@ -0,0 +1,50 @@
...
+				};
+
+				render() {
+					const {
+						label,
+						instanceId,
+						className,
+						help,
+					} = this.props;
+					const id = `inspector-${ customId }-control-${ instanceId }`;
```
tn3rb: where is `customId ` coming from? 
ie: why is it not in the list of constants extracted from the props like `instanceId` ?

docs/AA--Javascript-in-EE/higher-order-components/README.md
```diff
@@ -0,0 +1,11 @@
+From the [react documentation](https://reactjs.org/docs/higher-order-components.html):
+
+> A higher-order component (HOC) is an advanced technique in React for reusing comopnent logic.  HOCs are not a part of the REACT API, per se. They are a pattern that emerges from React's compositional nature.
+
+Within EE, we use higher order components to extract common patterns we see in use for various components to keep our architecture DRY (Don't Repeat Yourself principle).
+
+All the higher order components listed in this section of our documentation are exposed within WordPress via the `EventEspresso\core\domain\services\assets\CoreAssetManager::JS_HANDLE_EE_HOC_COMPONENTS` script handle that can be used when registering your own script implementing a hoc.  Individual HOC documentation will further elaborate on how to implement it.
```
tn3rb: I think `HOC ` should be capitalized everywhere so maybe cahnge this to:

> when registering your own script implementing a HOC.

docs/AA--Javascript-in-EE/higher-order-components/base-control.md
```diff
@@ -0,0 +1,125 @@
...
+
+If provided, this string will be used as the content for accessible help text describing the wrapped component.
+
+- Type: `String`
+- Required: No
+- Default: undefined
+
+## Usage
+
+Note: If this HOC is being used in an ee core component then please import it directly as opposed to using the package name.  This ensures that the built component bundle will have the necessary code from the hoc in the bundle.
```
tn3rb: capitalize HOC

> This ensures that the built component bundle will have the necessary code from the HOC in the bundle.

assets/src/components/form/select/model-selects/test/datetime.js
nerrad: ya its correct.  The reason why `DateTimeSelect` is not deconstructed is because it is the `default` export in `../datetime`.  So in `..datetime` there is:

```js
export default class DatetimeSelect extends Component { /*...*/ }
export const EditorDatetimeSelect
```

assets/src/higher-order-components/base-control.js
nerrad: It's coming from the argument passed into the HOC: `( customId = '' ) =>`

The signature of the HOC is `withBaseControl( customId = '' )( WrappedComponent)`

assets/src/components/form/select/model-selects/test/datetime.js
tn3rb: kk... didn't know imports worked like that... kewl thnx

assets/src/higher-order-components/base-control.js
tn3rb: ya I saw that it was an argument passed to the HOC but didn't realize that that meant it was automatically passed into the `render()` method, ie: I thought all incoming arguments just became part of the props and that you still had to extract them within each method.

so kewl... that's two new things I just learned 👍

nerrad: Well its not really "passed" in.  It's just that the argument is within scope of all the code contained within the hoc function. It's similar to if you did something like this:

```
const something = ( someArg ) => {
    return function() {
         console.log(someArg);
    }
};

something('hello there')();
//would print 'hello there'
```

Issue #557, opened 2018-06-28
Allow payment method's default front-end name to differ from its admin-only name


mnelson4:
For https://github.com/eventespresso/eea-paypal-smart-buttons/issues/7, we'd like the front-end name of the payment method to be different than the back-end name. (On the frontend we want something fairly short but not necessarily unique, like "PayPal Express Checkout", but on the back-end we need something unique so admins can tell "PayPal Express" apart from "PayPal Express Checkout with Smart Buttons").
For that

mnelson4:
related PR https://github.com/eventespresso/event-espresso-core/pull/558 merged in

PR #558, opened 2018-06-28
Fet/different default frontend name

mnelson4:
Addresses https://github.com/eventespresso/event-espresso-core/issues/557. Add a new method to `EE_PMT_Base` that allows payment methods to have a different front-end name than back-end name and make use of it. No EE core add-ons use it yet though, they're fine having the same default name for frontend and backend


tn3rb:
I'm kinda confused by this...
the entire purpose of having two name fields on the Payment Methods, ie: `PMD_name` and `PMD_admin_name`
was so that there could be a different name for the frontend vs the admin.

Can you plz clarify how `defaultFrontendName()` is different than just using `PMD_name` ?

PR #559, opened 2018-06-28
Refactor Model Select Build Options and move optionsEntityMap objects onto models

tn3rb:
The default `ModelSelect` `buildOptions` callback included a default `optionsEntityMap` that included data for every model and although the entire `buildOptions` callback could be overridden by any code implementing a `ModelSelect`, the `optionsEntityMap` itself could not. 

The shortcoming of this is that if a particular use of a selector wanted to display a different label, the only way to do so would be by overriding the entire `buildOptions` callback. Another more pedantic issue with the current setup is that model specific data (the entries in the default `optionsEntityMap`) should be part of the models themselves.

The work in this branch moves the individual option mappings into each of the models and then refactors `ModelSelect` to expect the `optionsEntityMap`  as a required property. As well, the `optionsEntityMap`  have been keyed so that multiple mappings can be creating and then selected during usage. An additional argument named `mapSelection` has been added which allows the `buildOptions` function to choose from the provided mappings.

## How has this been tested
existing unit tests


assets/src/components/form/select/build-options.js
```diff
@@ -1,65 +1,59 @@
...
+		optionsEntityMap[ mapSelection ] :
+		null;
+	// if not but default exists then use that
+	map = map === null && ! isUndefined( optionsEntityMap.default ) ?
+		optionsEntityMap.default :
+		null;
+	// if default doesn't exist then just grab first option
+	map = map === null ?
+		optionsEntityMap[ Object.keys( optionsEntityMap )[ 0 ] ] :
+		map;
```
nerrad: I don't think we should use a fallback to `default` (or the first option) if the requested `mapSelection` value doesn't have a match.  That could lead to hidden unexpected and hard to trace behaviour.  Instead I think it'd be better to maybe print a console warning here (using the `warn` module`) and just return an empty array (btw you can also unit-test for console behaviour... if you need help with that let me know.

If you make my suggested changes, this would effectively mean that EVERY provided map must be at _least_ indexed by a `default` key.  Since that is the default value for the `mapSelection` argument, that allows for correctly reducing to an option array if an optionMap is provided with no `mapSelection` argument.  At the same time, there will be helpful warnings given in the console (in development mode) if an invalid `optionMap` or `mapSelection` argument is provided.

assets/src/data/model/datetime/query.js
```diff
@@ -35,6 +39,19 @@ export const queryDataTypes = {
...
+		label: ( entity ) => {
+			return prettyDateFromDateTime( entity );
+		},
+	},
+	name: {
+		value: 'DTT_ID',
+		label: 'DTT_name',
+	},
+};
+
```
nerrad: As discussed in slack, I'd like to see all these `optionsEntityMap` moved from the current `model/{modelName}/query.js` location, into the `compnents/form/select/model-selects/{modelName}.js` file.

assets/src/components/form/select/build-options.js
```diff
@@ -22,21 +23,36 @@ const buildOptions = (
...
 ) => {
-	if ( isEmpty( entities ) || isEmpty( optionsEntityMap ) ) {
+	if ( isEmpty( entities ) ) {
+		warning(
+			false,
+			'A valid array of entities must be supplied ' +
+			'in order to build options for a ModelSelect component',
+			entities
+		);
+		return [];
```
nerrad: I don't think you should show a warning when entities are empty.  It's entirely feasible that a REST query could result in an empty set returned.  So you can check for empty entities and return and empty array right away but no need for warning imo.

assets/src/components/form/select/model-selects/datetime.js
```diff
@@ -32,7 +46,7 @@ export default class DatetimeSelect extends Component {
 		...model.defaultQueryData,
 		getQueryString: model.getQueryString,
 		selectLabel: __( 'Select Datetime', 'event_espresso' ),
-		addAllOptionLabel: __( 'All Datetimes', 'event_espresso' ),
+		optionsEntityMap: optionsEntityMap,
```
nerrad: with ES6 you can just write this as `optionsEntityMap` so:

```
...model.defaultQueryData,
getQueryString: model.getQueryString,
selectLabel: __( 'Select Datetime', 'event_espresso' ),
optionsEntityMap
```

assets/src/components/form/select/model-selects/event.js
```diff
@@ -31,6 +39,7 @@ export default class EventSelect extends Component {
 		...model.defaultQueryData,
 		getQueryString: model.getQueryString,
 		selectLabel: __( 'Select Event', 'event_espresso' ),
+		optionsEntityMap: optionsEntityMap,
```
nerrad: can just be `optionsEntityMap`

assets/src/components/form/select/model-selects/status.js
```diff
@@ -31,7 +39,7 @@ export default class StatusSelect extends Component {
 		...model.defaultQueryData,
 		getQueryString: model.getQueryString,
 		selectLabel: __( 'Select Status', 'event_espresso' ),
-		addAllOptionLabel: __( 'All Statuses', 'event_espresso' ),
+		optionsEntityMap: optionsEntityMap,
```
nerrad: can just be `optionsEntityMap`

assets/src/components/form/select/model-selects/ticket.js
```diff
@@ -29,7 +37,7 @@ export default class TicketSelect extends Component {
 		...model.defaultQueryData,
 		getQueryString: model.getQueryString,
 		selectLabel: __( 'Select Ticket', 'event_espresso' ),
-		addAllOptionLabel: __( 'All Tickets', 'event_espresso' ),
+		optionsEntityMap: optionsEntityMap,
```
nerrad: can just be `optionsEntityMap`

PR #561, opened 2018-06-28
Bug/use correct pm name function

mnelson4:

Fixes an issue I introduced today on https://github.com/eventespresso/event-espresso-core/pull/558 which is that when someone activates a payment method for the first time, we call a method that doesn't exist and it will cause a fatal error.

## How has this been tested

I activated a new payment method using payment methods pro. On master it has a fatal error but it works on this branch just fine

## Checklist

* [x] My code is tested.


tn3rb:
sry for missing something so painfully obvious

mnelson4:
## Testing
* [ ] Activate a completely new payment method which hasn't been active before (if you've had them all active, you can use payment methods pro to create a copy payment method, that tests the same code). There should be no fatal error. (On current master there will be)

PR #565, opened 2018-07-03
Allow for querying attendees by registration ID in js attendee model.

nerrad:

This pull modifies the attendee js model so that queries can be performed for the attendee related to the provided Registration ID.

## How has this been tested

* [x] ensure javascript unit tests are still running successfully.

## Checklist

* [x] My code is tested.

### DIFF
--- a/assets/src/data/model/attendee/index.js
+++ b/assets/src/data/model/attendee/index.js
@@ -28,6 +28,7 @@ export const queryDataTypes = {
 	forDatetimeId: PropTypes.number,
 	forTicketId: PropTypes.number,
 	forStatusId: PropTypes.oneOf( REGISTRATION_STATUS_IDS ),
+	forRegistrationId: PropTypes.number,
 	queryData: PropTypes.shape( {
 		limit: PropTypes.number,
 		orderBy: PropTypes.oneOf( Object.keys( orderByMap ) ),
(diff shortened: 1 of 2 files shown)


assets/src/data/model/attendee/index.js
```diff
@@ -91,13 +94,16 @@ export const whereConditions = ( {
...
+	forRegistrationId = parseInt( forRegistrationId, 10 );
 	// add query param for entity we are filtering for
 	if ( forTicketId !== 0 && ! isNaN( forTicketId ) ) {
 		where.push( `where[Registration.Ticket.TKT_ID]=${ forTicketId }` );
 	} else if ( forDatetimeId !== 0 && ! isNaN( forDatetimeId ) ) {
 		where.push( `where[Registration.Ticket.Datetime.DTT_ID]=${ forDatetimeId }` );
 	} else if ( forEventId !== 0 && ! isNaN( forEventId ) ) {
 		where.push( `where[Registration.EVT_ID]=${ forEventId }` );
+	} else if ( forRegistrationId !== 0 && ! isNaN( forRegistrationId ) ) {
+		where.push( `where[Registration.REG_ID]=${ forRegistrationId }` );
```
tn3rb: I would put this first since it should likely trump all of the other conditions which are ordered from most specific (narrowest scope) to the least specific (widest scope). Since the relationship between attendees and registrations is 1:1, it is the most specific and narrowest of scopes out of all of these. Otherwise this looks great.

nerrad: I'd agree if this was a simple if/else conditional but its not.  Since all the conditions are `else ifs` I don't see how putting it first will make things any more performant?

tn3rb: because if someone accidently adds both an Event ID AND a Registration ID, the way you have it would return all Attendees for the Event

nerrad: Clarified in slack chat that Brent was inferring that `forRegistrationId` would trump (override) any other ids that are present.

nerrad: > because if someone accidently adds both an Event ID AND a Registration ID, the way you have it would return all Attendees for the Event

I don't think it would because by default WHERE conditions are `AND` not `ANY`.   It's just be a pointless query (and unnecessary extra joins).

PR #566, opened 2018-07-04
Improve status model constants

nerrad:

Originally, the status model constants are singular matches from constant name to the value it represents.  One problem with this is it makes it harder to extract statuses by type.  In this pull I changed things so that the status constants are now grouped by model which provides the following improvements:

1. Less typing on imports.
2. Easier to do logic that gets all statuses for a model (i.e. using `lodash.values` you could do `const allRegStatuses = values( registrationStatus );`
3. the status model can also export all statuses in an array more easier.

## How has this been tested

* [x] ensure all existing unit tests pass without any fails.  This should not affect any test snapshots or anything.

## Checklist

* [x] My code is tested.

### DIFF
--- a/assets/src/data/model/status/constants.js
+++ b/assets/src/data/model/status/constants.js
@@ -12,13 +12,13 @@ export const STATUS_TYPE_PAYMENT = 'payment';
 export const STATUS_TYPE_REGISTRATION = 'registration';
 export const STATUS_TYPE_TRANSACTION = 'transaction';
 // email
-export const emailStatus = {
+export const EMAIL_STATUS_ID = {
 	DRAFT: 'EDR',
 	SENT: 'ESN',
 	EXPIRED: 'EXP',
 };
 // event
-export const eventStatus = {
+export const EVENT_STATUS_ID = {
 	ACTIVE: 'ACT',
 	REGISTRATION_CLOSED: 'CLS',
 	DELETED: 'DEL',
@@ -32,7 +32,7 @@ export const eventStatus = {
 	SECONDARY: 'SEC',
 };
 // message
-export const messageStatus = {
+export const MESSAGE_STATUS_ID = {
 	DEBUG: 'MDO',
 	EXECUTING: 'MEX',
 	FAIL: 'MFL',
@@ -42,7 +42,7 @@ export const messageStatus = {
 	SENT: 'MSN',
 };
 // payment
-export const paymentStatus = {
+export const PAYMENT_STATUS_ID = {
 	APPROVED: 'PAP',
 	CANCELLED: 'PCN',
 	DECLINED: 'PDC',
@@ -50,7 +50,7 @@ export const paymentStatus = {
 	PENDING: 'PPN',
 };
 // registration
-export const registrationStatus = {
+export const REGISTRATION_STATUS_ID = {
 	APPROVED: 'RAP',
 	CANCELLED: 'RCN',
 	DECLINED: 'RDC',
@@ -60,7 +60,7 @@ export const registrationStatus = {
 	WAIT_LIST: 'RWL',
 };
 // transaction
-export const transactionStatus = {
+export const TRANSACTION_STATUS_ID = {
 	ABANDONED: 'TAB',
 	COMPLETE: 'TCM',
 	FAILED: 'TFL',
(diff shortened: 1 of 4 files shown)


assets/src/data/model/status/constants.js
```diff
@@ -7,46 +12,67 @@ export const STATUS_TYPE_PAYMENT = 'payment';
 export const STATUS_TYPE_REGISTRATION = 'registration';
 export const STATUS_TYPE_TRANSACTION = 'transaction';
 // email
-export const STATUS_ID_EMAIL_DRAFT = 'EDR';
-export const STATUS_ID_EMAIL_SENT = 'ESN';
-export const STATUS_ID_EMAIL_EXPIRED = 'EXP';
+export const emailStatus = {
```
tn3rb: Might be better to give this a slightly more accurate name since `emailStatus` sounds singular. 
Maybe `emailStatusArray` or `allEmailStatus` or something like that???

This would also leave the `emailStatus` name available for client code to use to indicate a singular status which would be the natural choice when trying to name such a variable imho.

Same goes for the rest of these constants:

- `eventStatusArray` or `allEventStatus` or ???
- `messageStatusArray` or `allMessageStatus` or ???
- `paymentStatuArrays` or `allPaymentStatus` or ???
- `registrationStatusArray` or `allRegistrationStatus` or ???
- `transactionStatusArray` or `allTransactionStatus` or ???

assets/src/data/model/status/constants.js
```diff
@@ -7,46 +12,67 @@ export const STATUS_TYPE_PAYMENT = 'payment';
...
-export const STATUS_ID_TRANSACTION_OVERPAID = 'TOP';
+export const transactionStatus = {
+	ABANDONED: 'TAB',
+	COMPLETE: 'TCM',
+	FAILED: 'TFL',
+	INCOMPLETE: 'TIN',
+	OVERPAID: 'TOP',
+};
+
+export const ALL_STATUSES = [
```
tn3rb: for consistency... do we want to use ALL CAPS for a constant that is not a single value?

Or should the constants above for the status ID arrays be capitalized ??? 

ie: `EVENT_STATUS_ARRAY` 

as well... since there are filters now for JS, maybe we should slap a filter on this?

assets/src/data/model/status/constants.js
nerrad: I disagree, this is not an array it is an object.  When implemented in code the usage will look something like:

```
emailStatus.SENT
```

If there's cases where client code may want to name a variable to represent status the import can assign the constant to a different name a-la:

```
import { emailStatus as someOtherName }
```

So to sum up, I picked the name because its _typical_ usage makes more sense with that label.

assets/src/data/model/status/constants.js
nerrad: Convention in javascript is that usage of `ALL_CAPS` is a signal that the value is intended to be immutable because `const` in javascript simply prevents reassigning (not immutability).  So it's acceptable to use CAPS in this case as a signal that it is considered immutable.  I didn't capitalize the individual model status arrays because I figured they are an object representation that _may_ be mutable (i.e. new temporary status added to the object).  However the individual object keys ARE all caps to signal immutability for the strings.

Based on that description though, I'm open to changing `emailStatus` to `EMAIL_STATUS`.  This would have the added benefit of still allowing client code to do something like:

```
let emailStatus = EMAIL_STATUS.SENT
```

> FWIW, I think the *_STATUS_ID_ARRAY schema is the most accurate and most consistent format, but I'll let you decide in case you think of something even better.

As mentioned earlier I disagree for all the reasons I already gave in my previous comment.  They are objects intended for easily accessing the individual statuses (similar to constant).

> One final thought... since there are filters now for JS, maybe we should slap a filter on this array so it can be modified at runtime?

Ya good thought, however within javascript I'd prefer to really be hardline on not implementing a filter and/or action until we've got an actual use-case implementing it.  Reason being is we _may_ find that the implementation of the filter/action might end up being done differently than we initially thought (so as to limit its scope if necessary etc).

assets/src/data/model/status/constants.js
tn3rb: doh... right.. that's an object :facepalm:

assets/src/data/model/status/constants.js
tn3rb: kk good points... 

do we still want to be pedantic about this being an object containing status **ID**s, ie: `EMAIL_STATUS_ID` vs `EMAIL_STATUS`  ???

I'll leave that and whether or not to imply immutability by capitalizing the object names up to you.

Issue #509, opened 2018-06-07
GDPR - Remove phone number from the EE Contact when person data erase request is fulfilled.


Pebblo:
The 'Phone number' answer is not removed from the EE contact when using WP's new GDPR erase feature.

Before - http://take.ms/cds2A

After  - http://take.ms/3dQwD

To reproduce create an event that collects Address info and add a registration using dummy data, be sure to provide the phone number.

Open up the registration in EE and note the email address for the EE Contact.

To reproduce, go to Tools -> Erase Personal Data

Create a request for the above email and force it through.

Now refresh the registration/contact and the Phone number will still be present.

mnelson4:
the PR was merged into master

PR #567, opened 2018-07-04
Gdpr/phone not remoevd

mnelson4:
https://github.com/eventespresso/event-espresso-core/issues/509

EE core code

## How has this been tested

Make sure you have a registration with a phone number. Create a data erasure request for that user's email. Erase their data. Their phone number should be erased too

## Checklist

* [x] My code is tested.
* [ ] My code accounts for when the site is Maintenance Mode (MM2 especially disallows usage of models) --> I actually realized there's an issue with this, but the issue also exists in master, see https://github.com/eventespresso/event-espresso-core/pull/568


Pebblo:
I've confirmed this works as expected and the phone number is now removed.

@mnelson4 just to note, the change log item looks incorrect above.

mnelson4:
> @mnelson4 just to note, the change log item looks incorrect above.

ooOOups, ya. I'll fix that

mnelson4:
another review required since previous changes. I don't think re-testing is necessary though, right?

nerrad:
No I don't think retesting is needed.

PR #568, opened 2018-07-04
Bug/dont erase data in maintenance mode

mnelson4:
An issue discovered on https://github.com/eventespresso/event-espresso-core/pull/567#pullrequestreview-134457119, which is that if you try to use WP's privacy export or erasure tools while EE is in maintenance mode, you'll get a fatal error (because we're querying the EE tables when they might not be setup)

EE core code

## How has this been tested

* [ ] Create a data export request as usual. Then go into the DB and manually change the WP option `ee_maintenance_mode` to have value `2` (and if it doesn't exist, create the option). Then try a data export. There should be no error, but the EE data shouldn't appear in the export.
* [ ] Continuing from the previous test, also attempt a data erasure request. There should be no error, but EE data shouldn't be deleted either.
* [ ] Manually visit the EE maintnenace page. When you do this, EE will see no migrations are required, and will take the site out of maintenance mode. Try again to export and erase a registrant's data. They should both work fine again

## Checklist

* [x] My code is tested.


mnelson4:
oh my bad- I totally didn't notice the check for `models_can_query` in `EE_System` and just thought we'd run into a fatal error. So this actually isn't necessary at all. Sorry and thanks!

PR #581, opened 2018-07-16
Force curl to use TLS1.2 when connecting to PayPal.

Pebblo:
See:
https://eventespresso.com/topic/error-http_request_failed-curl-error-35-error14094410ssl-routinesssl3_read_/
https://eventespresso.com/topic/paypal-pci-issue/
https://eventespresso.com/topic/error-offsite-payment-method-was-not-configured-properly/

These users all seem to be using Bluehost and running into the same problem when connecting to PayPal. They have older version of cURL and OpenSSL which apparently can't be updated, running tests on the site using the TLS 1.2 plugin and also some custom code show the connection is using TLS1.2.

Discussed this with Josh a little in slack, we were both in favor of removing this completely but I did a little looking around and found Woo still have almost the exact same code, so I figure these still some setups were the default value wont work and this is still needed?

I saw Mikes comments in a couple of WP trac threads on this so this should probably be assigned to him. I also consider this High as users payments are being blocked.

### DIFF
--- a/core/EE_Payment_Processor.core.php
+++ b/core/EE_Payment_Processor.core.php
@@ -853,7 +853,7 @@ class EE_Payment_Processor extends EE_Processor_Base implements ResettableInterf
         if (strpos($url, 'https://') !== false && strpos($url, '.paypal.com') !== false) {
             // Use the value of the constant CURL_SSLVERSION_TLSv1 = 1
             // instead of the constant because it might not be defined
-            curl_setopt($handle, CURLOPT_SSLVERSION, 1);
+            curl_setopt($handle, CURLOPT_SSLVERSION, 6);
         }
     }
 }


Pebblo:
Mike, we discussed this a little more in Slack and this snippet came up:

https://github.com/woocommerce/woocommerce-gateway-stripe/commit/877dcabf5c60ef6e41f770ecf654274fa1fc0ce5

That's no longer used and Darren traced it back this comment:

https://github.com/woocommerce/woocommerce-gateway-stripe/issues/45#issuecomment-248846989

So, the question now is, should we just remove this completely or use the above?

mnelson4:
lemme think through all the different variables: 
* requests going to non-paypal-servers will continue to negotiate SSL/TLS version according to their openSSL version's default. (so for WP core, using the openSSL version's default value is probably best option because it usually makes the best choice)
* for servers using a version of openSSL that doesn't support TLS 1.2, it doesn't matter what value we set, requests to PayPal will never work from now on
* versions of CURL below 7.34.0 won't understand `CURL_SSLVERSION_TLSv1_2` (value of 6) for `CURLOPT_SSLVERSION` as that's when support it was added (see https://linux.die.net/man/3/openssl_version_number). 
* it appears some versions of openSSL when you set `CURLOPT_SSLVERSION` to `CURL_SSLVERSION_TLSv1` (value of 1) doesn't properly start using TLS 1.2, despite that's what it's supposed to do

So I think the summary is: 
* if we set `CURLOPT_SSLVERSION` to 6, with older versions of openSSL and CURL, when that value isn't yet handled, I'm not sure if it will work. But for newer versions of openSSL and CURL it will work
* if we set  `CURLOPT_SSLVERSION` to 1, it theoretically should always work for all versions of openSSL and CURL, except it didn't for all those experienced this issue

I think that's why https://github.com/woocommerce/woocommerce-gateway-stripe/commit/877dcabf5c60ef6e41f770ecf654274fa1fc0ce5 had the conditional: try to use a value of 6 for newer openSSL/CURL, a value of 1 for older openSSL/CURL. But like it was said on https://github.com/woocommerce/woocommerce-gateway-stripe/issues/45#issuecomment-248846989, PHP doesn't actually know what version of openSSL CURL is using, so we never know which version to use. However, the advice to just not set `CURLOPT_SSLVERSION` is general advice, when you don't know what versions of SSL/TLS the server supports. In this special case, we know which version is supports: TLS1.2. So if we can force TLS 1.2 or higher in communications with PayPal, we should.

So the main question is: what happens if we set `CURLOPT_SSLVERSION` to 6 while openSSL is on 1.0.1 (ie it supports TLS 1.2) but CURL is less than 7.34.0? If that works ok, this solution is fine. If that causes an error, this probably needs to be changed... now how to find out...

openSSL supports TLS1.2 as of 1.0.1

mnelson4:
Looking back at https://core.trac.wordpress.org/ticket/34924, I'm reminded that it seems setting `CURLOPT_SSLVERSION` to 1 actually DID help CURL 7.29 users (whereas NOT specifying any version caused a connection error).
So it seems setting `CURLOPT_SSLVERSION` to 1 fixes some problems (mentioned on https://core.trac.wordpress.org/ticket/34924) but causes others (mentioned at the beginning of this ticket).
Setting `CURLOPT_SSLVERSION` to 6 will either force CURL/openSSL to use TLS1.2 (not just an earlier version of TLS, which setting `CURLOPT_SSLVERSION` to 1 sometimes did) when CURL/openSSL supports it; or if CURL doesn't support it, it gets ignored and so openSSL is free to negotiate the TLS/SSL version to the best of its abilities.
So I'm in favour of this commit as-is.
Any holes in my logic?

mnelson4:
oh, plus this is exactly what WooCommerce is currently doing: https://github.com/woocommerce/woocommerce/blob/bfebd305654919629d5d8e32782ef0d249416991/includes/class-wc-https.php#L133 (I think @Pebblo mentioned that, but here's the link)

mnelson4:
FYI: in Slack @joshfeck said this was fine by him

mnelson4:
Ya ideally we’d test this with curl 
* [] <7.35 (I can do that quite easily on my bluehost account)
* [] >=7.35.

Not much need to test old versions of OpenSSL that don’t support TLS 1.2, as that absolutely needs to be updates in order to use PayPal.

joshfeck:
cURL version 7.47.0 tested, no errors
screen shot 2018-07-17 at 12 33 14 am

Pebblo:
I checked on one of my test sites (which already works fine before this change) and I have:

`curl 7.19.7 (x86_64-redhat-linux-gnu) libcurl/7.19.7 NSS/3.27.1 zlib/1.2.3 libidn/1.18 libssh2/1.4.2`

Works fine there. so good to go @mnelson4 ?

mnelson4:
Yeah @Pebblo I just tested this exact branch on my Bluehost site and it worked fine

PR #590, opened 2018-07-18
Tweak capability checks to allow readonly for registration form answers

joshfeck:

Registration form answers aren’t displayed for a restricted user. Only registration status, registration details and contact details are displayed if the user account has `ee_read_registration` and does not have `ee_edit_registration` capabilities. 

It was also pointed out that the inputs for changing the reg. status are displayed if the user does not have the `ee_edit_registration` cap. If they try to change the reg. status they get an error. So we could remove those from the view too.

No way, José.

## How has this been tested

Tested by ensuring a user account with both `ee_edit_registration` and `ee_read_registration` can still edit answers. Then logged into a user account with `ee_read_registration` and no `ee_edit_registration` and verified they can view (read but no ability to edit) the answers to registration form questions and no reg. status change inputs appear.
## Checklist

* [x] I will add a changelog entry for this pull request when merging
* [x] My code is tested.


nerrad:
Great stuff!

joshfeck:
This needs testing. 

To test: View a Event Espresso > Registration in the admin while logged in as a regular admin user or as an Event admin that has the `ee_edit_registration` capability. **There should be no change to the admin page** compared to master.

Then, switch to a user account that has all of the `ee_read_registration` type capabilities, including `ee_read_others_registration`, but does not have the `ee_edit_registration` type capabilities.

The expect result will be no inputs to change the Registration status, and no links to edit or update the registration answers for the user with no `edit_registration` capabilities.

joshfeck:
You'd actually need to `ee_edit_payments` capability in order to stop the Event Admin from making changes to the transaction. If you remove that cap can you add/edit payments on the Transactions admin page?

Issue #579, opened 2018-07-13
x_currency_code and Authorize.net AIM


lorenzocaum:
In the Authorize.net AIM payment gateway, I saw a few currencies like these:

AUD, USD, CAD, EUR, GBP, or NZD.

However, when testing, I don't see the x_currency_code parameter in use in the logs but I do see others: https://cl.ly/3P2T2M233Q0q

From this doc (https://www.authorize.net/content/dam/authorize/documents/AIM_guide.pdf), you can use x_currency_code to accept payments in another currency even if your merchant account has a default set. For example, you can have a default of USD on the merchant account and then you should be able to use the parameter to accept payments in another currency like EUR.

What happens right now (assuming your Authorize.net account is set to USD) is that if your site is set to a country in Europe, then you'll see Euros for the currency but then Authorize.net will run the transaction in the default currency of USD.

sethshoultes:
@knazart or @mnelson4 thoughts on this issue?

mnelson4:
yeah good point @lorenzocaum. I just checked the code and we're definetely not sending the currency code to AIM, or Authnet Accept actually. I think we've gotten by because most people set the same currency on their Authorize.net account as what's in EE. But if someone set that incorrectly, it would really stink if they're getting paid in the wrong currency!
So ya, I think we should address this, and it should be plenty easy to pass the currency code along.
But before we jump in and fix this, I should ask: could any site owners be relying on the current behaviour? Eg, their Authnet account is in Euros, but their site's location is still set to the default USA location, and they've just changed USA's currency code to €: so they're expecting to be paid in Euros and they currently are, and when we make this change they'll suddenly start getting paid in USD.
IMO this situation is possible, but it's probably still best to make this fix.

lorenzocaum:
That scenario is a possibility but I think the occurrence would be low as compared to most event organizers that expect the currency to be consistent with the selected country by default.

Using the currency that is linked to the base of operations through the general settings page would likely be the best choice for most event organizers.

@Pebblo Could you share a recommendation here for how we determine the currency to use?

knazart:
Yeah, not sending a currency in this situation is a no-no. But in the DOCs the `x_currency_code` parameter is marked as optional and that makes me remember that some gateways we encountered along the way, in fact, didn't allow to override/change the default (set in the merchant account) currency when creating a transaction through the API. I am not sure that this is the case here, but there is a possibility that this parameter was added/updated after this gateway was developed.

[...]

PR #591, opened 2018-07-18
pass currency code for AIM

mnelson4:

See https://github.com/eventespresso/event-espresso-core/issues/579

It seems AIM added support for passing currency back in 2013, see https://community.developer.authorize.net/t5/Integration-and-Testing/Multi-currency-support/td-p/34758

However, it's not clear if they allow you to pass a currency that's different from the Authorize.net account's currency. https://community.developer.authorize.net/t5/The-Authorize-Net-Developer-Blog/Authorize-Net-UK-Europe-Update/ba-p/35957 says 
> One gateway account is required for each currency

and https://community.developer.authorize.net/t5/Integration-and-Testing/Multi-currency-support/m-p/34765/highlight/true#M19188 says 

> Authorize.Net is fully live with GBP, Euros, CAD and USD but we currently do not support multi-currency in one account. If a merchant wants to accept multiple currencies, example GBP and EUR, the merchant will need two Authorize.Net Payment Gateway accounts one for GBP and another for EUR transacting merchants.

## How has this been tested

* [x] Make a payment with AIM while country is set to USA
* [x] Make a payment with AIM while country is set to Canada (it won't work because AIM sandbox only works with USD)
* [ ] I'd like the customer who thinks they're setup for multi-currency to validate this works for them before we make this change. Because like I noted on the related ticket, it's possible some users are relying on the existing behaviour (eg, their Authorize.net account is setup for CAD but their country in EE is still the USA default; so currently they're receiving funds in CAD but this change will make them start receiving funds in USD, or make payments stop entirely, which they might not expect.)

## Checklist

* [ ] My code is tested.


mnelson4:
@lorenzocaum would you like to test this? (See https://github.com/eventespresso/event-espresso-core/pull/591#issue-202341496 under "How has this been tested" for testing notes.)
It would be nice to have the user with the issue test it too... maybe I could cherry pick these commits into eventsmart after initial testing, so they can do that.

lorenzocaum:
Hi Mike, I ran the first two tests:

- Test Authorize.net with base of operations for United States
- Test Authorize.net with base of operations for another country. I used Canada.

Both worked as expected and on the second one, I saw reason code 39 (https://cl.ly/242l2x272g2B).

Lets work on the cherry pick for the commit to ES tomorrow.

mnelson4:
@lorenzocaum would you like to wait on the customer confirming this helps them before merging to master?

lorenzocaum:
Yes, that plan sounds good. I followed up with them yesterday and I'm waiting to hear back.

mnelson4:
On https://github.com/eventespresso/eventsmart.com-website/issues/64#issuecomment-410313267 @lorenzocaum noted the customer seems to have abandoned trying to use two currencies from one Authnet account. So the feature they wanted doesn't seem possible.
So, does this PR still have a purpose? I think so: if folks setup their Authnet account to use Euros, and then they setup EE to charge in USD, it will come as a surprise to attendees when they intended to pay $10 USD and are instead charged 10 Euros. On this branch, they will instead get an error. IMO it's better to have an error than unexpectedly pay double.
So @lorenzocaum let's merge this to EE master?

lorenzocaum:
Hi Mike, yes please merge this in because as you mentioned, it will still be helpful.

PR #592, opened 2018-07-18
even if the ee_error_js is already enqueued, we should still localize… Fixes #589

### LINKED ISSUE #589: Javascript error on pages with an error message and persistent notice
## Issue Overview

If there is a page where both `EE_Error::add_error()` and `EE_Error::add_persistent_admin_notice()` are used, there will be a javascript error saying `ReferenceError: ee_settings is not defined` pointing to line  `EE_Error.js:30`. This javascript error can cause other javascript (like media selector) to not work.

## Bug report or feature request?

* [x] Bug
* [ ] Feature
* [ ] Neither

## Environment Data:

Version of EE: 4.9.66.rc.000

Version of WordPress: 5.0-alpha-43320

PHP Version:   7.1

Browser used:  firefox

## Steps to Reproduce (for bugs)

To get a persistent admin notice, activate an add-on with a payment method, like the braintree gateway add-on. Then activate its payment method. Then deactivate the plugin/add-on (NOT the payment method). Then visit the payment method page (this is where EE will realize the payment method is missing even though it wasn't explicitly deactivated) and will create the persistent admin notice which you'll see on the NEXT page load.
To get an error notice, you can just add `EE_Error::add_error()` inside `EE_System::initialize()`. 
Once both of these are appearing, use your browser's web inspector to see the javascript error on pageload. You should see it in there.

## Expected Behaviour

There should be no javascript error.

## Current Behaviour

There is a javascript error.

## Related Information:

I think the problem stems from `PersistentAdminNoticeManager::enqueueAssets()` enqueues the script with handle `ee_error_js`, but its localization data is given the name `ee_dismiss`. When `EE_Error::_print_scripts()` runs (which I've double-checked, it does in this case), and it finds the script with handle `ee_error_js` has already been enqueued, and so doesn't localize `ee_settings`.
If that method were to still call `wp_localize_script('ee_error_js', 'ee_settings', array('wp_debug' => WP_DEBUG));`, even when `ee_error_js` was already enqueued, it would avoid the javascript error.

mnelson4:
… its script

Fixes #589 

## How has this been tested

See the ticket's "steps to reproduce" section

## Checklist

* [x] My code is tested.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,7 +29,7 @@ Headings should be (only add headings as needed):
 -  Fixed privacy erasure so contact phones are also erased ([567](https://github.com/eventespresso/event-espresso-core/pull/567))
 -  Fixed fatal error thrown when using WP4.5 ([571](https://github.com/eventespresso/event-espresso-core/pull/571))
 -  Fixed issue with [ESPRESSO_EVENT_ATTENDEES] shortcode to allow displaying attendees when specifying an archived ticket ([583](https://github.com/eventespresso/event-espresso-core/pull/583))
-
+-  Fixed a javascript error on admin pages showing both an EE error and persistent notice. ([592](https://github.com/eventespresso/event-espresso-core/pull/592))
 ## [4.9.64.p]
 
 ### Added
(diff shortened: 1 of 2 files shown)


mnelson4:
need to add checklist item...

PR #593, opened 2018-07-19
Add support for INR to PayPal Express and Pro - Fixes #585

### LINKED ISSUE #585: INR currency for India and PayPal Express
If you update your currency to India and then enable PayPal Express, then you'll see this message:

`This payment method does not support the currency set on your site (INR). Please activate a different payment method or change your site's country and associated currency.`

- You can pick up a merchant account in India with PayPal now. See this screen: https://www.paypal.com/in/webapps/mpp/account-selection
- The API doc has an entry for INR now: https://developer.paypal.com/docs/classic/api/currency_codes/

I think the source of that message is this line in PayPal Express: https://github.com/eventespresso/event-espresso-core/blob/master/payment_methods/Paypal_Express/EEG_Paypal_Express.gateway.php#L79

--
https://secure.helpscout.net/conversation/622381056/548487/

Pebblo:
Adds support for INR currency to PayPal Express and Pro.

These payment methods are a part of core.

## How has this been tested
The code change itself has been tested and doesn't throw any errors, but purchasing using INR has not.

## Checklist

* [x] My code is tested.


Pebblo:
Thanks @mnelson4, added the changelog entry.

CHANGELOG.md
```diff
@@ -17,6 +17,10 @@ See [our documentation](https://github.com/eventespresso/event-espresso-core/blo
 
 ## [$VID:$]
 
+### Added
+
+- Added support for Indian Rupees currency (INR) to PayPal Espresso and Pro ([593](https://github.com/eventespresso/event-espresso-core/pull/593)) 
+
```
mnelson4: lookgs good

Pebblo:
I'm not having any luck trying to test an INR payment.

I created 2 sandboxes for India (buyer and seller) and tried to make a payment, I can get through to the PayPal checkout but it stops with 'Things don’t appear to be working at the moment.'

@lorenzocaum do you have an Inida account you can test with?

lorenzocaum:
@Pebblo I created a sandbox account but it wasn't working because I used an email that had a "+" in it.

I created new accounts with another set of email addresses but during the payment process it asks for a CVV which isn't in the sandbox profile and after a couple of attempts, it shows this:

https://cl.ly/0P2G2G0i0p21

lorenzocaum:
I had a look at the logs through Event Espresso and I noticed that we are using US for the locale (https://cl.ly/0W1p0X2q1k0K):

https://developer.paypal.com/docs/classic/api/locale_codes/#supported-locale-codes

I noticed it in the URL so I manually changed the country code to `IN` and local code to `EN_GB` to match the information in the link above for India but no luck with getting checkout to work.

Pebblo:
That's the same problem I got.

Everything seems to be working fine until you log into the PayPal account to make a payment with and it just dies from there.

>I had a look at the logs through Event Espresso and I noticed that we are using US for the locale 

As far as I can tell thats just for the checkout locale and it part of some other factors EE uses to determine the locale:

https://developer.paypal.com/docs/classic/api/locale_codes/#localization-factors

@mnelson4 do you have any thoughts on this?

As far as EE is concerned I can't see any other changes needed and we could probably release this, right now it just seems to be an issue with the sandbox accounts?

mnelson4:
I came across some info, and using INR is actually tricky. See the attached document from PayPal.
[enabling-inr-payments-direct-website-ec.pdf](https://github.com/eventespresso/event-espresso-core/files/2250310/enabling-inr-payments-direct-website-ec.pdf) Only folks living IN India can pay in INR! So they suggest allowing everyone else in pay in a different currency. We, of course, don't actually support multiple currencies though. (So unless someone hacks EE core to somehow add that, they'd need to setup a separate site using a different currency, and redirect non-Indians to it. Unless someone has a better idea).
So, I think we're having trouble testing with INR because of the that, and there's some extra setup steps mentioned in that PDF that we haven't done.

So I think we should just test regular payments in a differnet currency still work, and maybe add some more documentation. Thoughts @lorenzocaum ?

lorenzocaum:
Okay, here is a plan to move forward with this.

1) Does this branch need to be updated with master? If so, could someone take care of that? 

2) Then please post a note here and I'll then reach out to the person that asked about this and ask them to test

Pebblo:
@lorenzocaum master has been merged into this branch, merging to master is blocked due to a review being required but this branch has what is basically a text change so can be tested as is.

lorenzocaum:
Hi Mike, I reached out the the customer that reported this for feedback. You can go ahead and merge this. Thanks!

PR #596, opened 2018-07-19
add changelogs to skeleton addons

mnelson4:

Adds changelogs to skeleton addons in tests/mocks/addons

## How has this been tested

It hasn't, but this file isn't included in EE core releases, and it's not used in add-ons anyways either.

## Checklist

* [ ] I have added a changelog entry for this pull request -> seeing how this isn't included in the released version, I don't think it should have a changelog entry
* [ ] My code is tested.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ Headings should be (only add headings as needed):
 - Fixed (bug fixes)
 - Security (vulnerability fixes)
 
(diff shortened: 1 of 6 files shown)


mnelson4:
Yeah I think linking to the docs would be good, so I took a stab at it above. I also added an exception to "every PR needs a changelog entry" which I think makes sense, but it may be up for debate.

Issue #360, opened 2018-04-26
Allow page builder requests to have template_tags.php loaded


joshfeck:
@sethshoultes mentioned this in Slack:
> i have both EE4 and Beaver Builder activated locally but i am getting a fatal error when trying to use the page builder on an event page:

```
Fatal error: Call to undefined function espresso_venue_id() in /x/wp-content/plugins/event-espresso-core-github/core/helpers/EEH_Schema.helper.php on line 70
```

I've seen something similar while working with the Elementor page builder plugin:

```
Fatal error: Uncaught Error: Call to undefined function espresso_list_of_event_dates() in /x/wp-content/plugins/event-espresso-core/widgets/upcoming_events/EEW_Upcoming_Events.widget.php:394 
Stack trace: 
#0 /x/wp-content/plugins/elementor/includes/widgets/wordpress.php(244): EEW_Upcoming_Events->widget(Array, Array) 
#1 /x/wp-content/plugins/elementor/includes/base/widget-base.php(462): Elementor\Widget_WordPress->render() 
#2 /x/wp-content/plugins/elementor/includes/base/widget-base.php(591): Elementor\Widget_Base->render_content() 
#3 /x/wp-content/plugins/elementor/includes/base/element-base.php(662): Elementor\Widget_Base->get_raw_data(true) 
#4 /x/wp-content/plugins/elementor/includes/base/element-base.php(662): Elementor\Element_Base->get_raw_data(true) 
#5 /x/wp-content/plugins/elementor/core/base/document.php(521): Elementor\Element_Base->get_raw_data(t in /x/wp-content/plugins/event-espresso-core/widgets/upcoming_events/EEW_Upcoming_Events.widget.php on line 394
```

One possible solution would be to tweak the conditionals that determine whether to include the template tags file.

The underlying issue is the page builders are trying to load content as if it was the front end of the site, but it’s not the front end of the site.
Here’s the conditional from EE_System.core.php:

```php
if (
    is_readable(EE_PUBLIC . 'template_tags.php')
    && ($this->request->isFrontend() || $this->request->isIframe() || $this->request->isFeed())
) {
    require_once EE_PUBLIC . 'template_tags.php';
}
```

so one idea could be to look at tweaking what’s defined as `isFrontend()`

and/or add a filter switch to that conditional to be used by a compatibility shim.

Pebblo:
Another report here: 

[...]

PR #600, opened 2018-07-20
just always load template tags to avoid problems in page builders

mnelson4:

https://github.com/eventespresso/event-espresso-core/issues/360
Fixes a fatal error while using many page builder plugins resulting from template tags only being loaded on the front-end

Appropriate question for once. It was suggested this could be handled by an add-on or shim or code snippet, and we may go for that. But these are the reasons I thought it best to fix this issue in EE core and just always include the template tags, even on admin requests:
* loading the template tags file takes about 1 millisecond, and the conditional to determine if we should include it also takes about 1 second. Because it takes just as long to check as it does to just include it, no speed is saved by checking in the admin. In fact, it's a hair-width quicker to NOT check and just always include it
* it will be a bit of a pain to list out all the page builders out there, this avoids that
* when new page-builders are introduced, we don't have to include it in the whitelist

## How has this been tested

* [ ] Use elementor plugin, add an upcoming events widget to a page and try to save it. Upon reloading the page, I got the error from the issue. But on this branch it works fine
* [ ] Seth said he reproduced the issue in Beaver Builder (not sure what steps he took to reproduce)
* [ ] A general scan of the EE admin would be good too, just in case this somehow has unexpected consequences in there
* [ ] Test a theme that customizes EE's default templates. They should continue to work as normal.

## Checklist

* [x] My code is tested.

### DIFF
--- a/core/EE_System.core.php
+++ b/core/EE_System.core.php
@@ -1161,15 +1161,9 @@ final class EE_System implements ResettableInterface
         // integrate WP_Query with the EE models
         $this->loader->getShared('EE_CPT_Strategy');
         do_action('AHEE__EE_System__core_loaded_and_ready');
-        // load_espresso_template_tags
-        if (($this->request->isFrontend()
-             || $this->request->isIframe()
-             || $this->request->isFeed()
-             ||  $this->request->isAjax()
-            ) && is_readable(EE_PUBLIC . 'template_tags.php')
-        ) {
-            require_once EE_PUBLIC . 'template_tags.php';
-        }
+        // always load template tags, because it's faster than checking if it's a front-end request, and many page
+        // builders require these even on the front-end
+        require_once EE_PUBLIC . 'template_tags.php';
         do_action('AHEE__EE_System__set_hooks_for_shortcodes_modules_and_addons');
     }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,9 @@ See [our documentation](https://github.com/eventespresso/event-espresso-core/blo
 
 ## [$VID:$]
 
+### Fixed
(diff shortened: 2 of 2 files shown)


core/EE_System.core.php
```diff
@@ -1161,15 +1161,9 @@ public function core_loaded_and_ready()
...
-             || $this->request->isIframe()
-             || $this->request->isFeed()
-             ||  $this->request->isAjax()
-            ) && is_readable(EE_PUBLIC . 'template_tags.php')
-        ) {
-            require_once EE_PUBLIC . 'template_tags.php';
-        }
+        // always load template tags, because it's faster than checking if it's a front-end request, and many page
+        // builders require these even on the front-end
+        require_once EE_PUBLIC . 'template_tags.php';
```
nerrad: should there still be a `is-readable` check though?

mnelson4: good question. Under what circumstances, besides someone hacked EE core by removing or moving that file, would the file NOT be readable?
I see the constant `EE_PUBLIC` is defined in `core/espresso_definitions.php` and is always `define('EE_PUBLIC', EE_PLUGIN_DIR_PATH . 'public' . DS);`, where `EE_PLUGIN_DIR_PATH` is always the plugin's directory path. (Maybe at one point it was dynamic, but from the code I read it doesn't get much more static than this.)
From what I read, that file will always be present unless someone has hacked EE core. Right? If so, I'm inclined to not protect against someone hacking EE core, and save a bit of processing to check if the file exists.

nerrad: works for me.

mnelson4:
Passing to @joshfeck as you created the related issue (although you may prefer to pass onto Seth). I put some suggested testing notes under the title "How this has been tested" above

joshfeck:
This fixes the issue reported earlier when trying to edit a page with Elementor, where the page has an embedded Event Espresso Upcoming Events widget.

PR #602, opened 2018-07-24
Add Calculated Fields class for Attendee model

tn3rb:
need to be able to retrieve attendee related data via the API that is not immediately available via the models. For example: Attendee Gravatars

## How has this been tested
only in usage in the Event Attendees Gutenberg block branch

### DIFF
--- a/core/libraries/rest_api/CalculatedModelFields.php
+++ b/core/libraries/rest_api/CalculatedModelFields.php
@@ -55,6 +55,7 @@ class CalculatedModelFields
         $event_calculations_class = $rest_api_calculations_namespace . 'Event';
         $datetime_calculations_class = $rest_api_calculations_namespace . 'Datetime';
         $registration_class = $rest_api_calculations_namespace . 'Registration';
+        $attendee_class = $rest_api_calculations_namespace . 'Attendee';
         return apply_filters(
             'FHEE__EventEspresso\core\libraries\rest_api\Calculated_Model_Fields__mapping',
             array(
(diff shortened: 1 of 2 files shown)


core/libraries/rest_api/calculations/Attendee.php
```diff
@@ -0,0 +1,65 @@
...
+     * @throws InvalidDataTypeException
+     * @throws InvalidInterfaceException
+     */
+    public static function userAvatar(array $wpdb_row, WP_REST_Request $request, Base $controller)
+    {
+        $attendee = Attendee::getAttendeeObject($wpdb_row);
+        if (! $attendee instanceof EE_Attendee) {
+            return '';
+        }
+        $avatar = get_avatar_url($attendee->email());
```
mnelson4: if we're already needing to dig into the array to get the `Attendee_CPT.ID` entry, we could save a little processing by instead grabbing the `Attendee_Meta.ATT_email` value, and passing that to `get_avatar_url()`, no?

core/libraries/rest_api/CalculatedModelFields.php
```diff
@@ -82,6 +83,9 @@ protected function generateNewMapping()
                 'Registration' => array(
                     'datetime_checkin_stati' => $registration_class,
                 ),
+                'Attendee' => array(
+                    'userAvatar' => $attendee_class,
```
mnelson4: this string `userAvatar` is what REST API clients will put in the querystring. The convention so far has been to have those be snake case. However, the method on the class can remain camelcase- `CalcualtedModelFields::retrieveCalculatedFieldValue()` takes care of translating the snake-case string into a camelcase one to determine which method to use.

tn3rb:
> the private method was unneeded

oh I left that in on purpose because I figured it would be handy if not necessary for any further additions to this class. No problem removing it though. thnx

tn3rb:
oh and went ahead and merged this since it does not interact with any other code

PR #608, opened 2018-07-28
improper use of str_replace

mnelson4:
… for individual characters like other functions. This is how string are normalized throughout most of EE, and it works assuming DS continues to just be a slash

 Fixed a syntax issue inside `EE_Config::register_ee_widget()`.
Noticed this mistake while on an experimental branch of EE (it was causing an error which prevented EE from finding the promotion addon's widget). However, on EE master, this didn't present a problem so it's lower priority (kinda coincidentally, the details are a bit dizzying and I think beside the point). But we may as well fix it: we accidentally passed a list of characters to PHP `str_replace()` in the first argument... except it doesn't accept a string of characters (like some other methods do), it just accepts a single needle, or an array of needles.

## How has this been tested

* [x] EE core widgets are still available and work
* [x] The promotion's widget is still available and works

## Checklist

* [ ] My code is tested.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@ See [our documentation](https://github.com/eventespresso/event-espresso-core/blo
 
 ### Fixed
 - Fixes a fatal error while using many page builder plugins resulting from template tags only being loaded on the front-end([600](https://github.com/eventespresso/event-espresso-core/pull/600))
-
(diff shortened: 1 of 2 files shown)


mnelson4:
Passing to Brent now, seeing how Darren's gone (probably more in his domain anyway, just originally passed to Darren in keeping with "Don't bother Brent when possible" policy)

mnelson4:
testing notes above, under "How has this been tested".
Passing to @sethshoultes to hand off to a tester

PR #614, opened 2018-07-30
Gutenberg Route Specific Block Loading

tn3rb:
Adds a new system that allows for the definition of "routes" that can then be compared against the current request to determine if they match. 
This can be used to control loading for any aspect of EE with regards to the current request. 
Blocks now define their supported routes instead of their supported post types and the new system is incorporated into the `BlockRegistrationManager` in order to compare those against the current request and stop Block loading if none of its supported routes match.

## How has this been tested
in usage only. Once initial review/code direction is agreed upon I will add unit tests.


core/domain/entities/route_match/specifications/admin/WordPressPostsListTable.php
```diff
@@ -0,0 +1,31 @@
...
+     *
+     * @since $VID:$
+     * @return boolean
+     */
+    public function routeMatches()
+    {
+        return strpos($this->request->requestUri(),'wp-admin/edit.php' ) !== false
+            && (
+                $this->request->getRequestParam('post_type') === 'post'
+                || $this->request->requestParamIsSet('post_type') === false
```
nerrad: In native wp posts lists table loading, it appears the page will invoke `wp_die` if either there is no `post_type` set in the request or the value is not in the array returned by `get_post_types( 'show_ui' => true )`.  I realize that this specification check is likely invoked before the native wp post list table check but I wonder if it'd be more accurate to remove this second `|| $this->request->requestParamIsSet( 'post_type' ) === false` check?

core/interfaces/FactoryInterface.php
```diff
@@ -0,0 +1,20 @@
...
+ */
+interface FactoryInterface
+{
+    /**
+     * @param string $fqcn  Fully Qualified Class Name
+     * @return object       Instantiated object
+     * @since $VID:$
+     */
+    public static function create($fqcn);
+}
```
nerrad: We already have `EventEspresso\core\domain\services\factories\FactoryInterface`  I see they differ slightly, but that seems like it could cause confusion?

core/services/route_match/RouteMatchSpecificationManager.php
```diff
@@ -0,0 +1,146 @@
...
+
+    /**
+     * Given the FQCN for a RouteMatchSpecification, will return true if the current request matches
+     *
+     * @param string $routeMatchSpecificationFqcn fully qualified class name
+     * @return bool
+     * @throws InvalidClassException
+     * @since $VID:$
+     */
+    public function currentRequestMatches($routeMatchSpecificationFqcn)
```
nerrad: I'd prefer if the name of this followed the pattern typically used for methods returning boolean types and was maybe something like `isCurrentRequestMatchingRoute`

core/services/route_match/RouteMatchSpecificationManager.php
```diff
@@ -0,0 +1,146 @@
...
+     * Handy method for development that returns
+     * a list of existing RouteMatchSpecification classes
+     * matching the current request that can be used to identify it.
+     * If no matches are returned (or nothing acceptable)
+     * then create a new one that better matches your requirements.
+     *
+     * @return array
+     * @since $VID:$
+     */
+    public function findSpecificationsMatchingCurrentRequest()
```
nerrad: I'd prefer either `findRouteSpecificationsMatchingCurrentRequest` or `findRoutesMatchingCurrentRequest` as a method name.

core/services/collections/Collection.php
```diff
@@ -35,20 +35,64 @@ class Collection extends SplObjectStorage implements CollectionInterface
      */
     protected $collection_interface;
 
+    /**
+     * a short dash separated string describing the contents of this collection
+     *
+     * @type string $collection_identifier
+     */
+    protected $collection_name;
```
nerrad: I realize that the comments say that this property is for describing the contents of this collection.  However, I think it'd also be useful to indicate in the phpdocs that the collection identifier is derived in prat from this property as well.

core/services/route_match/RouteMatchSpecificationFactory.php
```diff
@@ -0,0 +1,82 @@
...
+    private $loader;
+
+
+    /**
+     * RouteMatchSpecificationFactory constructor
+     *
+     * @param RouteMatchSpecificationDependencyResolver $dependency_resolver
+     * @param LoaderInterface                           $loader
+     */
+    public function __construct(RouteMatchSpecificationDependencyResolver $dependency_resolver, LoaderInterface $loader)
```
nerrad: I realize its still early and you might not want to act on my comment yet because of that, but this almost appears like a potential pattern that could occur for other factories that require a resolver.  So my thinking is there may be good use case for something like:

```php
abstract class FactoryWithResolver implements FactoryInterface {
    private $dependency_resolver;
    private $loader;
    public function __construct(DependencyResolverInterface $dependency_resolver, LoaderInterface $loader) {
        $this->dependency_resolver = $dependency_resolver;
        $this->loader = $loader;
    }

    public function createInstance($fqcn) {
        $this->dependency_resolver->resolveDependencies($fqcn);
        return $this->loader->getShared($fqcn);
    }
}
```

Child classes would still need to implement the static `create` method, but this could eliminate repeated boilerplate.  

Again, I know there's no demonstrated need for this additional abstraction yet so understood if you decide to ignore this comment for now.

core/EE_System.core.php
```diff
@@ -927,6 +933,25 @@ private function _maybe_brew_regular()
...
+     * @since $VID:$
+     * @throws Exception
+     */
+    public function loadRouteMatchSpecifications()
+    {
+        if ($this->request->isAdmin() || $this->request->isFrontend() || $this->request->isIframe()) {
+            try {
+                $this->loader->getShared(
+                    'EventEspresso\core\services\route_match\RouteMatchSpecificationManager'
+                );
```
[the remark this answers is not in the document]
tn3rb: > Should ajax requests be included as well 

hmmm... ya maybe I shouldn't have any restrictions on when the `RouteMatchSpecification`s load since they could conceivably be used for detecting specific AJAX routes as well assuming you know the right request params in advance. 

Will remove all restrictions.

core/domain/entities/route_match/RouteMatchSpecificationInterface.php
```diff
@@ -0,0 +1,22 @@
...
+ */
+interface RouteMatchSpecificationInterface
+{
+    /**
+     * returns true if current request matches specification
+     *
+     * @since $VID:$
+     * @return boolean
+     */
+    public function routeMatches();
```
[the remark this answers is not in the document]
tn3rb: ya that works... I think I initially had thought along the lines of `isRouteMatch` or  `isRouteMatching` and other variations that stank, but missed  `isMatchingRoute`.

Will change.

core/domain/entities/route_match/specifications/admin/WordPressPostsListTable.php
tn3rb: > In native wp posts lists table loading, it appears the page will invoke wp_die if either there is no post_type set in the request or the value is not in the array returned by get_post_types( 'show_ui' => true ).

mmmm... not seeing that in `WP_Posts_List_Table` nor in `WP_List_Table`, where is that happening?
 
`WP_Posts_List_Table` gets the post_type from the current `WP_Screen`, which we could do if we loaded the `RouteMatchSpecification` classes much later in the request, but that would obviously make them less usable for making loading decisions earlier on. It's possible we could work the use of `WP_Screen` into them via another method that can only be called after a certain hook point has run. Other implementers of `WP_List_Table`, such as `WP_Comments_List_Table` neither require `post_type` to be set as a request parameter nor as a property on `WP_Screen` (ie: the comments list table for standard WP Posts will not have `post_type` set at all)

>  I wonder if it'd be more accurate to remove this second || $this->request->requestParamIsSet( 'post_type' ) === false check?

Well in some cases the full url for the WP Posts list table is simply `http://local.wordpress.test/wp-admin/edit.php` with no post type specified, but other times, depending on where you navigate from (such as by using the `All`, `Published`, `Draft` links in the table header), it will be `http://local.wordpress.test/wp-admin/edit.php?post_type=post`.  Other post types appear to always display the `post_type` in the url, which subsequently gets added to the `WP_Screen` object. 

What we could do, is utilize the second parameter for the `RequestInterface::getRequestParam()` method which allows you to set a default value **if** the parameter is not set in the request parameters. So instead of:

```php
        return strpos($this->request->requestUri(),'wp-admin/edit.php' ) !== false
            && (
                $this->request->getRequestParam('post_type') === 'post'
                || $this->request->requestParamIsSet('post_type') === false
            );
```

we could simply do something like:

```php
        return strpos($this->request->requestUri(),'wp-admin/edit.php' ) !== false
            && $this->request->getRequestParam('post_type', 'post') === 'post';
```

which would essentially do the same thing (assume that no `post_type` request param means that `post_type` is equal to `posts`), so I can make that change.

core/services/route_match/RouteMatchSpecificationManager.php
tn3rb: will change to `findRouteSpecificationsMatchingCurrentRequest`

core/interfaces/FactoryInterface.php
tn3rb: oh.. hmmm... think I missed that... 
ya I agree that it could be confusing, so I'll get rid of `core/interfaces/FactoryInterface` in favour of `EventEspresso\core\domain\services\factories\FactoryInterface`

core/services/route_match/RouteMatchSpecificationFactory.php
tn3rb: ya I think that's a great idea. I think there's a few types of classes (like `EE_DMS_Core_*` or shortcodes) that have identical dependencies hardcoded in `EE_Dependency_Map` that could benefit from this. We could likely even abstract the `RouteMatchSpecificationDependencyResolver` out into a parent `DependencyResolver` class as well, since all such classes would likely require the same dependencies, and likely behave in a very similar fashion.

core/domain/entities/route_match/specifications/admin/WordPressPostsListTable.php
nerrad: > mmmm... not seeing that in WP_Posts_List_Table nor in WP_List_Table, where is that happening?

I'm not looking at those.  I'm looking at `wp-admin/edit.php` (and where the `$typenow` global is set in `wp-admin/admin.php`).

core/services/route_match/RouteMatchSpecificationManager.php
tn3rb: I think that convention primarily exists for methods that are **getters** for boolean properties, but is not so strict otherwise, and doesn't always have to mean the method is prefixed with `is` (ie: `has`, `should`, etc are also common), but there are also lots of examples of method/function names that don't follow that pattern because the word usage already makes it clear that they return a truthy value. For example: `class_exists()`, `SplObjectStorage::contains()`, etc. 

The more important underlying concept is that the method name is readable and this works best if the name can be used as both an assertion or a question, for example

```php
if ( class_exists( $fqcn ) ) {
// or
if ( $this->contains ) {
// or
if ( $this->isTrue ) {
```

what I don't like about `isCurrentRequestMatchingRoute` is that to me it only works as a question and not an assertion, and sounds weird when used in actual code, ie:

```php
if ( $this->isCurrentRequestMatchingRoute( $route ) ) {
```

I'll try to think if a better name that isn't awkward but still clearly sounds like an assertion/question

nerrad: I'm fine with any name that provides clarity when the method is implemented.  I personally don't think the name I suggested sounds weird in use (as per your example), the context _is_ a condition, i.e. written another way "If this is [the] current request matching [the] route then" (so the method name drops the "the" words.  I just think consistent usage of `is` prefix for these types of methods makes it immediately clear what their purpose is.  

Another suggestion that might work is `isRequestMatchingRoute`?

This is the last comment I'll make on this method name, I don't want to make another long thread out of this.

tests/testcases/core/services/commands/CommandBusTest.php
```diff
@@ -38,8 +38,8 @@ public function setUp()
         // need to override the existing alias for the CommandHandlerManagerInterface
         // or else the REAL class will still get used
         EE_Dependency_Map::instance()->add_alias(
-            'EventEspresso\core\services\commands\CommandHandlerManagerInterface',
-            'EventEspresso\tests\mocks\core\services\commands\CommandHandlerManagerMock'
+            'EventEspresso\tests\mocks\core\services\commands\CommandHandlerManagerMock',
+            'EventEspresso\core\services\commands\CommandHandlerManagerInterface'
```
nerrad: why was this changed around?  Does it mean the behaviour of `EE_Dependency_Map::instance()->add_alias` has changed?  Does that have a wider impact than the original purpose for this pull and thus needs more specific testing coverage?

core/services/dependencies/ClassAlias.php
```diff
@@ -0,0 +1,59 @@
...
+
+/**
+ * Class ClassAlias
+ * Simple value object for representing a class alias such as an interface
+ *
+ * @package EventEspresso\core\services\dependencies
+ * @author  Brent Christensen
+ * @since   $VID:$
+ */
+class ClassAlias
```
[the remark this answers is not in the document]
nerrad: Or maybe `ConcreteClassOf`?

core/EE_Dependency_Map.core.php
```diff
@@ -826,9 +826,13 @@ protected function _register_core_class_loaders()
         $request = &$this->request;
         $response = &$this->response;
         $legacy_request = &$this->legacy_request;
+        $self = &$this;
         // $loader = &$this->loader;
         $this->_class_loaders = array(
             // load_core
+            'EE_Dependency_Map'                            => function () use (&$self) {
+                return $self;
+            },
```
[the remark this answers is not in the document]
tn3rb: oh ya... forgot about that, will change...

core/services/dependencies/ClassAlias.php
```diff
@@ -0,0 +1,59 @@
...
+     *
+     * @param string $alias Interface specified by implementing class
+     * @param string $fqcn  Concrete class that satisfies interface
+     * @throws InvalidAliasException
+     */
+    public function __construct($alias, $fqcn)
+    {
+        if (! is_subclass_of($fqcn, $alias)) {
+            throw new InvalidAliasException($fqcn, $alias);
+        }
```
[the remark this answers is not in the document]
tn3rb: I can add more tests coverage, and autoloading is fine in this case.

core/services/dependencies/ClassAlias.php
```diff
@@ -0,0 +1,59 @@
...
+    private $fqcn;
+
+    /**
+     * DependencyAlias constructor.
+     *
+     * @param string $alias Interface specified by implementing class
+     * @param string $fqcn  Concrete class that satisfies interface
+     * @throws InvalidAliasException
+     */
+    public function __construct($alias, $fqcn)
```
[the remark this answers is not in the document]
tn3rb: why? 
nobody will look through the code at what it does prior to creating an object and their IDE should tell them what the parameters are.

core/services/dependencies/DependencyResolver.php
```diff
@@ -0,0 +1,177 @@
...
+
+    /**
+     * @var EE_Dependency_Map $dependency_map
+     */
+    private $dependency_map;
+
+    /**
+     * @var ClassAlias[] $aliases
+     */
+    protected $aliases = array();
```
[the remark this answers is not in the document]
tn3rb: Totally disagree since it is exactly what is intended. I think you may be misinterpreting the purpose of what this is for. 
I'll look at the phpdoc descriptions and see if I can clarify things, but I see no reason to rename it since this property is fully intended to be a list of aliases.

tests/includes/CoreLoader.php
```diff
@@ -164,10 +164,6 @@ public function setupDependencyMap()
             ),
             true
         );
-        EE_Dependency_Map::instance()->add_alias(
-            'EE_Session_Mock',
-            'EventEspresso\core\domain\services\session\SessionIdentifierInterface'
-        );
```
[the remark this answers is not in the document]
tn3rb: it's not used at all

core/EE_Dependency_Map.core.php
```diff
@@ -953,7 +957,6 @@ protected function _register_core_aliases()
             'EventEspresso\core\services\loaders\LoaderInterface'                          => 'EventEspresso\core\services\loaders\Loader',
             'CommandFactoryInterface'                                                      => 'EventEspresso\core\services\commands\CommandFactoryInterface',
             'EventEspresso\core\services\commands\CommandFactoryInterface'                 => 'EventEspresso\core\services\commands\CommandFactory',
-            'EventEspresso\core\domain\services\session\SessionIdentifierInterface'        => 'EE_Session',
```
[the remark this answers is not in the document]
tn3rb: I added an `is_subclass_of()` check to `core/services/loaders/ClassInterfaceCache.php` to verify that aliases being added there were indeed a class that can safely be substituted for the supplied FQCN. The session does not load until the WP `init` action so this alias can not be verified until then, therefore it should not be added until that time. In other words: adding the alias too early caused problems and it can not be used until the session is loaded anyways.
That particular interface however is not actually being used anywhere anymore and does not adequately represent the session class anyways, so it's removal as an alias is sound.

tests/mocks/core/services/dependencies/DependencyResolverMock.php
```diff
@@ -0,0 +1,45 @@
...
+     */
+    public function getNamespaceRoots()
+    {
+        return $this->namespace_roots;
+    }
+
+    /**
+     * Used to configure and/or setup any aliases or namespace roots required by the DependencyResolver
+     *
+     * @since $VID:$
```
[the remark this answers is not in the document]
tn3rb: my IDE adds this automatically, but I'll try to remember to remove from now on. Technically this is still correct.

tests/mocks/core/services/dependencies/composites/OofOuchOwieDecorator.php
```diff
@@ -0,0 +1,39 @@
+<?php
+
+namespace EventEspresso\tests\mocks\core\services\dependencies\composites;
+
+use EventEspresso\core\services\request\RequestInterface;
+
+/**
+ * OofOuchOwieDecorator
+ *
+ * @package EventEspresso\tests\mocks\core\services\dependencies\resolvers
```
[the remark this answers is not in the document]
tn3rb: will fix

tests/mocks/core/services/dependencies/composites/OofOuchOwie.php
```diff
@@ -0,0 +1,45 @@
...
+     */
+    protected $key;
+
+    /**
+     * @var RequestInterface $request
+     */
+    protected $request;
+
+    /**
+     * @param OofOuchOwieInterface $specification
```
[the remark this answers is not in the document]
tn3rb: will fix

tests/mocks/core/services/dependencies/composites/OofOuchOwieInterface.php
```diff
@@ -0,0 +1,19 @@
+<?php
+
+namespace EventEspresso\tests\mocks\core\services\dependencies\composites;
+
+/**
+ * OofOuchOwieInterface
+ *
+ * @package EventEspresso\tests\mocks\core\services\dependencies\resolvers
```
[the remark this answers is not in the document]
tn3rb: will fix

tests/testcases/core/services/commands/CommandBusTest.php
tn3rb: This was wrong. This was basically saying:

```
substitute all requests for EventEspresso\tests\mocks\core\services\commands\CommandHandlerManagerMock
with an instance of EventEspresso\core\services\commands\CommandHandlerManagerInterface
```

which of course is impossible and should have been the other way around.

This error was detected by the addition of the `is_subclass_of()` check to `core/services/loaders/ClassInterfaceCache.php`

core/services/dependencies/ClassAlias.php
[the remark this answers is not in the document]
tn3rb: No because the intent is to be an alias, which just means something is using another name.
Whether that means an interface or child class doesn't matter, we are just using this VO as a way to indicate that a class name is being swapped out for another.

Issue #494, opened 2018-05-31
Avoid PHP fatal error if PHP sessions are not available, notably on Pantheon servers


joshfeck:
See:
https://pantheon.io/docs/wordpress-known-issues/#php-sessions
and
https://pantheon.io/docs/wordpress-sessions/

tldr; Pantheon requires a plugin to enable PHP sessions. It seems that other plugins that require sessions do not throw a fatal (only a PHP warning) so this would be a nice bit of managing a worse-case scenario if the fatal could be avoided.

sethshoultes:
@mnelson4 is this something you can fix?

mnelson4:
I could investigate a bit, but session-related stuff is usually @tn3rb 's domain

mnelson4:
@joshfeck where does the fatal error happen? I've disabled sessions by adding `ini_set('session.use_cookies', '0');` to my localhost's `wp-config.php`, and I can't register for an event, but have yet to see a fatal error. Or got any tips on how to reproduce the issue?

joshfeck:
One way is get a free Pantheon hosting account (dev sites are free) and don't install their plugin that adds support for PHP sessions. 

This is the error I get if I deactivate that plugin:

```Warning: session_start(): user session functions not defined in /code/wp-content/plugins/event-espresso-core-reg/core/EE_Session.core.php on line 567 ```

```Fatal error: session_start(): Failed to initialize storage module: user (path: ) in /code/wp-content/plugins/event-espresso-core-reg/core/EE_Session.core.php on line 567 ```

mnelson4:
Ok thanks @joshfeck. Ya I'm just reproducing the fatal error on Pantheon's server.
What plugins do you know of that only produce a warning though? I actually didn't know any offhand that even used the session (that I could think of anyway), but found the "Wise Chat" did, but it also had a fatal error on Pantheon.
If you know of any that would probably help me to identify how to fix this

mnelson4:
aha, I've isolated the problem. Pull request imminent. (Although if you do know of any other plugins that don't have the fatal, just a warning, that would still be useful for comparison)

mnelson4:
links useful for reference:
https://www.drupal.org/project/civicrm_starterkit/issues/1978796
https://phppot.com/php/php-session-configuration/
https://wordpress.org/plugins/wp-native-php-sessions/

joshfeck:
[...]

PR #615, opened 2018-08-02
if there is a custom session save handler, and it causes a fatal erro…

mnelson4:
…r, avoid starting the session and inform admins of the problem and offer a solution. Once they have fixed the problem, allow them to retry the session again

https://github.com/eventespresso/event-espresso-core/issues/494
Fixes a fatal error when there is an invalid session save handler, like on Pantheon servers.
Well, detects when there is a fatal error (I don't think we can avoid the fatal error initially) and avoids calling the session code that triggers the fatal error on subsequent requests; and then instructs the admin how to fix the error. See https://drive.google.com/file/d/11HYwn-dG6XPvJB26-AcCXfa54VZx4xVo/view for the current UX.

## How has this been tested

* [ ] Try registering for an event on a "normal" server (not pantheon). It should work normally
* [ ] Add `ini_set('session.save_handler', 'user');` to the top of your `wp-config.php` file, or use a pantheon server. You will still get the fatal error once. Sorry. But the next request will give you a helpful error message, but otherwise WordPress will work normally...
* [ ] Click the link that says "click here to check if the problem is resolved." first. You should get the fatal error again once, then this same page again.
* [ ] Log out and visit the front-end. You should get a helpful error message that won't scare average users.
* [ ] Log in again. Follow the link in the error message to the plugin Pantheon created to fix the issue. Install it. Then click "click here to check if the problem is resolved." again. Everything should work fine again.

## Checklist

* [x] My code is tested.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ See [our documentation](https://github.com/eventespresso/event-espresso-core/blo
 ### Fixed
 - Fixes a fatal error while using many page builder plugins resulting from template tags only being loaded on the front-end([600](https://github.com/eventespresso/event-espresso-core/pull/600))
 - Fixed venue description on event page so line breaks are shown properly ([612](https://github.com/eventespresso/event-espresso-core/pull/612))
+- Fixed a fatal error when there is an invalid session save handler, like on Pantheon servers ([615](https://github.com/eventespresso/event-espresso-core/pull/615))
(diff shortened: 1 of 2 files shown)


tn3rb:
plz see #620

tn3rb:
closing this in favour of #620

mnelson4:
👍

PR #616, opened 2018-08-02
Sideload all language files (POT, PO and MO) rather than just the MO file

Pebblo:
This just makes it easier for translators, especially if using a tool such as [Loco Translate](https://wordpress.org/plugins/loco-translate/)

The changes are to a function within core used to load language files, this wouldn't be any use in an add-on.

### Important - test #617 before testing this branch, the changes here rely on that branch.

Load the site using English US and look within event-espresso-core-reg/languages/
You should now see at least event_espresso.pot 
(Master won't download that POT file)

Delete any MO/PO/POT files in that directory.

Change the site language to anything other then English US, (English UK is fine) and you should see `event_espresso.pot`, `event_espresso-{locale}.po` and `event_espresso-{locale}.mo`

Confirm the filesize is different for each file (meaning it's not just the same file downloaded 3 times).

## Checklist

* [x] My code is tested.


Pebblo:
Note the changes in this branch rely on the change in #617

core/EE_Load_Textdomain.core.php
```diff
@@ -27,7 +27,7 @@ class EE_Load_Textdomain extends EE_Base
      */
     public static function load_textdomain()
     {
-        self::_maybe_get_langfile();
+        self::_maybe_get_langfiles();
```
tn3rb: sorry, but you can't just rename a public file since it's possible that someone might be using this in their own addon or theme or whatever. It's unlikely that this class is being used by anyone else but it wouldn't be the first time I've thought that and been wrong if someone was using it.

So what you need to do is keep the existing function declarations as is and then just add your new functions to the class.
If the old functions are completely unwanted, then they can be forwarded to the new functions (ie: calling A calls B) and doing it wrong notices can be added to inform devs that the old function is deprecated and scheduled for eventual removal.

core/EE_Load_Textdomain.core.php
```diff
@@ -42,13 +42,13 @@ public static function load_textdomain()
...
     /**
-     * The purpose of this method is to sideload the lang file for the given WPLANG locale (if necessary).
+     * The purpose of this method is to sideload all of the lang files for EE, this includes the POT file and also the PO/MO files for the given WPLANG locale (if necessary).
      *
      * @access private
      * @static
      * @return void
      */
-    private static function _maybe_get_langfile()
+    private static function _maybe_get_langfiles()
```
tn3rb: same issue, can't rename public functions

tn3rb: sigh ... oops... sry... this does not apply to this function since it is private... my bad

core/EE_Load_Textdomain.core.php
tn3rb: ug... totally read this wrong.

joshfeck:
This is on hold until #617 is ready.

garthkoyle:
@Pebblo I'm not sure I'm testing this correctly. I've updated garth.eventespresso.com to Master, then checked out this branch. 

My `/event-espresso-core-reg/languages/` directory only has an index file. 

When I switch languages, no mo, po, or pot files are created, only the index file exists.

You can look at that on my sandbox site if you'd like. If I'm not groking the testing steps, then please correct me.

Pebblo:
@garthkoyle Hmm, that's odd.

Can reset my password on that site please, I tried but it couldn't send an email.

I'd switch back to master and set the site language to French, then confirm you get just a .mo file saved in `{ee-core-plugin-directory}/languages/`

garthkoyle:
Tony helped me clean up the site which had two versions of EE installed.

Once we did that, and changed the language, the three files appeared and they are different sizes.

```-rw-r--r-- 1 eeservice eeservice     284 Apr  3 14:47 event_espresso-fr_BE.mo
-rw-r--r-- 1 eeservice eeservice 1673517 Apr  3 14:47 event_espresso-fr_BE.po
-rw-r--r-- 1 eeservice eeservice 1818946 Apr  3 14:47 event_espresso.pot
-rw-rw-r-- 1 eeservice eeservice      30 Jun 29  2018 index.php```

Pebblo:
Thanks Garth.

Merging this in.

PR #617, opened 2018-08-02
Fix sideloader using incorrect _upload_from property name.

Pebblo:
If you try to use the sideloader class to load multiple files you need to set the `_upload_from` property, there's a setter for just that `$sideloader->set_upload_from()`

However, it's using the wrong property name so has no effect on what the sideloader does.

It's a fix to a core function.

## How has this been tested
We don't have code in core to test this changes in this specific branch so you'll need to add something like this to your site:

```      
$lang = get_locale();

// load sideloader
$sideloader_args = array(
    '_upload_to'     => EE_PLUGIN_DIR_PATH . 'languages/',
    '_download_from'   => 'something',
    '_upload_from'   => 'https://github.com/eventespresso/languages-ee4/blob/master/event_espresso.pot?raw=true',
    '_new_file_name' => 'event_espresso.pot',
);
$sideloader = EE_Registry::instance()->load_helper('Sideloader', $sideloader_args, false);
$sideloader->sideload();


// made it here so let's get the language files from the github repo, first the .mo file
$sideloader->set_upload_from('https://github.com/eventespresso/languages-ee4/blob/master/event_espresso-' . $lang . '.mo?raw=true');
$sideloader->set_new_file_name('event_espresso-' . $lang . '.mo');
$sideloader->sideload();

// now the .po file:
$sideloader->set_download_from('https://github.com/eventespresso/languages-ee4/blob/master/event_espresso-' . $lang . '.po?raw=true');
$sideloader->set_new_file_name('event_espresso-' . $lang . '.po');
$sideloader->sideload();
```

I add it to a page template just because it's easy, you can run it wherever, debug console for example.

However, before running it go to `event-espresso-core-reg/languages/`

Remove and po/mo/pot files that may be in there.

Run the above and make sure you get 3 files, a mo, po and POT. They should NOT all be the same exact same filessize (which would happen with master as the location to download the file from was never updated).

If you have WP_DEBUG enabled with the above code you'll get a couple of deprecated notices (core will also throw one once this is merged and it checks for language files, I'll update that shortly).

So, in this ticket your just confirming you get 3 files that are not the same file with different filenames and that a deprecated notice is thrown with the above code stating `set_upload_from` is deprecated.

## Checklist

* [x] My code is tested.


core/helpers/EEH_Sideloader.helper.php
```diff
@@ -71,9 +71,9 @@ public function set_upload_to($upload_to_folder)
     {
         $this->_upload_to = $upload_to_folder;
     }
-    public function set_upload_from($upload_from_folder)
+    public function set_upload_from($upload_from)
     {
-        $this->_upload_from_folder = $upload_from_folder;
+        $this->_upload_from = $upload_from;
```
tn3rb: might want to keep the parameter name as `$upload_from_folder` since it is actually a bit more descriptive since it tells anybody looking at this that it represents a folder name. Might be even better to rename the parameter to `$folder_name` so that the a call to the function becomes even more descriptive, ie: `set_upload_from($folder_name)`, whereas keeping it as is (`set_upload_from($upload_from)`) leaves it kind of cryptic as to what is expected.

So one extra job that will be required of you now that you have touched this file (ya opened a can of worms boy!) is to go through and bring it's formatting and phpdocs up to date. This primarily means adding phpdocs to all of the methods that don't have them since it looks like the formatting is already decent for the most part.

Pebblo: I think we spoke about this in slack a while back but `$_upload_from` would actually be the full path for the file you want to download, so `$upload_from_folder` is incorrect and `set_upload_from($folder_name)` even more so (although I completely understand why you would suggest it).

Personally I find most of the 'upload_' parameters used in the sideloader a little confusing as we aren't referring to an upload in a single step, we want to download_from into a temp directory, then then 'upload'/move to the location.

So to me something like this makes more sense:

```
$sideloader_args = array(
    '_upload_to'     => EE_PLUGIN_DIR_PATH . 'languages/',
    '_download_from'   => 'https://github.com/eventespresso/languages-ee4/blob/master/event_espresso.pot?raw=true',
    '_new_file_name' => 'event_espresso.pot',
);
```

Obviously any `download_` properties would have their own setter/getters etc.

Anyway, I digress :)

Sorry for the huge delay on this, I set it aside until I set up PHPStorm.... then never actually set it up until now.

Pebblo:
Huge delay on this because I wanted to get PHPStorm set up, never did and carried on with Sublime.

Then support got more in the way than dev time so....

Anyway, I don't know if you remember chatting about this in Slack Brent but I added a reply to the above anyway :)

core/helpers/EEH_Sideloader.helper.php
Pebblo: Oh, and something I wasn't 100% here is `@since $VID:$` for this method.

It simply won't work at all up until this change so I'm assuming I use the above rather than when it was actually added?

tn3rb: no you can skip adding the `@since` tag since that is used to indicate a new method added
any methods that do not have that tag can simply be considered legacy code

Pebblo:
Howdy Brent,

The last commit above is one I'm not sure about, I think there's likely a better way to handle it but I'll explain why I did it the way I have and go from there.

Taking my original example code from above, if you call the Sideloader passing the arguments like so:

`$sideloader_args = array(
    '_upload_to'     => EE_PLUGIN_DIR_PATH . 'languages/',
    '_upload_from'   => 'https://github.com/eventespresso/languages-ee4/blob/master/event_espresso.pot?raw=true',
    '_new_file_name' => 'event_espresso.pot',
);
$sideloader = EE_Registry::instance()->load_helper('Sideloader', $sideloader_args, false);
$sideloader->sideload();`

Then `init()` would try to set the `_upload_from` property directly, I was going to add a check for `_upload_from` in that loop and throw a deprecated notice etc but realized that the setters did this already, so I swapped out `init()` to call those dynamically rather than setting the properties itself.

That meant I needed a `_upload_from` property so that it would call the `set_upload_from` property if passed to it which then throws the doing it wrong notice etc.

Happy to change that to something else if preferred.

core/helpers/EEH_Sideloader.helper.php
```diff
@@ -14,15 +14,47 @@
 class EEH_Sideloader extends EEH_Base
 {
 
+    /**
+     * @since   4.1.0
+     * @var     string
+     * @access  private
+     */
```
tn3rb: awesome thnx for adding this 👍 

we don't need the `@access` tag however since `private` is clearly added before the property
any modern IDE will properly detect the property scope without the need of this tag, so this can be removed

core/helpers/EEH_Sideloader.helper.php
```diff
@@ -14,15 +14,47 @@
...
+     * @deprecated since version $VID:$
+     */
     private $_upload_from;
+
+    /**
+     * @since   $VID:$
+     * @var     string
+     * @access  private
+     */
+    private $_download_from;
```
tn3rb: mucho bettero 🌮

core/helpers/EEH_Sideloader.helper.php
```diff
@@ -188,4 +297,45 @@ public function sideload()
...
+        EE_Error::doing_it_wrong(
+            __CLASS__ . '::' . __FUNCTION__,
+            __(
+                'EEH_Sideloader::set_upload_from was renamed to EEH_Sideloader::set_download_from',
+                'event_espresso'
+            ),
+            '$VID:$'
+        );
+        $this->set_download_from($upload_from);
+    }
```
tn3rb: POIFECT 👌

core/helpers/EEH_Sideloader.helper.php
```diff
@@ -52,7 +85,8 @@ private function _init($init)
 
         foreach ($props as $key => $val) {
             if (EEH_Class_Tools::has_property($this, $key)) {
```
tn3rb: oops missed this in my first pass...

i realize you didn't add this, but we don't need this anymore since it was only acting as a shim for older PHP versions.

this can be replaced with `property_exists()`

tn3rb:
re:

> Then init() would try to set the _upload_from property directly, I was going to add a check for _upload_from in that loop and throw a deprecated notice etc but realized that the setters did this already, so I swapped out init() to call those dynamically rather than setting the properties itself.
> 
> That meant I needed a _upload_from property so that it would call the set_upload_from property if passed to it which then throws the doing it wrong notice etc.


ya I think that's fine...
using the setters to set the properties is much better than doing so directly and should be the favoured approach.

Directly setting properties is only preferential for dependencies that are injected in the constructor, ie:

```php
    public function __construct(DependencyClass $dependency)
    {
        $this->dependency = $dependency;
    }
```

but for dynamic values like those in the `$init` array passed to the sideloader, using setters is far superior because it allows you to control those properties as they are added, meaning you can do things like convert the format, validate and/or sanitize the values, or check if they are deprecated. So your solution is perfect imho.

core/helpers/EEH_Sideloader.helper.php
```diff
@@ -84,7 +77,7 @@ private function _init($init)
         $props = array_merge($defaults, $init);
 
         foreach ($props as $key => $val) {
-            if (EEH_Class_Tools::has_property($this, $key)) {
+            if (property_exists($this, $key)) {
                 $setter = 'set' . $key;
                 $this->$setter($val);
```
tn3rb: ohhhh... actually another possibly better way to do this would be to verify that the setter exists instead of checking for the property, ie:

```suggestion
            $setter = 'set' . $key;
            if (method_exists($this, $setter)) {
                $this->$setter($val);
```

since, although things would work absolutely fine the way you have things now, there's always the chance that someone could add a new property in the future that does not follow the correct setter naming schema (or not have a setter at all) and then try to pass that prop name within the `$init` array. So the `property_exists()` check would pass, but the subsequent attempt to call the setter would fail. So really we only care if the setter exists. This of course could then benefit from the addition of an error message to indicate that no setter exists for the provided property.

Sorry for bouncing this back again.

Pebblo:
Thanks Brent :)

I did some testing on this and spotty a typo, fixed above.

However, I've found that if I add a property that doesn't have a setter, no error is thrown.

I'm just using:

```
$sideloader_args = array(
    '_upload_to'     => EE_PLUGIN_DIR_PATH . 'languages/',
    '_upload_from'   => 'https://github.com/eventespresso/languages-ee4/blob/master/event_espresso.pot?raw=true',
    '_new_file_name' => 'event_espresso.pot',
    '_tonys_super_prop' => 'Yep',
);
$sideloader = EE_Registry::instance()->load_helper('Sideloader', $sideloader_args, false);
```

That's just thrown in a WP page template to test but no error and nothing is logged. If I throw some debug code in the else just before the new EE_Error::add_error() call that's output to the page.

Is there something I'm missing?

And, should I wrap that error in a wp_debug check? 

IIRC there was a massive notice/error thrown in the admin when the sideloader couldn't find the file(s) on github which was eventually wrapped in a debug check.

core/helpers/EEH_Sideloader.helper.php
```diff
@@ -69,20 +69,20 @@ private function _init($init)
 
         $props = array_merge($defaults, $init);
 
-        foreach ($props as $propery => $val) {
-            $setter = 'set' . $propery;
+        foreach ($props as $property => $val) {
+            $setter = 'set' . $property;
```
tn3rb: BAHAHA how long did we both look at that without noticing the typo? :facepalm:

Pebblo: Just a minute.... or 3 :)

Pebblo:
I ended up just confirming that it add_error worked when we normally sideload the language files for now, a notice is shown in the admin when an invalid property is set which I think is ok for the time being.

We've never had any reports of issues with people using the side-loader so tbh, I don't think any has as the setter simply didn't work before now.

joshfeck:
I tested this with the code snippet and now get all the correct files.

Issue #392, opened 2018-05-07
First errant conditional in `espresso_event_tickets_available()` template tag


joshfeck:
The function's signature has 3 params:
```
function espresso_event_tickets_available( $EVT_ID = 0, $echo = TRUE, $format = TRUE )
```

so it's echo'ed by default and format is true. If formatting is set to false then an array of tickets is returned. However, this conditional causes nothing to be displayed if the defaults are used:

```
if ( $echo && ! $format ) {
	echo $html;
	return '';
}
```

mnelson4:
PR is off for review

mnelson4:
related pull request was merged in

PR #619, opened 2018-08-02
if we are formatting it and requested it be echoed, by all means do. …

mnelson4:
…What strange is if they requested it to NOT be formatted but still requested it to be echoed. We dont echo anything. We could do a var_dump I suppose but that seems silly

https://github.com/eventespresso/event-espresso-core/issues/392
Fixes espresso_event_tickets_available() so it echoes out the tickets when default arguments are provided.

## How has this been tested

* [x] Use `espresso_event_tickets_available()` with an event ID. It should echo out a list of tickets for that event

## Checklist

* [x] My code is tested.


mnelson4:
@joshfeck wanna test this? Maybe approve the review again too (there was a conflict in the changelog only since Nazar last reviewed)

joshfeck:
Tested and verified the fix. This is ready for merging.

PR #620, opened 2018-08-02
Fix Session Start Fatal Errors when known Custom Session Save Handler is in use

tn3rb:
just a refactoring of #615


tn3rb:
### testing

* [ ] Try registering for an event on a "normal" server (not pantheon). It should work normally
* deactivate EE
* [ ] Add `ini_set('session.save_handler', 'user');` to the top of your `wp-config.php` file, or upload branch to a pantheon server. 
* reactive EE
* You will still get the fatal error once. Sorry. But the next request will give you a helpful error message, but otherwise WordPress will work normally...
* [ ] Click the link that says "click here to check if the problem is resolved." first. You should get the fatal error again once, then this same page again.
* [ ] Log out and visit the front-end. You should get a helpful error message that won't scare average users.
* [ ] Log in again. Follow the link in the error message to the plugin Pantheon created to fix the issue. Install it. Then click "click here to check if the problem is resolved." again. Everything should work fine again.

joshfeck:
I tested this locally and on a Pantheon server. The tests check out OK, the strange thing about this branch is I can still register for events on the Pantheon server even without installing the WP Native Session plugin. The registration was made while logged out so I'd avoid the on screen nag messages about no sessions. The on-screen nag messages about no sessions that are shown if logged in actually prevent advancing through the steps of SPCO.

joshfeck:
Would you expect the registration process to work as expected without the native PHP session plugin installed?

mnelson4:
I'm not sure what's the best UX. Yes, registrations can work in this branch even without sessions, but we're warning site owners it won't be working "normally". (Eg MER won't work, they'll miss some messages etc).
We could alternatively disable EE completely if that's better?

Issue #500, opened 2018-06-05
Message Queue triggers a fatal error if the site is in MM2


Pebblo:
Whilst testing #489 I've had a site in MM2 for a prolonged period and noticed that the message queue is triggering a fatal:

```
[05-Jun-2018 08:19:18 UTC] PHP Fatal error:  Uncaught EE_Error: Event Espresso Level 2 Maintenance mode is active. That means EE can not run ANY database queries until the necessary migration scripts have run which will take EE out of maintenance mode level 2. Please inform support of this error. in D:\xampp\htdocs\ee4\wp-content\plugins\event-espresso-core-reg\core\db_models\EEM_Base.model.php:2378
Stack trace:
#0 D:\xampp\htdocs\ee4\wp-content\plugins\event-espresso-core-reg\core\db_models\EEM_Base.model.php(1080): EEM_Base->_do_wpdb_query('get_results', Array)
#1 D:\xampp\htdocs\ee4\wp-content\plugins\event-espresso-core-reg\core\db_models\EEM_Base.model.php(954): EEM_Base->_get_all_wpdb_results(Array, 'ARRAY_A', NULL)
#2 D:\xampp\htdocs\ee4\wp-content\plugins\event-espresso-core-reg\core\libraries\messages\EE_Messages_Queue.lib.php(225): EEM_Base->get_all(Array)
#3 D:\xampp\htdocs\ee4\wp-content\plugins\event-espresso-core-reg\core\libraries\messages\EE_Messages_Processor.lib.php(230): EE_Messages_Queue->get_to_send_batch_and_send()
#4 D:\xampp\htdo in D:\xampp\htdocs\ee4\wp-content\plugins\event-espresso-core-reg\core\db_models\EEM_Base.model.php on line 2378
```

To reproduce you can add: `update_option('ee_maintenance_mode', 2);`

In functions.php or a plugin to set EE Mainteance mode lvl 2.

Then visit the front end of the site after a few mins and check your error log.

I don't have messages in the queue for generation/sending but I do have 'Failed Sending, Can Be Retried' (expected with my current local setup).

Not classed as critical as this has no effect on users or the expected behavior (users don't see the error on the front end/admin as its run on WP cron and messages don't send because there was a fatal but messages wouldn't send in MM2 anyway).

sethshoultes:
@mnelson4 and @Pebblo does this need to be tested? Are there testing notes?

mnelson4:
[...]

PR #622, opened 2018-08-03
dont try to send messages while in full maintenance mode

mnelson4:

https://github.com/eventespresso/event-espresso-core/issues/500
Fixes error when trying to send messages in full maintenance mode.

## How has this been tested

* [ ] verify that when you register for an event, messages get sent out as normal
* [ ] see testing instructions from Tony on https://github.com/eventespresso/event-espresso-core/issues/500 (although I wasn't able to reproduce the original issue (I actually didn't never saw the messages code get called while in maintenance mode; possibly it got coincidentally fixed, but more likely I just can't reproduce)

## Checklist

* [ ] My code is tested.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ See [our documentation](https://github.com/eventespresso/event-espresso-core/blo
 - Fixed an error when migrating from EE3 from attendee email index being too big ([611](https://github.com/eventespresso/event-espresso-core/pull/611))
 - Fixed venue description on event page so line breaks are shown properly ([612](https://github.com/eventespresso/event-espresso-core/pull/612))
 - Fixed caching loader identifier ([610](https://github.com/eventespresso/event-espresso-core/pull/610))
+- Fixed error when trying to send messages in full maintenance mode ([622](https://github.com/eventespresso/event-espresso-core/pull/622))
 
 ### Changed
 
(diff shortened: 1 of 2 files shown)


mnelson4:
hey @Pebblo while we're waiting for Darren to get back, can you double-check you can still reproduce this issue to confirm this fix is needed?

mnelson4:
Yeah, using this logic whenever `is_locked` is called makes sense. What about having a new method, `can_send_messages()` or something, that checks `is_locked` and `models_can_query()`, and we use that in place of `is_locked`? 
Placing this logic inside `is_locked` would be a bit unexpected for me, but I do like how DRY it is. But if you feel this logic is more at home in `is_locked()` than  new method named something like `can_send_messages()`, then I'll go with that.

nerrad:
In this case I think `is_locked` is okay because the purpose of `is_locked` is to check whether _something_ has locked the messages and preventing them from generating or sending (depending on the "type" passed in as an argument).  By putting the `models_can_query` checked in this method we're including that in the requirements for whether its considered "locked" or not (and also it'd be a requirement that can't be overridden).

mnelson4:
passing to @Pebblo for testing, see original PR description for testing notes

sethshoultes:
@Pebblo are you planning on testing this or would you like someone else to test?

Pebblo:
All good, no more fatal and messages send as normal when not in MM.

PR #625, opened 2018-08-06
Add localized status labels for statuses and a label value object

nerrad:

While working on the barcode scanner react app, I realized that there's places where the localized status label for a status code was needed (i.e. RAP = registration approved status).

This pull adds:

- new constants for statuses to the event, ticket, and datetime models.  These are the equivalent to the statuses found PHP side for those models and are added to the model files because they are not statuses stored in the status table.
- add missing `MRT` status constant for messages.
- add status constants to represent the native wp post type statuses
- add status constant to represent an 'unknown' status code.
- add label value object
- add status model helper file that includes `prettyStatus` and `prettyStatuses` methods for getting the localized label for a given status code (or codes).  This file also contains all the maps of status code to localized label (that currently are not exported).
- all covered by unit tests.

## How has this been tested

* [x] local unit tests pass

## Checklist

* [x] My code is tested.


docs/AA--Javascript-in-EE/value-objects/README.md
```diff
@@ -46,4 +46,5 @@ EE javascript currently implements the following value objects:
 | Name | Description |
 | ------- | ------------------- |
 [Money](money.md) | Represents an amount of money.
-[Currency](currency.md) | Represents a description of currency.
\ No newline at end of file
+[Currency](currency.md) | Represents a description of currency.
+[Label](label.md) | Represents a label singular and plural strings.
```
tn3rb: maybe:

> Represents singular and plural versions for a string.

docs/AA--Javascript-in-EE/value-objects/label.md
```diff
@@ -0,0 +1,168 @@
+## Label Value Object
+The label value object represents a label which consists of a singular string and plural string.  For example a shoe label where singular is `shoe` and plural is `shoes`.
+
+## Features
+
+The label value object exposes the singular and plural forms of the labels as their raw value but also allows for retrieving formatted strings via helper methods.
+
+## Installation/Loading
+
+The label vo is available via the `eejs` global on its `valueObjects` property.  So you can access it directly via `eejs.valueObjects.Label` provided the bundled `@eventespresso/vo` javascript has been loaded before your script. For instructions on different ways to do this you can [go here](README.md#Usage).
```
tn3rb: since "vo" is an acronym, it should be capitalized. 
As well, before using an acronym for the first time on its own, it should first be defined by introducing it after the full term that it represents. So for example, in the "Features" section above this, we would do something like the following:

> The label value object (VO) exposes the singular and plural forms...

which then allows us to use the acronym on its own since we have now defined it:

> The label VO is available...

docs/AA--Javascript-in-EE/value-objects/label.md
```diff
@@ -0,0 +1,168 @@
...
+
+#### `Label.FORMAT_SENTENCE_CASE`
+
+This static property on `Label` can be used to reference the sentence case format string.
+
+### Instance
+
+The following methods and properties exist after instantiating a `Label`.
+
+### _Methods_
```
tn3rb: Why are there two "Methods" headings/sections ???

This should be consolidated with the other Methods section above and/or that section should be moved down and consolidated with this one (which would make sense to put all methods after the properties section)

docs/AA--Javascript-in-EE/value-objects/label.md
```diff
@@ -0,0 +1,168 @@
...
+
+#### `Label.FORMAT_UPPERCASE`
+
+This static property on `Label` can be used to reference the uppercase format string.
+
+#### `Label.FORMAT_SENTENCE_CASE`
+
+This static property on `Label` can be used to reference the sentence case format string.
+
+### Instance
```
tn3rb: Is this heading necessary?

The structure is kinda weird...
The following are the main headings (`<h3>`) used in this doc:

- Constructor
- Methods
- Properties
- Instance
- Methods

I think it could be reduced and reorganized to just be:

- Constructor
- Properties
- Methods

and the line following this comment:

> The following methods and properties exist after instantiating a `Label`.

could maybe be moved to the bottom of the "Constructor" section as an intro to the next two sections.

docs/AA--Javascript-in-EE/value-objects/label.md
```diff
@@ -0,0 +1,168 @@
...
+
+// returns 'flock'
+console.log( myNewLabel.plural );
+```
+
+#### `asSentenceCase( singular = true )`
+
+Returns the value for the indicated property formatted in sentence case.  By default returns the value for the singular property, however if you pass in anything but `true` as the argument, then the plural property value is used and returns as sentence case.
+
+> Note, this strips any `-` in dashed labels.  So for instance if your label value was `something-else`, the value returned would be `Something Else`.
```
tn3rb: may want to note that it also strips underscores (see: https://lodash.com/docs/4.17.10#startCase)

docs/AA--Javascript-in-EE/value-objects/label.md
```diff
@@ -0,0 +1,168 @@
...
+
+**Example**:
+
+```js
+const myLabel = new Label( 'pack of gum', 'packs of gum' );
+
+// 'Pack Of Gum'
+console.log( myLabel.asSentenceCase() );
+
+// 'Packs of Gum'
```
tn3rb: should be: 

> 'Packs Of Gum'

docs/AA--Javascript-in-EE/value-objects/label.md
```diff
@@ -0,0 +1,168 @@
...
+console.log( myLabel.asSentenceCase( false ) );
+```
+
+#### `asLowerCase( singular = true )`
+
+Returns the value for the indicated property formatted in lower case.
```
tn3rb: in some places you use "lower case" and in others (like the Properties section) you use "lowercase",
either is correct, but we should stick with one or the other.

docs/AA--Javascript-in-EE/value-objects/label.md
```diff
@@ -0,0 +1,168 @@
...
+
+**Example**:
+
+```js
+const myLabel = new Label( 'HAPPY', 'conFuSed` );
```
tn3rb: should be:

> 'conFuSed'

with a single quote at the end, not a back-tick, which is borking the formatting for the rest of the doc

docs/AA--Javascript-in-EE/value-objects/label.md
```diff
@@ -0,0 +1,168 @@
...
+// 'happy'
+console.log( myLabel.asLowerCase() );
+
+// 'confused'
+console.log( myLabel.asLowerCase( false ) );
+```
+
+#### `asUpperCase( singular = true )`
+
+Returns the value for the indicated property formatted in upper case.
```
tn3rb: same issue as  "lower case" vs "lowercase", need to use one version only

docs/AA--Javascript-in-EE/value-objects/label.md
```diff
@@ -0,0 +1,168 @@
...
+
+**Example**
+```js
+const myLabel = new Label( 'dog', 'DOGS' );
+
+// Dog
+console.log( myLabel.asFormatted() );
+
+//dogs
+console.log( myLabel.asFormatted( false, Label.FORMAT_LOWERCASE ) );
```
tn3rb: add space at start of comment for consistency, ie:

> // dogs

docs/AA--Javascript-in-EE/value-objects/label.md
```diff
@@ -0,0 +1,168 @@
...
+
+//DOG
```
tn3rb: add space at start of comment for consistency, ie:

> // DOG

docs/AA--Javascript-in-EE/value-objects/label.md
nerrad: This is following the same format as used for the other value object docs.  The first section deals with methods/properties on the actual constructor object (i.e. the `Label`) whereas the second section deals with the the methods and properties on the _instance_ returned on construction.  Thus the two sections are deliberate.

docs/AA--Javascript-in-EE/value-objects/label.md
nerrad: No.  In javascript the `constructor` object is what is passed around.  So the `Constructor` heading refers to that and what is accessible on that object.  The `instance` heading refers to the methods and properties available on the instance after the object is constructed.  So they are _deliberately_ two sections.

tn3rb: ahh kk

Issue #423, opened 2018-05-11
Migrations: Change DMS so that EE3 events with "Deleted" status and their associated data do not get migrated


sethshoultes:
Codebase https://events.codebasehq.com/projects/event-espresso/tickets/10290

Per @joshfeck 

One site in particular had a series of recurring events that was made by mistake (one event every day for 20 years). The series was quickly deleted by clicking on the "Delete all events in this series" link. This doesn't really delete the events, it acts more like the Trash bin in WP. When they tried to migrate to EE4, it was extremely slow even though they had more than sufficient server resources.

I think before there's an EE4 REM add-on the migrations scripts need to ignore these deleted events because if there is a significant amount of deleted events in EE3 it can have a negative impact on the migration experience. They either have to wait a really long time to migrate unwanted data, or go through and perm. delete potentially hundreds or thousands of events that they thought were deleted.

Per @mnelson4 

I think this should just require changing `EE_DMS_4_1_0_events::_migration_step()`'s query to NOT migrate  events where `event_status="D"`, and maybe where `is_active='Y'`. Also need to make sure we don't migrate registrations, questions group relations, datetimes or anything else from those events

mnelson4:
PR was merged in

PR #629, opened 2018-08-07
Fix very slow migration from EE3 when there are deleted events

mnelson4:

https://github.com/eventespresso/event-espresso-core/issues/423
Fixes very slow migration from EE3 when there are deleted events, by preventing importing deleted events from EE3 (and their associated data, like attendees, question-group relations, datetimes, etc)

## How has this been tested

* [x] Permanently delete EE4 data from your site. Then isntall EE3. Add an event with a couple times and prices. Create a custom question and question group put it on the event. Register for the event. Then delete it. Create another event (which won't be deleted) with registrations too.
* [x] Deactivate EE3 and activate EE4 and run the migrations. The deleted events should not be migrated, and there should be no errors. The other events should be migrated normally.

## Checklist

* [ ] My code is tested.

### DIFF
--- a/core/data_migration_scripts/4_1_0_stages/EE_DMS_4_1_0_attendees.dmsstage.php
+++ b/core/data_migration_scripts/4_1_0_stages/EE_DMS_4_1_0_attendees.dmsstage.php
@@ -5,85 +5,55 @@
  * 3.1 Attendee table definition:
  * delimiter $$
 
-CREATE TABLE `wp_events_detail` (
-  `id` int(11) unsigned NOT NULL AUTO_INCREMENT,
-  `event_code` varchar(26) DEFAULT '0',
-  `event_name` varchar(100) DEFAULT NULL,
-  `event_desc` text,
-  `display_desc` varchar(1) DEFAULT 'Y',
-  `display_reg_form` varchar(1) DEFAULT 'Y',
-  `event_identifier` varchar(75) DEFAULT NULL,
-  `start_date` varchar(15) DEFAULT NULL,
-  `end_date` varchar(15) DEFAULT NULL,
-  `registration_start` varchar(15) DEFAULT NULL,
-  `registration_end` varchar(15) DEFAULT NULL,
-  `registration_startT` varchar(15) DEFAULT NULL,
-  `registration_endT` varchar(15) DEFAULT NULL,
-  `visible_on` varchar(15) DEFAULT NULL,
-  `address` text,
-  `address2` text,
-  `city` varchar(100) DEFAULT NULL,
-  `state` varchar(100) DEFAULT NULL,
-  `zip` varchar(11) DEFAULT NULL,
-  `phone` varchar(15) DEFAULT NULL,
-  `venue_title` varchar(250) DEFAULT NULL,
-  `venue_url` varchar(250) DEFAULT NULL,
-  `venue_image` text,
-  `venue_phone` varchar(15) DEFAULT NULL,
-  `virtual_url` varchar(250) DEFAULT NULL,
-  `virtual_phone` varchar(15) DEFAULT NULL,
-  `reg_limit` varchar(25) DEFAULT '999999',
-  `allow_multiple` varchar(15) DEFAULT 'N',
-  `additional_limit` int(10) DEFAULT '5',
-  `send_mail` varchar(2) DEFAULT 'Y',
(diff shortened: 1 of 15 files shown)


core/data_migration_scripts/4_1_0_stages/EE_DMS_4_1_0_attendees.dmsstage.php
```diff
@@ -5,85 +5,55 @@
...
+`hashSalt` VARCHAR(250) NULL DEFAULT NULL,
+PRIMARY KEY (`id`),
+INDEX `registration_id` (`registration_id`),
+INDEX `event_id` (`event_id`)
+)
+COLLATE='utf8_general_ci'
+ENGINE=InnoDB
+AUTO_INCREMENT=3
+;
+
```
nerrad: This seems like an odd change.  Why is the creation of `wp_events_detail` table removed, and the creation of `wp_events_attendee` added?

core/data_migration_scripts/4_1_0_stages/EE_DMS_4_1_0_event_venue.dmsstage.php
```diff
@@ -53,7 +62,11 @@ public function _migration_step($num_items = 50)
     public function _count_records_to_migrate()
     {
         global $wpdb;
-        $count = $wpdb->get_var("SELECT COUNT(id) FROM ".$this->_old_table);
+        $count = $wpdb->get_var(
+            "SELECT COUNT(ev.id) FROM ".$this->_old_table . " AS ev 
+            INNER JOIN " . $wpdb->prefix . 'events_detail AS e ON ev.event_id=e.id
```
nerrad: here you have `e.id` and `ev.event_id` yet in the select part of the query you are using `COUNT(ev.id)`.  Which should it be `COUNT(e.id)` or `COUNT(ev.event_id)`?

core/data_migration_scripts/4_1_0_stages/EE_DMS_4_1_0_attendees.dmsstage.php
mnelson4: Ah, ya this does look pretty weird. What you might not see from the summary is this is actually just a comment. When originally writing these DMSs, I put the relevant EE3 table definitions in comments for reference... but on this class which migrated attendees I had previously put the wrong EE3 table definition in the comment- the events table. So here I'm instead showing the old attendees table, as that's what this class migrates from. Does that make more sense?

core/data_migration_scripts/4_1_0_stages/EE_DMS_4_1_0_event_venue.dmsstage.php
mnelson4: Hmmm, thanks for following the code along this well.
`$this->_old_table` was the EE3 `events_venue_rel` table, so previously we were counting `events_venue_rel.id`.
With this change, we're also joininig to the events table so we can skip event-venue-relation rows for deleted events, but I think we should continue counting the same column, `events_venue_rel.id`, aliased as `ev.id`.
Does that make sense?

mnelson4:
Thanks for the in-depth review Darren! Passing back with attempted explanations of the code in question

core/data_migration_scripts/4_1_0_stages/EE_DMS_4_1_0_attendees.dmsstage.php
nerrad: Oh I see.  It's also breaking standards a bit because every line of a phpdoc block should have a `*` line delimiter (which the comments here don't have).  I'm not going to be picky on it though if you don't want to fix.

core/data_migration_scripts/4_1_0_stages/EE_DMS_4_1_0_event_venue.dmsstage.php
nerrad: I think I understand.  Essentially `ev.id` then is the event_venue relation id not the event_id.  If that's a correct understanding then ya this is fine.

mnelson4:
passing to Josh for testing seeing how he originally reported. Testing notes on my first comment as usual

joshfeck:
This worked really well. I actually have some testing data that goes back many years where a lot of events were moved to "Deleted" status. The only data that was migrated was the data related to non-deleted events. The migration went much more quickly. This is ready for merging.

PR #633, opened 2018-08-08
Namespace site license key request key (fixes #605)

### LINKED ISSUE #605: Make license key check request more resilient
For the backstory see this forum topic:
https://eventespresso.com/topic/api-timeout-calls-crashing-admin-on-high-traffic-site/

and this slack chat starting here:
https://eventespresso.slack.com/archives/C080L8G68/p1532457321000038

(Note to future readers: the slack chat may not be available due to free account constraints) 
So the tldr; version of that chat is we think there may be another plugin highjacking the request leading to runaway license key checks numbering into the tens of thousands for this particular site.
The key PUE looks for in the post request is `site_license_key`, so maybe one part of this is make that key more unique.

nerrad:

See #605 for details.  Essentially, this pull request modifies the form name field used for the support/site license key on the general options page so that its namespaced a bit more.  Then in pue-client, rather than listening for `site-license-key` on requests, its listening for this new namespaced key (`ee-site-license-key`).

The original request key was a bit too generic and its feasible this resulted in pue triggering multiple requests to our server as a result of a plugin and/or theme also submitting `site-license-key` somewhere in a request as a key.

This pull request also fixes a small bug for registered EE add-ons which were not using the `PUE_UPDATES_ENDPOINT` constant if present.  This affected testing with that constant.

## How has this been tested

I have yet found a way to reproduce the issue prompting #605 so there's no way currently for me to verify this actually fixes the issue some users report.  So testing has focused mainly on ensuring existing pue functionality is not broken.

* [x] Verify license key validation still works as expected on saving the General Options page.
* [x] Verify that license key validation for both ee core and any add-ons works as expected (in testing verify that you are able to download "updates" from staging.eventespresso.com).

## Checklist

* [ ] My code is tested.

### DIFF
--- a/core/domain/services/pue/Config.php
+++ b/core/domain/services/pue/Config.php
@@ -58,7 +58,7 @@ class Config
 
     public function optionKey()
     {
-        return 'site_license_key';
+        return 'ee_site_license_key';
     }
 
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@ See [our documentation](https://github.com/eventespresso/event-espresso-core/blo
 - Fixed Authorize.net AIM so payment currency is sent ([591](https://github.com/eventespresso/event-espresso-core/pull/591))
 - Fixed DuplicateCollectionIdentifierException errors when converting old PersistentAdminNotice Fixes ([505](https://github.com/eventespresso/event-espresso-core/pull/505))
 - Fixes a syntax issue inside `EE_Config::register_ee_widget()` ([608](https://github.com/eventespresso/event-espresso-core/pull/608))
+- Namespace `site-license-key` in requests (to `ee-site-license-key`) to hopefully fix issue some users see with multiple pue license key requests happening. ([633](https://github.com/eventespresso/event-espresso-core/pull/633))
 
 ### Changed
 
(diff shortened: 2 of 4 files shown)


core/domain/services/pue/Config.php
```diff
@@ -58,7 +58,7 @@ public function checkPeriod()
 
     public function optionKey()
     {
-        return 'site_license_key';
+        return 'ee_site_license_key';
```
mnelson4: we don't need to migrate over the old option, right? The old option (which may have been hijacked by another plugin) just isn't used anymore, right? What will the code do when it can't find the old `site_license_key` WP option?

nerrad: The name for this is actually a bit misleading.  When PUE was first implemented, there was actually a specific option using this key in the option table and thus saving the form with the input used the option key as the input `name` attribute.  Since then, we still have a form input using this value as its `name` attribute, but the actual value is saved in the database as a part of `Network_Config`.

mnelson4: ahhh, gotcha. So changing this only affects the form, not how the data is stored. Sounds good.

sethshoultes:
@joshfeck can you test this one please?

joshfeck:
Tested and verified, ready for merging.

PR #634, opened 2018-08-08
Add checkin model to `@eventespresso/model` bundle

nerrad:

This pull add the *Checkin* model to `@eventespresso/model` module.  Along with this:

- `buildOptions` in `build-options.js` has been tweaked so the `optionsEntityMap` passed by the model can provide a function for the `mapSelection` to build the options instead of just the map.  This was added in case there are specific types of option arrays a model wants to build.  The new checkin model is using this for its `default` options map.
- jest config has been tweaked so that it `@eventespresso/model` usage is correctly mapped to the model package.

## How has this been tested

* [x] unit test coverage

## Checklist

* [x] My code is tested.

### DIFF
--- a/assets/src/data/model/status/helpers.js
+++ b/assets/src/data/model/status/helpers.js
@@ -221,12 +221,12 @@ const STATUS_TRANSLATION_MAP_DATETIME = {
  */
 const STATUS_TRANSLATION_MAP_CHECKIN = {
 	[ CHECKIN_STATUS_ID.STATUS_CHECKED_IN ]: new Label(
-		__( 'check in', 'event_espresso' ),
-		__( 'checked in', 'event_espresso' )
+		__( 'check-in', 'event_espresso' ),
+		__( 'check-ins', 'event_espresso' )
 	),
 	[ CHECKIN_STATUS_ID.STATUS_CHECKED_OUT ]: new Label(
-		__( 'check out', 'event_espresso' ),
-		__( 'checked out', 'event_espresso' )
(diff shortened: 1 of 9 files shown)


assets/src/data/model/status/helpers.js
```diff
@@ -213,6 +214,25 @@ const STATUS_TRANSLATION_MAP_DATETIME = {
...
+ */
+const STATUS_TRANSLATION_MAP_CHECKIN = {
+	[ CHECKIN_STATUS_ID.STATUS_CHECKED_IN ]: new Label(
+		__( 'check in', 'event_espresso' ),
+		__( 'checked in', 'event_espresso' )
+	),
+	[ CHECKIN_STATUS_ID.STATUS_CHECKED_OUT ]: new Label(
+		__( 'check out', 'event_espresso' ),
+		__( 'checked out', 'event_espresso' )
+	),
```
nerrad: The above two implementations are not strictly _singular_ and _plural_.  I thought it might be good to provide the singular version which is imperative and a "plural" version which is declarative.  However, I'm on the fence with this as it could be unexpected.  Perhaps plural versions should simply be `check ins` and `check outs`?

nerrad: ya I agree with your thoughts and it'd be nice to have some additional label type vos in the future if/when needed.  For now I'll revert what I did here.  As per further discussion in slack, I will also change the singular/plural versions  to be `check-in` and `check-out`.  Since the label is intended to be more of a "noun" representing the check in "status" we think that'd be more appropriate.

Issue #575, opened 2018-07-10
Improve sanitizing of Event Slug option


joshfeck:
Found while troubleshooting the issue that led to #574. It appears that instead of stripping accents from the Event slug option field, it's trying to convert them. e.g. the following screenshots:

![user input](https://user-images.githubusercontent.com/891156/42523803-ce10d2c6-843c-11e8-97d4-69a5e1a7f320.png)
_user input_

![broken URL after saving](https://user-images.githubusercontent.com/891156/42524002-4188b85e-843d-11e8-9d13-9bdaa8811462.png)
_after saving_

The effects are more than cosmetic (go to view event pages and you'll get redirected to the home page).

mnelson4:
the problem is happening inside WordPress' `sanitize_title_with_dashes()` function...

mnelson4:
well, that's where it's becoming crazy-looking letters. But I think that's intended. I think the problem might be happening somewhere else...

mnelson4:
however, I see when I save a post with accents in its name, it just removes the accents...

mnelson4:
kk I studied how WP core does it, and have a PR...

mnelson4:
hmmm, although there's the more general problem: what to do with unicode in the slug. I see CPTs with unicode slugs created with the CPT UI plugin work fine though...

mnelson4:
aha, I see more of what's going on. When we save the EE event CPT slug, it gets percent-encoded, which seems to be the thing to do, according to this stack overflow question: https://stackoverflow.com/a/2742985/1493883.
Plus, when someone goes to visit a URL with those funny-looking percent encoded characters, the browser automatically converts them to the unicode characters so it looks good... but WordPress doesn't recognize it anymore. I want to track down that WP code and give it an earload... or at least confirm it's where the problem resides. (Assuming it is, we may want to, for now, just do what CPT UI pluginin does with slugs... absolutely nothing, just pass them onto WP and let it deal with them.)

tn3rb:
Option 3 should be the right answer, but if it is not working due to issues in WP core, then we may need to temporarily drop down to option 2 and just do nothing for the time being.
If and/or when the issue is resolved in WP core, then we can return to percent encoding the URLs. We may want to also investigate the issue in WP core to help find a solution.

tn3rb:
[...]

PR #635, opened 2018-08-08
Fixed event slug sanitization for accents and unicode characters

mnelson4:

https://github.com/eventespresso/event-espresso-core/issues/575
Fixed event slug sanitization for accents and unicode characters. Before they would get percent-encoded, which should have been ok, but WP wasn't handling them correctly (see https://core.trac.wordpress.org/ticket/44759). So, our solution is to leave them as unicode characters, which is what https://wordpress.org/plugins/custom-post-type-ui/ and WooCommerce do. We do attempt do a bit of sanitization on the admin's submitted slug though- we do the same sanitization as what's done on post slugs (run it through `sanitize_title()`), except we undo the percent encoding of unicode characters.
`sanitize_title()` does remove accents, which seems unnecessary to me, but if site owners take issue with that, they would have already taken issue with that in post titles (and they could use this code snippet I created: https://github.com/eventespresso/ee-code-snippet-library/blob/master/templates/mn_allow_accents_in_slugs.php), in which case those improvements would apply to our slug too, because we're using the same filter.
Note: I asked about why WordPress removes accented characters from slugs in the first place on stack overflow: https://wordpress.stackexchange.com/questions/310894/why-dont-wordpress-post-slugs-allow-accents

## How has this been tested

* [x] Go to events -> templates and change the events slug to something with accented characters, like `Áventos` and save it. The accents should be removed, and most importantly, you should still be able to view events on the front-end
* [x] Now change the events slug to have other unicode characters in it, like `中文` and save it. They should be left intact, and the new event permalinks should stil lwork
* [x] Try to make an invalid events slug by adding every crazy character and bit of hacky HTML you can, and save. All that bad stuff should get thrown out, and you should just be left with a valid slug
* [x] Try to save a slug with HTML or javascript or CSS or PHP or whatever crazy hacky thing you can think of. That should all be removed as hoped-for. Also, try to put a space in a slug. That should be turned into a dash as usual.

## Checklist

* [ ] My code is tested.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,7 +31,7 @@ See [our documentation](https://github.com/eventespresso/event-espresso-core/blo
 - Fixed Authorize.net AIM so payment currency is sent ([591](https://github.com/eventespresso/event-espresso-core/pull/591))
 - Fixed DuplicateCollectionIdentifierException errors when converting old PersistentAdminNotice Fixes ([505](https://github.com/eventespresso/event-espresso-core/pull/505))
 - Fixes a syntax issue inside `EE_Config::register_ee_widget()` ([608](https://github.com/eventespresso/event-espresso-core/pull/608))
-
+- Fixed URL validation when URL was for a site denying access to our HTTP client ([628](https://github.com/eventespresso/event-espresso-core/pull/628))
 ### Changed
 
 - Updated js build process to use Babel 7 ([578](https://github.com/eventespresso/event-espresso-core/pull/578))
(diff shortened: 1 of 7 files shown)


mnelson4:
Sorry Josh, passing your way again, just because you're so good at reporting issues! (I think it's fine to pass it on to someone else if you like). Testing notes on my first comment.

joshfeck:
This works great and can be merged.

mnelson4:
great; Travis is having issues so I'm running tests locally and am taking off. If it all looks good I'll merge when I get back

mnelson4:
there were some build issues, but unit tests pass locally

Issue #630, opened 2018-08-08
Promotions end point throwing a fatal error


Pebblo:
Reported here: https://eventespresso.com/topic/getting-internal-server-error-500-when-querying-promotions/

**Describe the bug**
If you navigate to the rest api promotions endpoint:

/wp-json/ee/v4.8.36/promotions

You'll get a fatal error:

```
PHP Fatal error:  Uncaught Error: Call to a member function getTimezone() on null in /wp-content/plugins/event-espresso-core-reg/core/libraries/rest_api/controllers/model/Read.php:691
Stack trace:
#0 /wp-content/plugins/event-espresso-core-reg/core/libraries/rest_api/controllers/model/Read.php(579): EventEspresso\core\libraries\rest_api\controllers\model\Read->createBareEntityFromWpdbResults(Object(EEM_Promotion), Array)
#1 /wp-content/plugins/event-espresso-core-reg/core/libraries/rest_api/controllers/model/Read.php(360): EventEspresso\core\libraries\rest_api\controllers\model\Read->createEntityFromWpdbResult(Object(EEM_Promotion), Array, Object(WP_REST_Request))
#2 /wp-content/plugins/event-espresso-core-reg/core/libraries/rest_api/controllers/model/Read.php(84): EventEspresso\core\libraries\rest_api\controllers\model\Read->getEntitiesFromModel(Object(EEM_Promotion), Object(WP_REST_Request))
#3 /wp-content/plugins/event-espresso-core-reg/core/libraries/rest_api/controllers/model/Read.php on line 691
```
(Note you'll need to be authenticated)

**Additional context**
This is an issue with promotions, however the REST API is in core so I think this should be here?

joshfeck:
Some additional info in case it's helpful to know:

If you send a request for the promotion via the line item you get a similar error:
e.g.
/wp-json/ee/v4.8.36/transactions/2037/line_items

error logged is:
```
PHP Fatal error:  Uncaught Error: Call to a member function getTimezone() on null in /srv/www/wordpress-default/public_html/wp-content/plugins/32-core/core/libraries/rest_api/controllers/model/Read.php:691
Stack trace:
#0 /srv/www/wordpress-default/public_html/wp-content/plugins/32-core/core/libraries/rest_api/controllers/model/Read.php(579): EventEspresso\core\libraries\rest_api\controllers\model\Read->createBareEntityFromWpdbResults(Object(EEM_Promotion), Array)
[...]

PR #638, opened 2018-08-09
Fixed null datetimes in REST API

mnelson4:

https://github.com/eventespresso/event-espresso-core/issues/630
Fixed null datetimes in REST API (before they threw a fatal error). If a datetime column has NULL for its value, the model's value will be null, and the REST API will also return null.

## How has this been tested

* [x] Activate the promotions add-on, and create a promotion with no start or end time. Then send an authenticated query to `GET /wp-json/ee/v4.8.36/promotions`. It should no longer have an error, instead the promotion's start and end times should be `null`
## Checklist

* [x] My code is tested.

### DIFF
--- a/core/db_models/fields/EE_Datetime_Field.php
+++ b/core/db_models/fields/EE_Datetime_Field.php
@@ -763,8 +763,8 @@ class EE_Datetime_Field extends EE_Model_Field_Base
         $default_raw = $this->get_default_value();
         if ($default_raw instanceof DateTime) {
             return $default_raw;
-        } elseif( is_null($default_raw)) {
-          return $default_raw;
+        } elseif (is_null($default_raw)) {
+            return $default_raw;
         } else {
             return new DbSafeDateTime(
                 $this->get_default_value(),
--- a/core/db_models/fields/EE_Datetime_Field.php
+++ b/core/db_models/fields/EE_Datetime_Field.php
@@ -750,6 +750,28 @@ class EE_Datetime_Field extends EE_Model_Field_Base
         }
     }
 
+    /**
+     * Gets the default datetime object from the field's default time
+     * @since $VID:$
+     * @return DbSafeDateTime|null
+     * @throws InvalidArgumentException
+     * @throws InvalidDataTypeException
+     * @throws InvalidInterfaceException
+     */
+    public function getDefaultDateTimeObj()
+    {
+        $default_raw = $this->get_default_value();
+        if ($default_raw instanceof DateTime) {
+            return $default_raw;
+        } elseif( is_null($default_raw)) {
+          return $default_raw;
+        } else {
+            return new DbSafeDateTime(
+                $this->get_default_value(),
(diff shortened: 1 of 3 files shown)


core/libraries/rest_api/controllers/model/Read.php
```diff
@@ -688,19 +688,29 @@ protected function createBareEntityFromWpdbResults(EEM_Base $model, $db_row)
...
-                    $field_obj,
-                    $field_value,
-                    $this->getModelVersionInfo()->requestedVersion()
-                );
+                if (is_null($field_value)) {
+                    $gmt_date = $local_date = ModelDataTranslator::prepareFieldValuesForJson(
+                        $field_obj,
+                        $field_value,
+                        $this->getModelVersionInfo()->requestedVersion()
+                    );
```
nerrad: So to be clear, if the field_value is null, then this will use the current_time for both `local_date` and `gmt_date`?  If so, why do we want that?

mnelson4: Hmmm, it depends on whether the field allows `null`s or not. If the field does allow `null`, both `$gmt_date` and `$local_date` will be `null`, which I think makes sense.
But if the field doesn't allow `null`s, I think you're right: they'll both end up being the current time. But that would be an unusual circumstance: the field says it doesn't allow null (and that should mean the database column doesn't allow it either) and yet somehow we've got a null here anyway... in that case, I'd be inclined to have an exception, because I can't think of a case where the we're preparing a non-nullable datetime field whose value from the DB is null. So I'm inclined to have an exception if that happens...

mnelson4: ok, actually it wasn't too hard to handle, so no exception.
So I can't think of a time when the model field won't allow nulls, but the value from the DB is null anyway, but even if that happens, let's set the value to be the field's default. And even then, the GMT field should be the default time in GMT, and the non-GMT field should be the default time in the site's timezone.

core/db_models/fields/EE_Datetime_Field.php
```diff
@@ -750,6 +750,28 @@ public function get_default_value()
...
+     * @throws InvalidArgumentException
+     * @throws InvalidDataTypeException
+     * @throws InvalidInterfaceException
+     */
+    public function getDefaultDateTimeObj()
+    {
+        $default_raw = $this->get_default_value();
+        if ($default_raw instanceof DateTime) {
+            return $default_raw;
+        } elseif( is_null($default_raw)) {
```
nerrad: improper spacing after opening `(` bracket here.

nerrad:
I approved, but there's some lint errors needing fixed.

mnelson4:
right thanks Darren. Darn I totally missed that. Ok passing off for testing

Pebblo:
Tested on a site I know already throws a fatal on the above endpoint and with this branch, no more fatal.

Good to go.

Issue #574, opened 2018-07-10
Improve rewrite slug escaping


joshfeck:
Originally reported here:
https://eventespresso.com/topic/warning-message-at-the-top-of-the-menu/

Steps to replicate the issue:
- Activate WP_DEBUG
- Switch Site Language to Français
- Refresh permalinks
- View a single post (of any post type), where you'll see PHP warnings like these:

```Warning: preg_match(): Unknown modifier '0' in /srv/www/plugins/public_html/wp-includes/class-wp.php on line 219

Warning: preg_match(): Unknown modifier '0' in /srv/www/plugins/public_html/wp-includes/class-wp.php on line 218
```

I did some troubleshooting and tracked this down to the translation of the `event-type` text string, which is basically an unused taxonomy for the Event Espresso event post type. The [French translation of the string](https://translate.eventespresso.com/projects/event-espresso-4/fr/event-espresso-fr_fr/?filters%5Bstatus%5D=either&filters%5Boriginal_id%5D=14199&filters%5Btranslation_id%5D=82540) has a space in place of the dash:  `type d'événement`

So it turns out that the escaping used on rewrite slugs should not be `esc_html()`, but instead use something like `sanitize_title()`.  

As mentioned in chat, there's a slight chance this could be a breaking change if there's custom code out there actually relying on an improperly escaped rewrite slug. Probably a very slim chance though.

mnelson4:
PR merged in

PR #641, opened 2018-08-10
Fixed translated event taxonomy slugs so they stay valid

mnelson4:

https://github.com/eventespresso/event-espresso-core/issues/574
Fixed translated event taxonomy slugs so they stay valid (ie, don't get any invalid characters that aren't permitted in a URL). I couldn't get the old slugs to work anyway (eg in French), so I don't think harm in this change. Also, I opted to have the translated slugs NOT include accents, as I couldn't the URLs to work when they had accents. (Plus this is what WP does with post slugs anyways, see https://wordpress.stackexchange.com/questions/310894/why-dont-wordpress-post-slugs-allow-accents?noredirect=1#comment459608_310894)

## How has this been tested

* [ ] Create an event category with the slug "sluggy", then go to `/event-category/sluggy`. Your event should appear in the list. Then switch the site language to French, and refresh permalinks. Then go to `/cas-category/sluggy` and your event should still appear
* [ ] With your site still in French, go to `/type-devenement/single-event`, it should show all events (FYI, the event type "single-event" isn't modifiable, but nor is it really used because we haven't told anyone about it)
* [ ] 

## Checklist

* [ ] My code is tested.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@ See [our documentation](https://github.com/eventespresso/event-espresso-core/blo
 - Fixed DuplicateCollectionIdentifierException errors when converting old PersistentAdminNotice Fixes ([505](https://github.com/eventespresso/event-espresso-core/pull/505))
 - Fixes a syntax issue inside `EE_Config::register_ee_widget()` ([608](https://github.com/eventespresso/event-espresso-core/pull/608))
 - Fixed URL validation when URL was for a site denying access to our HTTP client ([628](https://github.com/eventespresso/event-espresso-core/pull/628))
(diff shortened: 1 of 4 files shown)


tn3rb:
restarted those two failed tests which looked like Travis box setup issues, they should pass

mnelson4:
I realized why the accented characters were causing problems- same issue as on https://app.zenhub.com/workspace/o/eventespresso/event-espresso-core/issues/635. So it needed the same fix- just `sanitize_title()` then `urldecode`. But in doing that, I realized there was an opportunity to make things a bit DRYer.
Also, these changes will overwrite what's in https://github.com/eventespresso/event-espresso-core/pull/635 (but Josh JUST tested that, so I think we may as well continue with that, in case this takes a bit longer)

PR #644, opened 2018-08-13
Added ‘format: money’ to schema for EE_Money_Field fields

nerrad:

For our javascript model entities (see #637), we're using the json-schema provided by our REST endpoint to generate the model entity constructors that automatically validate incoming data against the schema.  One thing we'd like to do is internally in these model entities, convert the raw value to a `moment` instance for datetime fields and to a `money` vo instance (for money fields).  For DateTime fields we are able to use the `format` key on the schema to know its "type".  However, there's no current way to know that an `integer` or `number` type in the schema represents a "money" value.  

This pull proposes adding `format:money` to the `EE_Money_Field` schema so that clients can derive that the value is derived from that field.  

## How has this been tested

* [x] Tested via checking the response on a schema request for money fields.  There should be a `format:money` property on the `properties.pretty` element.  So for example the schema for registrations would have the following object in its schema:

```json
"REG_paid" : {
    "properties": {
         "pretty": {
            "format": "money"
         }
    }
}
```

## Checklist

* [x] My code is tested.


mnelson4:
ya that sounds like a useful and simple change

Issue #394, opened 2018-05-07
Fix error when editing active plugins via the built-in editor


joshfeck:
Steps to replicate with EE4 master activated:
1) Go to the WP > Plugins > Editor page and try to make an edit to any **active** plugin.
2) Click "Update File"

The unexpected result will be this message:

```
Your PHP code changes were rolled back due to an error on line 56 of file wp-content/plugins/32-core/caffeinated/modules/recaptcha_invisible/InvisibleRecaptcha.php. Please fix and try saving again.
```
```
Uncaught TypeError: Argument 2 passed to EventEspresso\caffeinated\modules\recaptcha_invisible\InvisibleRecaptcha::__construct() must be an instance of EE_Session, none given in wp-content/plugins/32-core/caffeinated/modules/recaptcha_invisible/InvisibleRecaptcha.php:56
Stack trace:
#0 [internal function]: EventEspresso\caffeinated\modules\recaptcha_invisible\InvisibleRecaptcha->__construct(Object(EE_Registration_Config))
#1 wp-content/plugins/32-core/core/EE_Registry.core.php(1262): ReflectionClass->newInstanceArgs(Array)
#2 wp-content/plugins/32-core/core/EE_Registry.core.php(773): EE_Registry->_create_object('EventEspresso\\c...', Array, '', false)
#3 wp-content/plugins/32-core/core/services/loaders/CoreLoader.php(116): EE_Registry->create('EventEspresso\\c...', Array, true, false, false, '')
#4 wp-content/plugins/
```

mnelson4:
note: `wp_ajax_edit_theme_plugin_file` is what calls the extra asynchronous request to check if a fatal error was added.

mnelson4:
note: `wp_ajax_edit_theme_plugin_file` is what calls the extra asynchronous request to check if a fatal error was added.

mnelson4:
What's really odd: when I use `wp_remote_get` as a "loopback" request, it doesn't work. But if I use CURL directly, like on https://stackoverflow.com/questions/5412069/can-i-do-a-curl-request-to-the-same-server, it works fine. So VVV and laragon setups have some setup, combined with something in how wp's `wp_remote_get` works, that's causing the loopback requests to fail...

mnelson4:
aha- I think I pinpointed the problem being `Requests_Transport_cURL::stream_headers()`. If I disable that method from being invoked, the loopback request works fine

mnelson4:
[...]

PR #645, opened 2018-08-13
Fixed error when editing active plugins via the built-in editor

mnelson4:

https://github.com/eventespresso/event-espresso-core/issues/394.
I actually never managed to get WordPress' built-in plugin editor to work with either VVV or laragon (which was really strange, because doing direct loopback cURL requests worked fine, just not with WordPress' `wp_remote_get()`), and only managed to reproduce on a temporary site from poopy.life. So I wasn't able to isolate why exactly `InvisibleRecaptcha` wasn't getting the session passed in. But from my testing, simply setting a default value for the session solved the error. So this is really just a band-aid, but considering how much time I've spent on this and it's relatively minor value, it seemed like a good move forward.

## How has this been tested

* [ ] Activate EE and setup the invisible recaptcha. Then use WordPress' built-in plugin editor to edit an active plugin. It should work fine. (Note: you may get an error using WordPress' built-in plugin editor like me, in which case you should try from a server where that works, without EE active, first. Eg poopy.life installs work)
* [ ] Verify you can still register as normal with the invisible recaptcha active

## Checklist

* [ ] I have added a changelog entry for this pull request -> we are discontinuing this, as I understand it
* [x] My code is tested.


mnelson4:
@tn3rb you might have a better solution for this (specifically, you might know why the `EE_Session` wouldn't get injected into `InvisibleRecaptcha` on these special loopback requests), but I was thinking this is at least a working band-aid, and this is a relatively minor bug anyways.

tn3rb:
sorry but I can't accept this at all because it makes a required class dependency into an optional parameter. 
I understand that it appears to fix the issue, but this is not an acceptable way to go about it.

I took a look into the issue and created #650 to address the issue properly.
That PR has been assigned to you for review.
This one should be closed unless you have other questions comments to discuss here.

mnelson4:
yeah no worries, it was just a band-aid fix. I couldn't come up with anything better because of my difficulty reproducing this issue. Glad you could trace it better.

PR #647, opened 2018-08-15
Fixes the 1-click upgrade from Decaf so that a prompt to upgrade is displayed when a correct support key is entered

### LINKED ISSUE #609: Fix the 1-click upgrade from Decaf
Many users are reporting they've purchased an Event Espresso support license, entered a support key, and are not getting a prompt to upgrade to Caffeinated. So I've been advising them to manually do the upgrade.

I've replicated this issue on my test site at josh.eventespresso.com by deactivating EE4 regular, installing and activating EE4 decaf, then saving the General Settings page with the correct support key in place. No prompt to upgrade is displayed.

This might not be an issue with core and instead an issue on the server side (on ee.com's server), but I'm reporting this here so it can be triaged.

knazart:
This fixes #609 
A prompt to upgrade to Caffeinated from Decaf was not being displayed when a valid support key was entered. Work here should fix that.

## How has this been tested
* [x] Install and activate EE4 decaf (it should include the same changes done here). Then save the General Settings page with the correct support key in place. A prompt (as an admin notification) to upgrade should be displayed.

## Checklist

* [x] My code is tested.


joshfeck:
Awesome. The fix works so this is ready for merging.

PR #648, opened 2018-08-15
Added InvalidArgumentError Exception

nerrad:

This pull adds a `InvalidArgument` exception for use when asserting whether arguments provided to a function or class method are valid.  The argument might still be the right expected type but not be the expected value.

For example, in the DateTime value object I'm writing, I want to assert the the provided timezone string is a valid timezone string (and parseable via the library in use for manipulating dates and times).  For the assertion method, I want to throw `InvalidArgument` when the timezone string is not valid.

## How has this been tested

* [x] Verify no existing tests are broken

## Checklist

* [x] My code is tested.

### DIFF
--- a/assets/src/eejs/exceptions/invalid-argument.js
+++ b/assets/src/eejs/exceptions/invalid-argument.js
@@ -1,13 +1,12 @@
 /**
  * InvalidArgument
- * Usage: throw new eejs.InvalidArgument('some message', [parameter, argument])
+ * Usage: throw new eejs.InvalidArgument('some message'[, argument])
  *
  * Typically this error is thrown when a function or method is called with an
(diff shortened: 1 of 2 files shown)


assets/src/eejs/exceptions/invalid-argument.js
```diff
@@ -0,0 +1,23 @@
...
+ */
+export default class InvalidArgument extends Error {
+	constructor( ...args ) {
+		super( ...args );
+		if ( Error.captureStackTrace ) {
+			Error.captureStackTrace( this, InvalidArgument );
+		}
+		this.message = this.message !== '' ?
+			'Invalid argument provided. ' + this.message :
+		'Invalid argument provided.';
```
tn3rb: oooo actually this needs to be indented one more space

PR #651, opened 2018-08-16
Allow for null gateway log objects when creating payment logs

Pebblo:
As much as I really don't like payment logs that don't pass over a payment object when creating the log, we need to allow for them and more specifically for the current problem allow for them to be null.

In #545 I worked with Mike to make a change to core that allowed for transaction objects to be passed during the log calls which helped, however, in doing so we also changed the log calls to **always** require either an EEI_Payment, EEI_Transaction or EEI_Payment_Method object to be passed.

See the change here: https://github.com/eventespresso/event-espresso-core/pull/545/commits/3f216631cf3068bab540df9cc8cae7b468448144#diff-a8effa8ccc4d96fc1dc72d8b189ade94

Unfortunately, that's not always the case and previously it was fine because if it wasn't an EEI_Payment object, it was always logged as a payment method log, now we have no else clause.

To create a payment log entry, if EE is passed a Payment object it will link it correctly to the log entry, if its passed a Transaction object, it links it correctly, if its passed anything else, be it a Payment method object or null, it will now log it as a generic 'Payment method' log entry linked to the current payment method.

Doing so doesn't really help with troubleshooting (no less than if the log entry actually passed a payment method object to the call), but it prevents any errors with odd log calls.

Core fix.

## How has this been tested
Use the vanco payment method, it doesn't need to be set up correctly but when the PM attempts to pull the session_id from Vanco it creates log entries and passes null as an object to link it to.

So add the PM, add some dummy details, attempt to make a payment.

Master will throw a fatal error, this branch will not.

## Checklist

* [ ] My code is tested.


mnelson4:
also, in order to get this moving along a bit quicker, I tested it and can confirm the old bug existed and that it's fixed

mnelson4:
up to you @Pebblo if you want to pass this off to someone more to test

Pebblo:
Thanks, Mike!

I ran through some additional testing and didn't run into any issues with Stripe, Vanco or PayPal Express logging.

I'll merge this in.

joshfeck:
@Pebblo a friendly reminder: when you merge a PR please also remove the status:merge label and add the PR to current milestone.

Issue #595, opened 2018-07-19
"You need to select a ticket quantity" notice only displays for logged in users


joshfeck:
Reported here:
https://eventespresso.com/topic/error-message-when-zero-0-tickets-selected-not-displaying/
and confirmed locally

**Describe the bug**

No error message is displayed if you're logged out of the site and do not select a ticket quantity and just click the submit the button below the ticket selector.
**To Reproduce**

**Expected behavior**

There should be an error message, and there was one prior to Event Espresso 4.9.59.p

**Environment:**
 - Version of Event Espresso:  Current master, and current releases since 4.9.59.p

**Additional context**

This looks to be related to the work done in https://events.codebasehq.com/projects/event-espresso/tickets/11406

sethshoultes:
@mnelson4 can you fix this? If not, please pass to @tn3rb .

joshfeck:
@mnelson4 I'm not sure if you've looked into this, but one other thing I found while investigating this is it appears to be something going awry related to part of the `getStoredNotices()` method in EE_Error.core.php. I did a `var_dump($session_data);` inside that method's second conditional and it always returns an empty array when I'm logged out of the site and submit a ticket selector form with no selections. 
If I do a 
```var_dump(EE_Session::instance()->set_session_data(
  array(EE_Error::OPTIONS_KEY_NOTICES => $notices)
);```
just before the return in the `storeNotices()` method (next method down in the same class) it does return true, so it seems that something happening between these two methods. In chat, Darren mentioned it could be the notices aren’t getting persisted to the database.

**Comment by joshfeck** [comment]
Another clue is everything works as expected if MER is activated. So you can be logged out of the site, hit the ticket selector submit with 0 ticket selections, and the error notice **does** display. 

From what I can tell, the session vars aren't getting saved to the db in the above reported scenario because of the checks at the beginning of the _save_session_to_db method. Note also the comment there:
```
// and unless we're deleting the session data, don't save anything if there isn't a cart
```

So in this case, there isn't a cart, so the error notices will not get saved to the db as session vars.

joshfeck:
Brent mentioned in chat that "we could maybe add some logic to the session to also check for notices when saving since I don't think there's many other (if any) cases where notices get saved without a cart -- so it wouldn't be increasing the usage of the session much -- otherwise those notices have the potential to go to someone else"

mnelson4:
> So in this case, there isn't a cart, so the error notices will not get saved to the db as session vars.
[...]

PR #652, opened 2018-08-17
Fixed "You need to select a ticket quantity" error so it displays to logged-out users

mnelson4:

https://github.com/eventespresso/event-espresso-core/issues/595
When a logged-out selected to register for an event, but forgot to select a ticket quantity, the error message wouldn't display for them (it did for logged-in users still). This fixes is so it will also be displayed to them.
## How has this been tested

* [x] Log out from the site. Go to an event page. Click "Register Now" *without* selecting a quantity. You should see a helpful error message pointing you need to select a ticket quantity (on master you wouldn't)
* [x] Log in and repeat the above test. Double-check I didn't bork that somehow
* [ ] Other tests you'd suggest @tn3rb? (eg, if I accidentally made it so the session ALWAYS got saved, what symptoms would we see?)

## Checklist

* [ ] I have added a changelog entry for this pull request -> not doing this anymore, as I understand it
* [ ] My code is tested.


joshfeck:
I tested this and the results of the testing are good. This is ready for merging.

mnelson4:
resolves #595

mnelson4:
fixes #595

mnelson4:
utterly demolishes, salts the fields of, and cuts off the posterity of (but doesn't automatically close) #595

PR #660, opened 2018-08-23
Added a filter so the system status page's content can be modified

mnelson4:

Added a filter so the system status page's content can be modified.
While debugging https://github.com/eventespresso/eventsmart.com-website/issues/108 I found myself wanting some more info from a site. I could dig around directly in the DB and write custom scripts, but I thought it would be handy if add-ons (like the saas add-ons) could add more info to the system status page. Whatever info we might think helpful when debugging issues on that site. (In my particular case, I want to see exactly what add-ons are actually still registered and their versions; right now it's not very accessible).

## How has this been tested

* [x] Just verify the normal system status page didn't get broken

## Checklist

* [ ] My code is tested.

### DIFF
--- a/core/db_models/EEM_System_Status.model.php
+++ b/core/db_models/EEM_System_Status.model.php
@@ -38,7 +38,9 @@ class EEM_System_Status
      */
     public function get_system_stati()
     {
-        return array(
+        return apply_filters(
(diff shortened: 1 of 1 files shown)


joshfeck:
Tested and verified that the Event Espresso > Maintenance > System Information page remains the same.

PR #667, opened 2018-08-26
Fixed EEM_Question filter name so it's not a duplicate of existing filter

tn3rb:
`EEM_Question::__construct()` contains the following two filters that are **BOTH** using `FHEE__EEM_Question__construct__allowed_question_types` for their name:

```php
$this->_allowed_question_types = apply_filters(
            'FHEE__EEM_Question__construct__allowed_question_types',
            array( /* array content */ )
        );
        $this->_question_descriptions = apply_filters(
            'FHEE__EEM_Question__construct__allowed_question_types',
            array( /* array content */ )
        );
```

This is super problematic because any changes made using that filter will affect both arrays of data.

This appears to have been introduced during a merge to master in this commit:
https://github.com/eventespresso/event-espresso-core/blob/c8ef5adedfadddc6793e6e3597d1c7fa6fa07661/core/db_models/EEM_Question.model.php

So fortunately it looks like it has never been an issue.

## Checklist

* [x] My code is tested.

### DIFF
--- a/core/db_models/EEM_Question.model.php
+++ b/core/db_models/EEM_Question.model.php
@@ -131,7 +131,7 @@ class EEM_Question extends EEM_Soft_Delete_Base
             )
         );
         $this->_question_descriptions = apply_filters(
-            'FHEE__EEM_Question__construct__allowed_question_types',
+            'FHEE__EEM_Question__construct__question_descriptions',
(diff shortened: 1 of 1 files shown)


mnelson4:
thanks

Issue #658, opened 2018-08-22
Remove database field names appended to Registrations CSV report column headings


joshfeck:
The Registrations CSV reports have column headings that include the database field name (wrapped in square brackets). These seem to be unneeded and customers have asked about how to remove them. 

I've put together a little snippet that removes them on-the-fly via a filter hook:

https://gist.github.com/joshfeck/a4d95889cf4b9bfbe8d09e9eb9d8f590

It'd be cool if those were not included by default, but if possible, we'll need to take care to preserve some backwards compatibility to avoid breaking customizations like this one:
https://github.com/eventespresso/ee-code-snippet-library/blob/master/admin/tw_ee_registration_report_limit_columns_shown.php

sethshoultes:
I don't remember seeing the database field names in the CSV before. Is that something that was added recently?

@mnelson4 this something you can fix really quick?

mnelson4:
Ya that should be a pretty quick fix

PR #675, opened 2018-08-30
Changed CSV reports so headers don't include square brackets with field name

mnelson4:

https://github.com/eventespresso/event-espresso-core/issues/658
Changed CSV reports so headers don't include square brackets with field name. (That was originally there for CSV import/exports, but chances are, when we re-implement that, we may do it entirely differently anyway).

## How has this been tested

* [ ] View all the CSV reports (all attendee report, all registrations report, filtered registrations report). Column headers should never have square brackets in them.

## Checklist
* [ ] My code is tested.


mnelson4:
thanks Darren. Passing off to @joshfeck for testing because he reported the issue

Pebblo:
@mnelson4

If any users are using the functions we've been providing to customize the allowed fields and/or order of fields in the CSV, this change is going to break it? For example:

https://github.com/eventespresso/ee-code-snippet-library/blob/master/admin/tw_ee_registration_report_limit_columns_shown.php

https://github.com/eventespresso/ee-code-snippet-library/blob/master/admin/tw_registration_report_exclude_fields.php

https://github.com/eventespresso/ee-code-snippet-library/blob/master/admin/registration_report_limit_columns_shown.php

It does depend on the fields you are targeting, but either way, it's going to break those snippets until they are updated.

Can we filter the field name so we can at least revert this change easily for those users?
Or, can that be done with the current filters?

mnelson4:
Good point @Pebblo, thanks for thinking about that.
I think you can technically do that with the current filters, but it would be troublesome. Probably best to add a filter

mnelson4:
passing to @Pebblo to review the addition of the filter, see if it will meet your needs. You might like to test it too, or pass to Josh

Pebblo:
All good from me, however, when this was posted in support chat Josh mentioned another method which I thought was worth bringing up to see what you think.

Rather than the above filter to switch back to using the DB field, he suggested hooking in at a late priority and removing them then, that way we keep all of the functionality we have now with the existing filters but get clean CSV headers.

Would that be better @mnelson4?

mnelson4:
> Rather than the above filter to switch back to using the DB field, he suggested hooking in at a late priority and removing them then, that way we keep all of the functionality we have now with the existing filters but get clean CSV headers.

That's nice for folks who currently have customization. About what percentage of users would you guess have customizations? 

The main downside is that solution is it makes the code more complicated in the future, and quite strange for any future customizations (the column name they see in the CSV file is not what they need to name the column while customizing the column).

If it's somewhere around only 1% of folks using the report that have customizations that will need to be updated, I'd be inclined to leave this as-is.

Or, we could check if anyone is currently using the filter to change the report, and if so, use the old column names again.

Pebblo:
> About what percentage of users would you guess have customizations?

Tbh it's really difficult to say, maybe ~5%

It's enough to know that I've given out links to the various functions we have to enough people for it to cause us some grief on the forums if we change it now. My main concern is the posts that start with something like "The client wants to remove X,Y,Z from the CSV", which will be straight back to us for changing the columns.

I think this:

> Or, we could check if anyone is currently using the filter to change the report, and if so, use the old column names again.

would be best.

We can then work on updating the gists etc to stop using the column names and all works well, right?

mnelson4:
Ok I think I managed to turn on an extra 1% of my brain and found a slick solution that:

* preserves backward compatibility with all the code snippets I could find
* allows new code snippets to use the current column header names (instead of the older ones, which my original plan would have required)
* even if soon-to-be legacy code snippets are used, we'll still remove the ugly square brackets from column headers (which you would think would conflict with the first bullet point, but magically it doesn't)

Here's what I did (it's unfortunately clever):
* replaced the old filter `FHEE__EE_Export__report_registrations__reg_csv_array` (whose name needed to be changed eventually anyway, it indicates the called from somewhere else) with a new filter `FHEE__EventEspressoBatchRequest__JobHandlers__RegistrationsReport__reg_csv_array`
* we hook the function `ee_deprecated_registrations_report_csv_legacy_fields` into the new function. 
* That function checks if any code is trying to use the now-removed filter. If so, it updates the row to temporarily use the square brackets in the column headers (array indexes) again. The array now looks exactly how it did before this PR
* It runs the array through the legacy filter (which old code snippets still hook into, and update exactly like they did before)
* once all those legacy filter callbacks are done, loops over the array and removes the ugly square brackets again. Not only because they're ugly, but new code snippets won't be expecting them

My favourite part of this is that because this new filter is in `EE_Deprecated`, all the backward compatibility code is isolated in there and can be deleted once we feel like cleaning house again.

I ran this with all the code snippets I could find and it worked. @Pebblo do you want to double-check it works with the code snippets of which you're aware?

Pebblo:
THIS. IS. AWESOME!

I threw a bunch of functions at it that add details, reorder, exclude, a bit of everything really and all using the original filter.... all worked fine!

I then threw a few some of the same functions at the new filter without using the field_name in the brackets just to make sure it actually worked (no offence Mike :) )... no surprise that all worked as exepcted.

Hat tip @mnelson4 as this is a great solution to an awkward problem.

I've added my review and code wise it seems fine, however I'll admit I'm not up to par on the EE coding standards so when I see inconsitent spacing with comments like [THIS](https://github.com/eventespresso/event-espresso-core/pull/675/commits/145ef7d72041b03900ea960b83d2b2e016a0402a#diff-043f114ab070627e8054b4a519f1717aR1426) and [THIS](https://github.com/eventespresso/event-espresso-core/pull/675/commits/145ef7d72041b03900ea960b83d2b2e016a0402a#diff-043f114ab070627e8054b4a519f1717aR1453) (note the spacing around some comments/code and not others) I'm not sure if thats a problem, although I'm just nit picking as I'm unsure, to me everything looks fine.

tn3rb:
just a reminder that all code reviews should be performed by Darren or myself (or Nazar for gateways).

mnelson4:
thanks @Pebblo for checking it works with your code snippets.

Thanks @tn3rb for the reminder; I was on the fence whether or not this needed another review (I think at one point we were pretty loose with re-reviewing but that might just be my imagination :smiley: ).

Passing to @nerrad for another review then.

core/EE_Deprecated.core.php
```diff
@@ -1381,4 +1381,102 @@ public static function is_update_available($basename = '')
...
+ * Filters the CSV row to make it appear like the old labels (which were "$pretty_name[$field_name]").
+ * @since $VID:$
+ * @param $csv_row_data
+ * @param $reg_row
+ * @return array for CSV row
+ * @throws EE_Error
+ * @throws InvalidArgumentException
+ * @throws \EventEspresso\core\exceptions\InvalidDataTypeException
+ * @throws \EventEspresso\core\exceptions\InvalidInterfaceException
+ */
```
nerrad: add a `@deprecated` php doc tag please.  I do realize that there is other code in this file that doesn't have that tag (which is unfortunate).  The tag helps us with knowing when something was deprecated.

core/EE_Deprecated.core.php
```diff
@@ -1391,19 +1393,35 @@ public static function is_update_available($basename = '')
 );
 /**
  * Filters the CSV row to make it appear like the old labels (which were "$pretty_name[$field_name]").
+ * @deprecated since $VID:$. This only exists for backward compatibility with code snippets.
```
nerrad: This should be:

```
* @deprecated $VID:$ ...description
```
I realize that you may have gone by the deprecated tag on the `ee_deprecated_update_attendee_registration_form_old` but that is also wrong there.

mnelson4:
ah kk!

mnelson4:
@tony the main change you last tested is that we're adding an error message when folks are using the old filter, but from my testing it doesn't interfere with downloading the CSV files or corrupt them at all. 
Could you test with one of your code snippets? If its ok with one the rest should be fine too.

mnelson4:
> so I've removed those and committed (seemed daft to pass it back just for that and then retest), then tested.

Thanks for taking the initiative, but it's fishy that was necessary. I realized we were actually escaping the HTML twice so instead put the quotes back in and avoided escaping HTML the extra time.

> As this is one of our own add-ons, would it be better holding off of the notice for a little, push an update for the Promotions add-on and then add it?

Oh good catch. Yes I think we should delay the notice a little bit, and update the promotions add-on.

@nerrad you should probably re-review, sorry.

joshfeck:
This still needs further review.

mnelson4:
argh sorry obviously I thought I assigned Darren, and obviously not.
Kk this is ready for testing again

Pebblo:
@mnelson4 shouldn't [$VID:$](https://github.com/eventespresso/event-espresso-core/pull/675/commits/24aa0dfed1ac738f34ec19211b8e059289ccf604#diff-043f114ab070627e8054b4a519f1717aR1423) also be set to `4.9.75.p`?

Otherwise wont we get that messages showing up in 4.9.75.p but showing it was added in 4.9.69 (next version of EE at time of writing)?

Pebblo:
Other than the above (if its an issue) this is good to go.

I've retested this and it all works fine, the message shows up when I set my local version of EE to use 4.9.76.rc.010

mnelson4:
> @mnelson4 shouldn't $VID:$ also be set to 4.9.75.p?

Good question. As I understand it, the reason we can provide the two versions to `EE_Error::doing_it_wrong()` is so they can be different (otherwise we should have only allowed passing in one version string). We are deprecating something now (so developers will see it's deprecated now), but we don't want to start bothering regular users (with WP_DEBUG on) until later.

Thanks! Merging

PR #676, opened 2018-08-30
Add instanceof checks to avoid fatal errors on txn page

joshfeck:
There were a number of fatal errors being thrown if I tried to view a txn that had been deleted (or supplied an invalid txn ID in the URL to the txn page). These additional checks avoid the fatal errors.

## Checklist

* [x] My code is tested.

### DIFF
--- a/admin_pages/transactions/Transactions_Admin_Page.core.php
+++ b/admin_pages/transactions/Transactions_Admin_Page.core.php
@@ -470,7 +470,9 @@ class Transactions_Admin_Page extends EE_Admin_Page
         $this->_session = $this->_transaction instanceof EE_Transaction
             ? $this->_transaction->get('TXN_session_data')
             : null;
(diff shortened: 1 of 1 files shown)


joshfeck:
Yep, there's already code in there to show a message if there's no valid transaction, it just didn't show because the fatal errors would happen before it could be displayed. The message is: 
> An error occurred and the details for the transaction with the ID # %d could not be retrieved.

When this PR gets tested the user should see the above message.

joshfeck:
This needs testing.

To test, load up a transaction from Event Espresso > Transactions. Then add a few digits to the txn Id in the URL. For example if the TXN ID is 1299, change the URL to have TXN ID 1299999. With master you'll get a fatal error (or a white screen if no WP_DEBUG). With this branch, you'll get an error message explaining an error occurred.

Pebblo:
This works fine, the error message displays fine and normal transactions still work as expected.

PR #680, opened 2018-09-03
Improve REST Schema to include missing fields

nerrad:

The REST API JSON schema for resource endpoints is missing `link`, `_links`, and `_calculated_fields` from the response.  The work in this pull adds them.

## How has this been tested

* [x] Add unit tests and verified they passed.
* [x] Checked the schema endpoint for various resources (manual requests) and verified the response is as expected (ie. `OPTION http://ee.test/wp-json/ee/v4.8.36/events`)

## Checklist

* [ ] My code is tested.

### DIFF
--- a/core/entities/models/JsonModelSchema.php
+++ b/core/entities/models/JsonModelSchema.php
@@ -190,30 +190,30 @@ class JsonModelSchema
                         ),
                         'readonly' => true
                     ),
-                    'additionalProperties' => array(
-                        'type' => 'array',
-                        'description' => esc_html(
-                            'Additional property keys are the link to relations resource collection.',
-                            'event_espresso'
-                        ),
-                        'items' => array(
-                            'type' => 'object',
-                            'properties' => array(
-                                'href' => array(
-                                    'description' => esc_html(
-                                        'Link to relations collection for this entity resource.',
-                                        'event_espresso'
-                                    ),
-                                    'type' => 'string',
-                                ),
-                                'single' => array(
-                                    'type' => 'boolean'
-                                )
-                            ),
-                            'additionalProperties' => false
-                        )
-                    )
                 ),
+                'additionalProperties' => array(
+                    'type' => 'array',
+                    'description' => esc_html(
+                        'Additional property keys are the link to relations resource collection.',
+                        'event_espresso'
+                    ),
+                    'items' => array(
+                        'type' => 'object',
+                        'properties' => array(
+                            'href' => array(
+                                'description' => esc_html(
+                                    'Link to relations collection for this entity resource.',
+                                    'event_espresso'
+                                ),
+                                'type' => 'string',
+                            ),
+                            'single' => array(
+                                'type' => 'boolean'
+                            )
+                        ),
+                        'additionalProperties' => false
+                    )
+                )
             ),
             '_calculated_fields' => $this->fields_calculator->getJsonSchemaForModel($this->model)
         );
--- a/core/entities/models/JsonModelSchema.php
+++ b/core/entities/models/JsonModelSchema.php
@@ -123,6 +123,45 @@ class JsonModelSchema
                 : EEH_Inflector::pluralize_and_lower($model_name);
             $schema['properties'][ $model_name_for_schema ] = $relation->getSchema();
             $schema['properties'][ $model_name_for_schema ]['relation_model'] = $model_name;
+
+            // links schema
+            $links_key = 'https://api.eventespresso.com/' . strtolower($model_name);
+            $schema['properties']['_links']['properties'][ $links_key ] = array(
+                'description' => esc_html__(
+                    'Array of objects describing the link(s) for this relation resource.',
+                    'event_espresso'
+                ),
+                'type' => 'array',
+                'readonly' => true,
+                'items' => array(
+                    'type' => 'object',
+                    'properties' => array(
+                        'href' => array(
+                            'type' => 'string',
+                            'description' => sprintf(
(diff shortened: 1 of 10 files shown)


core/entities/models/JsonModelSchema.php
```diff
@@ -130,14 +140,90 @@ public function getInitialSchemaStructure()
...
+                                'href' => array(
+                                  'type' => 'string'
+                                ),
+                            ),
+                            'additionalProperties' => false
+                        ),
+                        'readonly' => true
+                    ),
+                ),
+                'additionalProperties' => array(
```
mnelson4: we could probably define all the other links, seeing how we know the related models. Do you think that would be worth adding? (Or maybe adding later?)

core/libraries/rest_api/CalculatedModelFields.php
```diff
@@ -45,7 +45,6 @@ public function mapping($refresh = false)
     {
         if (! $this->mapping || $refresh) {
             $this->mapping = $this->generateNewMapping();
-            $this->mapping_schema = $this->generateNewMappingSchema();
```
mnelson4: Having the method `mapping()` generate and cache the `mapping_schema` seemed like an unexpected side-effect to me. It may also have been an unnecessary one, because there are requests when we needed to use `mapping()` to calculate a field as requested, but not need the schema.
So, it seemed to make more sense to me that we only generate and cache `mapping_schema` when `getJsonSchemaForModel()` got called.

mnelson4:
Passing back to you Darren with some comments, and some tweaks of my own that you may have feedback on

core/entities/models/JsonModelSchema.php
```diff
@@ -112,6 +123,45 @@ public function getModelSchemaForRelations(array $relations_on_model, array $sch
                 : EEH_Inflector::pluralize_and_lower($model_name);
             $schema['properties'][ $model_name_for_schema ] = $relation->getSchema();
             $schema['properties'][ $model_name_for_schema ]['relation_model'] = $model_name;
+
+            // links schema
+            $links_key = 'https://api.eventespresso.com/' . strtolower($model_name);
+            $schema['properties']['_links']['properties'][ $links_key ] = array(
```
nerrad: Your suggestion to implement the actual relation links was a good one so I went ahead and did this here.

core/entities/models/JsonModelSchema.php
```diff
@@ -129,15 +178,71 @@ public function getInitialSchemaStructure()
             '$schema'    => 'http://json-schema.org/draft-04/schema#',
             'title'      => $this->model->get_this_model_name(),
             'type'       => 'object',
-            'properties' => array(),
+            'properties' => array(
```
nerrad: I realized I added `link`, `_links`, and `_calculated_fields` to the wrong index on the schema array and they needed to be nested in the top-level `properties` element.  Hence, the change here.

core/libraries/rest_api/calculations/Attendee.php
```diff
@@ -2,14 +2,8 @@
...
-use EE_Attendee;
-use EE_Error;
-use EEM_Attendee;
-use EventEspresso\core\exceptions\InvalidDataTypeException;
-use EventEspresso\core\exceptions\InvalidInterfaceException;
-use EventEspresso\core\libraries\rest_api\calculations\Base as Calculations_Base;
-use EventEspresso\core\libraries\rest_api\controllers\model\Base;
-use InvalidArgumentException;
+use EventEspresso\core\libraries\rest_api\calculations\Base as AttendeeCalculationsBase;
+use EventEspresso\core\libraries\rest_api\controllers\model\Base as AttendeeControllerBase;
```
nerrad: I had to make these modifications to every Calculations class because of a related fatals occurring in unit tests run in PHP5.6 and PHP5.5 (exposed on travis).  I suspect this is because of the new schema generation being tested for `_calculated_fields` and thus something like this wasn't exposed prior to this branch. 

I suspect this is related to [this php bug](https://bugs.php.net/bug.php?id=66773) and _may_ also be related to the `class_implements` check I'm doing which surfaces that bug.  At any rate, the solution is to have unique aliases across the static classes as I implemented.  So you'll see this in all the Calculations classes.

nerrad:
I think its worthwhile (and should be done) to move the calculated fields objects to non static classes and also to use proper injection into classes using them (or maybe inject some sort of `CalculatedFieldsFactory` that would serve the needed instance for a given model).

However,

- If you think this would take longer than a couple days to implement, I'd prefer it be done in a separate pull and we get this merged into master and then upstream.  Otherwise the work that is _needed_ from this branch will get held back by the refactor.
- @tn3rb did some work in Gutenberg/Master I think (against my cautions) in the calculated fields classes.  So whatever we do in a refactor, we'll have to note that after the merge upstream to `Gutenberg/Master` we'll need to refactor the work he did there as well.
- I'm wary of breaking potential back compat for third parties that might have implemented a custom Calculated Fields class (although admittedly its unlikely).  I don't think this means we don't do the refactor, but I wonder if it does mean we should make sure we document the change in developers.eventespresso.com?

mnelson4:
KK I created a separate pull request for that: https://github.com/eventespresso/event-espresso-core/pull/681.
If it flies in the next day or two, we can go with it. If not, we can fall back to this

nerrad:
Ya I'd like to see us try and get #681 merged, it's the better way to go imo.

mnelson4:
the alternative implementation of https://github.com/eventespresso/event-espresso-core/pull/681 has been accepted, so I think we should close this

PR #681, opened 2018-09-04
Improve REST Schema to include missing fields and call field calcualtor classes non-statically

mnelson4:
This is an alternative implementation of https://github.com/eventespresso/event-espresso-core/pull/680.

It achieves the same thing, plus calls the children of `EventEspresso\core\libraries\rest_api\calculations\Base` non-statically, meaning its methods don't need to be static. This helps avoid a bit of duplicated code, and makes inheritance much more straightforward.

I did not, however, change any existing static methods to be non-static, fearing that might break backward compatibility, in case 3rd party code is directly calling those classes... which I think is highly unlikely. So @nerrad if you think it's acceptable to change those methods (eg `EventEspresso\core\libraries\rest_api\calculations\Attendee::userAvatar()`) to be non-static, I am too.

Even if we do that, I don't think 3rd party code that wrote their own calculator classes will be broken by these changes (ie, if someone else made a class to calculate fields, and it had static methods, they can continue to be static, like they are currently on this branch.)

The REST API JSON schema for resource endpoints is missing `link`, `_links`, and `_calculated_fields` from the response.  The work in this pull adds them.

## How has this been tested

* [x] Add unit tests and verified they passed.
* [x] Checked the schema endpoint for various resources (manual requests) and verified the response is as expected (ie. `OPTION http://ee.test/wp-json/ee/v4.8.36/events`)
* [x] Double check the mobile apps continue to work as expected

## Checklist

* [ ] My code is tested.


core/libraries/rest_api/CalculatedModelFields.php
```diff
@@ -67,7 +70,8 @@ protected function generateNewMapping()
         );
         foreach ($models_with_calculated_fields as $model_name) {
             $calculated_fields_classname = $namespace . $model_name;
-            foreach (array_keys(call_user_func(array($calculated_fields_classname, 'schemaForCalculations'))) as $field_name) {
+            $calculator = LoaderFactory::getLoader()->getShared($calculated_fields_classname);
```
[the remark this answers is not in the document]
mnelson4: > I realize this is additional work but my understanding is whenever we refactor things we're supposed to bring it up to date with current loading best-practices we're using. I know its an even further break of back compat, but it seems to me we've established its okay?

I don't think it's all-or-nothing, because at some point it becomes unreasonable to keep refactoring. Having said that, I think your suggestions are good.
So far we haven't actually broken backward compatibility in this branch. However, Brent has previously ok'd changing constructor arguments that can be managed by the loader, seeing how, if you're using the loader, changing those doesn't break anything.

So I'll take a stab at that.

But what do you think of changing the static methods on the calculators to being non-static? That will break backward compatibility, but pretty up the code some. It's not essential so I didn't do it yet.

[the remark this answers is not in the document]
nerrad: > I don't think it's all-or-nothing, because at some point it becomes unreasonable to keep refactoring.

My comments were based on my observations of our lead developer's direction in various reviews he's done of my code.

> But what do you think of changing the static methods on the calculators to being non-static? That will break backward compatibility, but pretty up the code some. It's not essential so I didn't do it yet.

Ya breaks back compat but I think its highly unlikely anyone is using those directly.  Let's change the static methods on the calculators to be non-static.
IF we wanted to preserve back compat, we could simply have the static methods left in place temporarily and wrapping constructing the calculator using `LoaderFactory` with a doing_it_wrong notice (and calling the new replacement methods).  But I'm ambivalent on whether we actually do that or not.

mnelson4:
ok I think the above work implements your suggestions.
A couple notes: 
* I also added `CalculatedModelFieldsFactory::createFromClassname()` because `CalculatedModelFields::generateNewMapping()` creates a "mapping" array containing classnames, and changing that to instead to use model names is a backward incompatible change that I think is much more likely to be a problem. (Because 3rd party add-ons might be adding entries to that array, and those entries are currently classnames).
* I didn't see much point in `CalculatedModelFieldsFactory` implementing `FactoryInterface`, especially when `FactoryInterface` demands a static method and this PR is all about removing static methods
* I inject `Loader` into `CalculatedModelFieldsFactory` instead of `LoaderInterface` because that's what I needed to do to get it to work, and what I saw `EventEspresso\core\CPTs\CptQueryModifier` did. But, in principle, I'd prefer to typehint only for the interface than the concrete class

core/libraries/rest_api/calculations/CalculatedModelFieldsFactory.php
```diff
@@ -0,0 +1,64 @@
...
+ */
+class CalculatedModelFieldsFactory
+{
+    private $loader;
+
+    /**
+     * CalculatedModelFieldsFactory constructor.
+     * @param Loader $loader
+     */
+    public function __construct(Loader $loader)
```
nerrad: This should still typehint for `LoaderInterface`.  `EE_Dependency_Map` can still register the actual loader for injection but you don't want to typehint the concrete here.

core/libraries/rest_api/controllers/model/Read.php
```diff
@@ -43,25 +48,29 @@ class Read extends Base
...
+     * @param string $model_name
+     * @return WP_REST_Response|WP_Error
+     * @throws InvalidArgumentException
+     * @throws InvalidDataTypeException
+     * @throws InvalidInterfaceException
      */
     public static function handleRequestGetAll(WP_REST_Request $request, $version, $model_name)
     {
-        $controller = new Read();
+        $controller = LoaderFactory::getLoader()->getShared('EventEspresso\core\libraries\rest_api\controllers\model\Read');
```
nerrad: This changes things in that everytime `handleRequestGetAll` is called it is now returning a shared instance of `Read`.  Is that intentional?  Are there any cached things on this class that could cause problems in a long running php worker environment that might use the api internally (such as on ES with gearman)?

**Note: this comment applies everywhere similar behaviour has changed in your work**

mnelson4: Yeah it was intentional, but probably an error. There's some debug and response header info that's put onto the read controller, so we should probably be creating a new one instead. So thanks for checking

core/libraries/rest_api/calculations/CalculatedModelFieldsFactory.php
mnelson4: ahhh I see

mnelson4:
kk back at you @nerrad, I applied your suggestions

mnelson4:
oh oups, thanks for noticing, and thinking of running those big rest tests. They took a while but worked fine.

If this all looks good then, next I'd like to pass off for testing (I think support should double-check the mobile apps are unaffected)

PR #683, opened 2018-09-07
Avoid fatal error when invalid ID is in ticket selector shortcode fixes #655

### LINKED ISSUE #655: ESPRESSO_TICKET_SELECTOR throws an exception for all users if the event ID doesn't exist.
**Describe the bug**
If you add `[ESPRESSO_TICKET_SELECTOR event_id=999999]` to a page and view said page both logged in and out of the site.

Both outputs show the exception with various details, if you have WP_Debug enabled you'll get a 'pretty' output but shouldn't this exception only be thrown to admins, or at least editors rather than all users?

**Expected behavior**
iirc the exception should not be displayed to all users, if the event doesn't exist it should return an empty string.

**Screenshots**
See here - http://take.ms/M95vZ
(left logged in, right logged out)

**Environment:**
 - Version of Event Espresso:  master
 - Version of WordPress: 4.9.8
 - Browser: any
 - PHP Version 7.2

**Additional context**
Reported here: https://eventespresso.com/topic/ee-will-break-pages-with-missing-ee-events/

joshfeck:
Fixes issue #655 Only throw exception if WP_DEBUG is true and current user has the ability to edit the page

Otherwise just print a message.

See issue #655 

## How has this been tested

put ticket selector shortcodes on posts & pages, some with valid IDs and some with invalid IDs. Viewed the page with WP_DEBUG set to true and set to false. Viewed the page while logged in and viewed the page while logged out. 
## Checklist

* [x] My code is tested.

### DIFF
--- a/core/domain/entities/shortcodes/EspressoTicketSelector.php
+++ b/core/domain/entities/shortcodes/EspressoTicketSelector.php
@@ -71,21 +71,31 @@ class EspressoTicketSelector extends EspressoShortcode
         $event_id = isset($event_id) ? $event_id : 0;
         $event = EE_Registry::instance()->load_model('Event')->get_one_by_ID($event_id);
         if (! $event instanceof EE_Event) {
-            new ExceptionStackTraceDisplay(
-                new InvalidArgumentException(
-                    sprintf(
-                        esc_html__(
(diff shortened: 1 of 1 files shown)


joshfeck:
This is ready for testing.

- [ ] Put ticket selector shortcodes on posts & pages, some with valid IDs and some with invalid IDs

- [ ] View the posts & pages with WP_DEBUG set to true and view with WP_DEBUG set to false

- [ ] View the pages & posts while logged in and view while logged out

The expected results from doing the above steps is you'll not get fatal errors and the displayed error messages will be appropriate to the intended audience (admin vs. non-admin)

PR #685, opened 2018-09-11
Improve registration CSV reporting redirections

mnelson4:

The URL to download a registration CSV used to contain a query parameter of `return_url`. This was usually fine, but on ES and MENW there was some code that was sometimes prematurely causing a redirect. (The registrations admin/batch system code passes around the URL we ultimately want to redirect to in this variable, we don't immediately redirect to it). So the variable was namespaced to `ee_return_url` to avoid conflicts with other plugins or whatever was causing it.
Also, the link had two consecutive ampersands (eg `&&`), like noticed on https://github.com/eventespresso/eventsmart.com-website/issues/177#issuecomment-420354384
Lastly, we weren't calling `die` after `wp_redirect` which meant more code could be executed (and that other code was, indeed, causing a redirect).

## How has this been tested

* [x] go to the registration list table. View the link of the "Filtered CSV REport". The should be no consecutive ampersands, and it should work as expected. Especially check that the filters are still being applied as before (eg, if you were previously filtering registrations by month, the report should only show registrations for that month)
* [x] verify the contact CSV report still works too
* [x] also verify the event check-in CSV report works fine

## Checklist

* [ ] My code is tested.


admin_pages/registrations/EE_Registrations_List_Table.class.php
```diff
@@ -143,19 +143,22 @@ protected function _set_properties()
...
+            array_flip(
+                array(
+                    'page',
+                    'action',
+                    'default_nonce',
+                )
+            )
+        );
+        if (!empty($filters)) {
+            $this->_bottom_buttons['report_filtered']['extra_request']['filers'] = $filters;
```
nerrad: the last array key is `filers` should it be `filters`?

mnelson4:
thanks good catch Darren.
I also added all the other work that appears to fix https://github.com/eventespresso/eventsmart.com-website/issues/177#issuecomment-420354384 onto here and renamed this ticket. (You may have input about the cause of that issue, which may affect whether these changes are needed or not)

modules/batch/EED_Batch.module.php
```diff
@@ -163,11 +165,28 @@ public function enqueue_scripts_styles_batch_file_create()
...
+     */
+    protected function getReturnUrl()
+    {
+        if(isset($_REQUEST['ee_return_url'])) {
+    return $_REQUEST['ee_return_url'];
+        } elseif(isset($_REQUEST['return_url'])) {
+            return $_REQUEST['return_url'];
+        }else{
+            return site_url();
+        }
```
nerrad: looks like there's some coding style and indentation issues here.

mnelson4:
We found another fix for https://github.com/eventespresso/eventsmart.com-website/issues/177#issuecomment-420354384 (this wouldn't have actually fixed it anyway).

So the commit https://github.com/eventespresso/event-espresso-core/pull/685/commits/63b4ce9154778b6f7bb0d5f1cff382c0e5874965 is an unnecessary refactor. It might have been a good idea but is quite a bit of work to change and test, so I'd rather not do that now, especially when it doesn't actually solve any known problem.

So I reverted that commit. So now this is a pretty straightforward PR: it just avoids an extra `&` in the report generation URL, so this won't require nearly as much testing either.

Pebblo:
Looks like the PR has been tested: https://github.com/eventespresso/event-espresso-core/pull/685#pullrequestreview-161577152

Can it be merged or does it need additional testing?

mnelson4:
huh sorry, somehow I overlooked this. Yeah this is totally ready to go.

PR #686, opened 2018-09-11
Add EE to block categories before tests run - wip

tn3rb:
does what the title says. Wasn't *really* needed since we aren't testing the full block, but wanted to add this in before I forgot about it since it's likely we will need this kind of logic for other blocks. I was thinking this could maybe be moved to another file that can then be loaded by test files that need it.

## How has this been tested
via unit tests

## Checklist

* [x] My code is tested.

### DIFF
--- a/assets/src/blocks/event-attendees/test/index.js
+++ b/assets/src/blocks/event-attendees/test/index.js
@@ -2,11 +2,18 @@
  * External dependencies
  */
 import { shallow } from 'enzyme';
+import { getCategories, setCategories } from '@wordpress/blocks';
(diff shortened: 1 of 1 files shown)


nerrad:
Just a note that I approved this because technically there isn't anything wrong with it... but I still don't think it should be merged.  So essentially I'm saying its up to you whether you merge it or not.

tn3rb:
ya maybe I'll try another test that renders the full block..will put on hold for now

tn3rb:
this isn't needed atm

PR #692, opened 2018-09-14
Fix the Dompdf integration so that it supports PHP version 7.2

knazart:
The Dompdf integration will invoke quite a few PHP7.1+ specific warnings like `A non-numeric value encountered...`, when run by php7.1+. It also uses a few deprecated methods.

Work in this PR is considered as a code refactoring for the Dompdf integration so that it's complaint with php version up to 7.2. In places where the change is required I mostly try to apply the same refactoring that was done in the newest version of Dompgf on that particular peace of code.

## How has this been tested
* [ ] While running your server on php7.2, try generating a PDF in any way possible. For example, try downloading a PDF of the order receipt after a successful event registration. There should be no PHP Warnings or Deprecated messages in the logs.

## Checklist

* [ ] My code is tested.

### DIFF
--- a/core/third_party_libs/dompdf/include/block_frame_reflower.cls.php
+++ b/core/third_party_libs/dompdf/include/block_frame_reflower.cls.php
@@ -22,13 +22,13 @@ class Block_Frame_Reflower extends Frame_Reflower {
    * @var Block_Frame_Decorator
    */
   protected $_frame;
-  
+
   function __construct(Block_Frame_Decorator $frame) { parent::__construct($frame); }
 
   /**
    *  Calculate the ideal used value for the width property as per:
    *  http://www.w3.org/TR/CSS21/visudet.html#Computing_widths_and_margins
-   *  
+   *
    *  @param float $width
    *  @return array
    */
@@ -46,7 +46,7 @@ class Block_Frame_Reflower extends Frame_Reflower {
 
     $left = $style->length_in_pt($style->left, $w);
     $right = $style->length_in_pt($style->right, $w);
-    
+
     // Handle 'auto' values
     $dims = array($style->border_left_width,
                   $style->border_right_width,
@@ -65,7 +65,7 @@ class Block_Frame_Reflower extends Frame_Reflower {
       $absolute = false;
     }
 
-    $sum = $style->length_in_pt($dims, $w);
+    $sum = (float)$style->length_in_pt($dims, $w);
 
     // Compare to the containing block
     $diff = $w - $sum;
@@ -105,7 +105,7 @@ class Block_Frame_Reflower extends Frame_Reflower {
           $width = $diff;
 
         } else if ( $left === "auto" ) {
-          
+
           if ( $lm === "auto" )
             $lm = 0;
           if ( $rm === "auto" )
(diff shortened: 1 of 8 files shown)


knazart:
Right now the work/updates here covers just some (most used in my opinion) of the code but certainly not all of it. Each file of this integration has to be revisioned/edited to cover all possible places where the code is "outdated".
@nerrad should I go though all the files ? I think I covered all of the deprecated stuff, so mostly just adding type casting would be required. I try copying the changes from the newest version of Dompdf converter where the update in the code is required.

knazart:
> Being this is a third party library, ideally it'd be great to just update the library

Well it seems that the minimum required php version for it is the same as for EE - 5.4.0 or higher. Although I didn't just update the library because we do have a few changes/commits done directly on this lib. Maybe we could re-apply our modifications onto the updated lib, especially if some of the fixes we did were already applied in the later versions of the lib, but it's likely that we are going to stumble upon new and deprecated methods which will require us to revision EE core code that uses that lib. How hard that revision would be depends on how formally vs granularly (method overrides, class extending/implementing) that lib is used. And then there's complete retesting and maybe fixing new bugs. So I'm trying to take the quicker route for the sake of ES.
Btw, the last time this library was updated was in 2014.

> I like that you mentioned you implemented changes you saw in their files... so that at least means they should be preserved somewhat.

Correct, that was the intent.

So should I proceed with updating all other files with at least the type casting and then apply those changes to ES, or try updating this library completely ?

nerrad:
If the minimum version required for dompdf is 5.4 then I would say we should try updating the library completely.  I agree there may still be some fixes needed but I'd rather start from a common base.  Do it in a separate branch from this in case something comes up on ES where it becomes more urgent to get something out.

@mnelson4 do you have any other thoughts to add here (I know you've worked with the dompdf library in the past as well).

mnelson4:
yeah I'd be inclined to update dompdf to the latest version rather than forking it, especially if EE already requires PHP 5.4 anyway. May as well keep it up-to-date (it's possible there could be security fixes in it too).
I was going to do that on https://events.codebasehq.com/projects/event-espresso/tickets/7678, but got started thinking about switching to a JS-based HTML-to-PDF converter, but that initiative got de-prioritized so that it's been effectively canceled.

mnelson4:
...although, if @knazart's modifications already resolved the PHP 7.2 issues, and they're pretty innocuous, I'm ok with just sticking with them. 
But if it starts to look like modifying our existing dompdf version to work with PHP 7.2 will take more work, I think we should at least try updating dompdf.

knazart:
> But if it starts to look like modifying our existing dompdf version to work with PHP 7.2 will take more work...

Yes, it most definitely will. Work here covers most of the stuff that was triggered on ES and then some other that I was able to trigger myself. There are twice as much more methods with those same issues there. That is why I didn't proceed and began thinking about updating the integration all together.

> Do it in a separate branch from this in case something comes up on ES where it becomes more urgent to get something out.

Totally agree with that one.

I'll start with updating dompdf and then, if needed, will try to reapply the fixed that we did from the history of earlier commits.

mnelson4:
:+1: although is @sethshoultes aware of this? Updating libraries like this can sometimes be time consuming if the changes require refactoring

knazart:
This is now mostly a rush because we just moved to php7.2 and all the errors started popping up on ES. But yeah, maybe I'll create a ticket for this to be triaged.

mnelson4:
well it doesn't need another ticket, I'm just thinking @sethshoultes should be aware when he's deciding what else to assign to you.

knazart:
Work in PR #699 that was merged to master also cowers the changes/updates done here, so this pull request can be discarded.
Closing.

Issue #694, opened 2018-09-18
Skeleton payment methods use wrong pue slug


mnelson4:
## Issue Overview

The skeleton payment method has `espresso_new_payment_method` for the PUE slug, but it should probably instead be `eea-new-payment-method`. We should also check the find-and-replace creates it properly. Eg if you replace `New_Payment_Method` with `Foo_Bar`, the slug will become `eea-foo-bar`. If that doesn't work, we may want to just leave this string blank

## Bug report or feature request?

* [x] Bug
* [ ] Feature
* [ ] Neither

mnelson4:
PR merged

PR #695, opened 2018-09-19
Fixed pue slug in the skeleton new payment method add-on

mnelson4:

https://github.com/eventespresso/event-espresso-core/issues/694
Instead of having the default pue slug be `espresso_new_payment_method`, its `eea-new-payment-method`. I just checked and the renamer does successfully keep those dashes, so if you named the plugin `Foo_Bar`, the pue slug would be `eea-foo-bar`.
Alternatively we thought we could leave this blank, but I personally prefer having this default to the standard in order to reduce boilerplate we need to write.

## How has this been tested

* [] I used the renamer (provided it with `?search=New_Payment_Method&replace=Foo_Bar`) and found the `pue_plugin_slug` was successfully renamed (to `eea-foo-bar`)

## Checklist

* [ ] My code is tested.


mnelson4:
skipping testing as this only affects the skeleton add-on, not released code

PR #696, opened 2018-09-19
Add filter hooks to allow customizing the display of sub line items o…

joshfeck:
This PR adds two new filter hooks to allow for customizing the output of the sub line items

### DIFF
--- a/core/libraries/line_item_display/EE_SPCO_Line_Item_Display_Strategy.strategy.php
+++ b/core/libraries/line_item_display/EE_SPCO_Line_Item_Display_Strategy.strategy.php
@@ -367,6 +367,10 @@ class EE_SPCO_Line_Item_Display_Strategy implements EEI_Line_Item_Display
         if ($parent_line_item instanceof  EE_Line_Item
             && $line_item->children() === array()
             && $line_item->name() === $parent_line_item->name()
+            && apply_filters(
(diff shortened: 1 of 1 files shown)


core/libraries/line_item_display/EE_SPCO_Line_Item_Display_Strategy.strategy.php
```diff
@@ -401,6 +405,13 @@ private function _sub_item_row(EE_Line_Item $line_item, $options = array(), EE_L
         $html .= EEH_HTML::td();
         // end of row
         $html .= EEH_HTML::trx();
+        $html = apply_filters(
+            'FHEE__EE_SPCO_Line_Item_Display_Strategy___sub_item_row__filter_output',
```
tn3rb: to more accurately follow our naming convention, this filter name should simply be:

```php
'FHEE__EE_SPCO_Line_Item_Display_Strategy___sub_item_row__html'
```

joshfeck:
@sethshoultes this is ready to be tested. 

Testing notes:

Create a ticket that has a price modifier, then register for the event. The ticket price & sub item display should look exactly the same as what's displayed with ee4 master. 

Then, to test one of the filters that was added, you can add the following to a functions plugin:

```add_filter(
    'FHEE__EE_SPCO_Line_Item_Display_Strategy___sub_item_row__hide_main_sub_line_item',
    '__return_false'
);```

Which should change the display so it also shows the ticket price before the price modifier was added.

PR #700, opened 2018-09-20
Cast translated strings to strings before using as array keys

mnelson4:

Compatibility with WP trac ticket: https://core.trac.wordpress.org/ticket/41305#comment:24

## How has this been tested

* [ ] Retest all the things

## Checklist

* [ ] My code is tested.

### DIFF
--- a/core/CPTs/EE_CPT_Strategy.core.php
+++ b/core/CPTs/EE_CPT_Strategy.core.php
@@ -126,7 +126,7 @@ class EE_CPT_Strategy extends EE_Base
         $_CPT_endpoints = array();
         if (is_array($this->_CPTs)) {
             foreach ($this->_CPTs as $CPT_type => $CPT) {
-                $_CPT_endpoints [ $CPT['plural_slug'] ] = $CPT_type;
+                $_CPT_endpoints [ (string)$CPT['plural_slug'] ] = $CPT_type;
             }
         }
         return $_CPT_endpoints;
--- a/core/EE_Config.core.php
+++ b/core/EE_Config.core.php
@@ -1267,7 +1267,7 @@ final class EE_Config implements ResettableInterface
             EE_Error::add_error($msg . '||' . $msg, __FILE__, __FUNCTION__, __LINE__);
             return false;
         }
-        EE_Config::$_module_route_map[ $key ][ $route ] = array('EED_' . $module, $method_name);
+        EE_Config::$_module_route_map[ (string)$key ][ (string)$route ] = array('EED_' . $module, $method_name);
         return true;
     }
 


mnelson4:
Passing to Seth for triage before getting others involved. But I think it's good for us to get this into EE core asap in case https://core.trac.wordpress.org/ticket/41305  gets released. 
After this, it will also be good to test our add-ons with this branch. I suspect add-ons that add other CPTs (eg people add-on) will also need to be updated

mnelson4:
Actually, I quickly checked people, calendar, and promotions and they've all been ok with this change

core/CPTs/EE_CPT_Strategy.core.php
```diff
@@ -126,7 +126,7 @@ private function _set_CPT_endpoints()
         $_CPT_endpoints = array();
         if (is_array($this->_CPTs)) {
             foreach ($this->_CPTs as $CPT_type => $CPT) {
-                $_CPT_endpoints [ $CPT['plural_slug'] ] = $CPT_type;
+                $_CPT_endpoints [ (string) $CPT['plural_slug'] ] = $CPT_type;
```
nerrad: should there be validation that there is a `'plural_slug'` index in the array as well?

mnelson4: sure we could add that

mnelson4:
passing to @sethshoultes for testing or assigning to someone else for testing

mnelson4:
came across a few more spots we need to cast translated strings to regular strings in order to work with 41305 without warnings, so this should have another review. Sorry!

core/data_migration_scripts/EE_DMS_Core_4_6_0.dms.php
```diff
@@ -543,18 +543,18 @@ public function add_default_admin_only_payments()
...
+                    (string) __("Invoice", 'event_espresso')     => __(
                         "Invoice received with monies included",
                         'event_espresso'
                     ),
-                    __("Money Order", 'event_espresso') => '',
-                    __("Paypal", 'event_espresso')      => __("Paypal eCheck, Invoice, etc", 'event_espresso'),
-                    __('Other', 'event_espresso')       => __('Other method of payment', 'event_espresso'),
+                    (string) __("Money Order", 'event_espresso') => '',
+                    (string) __("Paypal", 'event_espresso')      => __("Paypal eCheck, Invoice, etc", 'event_espresso'),
+                    (string) __('Other', 'event_espresso')       => __('Other method of payment', 'event_espresso'),
```
nerrad: yikes, using translations as array keys seems super fragile.  I'm also wondering how well this works across php versions.  I understand there's some limitations on what we can do here for back compat reasons...

core/helpers/EEH_MSG_Template.helper.php
```diff
@@ -298,7 +298,7 @@ public static function get_shortcodes(
                     $_field = $field;
                 }
                 if (isset($_field)) {
-                    $localized[ $_field ] = $shortcodes;
+                    $localized[ (string) $_field ] = $shortcodes;
```
nerrad: gosh we repeat the seemingly fragile pattern I noted above elsewhere too :)  I'm looking at this whole code block and shuddering lol.  Refactoring is outside the scope of this issue though so for now I guess it'll have to do.

core/libraries/batch/JobHandlers/RegistrationsReport.php
```diff
@@ -257,10 +270,11 @@ public function get_csv_data_for($event_id, $offset, $limit, $question_labels, $
                 $reg_csv_array = array();
                 if (! $event_id) {
                     // get the event's name and Id
-                    $reg_csv_array[ __('Event', 'event_espresso') ] = sprintf(
+                    $reg_csv_array[ (string) __('Event', 'event_espresso') ] = sprintf(
+                        /* translators: 1: event name, 2: event ID */
                         __('%1$s (%2$s)', 'event_espresso'),
```
nerrad: what's the point of even localizing this?  If all it contains is some brackets and dynamic values?

core/data_migration_scripts/EE_DMS_Core_4_6_0.dms.php
mnelson4: I suppose one problem with using translated strings as array keys is in case the translations are coincidentally the same. Yeah, I should have just used sub-arrays with 2 elements instead. But ya, I don't think we can change this without breaking backwards compatibility

core/helpers/EEH_MSG_Template.helper.php
mnelson4: 🙈

core/libraries/batch/JobHandlers/RegistrationsReport.php
mnelson4: I think translators  might want to reorder them, or use something besides brackets. (Eg RTL languages might want to put them in the other order?) I'm not positive on exact cases but the justifications for it are pretty weak. But I'm inclined to leave it, especially considering folks might be using that translation to filter the value too.

mnelson4:
thanks Darren, I think those are good points to bring up. But I don't think those should hold this up (thanks), I'm going to pass back for testing.

sethshoultes:
@Pebblo can you this one, please?

joshfeck:
fwiw trac ticket 41305 got punted to WP 5.1

Pebblo:
Mike helped me setup Laragon to test this and it seems to be working.

I run into an issue with the payment notification emails causing the page to lock up  but Mike doesn't run into the same problem. The email is sent and the payment is applied from the admin, it just doesn't remove the 'spinner', no errors etc.

Registration emails work, if I disable notifications the update works so it seems specific to payment related emails.

@mnelson do you think this is likely just an issue with laragon or do I need to dig deeper on this?

mnelson4:
> @mnelson do you think this is likely just an issue with laragon or do I need to dig deeper on this?

I just looked into it again, and actually I noticed I get a similar issue with Laragon's payment emails on EE master branch. For some reason, generating payment emails is deathly slow (like it takes several minutes). One clue is that it's not necessarily sending them that is slow, because once they're generated, sending them again is fast. That's a bit of a mystery I'm investigating.

Either way, I'm confident it's a Laragon/config issue that's causing those emails to be generated so slowly, so the string casting changes on this branch.

Pebblo:
So, I've tested this a good few times now with messing around with Laragon and with the wp 41305 patch applied, this works great. (The issue with the emails is Laragon and as Mike said, if I sit and wait they eventually send).

I then retested using WP release and this branch, all good on that.

@mnelson4 Apologies for the delay on this, I think this is ready to go!

mnelson4:
Yeah that's annoying how those messages are so slow. I haven't figured out why I was getting that (or why it's magically gone away for me).
If you want to debug it further, you could try setting up xdebug profiling, but that's a whole other story (I did that only to then have the problem magically go away).

PR #705, opened 2018-09-24
Fix Session Exception Triggered by Error Notices Fixes #688

### LINKED ISSUE #688: Fatal error: Uncaught TypeError: Argument 1 passed to EE_Session::__construct() must implement interface
**Summary**
This was seen in logentries for Event Smart:

`PHP Fatal error: Uncaught TypeError: Argument 1 passed to EE_Session::__construct() must implement interface EventEspresso\core\services\cache\CacheStorageInterface, null given, called in /bacon/wp-content/plugins/event-espresso-core/core/EE_Session.core.php on line 194 and defined in /bacon/wp-content/plugins/event-espresso-core/core/EE_Session.core.php:212`

Stack trace: 

```
…d at /bacon/wp-content/plugins/event-espresso-core/core/EE_Session.core.php (194)
…d at /bacon/wp-content/plugins/event-espresso-core/core/
EE_Session.core.php (270)
…led at /bacon/wp-content/plugins/event-espresso-core/core/
EE_Error.core.php (760)
…led at /bacon/wp-content/plugins/event-espresso-core/core/
EE_Error.core.php (925)
…led at /bacon/wp-content/plugins/event-espresso-core/core/
EE_Error.core.php (865)
…bacon/wp-content/plugins/event-espresso-core/core/
EE_Front_Controller.core.php (408)
…oller::display_errors called at /wordpress-4.9.8/wp-includes/
class-wp-hook.php (286)
…P_Hook::apply_filters called at /wordpress-4.9.8/wp-includes/
class-wp-hook.php (310)
in WP_Hook::do_action called at /wordpress-4.9.8/wp-includes/plugin.php (515)
…do_action_ref_array called at /wordpress-4.9.8/wp-includes/
class-wp-query.php (3097)
in WP_Query::the_post called at /wordpress-4.9.8/wp-includes/query.php (805)
in the_post called at /bacon/wp-content/themes/sparkling/archive.php (27)
in include called at /wordpress-4.9.8/wp-includes/template-loader.php (74)
in require_once called at /wordpress-4.9.8/wp-blog-header.php (19)
in require called at /wordpress-4.9.8/index.php (17)
```

I noticed that the URL where it was trigged was on an embedded event category page: https://cl.ly/9f1949f5e410

**Steps to reproduce**
I'm not sure how to reproduce this issue.

tn3rb:
Fixes #688 

So what's happening in the above issue is that errors are getting triggered in various locations in the code base and we now have some logic in EE_Error that attempts to save the current user's notices based on their WP user id if they are logged in, or to their active session if one exists. The check for an active session in `EE_Session::isLoadedAndActive()` calls `EE_Session::instance()` which attempts to instantiate the session if one does not exist. But in order to properly instantiate the session you need to provide all of the required dependencies which there is no need to supply when merely checking if the session is active. `EE_Session::instance()` had no safeguards in place to verify that those dependencies were provided before attempting to instantiate the session, which would result in the TypeError exception. The simple solution is to add some instance checks before attempting instantiation, so the session will not be created unless all of the required dependencies are provided.

## How has this been tested
Attempted to write a unit test, but working with sessions during tests is super painful and would have required all kinds of crazy hoop jumping. It's obvious that the exception can no longer occur via this particular code path.

## Checklist

* [ ] My code is tested.


lorenzocaum:
Hi Brent, I ran through a registration on this branch and then I added extra characters to the reg URL and went to that page. It showed that transaction information wasn't available.

Then I went to an older transaction on the testing site and used the front-end link to go to registration checkout. I added extra characters to the URL and then browsed to it.

Afterward, I checked for errors on the site error log and the debug.log and I didn't see anything.

Please merge to master.

PR #706, opened 2018-09-24
Remove note about filters, fixes #702

### LINKED ISSUE #702: Remove please note help text from Custom Display Order setting
**This is not specific to a recent version of Event Espresso**

**Summary**
A few years ago we introduced the custom display order setting to rearrange the event details. Prior to that being available, we used some filters to help with the change and this message was added:

image 2018-09-22 at 11 27 01 am

Can we have that removed since its been a few years?

**Steps to reproduce**
This can be seen in the Templates tab in Event Espresso.

joshfeck:
This PR includes a change to remove notes about using filters and how they can interfere with the custom display order settings.


Pebblo:
Reviewed and tested this, text is removed and looks fine.

Merging in.

PR #708, opened 2018-09-25
Fix memory timeout for sites with a brazillion tags

tn3rb:
Fixes https://github.com/eventespresso/eventsmart.com-website/issues/45

So previously, for tag archive requests, we would first query the db and assemble an array for EVERY SINGLE EXISTING TAG associated with EE_Events or EE_Venues via a single query. The site in question on EventSmart has nearly 700 tags and that query would timeout. Not exactly sure why (j/k). So the work in this branch, replaces that maga query with one that uses the tag from the request and only returns an EE_Term object if that tag is associated with any EE_Events or EE_Venues.

## How has this been tested
via browser only

## Checklist

* [x] My code is tested.


tn3rb:
### testing

 - create and/or edit an event and add a tag to it
 - visit the tag archive on the frontend for that tag ie: domain.com/tag/{your-tag}
 - [ ] confirm that tagged event appears in tag archive
 - now add same tag to other events
 - [ ] confirm that all tagged events appear in tag archive
 - now do the same for venues

lorenzocaum:
I tested this and it looks good. @sethshoultes could you test as well please?

PR #711, opened 2018-09-27
Fix Potential Recursion during Reg Status Updates

tn3rb:
Josh was dealing with a customer that had some data corrupted by a botched migration. Then whenever they would attempt to activate the Wait List add-on, their site would time out and issue a 504 error. Josh emulated their site setup locally and copied their db data so that he could recreate the error locally. He was eventually able to identify that the error only occurred for two events and when it did, the ticket and datetime sold values would sky rocket indicating that some recursion was happening somewhere.

I was then able to recreate the issue locally as well, simply by setting the reg limit and ticket quantities for an event with an active wait list to infinite. Updating the event triggered the recursion. Here's a breakdown of the actual process:

- any visit to an event page of any sort (archive, single, admin page, etc) triggers a sold out status check
- if the Wait List (WL) add-on is active and spaces are available, then the WL will pull all RWL (Wait List) status registrations from the db
- if auto-promote is on, then the WL will bump the first RWL reg to either RPP or RAP (depending on the event default or the ticket price, etc)
- in `EE_Registration::set_status()` we compare the current set reg status with the incoming new reg status and determine whether there was a status change at all, and if so, whether to reserve or release a registration space.
- for no change we do nothing, but for status changes...
 - if the new status is RAP, we reserve a space, which means incrementing sold values and decementing reserved values.
 - if releasing a registration space we simply decrease sold counts.
- changing sold counts on tickets and datetimes triggers *another* sold out status check
- this starts the process described above all over again... which *should* be fine, because we need to recalculate after every change to determine if there is still space available and whether to continue promoting RWL registrations

but here's where the real issue was...

- when the WL would pull all RWL registrations from db on the second iteration... it would see the *exact* same list of registrations as it did before... despite us having already bumped the first reg from RWL to RAP or RPP

**WHY?**

- because in `EE_Registration::set_status()` we were reserving or releasing registration spaces *BEFORE* we actually changed the internal value for the registration status. So the registration that was supposed to have been promoted from RWL to RPP or RAP had not actually had its status changed yet and appeared to still be on the Wait List. But hey... we've got spaces available so let's promote that registration!

Here's the existing code:

```php
// reg status change detected
// TO approved
if ($new_STS_ID === EEM_Registration::status_id_approved) {
    // reserve a space by incrementing ticket and datetime sold values
    $this->_reserve_registration_space();
    // OR FROM  approved
} elseif ($old_STS_ID === EEM_Registration::status_id_approved) {
    // release a space by decrementing ticket and datetime sold values
    $this->_release_registration_space();
}
// update status
parent::set('STS_ID', $new_STS_ID, $use_default);
```

the fix is as simple as moving the setting of the internal reg status to be ABOVE the point where we start modifying ticket and datetime sold counts, and ultimately trigger additional sold out status and WL checks:

```php
// reg status change detected
// update status
parent::set('STS_ID', $new_STS_ID, $use_default);
// NOW IT'S SAFE TO CHANGE OTHER DATA after reg status change
// TO approved
if ($new_STS_ID === EEM_Registration::status_id_approved) {
    // reserve a space by incrementing ticket and datetime sold values
    $this->_reserve_registration_space();
    // OR FROM  approved
} elseif ($old_STS_ID === EEM_Registration::status_id_approved) {
    // release a space by decrementing ticket and datetime sold values
    $this->_release_registration_space();
}
```

The work in this branch is basically the above fix.

## How has this been tested
browser testing only

## Checklist

* [x] My code is tested.


core/db_classes/EE_Registration.class.php
```diff
@@ -160,6 +160,9 @@ public function set_status($new_STS_ID = null, $use_default = false, ContextInte
             && ! empty($new_STS_ID) // as well as the new status
             && $this->ID() // ensure registration is in the db
         ) {
+            // update internal status first
+            parent::set('STS_ID', $new_STS_ID, $use_default);
```
nerrad: I agree, this looks like the way things _should_ have been originally.  However, this does break back compat for the existing actions in this method that pass along this instance because prior to this change, the actions would have been fired _before_ the internal status had changed on the instance but after this change the actions are fired _after_ the internal status change.  Granted, we'd expect that client code would be utilizing the status variables passed along to the action callback but it's still a potential gotcha that I thought worth mentioning.

Does it mean we don't do this fix?  No.  But I think it may be worth emphasizing in any changelog notes (maybe even a brief developers.eventespresso.com blog post).  As well, I'm assuming you've done a search through all our official code (core and add-ons) to see if anything hooking into these actions would be affected.

tn3rb: Both Michael and I searched our local codebases and could not find ANY uses of either of those two hooks so it's unlikely that others are using them if we are not doing so ourselves. That said, I can make add a brief post to http://developer.eventespresso.com to give a heads up. I don't think the changelog entry should stray too far from convention because we don't want to make a precedent out of writing overly verbose entries.

tn3rb:
## testing

- open the event editor for an event and examine its datetime and ticket sold quantities as well as the listed number of registrations
- if the numbers appear to be off in any way, then open another event editor until you find an event where everything looks correct
- if you can't find an event that has totally accurate datetime and ticket sold and reg values, then:
    - create a new event with one or more ticket and datetime options
    - add some registrations on the front end 
    - verify that the numbers are all correct
- keep the event editor open for that event
- in another tab, open the registrations admin list table for the same event that is open in the editor
- select one of more registrations and change their status
- refresh the event editor page and note any changes to datetime and ticket sold quantities 
- if changing registrations to "Approved" then sold counts should increase accordingly
- if changing "Approved" registrations to any other status, then sold counts should decrease accordingly
- continue experimenting with changing registrations statuses and monitor sold counts in the editor
- if you notice any discrepancies, then plz try to immediately document the previous steps you had taken, then try to recreate the issue
- once your brain is thoroughly numb from the above process, feel free to experiment with changing registration statuses in other ways
- [ ] confirm your brain is numb but changing reg statuses appears to work correctly

sethshoultes:
@joshfeck and @Pebblo can I get your help with testing?

Pebblo:
Using the testing notes above, I spent some time on **master** and I couldn't trigger any kind of weird updates to the counts.

I then tested this on **BUG/reg-status-change-recursion** and again couldn't trigger any issues.

I then started just messing around doing various updates to registrations in the admin and couldn't trigger any weird stuff at all.

----

Because I couldn't trigger the problem to then see if it was fixed I switched to using the Waitlist add-on  to do some testing as thats where Josh found the problem, with that add-on I could trigger the wrong value updates.

(Testing done on **BUG/reg-status-change-recursion**)

Setup an event with 10 waitlist registrations, set them to Auto promote and set the DRS to Approved.

I used 2 separate paid tickets and a single datetime on the event, added a few registrations to both tickets and paid for each using Stripe... all updated correctly. (I had about 20 altogether)

Change the datetime limit to match whatever sold value you have to sell out the event and update.

Confirm all of your sold values etc and add 5 registrations to the event, they will be waitlist registrations.

Now update the datetime limit to something higher than your sold limit to trigger the auto promotion, when you update the event you should see a wall of 'Auto promotion' notices'. - http://take.ms/IFzTI

In this branch the sold values remain correct, in master they end up being incorrect.

So it looks like this prevents the sold values from being updated, but something is still going nuts with the waitlist add-on?

Pebblo:
Hey @tn3rb does the above cover this or do we need more testing?

tn3rb:
@Pebblo 
that should be fine... the wall of notices is because there is a corresponding fix needed in the Wait List add-on to prevent extra notices from being generated. The registrations statuses get changed correctly and the sold counts are all correct, but it currently generates those extra notices. I'll push a fix for that soon. It sounds like this is good to go though, unless @joshfeck wants to take a swing at this too...

PR #714, opened 2018-09-28
Add offers.availability to event structured data 

joshfeck:
This PR adds the `offers.availability` property if tickets are available or sold out. The following are the two applicable values for Event Espresso tickets:

https://schema.org/InStock
http://schema.org/SoldOut

(If one day EE adds the ability to do a true "pre-order" then we can add the PreOrder value too)

## Checklist

* [x] My code is tested.

Testing notes:

Each ticket status should be tested:
On Sale, Archived, Sold Out, Pending, and Expired
The offers.availability property should be output for "On Sale" and "Sold Out" tickets only. 

This tool can be used to validate the structured data while testing:
https://search.google.com/structured-data/testing-tool/u/0/


joshfeck:
This needs some testing Seth. To test you'll need events with the following ticket statuses, and check the source of the event page to make sure the JSON+LD schema has the following:

On Sale (this branch adds `availability: https://schema.org/InStock`)
Archived (archived tickets shouldn't be shown in the JSON+LD schema)
Sold Out (this branch adds `availability: http://schema.org/SoldOut`)
Pending (there should be no change to the JSON+LD for tickets that are not on sale yet)
Expired (there should be no change to the JSON+LD for tickets that are no longer on sale)

Issue #721, opened 2018-10-09
Fix [GOOGLE_MAP_IMAGE] shortcode used in messages


joshfeck:
It seems the map image no longer renders when used in a message template.

joshfeck:
It appears that the API key isn't included in the URL when it's an eventsmart site.

sethshoultes:
The images seem to be working for me as well:
![your ticket for event b-2 2018-10-10 06-58-50](https://user-images.githubusercontent.com/235315/46737911-00f55080-cc5a-11e8-9134-1855d1115308.jpg)
![events by espresso wordpress 2018-10-10 06-58-09](https://user-images.githubusercontent.com/235315/46737913-00f55080-cc5a-11e8-9c7d-d7cb4e286325.jpg)

sethshoultes:
Do you want a ticket to look into this on Event Smart?

joshfeck:
I think we're going to need an issue on the Event Smart repo. It doesn't look like the API key is getting set if it's an html ticket. It might be good to check email messages there too.

joshfeck:
I looked into this a bit further and this does appear to be a bug in EE4 core, and the bug is exposed by Eventsmart.

What's happening is we use the `FHEE__EEH_Maps__espresso_google_maps_js__api_key` filter hook on Eventsmart to pass in a Google Maps API key. See https://github.com/eventespresso/eventsmart.com-website/blob/master/plugins/ee-saas-solution/EE_Saas_Restriction_Mods.class.php#L130

This works great on a front end page(like an /events/ or /venues/ page), but on a ticket page (where the URL has ?ee=ee-txn-tickets-url) the function hooked to FHEE__EEH_Maps__espresso_google_maps_js__api_key isn't called.

nerrad:
This ended up just requiring a fix in the EventSmart Repo (you can see https://github.com/eventespresso/eventsmart.com-website/pull/243 for details) so I'm closing out this issue.

PR #722, opened 2018-10-10
Change Google map image to map link in default email templates, resolves #383

### LINKED ISSUE #383: Remove [GOOGLE_MAP_IMAGE] from default message templates
I did a support token today and found that I could get messages to not fail by removing the [GOOGLE_MAP_IMAGE] shortcode from the Registration Approved template. 

It could be replaced with [GOOGLE_MAP_LINK] or removed entirely. I think it'd be best if the default templates avoided deliverability issues. Also, with the recent changes to Google Maps API (billing) it might be a better default to help reduce API usage for sites with larger volume of registrations.

joshfeck:
Include a link to a google map in place of an image of a google map in default email templates. This is a simple change, but we might want to change how the links are displayed (maybe use some link text and use the map URL shortcode instead)? 

Open to suggestions!

## Checklist

* [ ] My code is tested.

### DIFF
--- a/caffeinated/core/libraries/messages/defaults/default/email_registration_event_list_attendee.template.php
+++ b/caffeinated/core/libraries/messages/defaults/default/email_registration_event_list_attendee.template.php
@@ -47,7 +47,7 @@
                         <li>[VENUE_ADDRESS]</li>
                         <li>[VENUE_CITY]</li>
                         <li>[VENUE_STATE], [VENUE_ZIP]</li>
-                        <li>[GOOGLE_MAP_IMAGE]</li>
+                        <li>[GOOGLE_MAP_LINK]</li>
                     </ul>
                     <!-- social &#038; contact -->
                     <table class="social" width="100%" bgcolor="">
--- a/core/libraries/messages/defaults/default/email_registration_event_list.template.php
+++ b/core/libraries/messages/defaults/default/email_registration_event_list.template.php
@@ -48,7 +48,7 @@
                             <li>[VENUE_ADDRESS]</li>
                             <li>[VENUE_CITY]</li>
                             <li>[VENUE_STATE], [VENUE_ZIP]</li>
-                            <li>[GOOGLE_MAP_IMAGE]</li>
+                            <li>[GOOGLE_MAP_LINK]</li>
                         </ul>
                         <table class="social" width="100%" bgcolor="">
                             <tbody>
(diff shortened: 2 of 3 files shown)


garthkoyle:
One of the considerations we have to make @sethshoultes is that each api request we make to Google Maps can cost $. We're going to spend $100-$200/mo on Event Smart for all our api requests: 1) venue maps on the event detail page or venue detail page, 2) registration confirmation emails, 3) tickets. 

I do like the maps in those locations though because it gives some people detail they might have a hard time getting if they are offline. So, if #721 is not actually broken, then I could be convinced to leave the map for now.

sethshoultes:
Should we go back to making Google Maps a paid feature on Event Smart? Or maybe allow users to obtain and use their own API key on Event Smart? Or remove the map image on Event Smart and replace it with the link?

garthkoyle:
Event Smart is hard enough to set up already. I wouldn't go the route of having people provide their own API keys.

Venues were a paid feature. It's a nice feature to give people, but it's not free anymore. I think we should just eat the cost of the API keys for now and see if we can grow Event Smart subscriber base by offering venues for free rather than restricting its use. Just my opinion right now.

**That said, maybe we don't need to provide the map on the ticket template?**

joshfeck:
> That said, maybe we don't need to provide the map on the ticket template?

The map still could be available for the end user where they'd place a shortcode where they want the map, but maybe not already added by default.

joshfeck:
> What about adding both an image and link?

The intent behind this PR is to help **ensure email deliverability** for the default templates. I've done support tokens where the issue was the email did not send until I removed the GOOGLE_MAP_IMAGE shortcode. 

So in a nutshell, the goal here is to set the defaults up for allowing success right out of the box. The end user can add the GOOGLE_MAP_IMAGE to the template, but the defaults can be changed so they no longer include the image shortcode.

sethshoultes:
@joshfeck that makes sense. Does this change also affect the ticket templates?

joshfeck:
This change doesn't affect ticket templates. 

I'm working on some better link text and will push some commits soon.

joshfeck:
Added some link text and went with "Map and Directions" because:
1) There's already translations for "Map and Directions"
2) There's not only a map, but also directions there too

joshfeck:
Is this ready for merging then @sethshoultes ?

sethshoultes:
Yes, sir.

PR #727, opened 2018-10-12
Bug/improve tickets on sale

joshfeck:

The problem this pull request attempts to solve is currently the tickets_on_sale() method ignores any and all ticket types that aren't the first to go on sale or the last to go on sale. 

Real world example: An event has a ticket where its sale end date is passed, a ticket that's currently on sale, and a ticket that is yet to go on sale. The current `tickets_on_sale()` method returns `false` with this scenario. 

I reject the premise of your question. 
## How has this been tested

This was tested by setting up an event as described above. Then I added:
```
if( !$event->tickets_on_sale() ) {
    continue;
}
```
to the Events table template add-ons template to see if the event would display in the list. The event does not display in the list when switched to Master. 

## Checklist
* [x] Darren actually wrote this code, I just did the pull request
* [x] My code is tested.

### DIFF
--- a/core/db_classes/EE_Event.class.php
+++ b/core/db_classes/EE_Event.class.php
@@ -1161,6 +1161,23 @@ class EE_Event extends EE_CPT_Base implements EEI_Line_Item_Object, EEI_Admin_Li
 
 
     /**
+     * This returns the number of different ticket types currently on sale for this event.
+     *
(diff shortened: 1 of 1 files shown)


joshfeck:
Open to suggestions on that new method name (and the docblock too) because naming things.

PR #736, opened 2018-10-16
Fixed fatal error and currency override on Countries settings page

### LINKED ISSUE #734: Fatal on Countries settings page if no country selected
**Describe the bugs**
1. You will get a fatal error when you go to the General Settings >> Countries tab while there is no Country selected in the General Settings >> Your Organization.
2. Currency settings are reset when a new country is added to the Country Dropdown Select Lists.

**To Reproduce**
- Bug No. 1:
  - Go to the Event Espresso - General Settings >> Your Organization and select a blank option on the Country setting.
  - Now go to the Countries tab. You will get a Fatal error like:
    ```
    Fatal error: Uncaught Error: Call to a member function name() on null in .../wp-content/plugins/event-espresso-core/admin_pages/general_settings/General_Settings_Admin_Page.core.php:612
    ```
- Bug No. 2:
  - Go to the Event Espresso - General Settings >> Your Organization and select United States as the Country setting.
  - Then go to the Countries tab and select some other country, for example Malaysia. Note the Currency Code and Sign (should be MYR and RM).
  - Select Yes on the Country Appears in Dropdown Select Lists setting and Save Country Details.
  - Go back to Your Organization tab and select Malaysia as your country.
  - When you go to the Countries tab again Malaysia will be selected but the currency will be wrong - in this scenario it will be USD. And if you go to any event on the front end the prices will be in USD also, and not in MYR.

**Environment:**
 - Version of Event Espresso:  4.9.69.rc.015
 - Version of WordPress: 4.9.8
 - Browser Firefox
 - PHP Version 7.2

tn3rb:
Fixes #734 
The fatal error occurred when you would select nothing for a site country on the "Your Organization" tab in the General Settings, because we were only performing an `isset()` check for the `EE_Organization_Config->CNT_ISO` value on the "Countries" tab when deciding whether to use that value or the default. Since an empty string `""` is considered "set", we would then try to retrieve the country with an ISO value of "" and then access properties for the returned `EE_Country` object, which of course was actually null. Using an additional `empty()` check and then defaulting to `'US'` for the site country ISO when nothing is set prevents the error.

The second issue was that disabling the currency inputs removed them from the submitted form inputs and our logic for processing the forms would also perform an `isset()` check and then use default values if the form inputs were not found. Only setting and saving the currency settings when the inputs are detected solved this issue. Not sure how this one got past testing before because I know I had saved the inputs during my testing, so I must have just missed that the currency values were changing. 

OOOOOooo... I just realized why I never noticed this issue... I had selected "United States Minor Outlying Islands" as an alternate country to test that the currency settings were disabled, but that "country" setting obviously also uses "USD" for currency, so overriding "USD" with "USD" would be hard to notice :facepalm:

## How has this been tested
browser based monkey testing only

## Checklist

* [x] My code is tested.

### DIFF
--- a/admin_pages/general_settings/General_Settings_Admin_Page.core.php
+++ b/admin_pages/general_settings/General_Settings_Admin_Page.core.php
@@ -1055,30 +1055,45 @@ class General_Settings_Admin_Page extends EE_Admin_Page
         $cols_n_values['CNT_name'] = isset($this->_req_data['cntry'][ $CNT_ISO ]['CNT_name'])
             ? sanitize_text_field($this->_req_data['cntry'][ $CNT_ISO ]['CNT_name'])
             : null;
-        $cols_n_values['CNT_cur_code'] = isset($this->_req_data['cntry'][ $CNT_ISO ]['CNT_cur_code'])
-            ? strtoupper(sanitize_text_field($this->_req_data['cntry'][ $CNT_ISO ]['CNT_cur_code']))
-            : 'USD';
-        $cols_n_values['CNT_cur_single'] = isset($this->_req_data['cntry'][ $CNT_ISO ]['CNT_cur_single'])
-            ? sanitize_text_field($this->_req_data['cntry'][ $CNT_ISO ]['CNT_cur_single'])
-            : 'dollar';
-        $cols_n_values['CNT_cur_plural'] = isset($this->_req_data['cntry'][ $CNT_ISO ]['CNT_cur_plural'])
-            ? sanitize_text_field($this->_req_data['cntry'][ $CNT_ISO ]['CNT_cur_plural'])
-            : 'dollars';
-        $cols_n_values['CNT_cur_sign'] = isset($this->_req_data['cntry'][ $CNT_ISO ]['CNT_cur_sign'])
-            ? sanitize_text_field($this->_req_data['cntry'][ $CNT_ISO ]['CNT_cur_sign'])
-            : '$';
-        $cols_n_values['CNT_cur_sign_b4'] = isset($this->_req_data['cntry'][ $CNT_ISO ]['CNT_cur_sign_b4'])
-            ? absint($this->_req_data['cntry'][ $CNT_ISO ]['CNT_cur_sign_b4'])
-            : true;
-        $cols_n_values['CNT_cur_dec_plc'] = isset($this->_req_data['cntry'][ $CNT_ISO ]['CNT_cur_dec_plc'])
-            ? absint($this->_req_data['cntry'][ $CNT_ISO ]['CNT_cur_dec_plc'])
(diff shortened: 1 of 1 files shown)


admin_pages/general_settings/General_Settings_Admin_Page.core.php
```diff
@@ -563,6 +563,19 @@ protected function _update_admin_option_settings()
     /*************        Countries        *************/
 
 
+    /**
+     * @return string
+     */
+    protected function getCountryIsoForSite()
+    {
+        return isset(EE_Registry::instance()->CFG->organization->CNT_ISO)
```
nerrad: No need to use `isset` because `empty` check includes that check.
```suggestion
        return ! empty(EE_Registry::instance()->CFG->organization->CNT_ISO)
```

admin_pages/general_settings/General_Settings_Admin_Page.core.php
```diff
@@ -563,6 +563,19 @@ protected function _update_admin_option_settings()
...
+               && ! empty(EE_Registry::instance()->CFG->organization->CNT_ISO)
```
nerrad:
```suggestion

```

admin_pages/general_settings/General_Settings_Admin_Page.core.php
nerrad: no idea where those `span class="pl-s1"` tags came from :/

admin_pages/general_settings/General_Settings_Admin_Page.core.php
```diff
@@ -563,6 +563,18 @@ protected function _update_admin_option_settings()
     /*************        Countries        *************/
 
 
+    /**
+     * @return string
+     */
+    protected function getCountryIsoForSite()
+    {
+span class="pl-s1">         return ! empty(EE_Registry::instance()->CFG->organization->CNT_ISO)
```
tn3rb: hmmm..still adding `span class="pl-s1">`

```php
         return ! empty(EE_Registry::instance()->CFG->organization->CNT_ISO)
```

tn3rb:
## testing

- follow the bug reproduction steps outlined by Nazar in #734
- [ ] confirm that no errors occur and data saves as expected
- feel free to experiment with country and currency settings in this branch and attempt to find other issues

sethshoultes:
@lorenzocaum can you take this for a spin, please?

knazart:
Yes, this seems to work as expected now.

Just have one question: Is it expected for the "Select Country" label for the countries drop-down to disappear when you "Save Country Details" ? Not trying to be picky, just didn't notice this before.

sethshoultes:
We should probably get this merged.

PR #738, opened 2018-10-17
Add detection for WP JSON API requests

tn3rb:
While feverishly writing code as part of his attempt to take over the world, @nerrad discovered that `Request::isFrontend()` returns `true` for WP REST API endpoint requests. Turns out there was no detection in place within `EventEspresso\core\domain\services\contexts\RequestTypeContextDetector::detectRequestTypeContext()` for regular WP API requests! This branch adds such detection.

To foil @nerrad's attempt to take over the world

## How has this been tested
added unit testcase for RequestTypeContextDetector

## Checklist

* [x] My code is tested.


tests/mocks/core/services/request/RequestMock.php
```diff
@@ -0,0 +1,211 @@
...
+    public $is_bot;
+
+
+    public function __construct(array $get, array $post, array $cookie, array $server, $ip_address = '0.0.0.0')
+    {
+        $this->get = $get;
+        $this->post = $post;
+        $this->cookie = $cookie;
+        $this->server = $server;
+        $this->request = array_merge($this->get, $this->post);
```
nerrad: Is it necessary to set _all_ the above properties (other than ip_address) here?  Won't that be taken care of by `parent::__construct`?

tests/mocks/core/services/request/RequestMock.php
```diff
@@ -0,0 +1,211 @@
...
+        $this->ip_address = $this->visitorIp($ip_address);
```
nerrad: This will immediately get overridden in the call to parent::__construct won't it?  Should it be moved below that call?

tests/mocks/core/services/request/RequestMock.php
```diff
@@ -0,0 +1,211 @@
...
+    /**
+     * @param string $key
+     * @param string $value
+     * @param bool   $override_ee
+     * @return    void
+     */
+    public function setRequestParam($key, $value, $override_ee = false)
+    {
+        $this->request[ $key ] = $value;
+    }
```
nerrad: This basically means any calls to this method in tests using the mock are completely ignoring the conditions present in the actual code.  IMO it makes the test fragile.  I think it'd better to not override this method.

tests/mocks/core/services/request/RequestMock.php
```diff
@@ -0,0 +1,211 @@
...
+    /**
+     * returns the value for a request param if the given key exists
+     *
+     * @param string $key
+     * @param null   $default
+     * @return mixed
+     */
+    public function getRequestParam($key, $default = null)
+    {
+        return isset($this->request[ $key ]) ? $this->request[ $key ] : $default;
```
nerrad: Does overriding the original method mean that the private method `RequestParameterDrillDown` is not covered by tests?

- This applies also to `requestParamIsSet`.
- This applies also to `getMatch`.

tests/mocks/core/services/request/RequestMock.php
```diff
@@ -0,0 +1,211 @@
...
+     * and return the value for the first match found
+     * wildcards can be either of the following:
+     *      ? to represent a single character of any type
+     *      * to represent one or more characters of any type
+     *
+     * @param string     $pattern
+     * @param null|mixed $default
+     * @return false|int
+     */
+    public function getMatch($pattern, $default = null)
```
nerrad: Any tests against this method aren't really testing actual production code.  It's just testing this mock method.

tests/mocks/core/services/request/RequestMock.php
```diff
@@ -0,0 +1,211 @@
...
+     * check if a request parameter exists whose key matches the supplied wildcard pattern
+     * wildcards can be either of the following:
+     *      ? to represent a single character of any type
+     *      * to represent one or more characters of any type
+     * returns true if a match is found or false if not
+     *
+     * @param string $pattern
+     * @return false|int
+     */
+    public function matches($pattern)
```
nerrad: Interestingly enough, in looking at this code it seems to suggest production code could actually be modified to call `getMatch` instead of `requestParameterDrillDown`.

tests/mocks/core/services/request/RequestMock.php
```diff
@@ -0,0 +1,211 @@
...
+    /**
+     * remove param
+     *
+     * @param string $key
+     * @param bool   $unset_from_global_too
+     */
+    public function unSetRequestParam($key, $unset_from_global_too = false)
+    {
+        unset($this->request[ $key ]);
+    }
```
nerrad: So are tests covering the `$unset_from_global` parameter?

tests/mocks/core/services/request/RequestMock.php
```diff
@@ -0,0 +1,211 @@
...
+
+
+    /**
+     * @return string
+     */
+    public function visitorIp($visitor_ip = '0.0.0.0')
+    {
+        return $visitor_ip;
+    }
```
nerrad: are tests covering detection of ip used in production code (via the `$_SERVER` key detection in the parent class)?

tests/mocks/core/services/request/RequestMock.php
```diff
@@ -0,0 +1,211 @@
...
+
+
+    /**
+     * @return string
+     */
+    public function requestUri()
+    {
+        return isset($this->server['REQUEST_URI']) ? $this->server['REQUEST_URI'] : '';
```
nerrad: Is this method in production code covered in tests anywhere?  The logic here is significantly different than that in `core/services/request/Request`

tests/mocks/core/services/request/RequestMock.php
```diff
@@ -0,0 +1,211 @@
...
+    }
+
+
+    /**
+     * @param string $user_agent
+     */
+    public function setUserAgent($user_agent = '')
+    {
+        $this->user_agent = $user_agent;
```
nerrad: Is there a test that covers the conditionals found in the parent `setUserAgent` method?

tests/mocks/core/services/request/RequestMock.php
tn3rb: no because I had to override the actual properties in this class because the others are private and we don't want to change those because we don't want ppl making child classes and messing with the Request's internals.
I wouldn't do stuff like this for a regular class, but for a mock I don't really care as long as it accomplishes its goal

tests/mocks/core/services/request/RequestMock.php
tn3rb: The entire point of the mock is to remove as much of the mocked class from tests as possible, that's why this is overriding the complex logic that is in the real `Request` class with something more straightforward. It does mean that you have to be aware of what you are doing and ensuring that you are **only** testing your targeted code as opposed to performing integration tests which is what a lot of our "unit" tests have unfortunately been.

You may have noticed that the `RequestTypeContextDetectorTest` test case in this branch does not extend `EE_UnitTestCase` nor `WP_UnitTestCase`, but extends `PHPUnit_Framework_TestCase` directly because it is intended to be a pure unit test and not an integration test.

Bottom line is that the intent here is solely to unit test the `RequestTypeContextDetector` class and not the `Request` class. 
IF your intent is to perform integration tests, then you should not use mocks of any classes involved.

tests/mocks/core/services/request/RequestMock.php
tn3rb: The purpose of the `RequestMock` class is NOT for testing the `Request` class. Its purpose is so that you can remove the effects of the `Request` class from other tests. In this case we are removing the effects of the `Request` class from the tests in the `RequestTypeContextDetectorTest` test case.

The private `RequestParameterDrillDown` method is (indirectly) covered in the tests in `EventEspresso\tests\testcases\core\request_stack\RequestTest`, specifically by the following tests:

- `testGetWithDrillDown()`
- `testIsSetWithDrillDown()`
- `testMatchesWithDrillDown()`
- `testGetMatchWithDrillDown()`

tests/mocks/core/services/request/RequestMock.php
tn3rb: You wouldn't use the `RequestMock` class to test this method. The point of simplifying this method is to remove as much complexity and/or side effects and/or influence of the mocked class from your tests on logic that would normally utilize the mocked class.

Consider the following:

If we wanted to be absolutely sure that our tested logic was behaving as expected, then it would make most sent to be absolutely sure of what was being returned from any external functions our tested logic was using. This would be the ideal way to go about this:

```php
    // in RequestMock

    /**
     * @var mixed $get_match_response
     */
    private $get_match_response;
    
    /**
     * @param mixed $get_match_response
     */
    public function setGetMatchResponse($get_match_response)
    {
        $this->get_match_response = $get_match_response;
    }

    /**
     * @param string     $pattern
     * @param null|mixed $default
     * @return mixed
     */
    public function getMatch($pattern, $default = null)
    {
        return $this->get_match_response;
    }
```

Now we have absolute total control over what our tested class receives when it calls `RequestMock::getMatch()`.

Want to see what happens when our tested class receives `null` after calling `RequestMock::getMatch()` ?

```php
// in actual unit test method
$request_mock->setGetMatchResponse(null);
$result = $tested_class->someMethodThatUsesGetMatch();
// now assert that result is as expected
```

I didn't go that far because this requires more work and I felt that the current tests were sufficient, but if I wanted to be totally methodical/surgical/precise then something like the above where you specifically control the responses of your mocked methods is far superior.

tests/mocks/core/services/request/RequestMock.php
tn3rb: well you're not wrong...

but look at the actual code involved:

```php
    public function getMatch($pattern, $default = null)
    {
        return $this->requestParameterDrillDown($pattern, $default, 'match');
    }

    public function matches($pattern)
    {
        return $this->requestParameterDrillDown($pattern, null, 'match') !== null;
    }
```

if you changed `matches()` to call `getMatch()` instead, you would **still** be calling `requestParameterDrillDown()`, but just indirectly... 
so what's the point? You gain nothing from doing that.

tests/mocks/core/services/request/RequestMock.php
tn3rb: Again, the `RequestMock` class is NOT for testing the `Request` class. 
Testing for the `Request::visitorIp()` method can be found in `EventEspresso\tests\testcases\core\request_stack\RequestTest::testIpAddress()`.
You'll notice in that test, I capture and reset the value of the `$_SERVER` global before and after running the test, much like I explained in one of my comments above.

tests/mocks/core/services/request/RequestMock.php
tn3rb: Again, the `RequestMock` class is NOT for testing the `Request` class.
It appears there is no test for `requestUri()` in `EventEspresso\tests\testcases\core\request_stack\RequestTest` but that has no bearing on what is in the `RequestMock` class.

tests/mocks/core/services/request/RequestMock.php
tn3rb: It appears there is no test for `setUserAgent()` in `EventEspresso\tests\testcases\core\request_stack\RequestTest` but that has no bearing on what is in the `RequestMock` class.

tests/mocks/core/services/request/RequestMock.php
tn3rb: Originally I had no constructor at all for the `RequestMock` class, but the constructor in `Request` failed to set the overridden properties on  `RequestMock` ie; `$this->post = $post;` in `Request::__construct()` set the value for `Request->post` but NOT the value for `RequestMock->post`.

I was surprised by this as you may be now, but it was necessary to setup the `RequestMock` class the way it is to prevent it's properties from all being `null`.

nerrad:
I've confirmed that the new request detection method for WordPress API requests works as expected against the work I'm going in a GB related branch.  After talking in slack with @tn3rb, he's confident this is sufficient for merging this into master.

PR #743, opened 2018-10-18
introduce default date coercing

nerrad:

Creating a DateTime (or DbSafeDateTime) object with `0000-00-00 00:00:00` results in a internal timestamp of `-0001-11-30 00:00:00.000000`.  This results in potential downstream problems when we go to output this because even when you output back to mysql date format you’ll get `-0001-11-30 00:00:00` which is DIFFERENT than what was sent in.

To resolve this, if we coerce those date and time strings to `0000-01-01 00:00:00` it will result in an internal timestamp of `0000-01-01 00:00:00.000000` which is better because it represents a more legible value as well as is consistent with what the object was created with.

Downstream this is more friendly when dates are converted to ISO strings to pass along on REST api packages etc.

## How has this been tested

This pretty much impacts anywhere dates are displayed, so along with existing phpunit tests (which shouldn't break), testing should just involve interacting with any EE functionality that touches upon dates and make sure there is no funkiness.  Funkiness would be things like:

- errors displayed
- unexpected dates or date changes

Any issues with the changes I made should show up right away.

Our automated tests in travis should also reveal any issues across the php versions we support with this change.

Note: see also https://wiki.php.net/rfc/parameter-no-type-variance which allows us to override `DateTime::createFromFormat` with confidence with regards to cross version PHP compat.

## Checklist

* [x] My code is tested.

### DIFF
--- a/core/db_models/fields/EE_Datetime_Field.php
+++ b/core/db_models/fields/EE_Datetime_Field.php
@@ -573,17 +573,11 @@ class EE_Datetime_Field extends EE_Model_Field_Base
         if (empty($datetime_string)) {
             $DateTime = new DbSafeDateTime(\EE_Datetime_Field::now, $this->get_UTC_DateTimeZone());
         } else {
-            $DateTime = DateTime::createFromFormat(
+            $DateTime = DbSafeDateTime::createFromFormat(
                 EE_Datetime_Field::mysql_timestamp_format,
                 $datetime_string,
                 $this->get_UTC_DateTimeZone()
             );
-            if ($DateTime instanceof \DateTime) {
-                $DateTime = new DbSafeDateTime(
-                    $DateTime->format(\EE_Datetime_Field::mysql_timestamp_format),
-                    $this->get_UTC_DateTimeZone()
-                );
-            }
         }
 
         if (! $DateTime instanceof DbSafeDateTime) {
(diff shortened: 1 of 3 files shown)


nerrad:
@tn3rb this needs another review.  Turns out the tests caught a flaw in the original code... good we have these tests :)

After the fix, there was only one test needing modified and that's just because expectations changed.

nerrad:
looks like there's still failing tests in different PHP versions than what I have locally so I'll have to figure that out now. sigh.  Taking back from you Brent.

core/db_models/fields/EE_Datetime_Field.php
```diff
@@ -573,17 +573,11 @@ public function prepare_for_set_from_db($datetime_string)
...
                 EE_Datetime_Field::mysql_timestamp_format,
                 $datetime_string,
                 $this->get_UTC_DateTimeZone()
             );
-            if ($DateTime instanceof \DateTime) {
-                $DateTime = new DbSafeDateTime(
-                    $DateTime->format(\EE_Datetime_Field::mysql_timestamp_format),
-                    $this->get_UTC_DateTimeZone()
-                );
-            }
```
tn3rb: kinda silly we didn't think to do this earlier :\

nerrad: yeah but also understandable.

nerrad:
Thanks @garthkoyle that testing is sufficient!

PR #749, opened 2018-10-20
Populate States/Provinces for the "Your Organization" tab Country input Fixes #739

### LINKED ISSUE #739: Improvements to General Settings > Your Organization > Country field and General Settings > Countries tab
These items need to happen:

A) The General Settings > Your Organization page’s Country input needs to load in all countries (not just active ones)
This is an improvement over the current UI because currently the end user needs to:
1) Go to the Countries tab
2) Select a country
3) Change Country Appears in Dropdown Select Lists == Yes
4) Click Save
5) Go back to General Settings > Your Organization
6) then select the Country

We can eliminate steps 1-5 if all countries load in that input. 
Another reason for this improvement is because, at a quick glance, having only United States and Canada available there by default suggests that's all that EE4 supports. See also:
https://wordpress.org/support/topic/only-for-us-and-canada/

B) The General Settings > Your Organization page’s State input needs to load in States for the currently selected Country. 
This is an improvement over the current UI because currently, for example, if you have United States selected, you have to scroll through Canadian Provences and/or which ever other activated country states are activated on the site

C) The Country value that is saved in that Country input needs to automatically be activated (in other words, its Country Appears in Dropdown Select Lists gets sets to Yes). This was previously requested by Mike in https://events.codebasehq.com/projects/event-espresso/tickets/7410

D) Related to B above, get an XML file or similar ready to auto-populate states for the selected country. See also Mike's comment from this P2:
http://p2.eventespresso.com/2016/02/19/discussion-about-countries-and-states-provinces-pack-hey/#comment-15456

E) Place a metabox at the top of the General Settings > Countries page to include a deactivation form with checkbox list of all currently activated countries. This will allow for quickly deactivating any unneeded countries. Only activated countries would be included in the list, once a country is deactivated it will no longer be included in the form. For the sake of clarity, Country Appears in Dropdown Select Lists set to Yes means the country is activated.

tn3rb:
Fixes #739
There was no easy way to sideload States/Provinces for Countries. This is a little rough around the edges but is capable for downloading, parsing, and saving sub region data for all of the countries with data listed here: https://github.com/astockwell/countries-and-provinces-states-regions/tree/master/countries

## How has this been tested
BBM (browser based monkey) testing only 

## Checklist

* [x] as far as I know it works!!!

### DIFF
--- a/admin_pages/general_settings/OrganizationSettings.php
+++ b/admin_pages/general_settings/OrganizationSettings.php
@@ -28,6 +28,7 @@ use EventEspresso\core\exceptions\InvalidFormSubmissionException;
 use EventEspresso\core\exceptions\InvalidInterfaceException;
 use EventEspresso\core\libraries\form_sections\form_handlers\FormHandler;
 use EventEspresso\core\libraries\form_sections\strategies\filter\VsprintfFilter;
+use EventEspresso\core\services\loaders\LoaderFactory;
 use InvalidArgumentException;
 use LogicException;
 use ReflectionException;
@@ -494,82 +495,13 @@ class OrganizationSettings extends FormHandler
         if ($country instanceof EE_Country) {
             $country->set('CNT_active', 1);
             $country->save();
-            $this->getCountrySubRegions($country);
+            /** @var \EventEspresso\core\services\address\CountrySubRegionDao $countryDao */
+            $countryDao = LoaderFactory::getLoader()->getShared(
+                'EventEspresso\core\services\address\CountrySubRegionDao'
+            );
+            $countryDao->saveCountrySubRegions($country);
         }
         return true;
-    }
-
-
-    /**
-     * @param EE_Country $country
-     * @return string
-     * @throws EE_Error
-     * @throws InvalidArgumentException
-     * @throws InvalidDataTypeException
-     * @throws InvalidInterfaceException
-     * @throws ReflectionException
-     */
-    private function getCountrySubRegions(EE_Country $country)
-    {
-        $CNT_ISO = $country->ID();
-        $has_sub_regions = EEM_State::instance()->count(array(array('Country.CNT_ISO' => $CNT_ISO)));
-        if ($has_sub_regions) {
-            return;
-        }
-        $json = $this->getCountryJson(
-            'https://raw.githubusercontent.com/astockwell/countries-and-provinces-states-regions/master/countries.json'
-        );
-        if (is_array($json)) {
-            foreach ($json as $country) {
-                if ($country->code === $CNT_ISO) {
-                    if($country->filename !== null) {
-                        $sub_regions = $this->getCountryJson(
-                            'https://raw.githubusercontent.com/astockwell/countries-and-provinces-states-regions/master/countries/'
-                            . $country->filename . '.json'
-                        );
-                        if (is_array($sub_regions)) {
-                            foreach ($sub_regions as $sub_region) {
-                                $abbrev = str_replace(
-                                    $CNT_ISO . '-',
-                                    '',
-                                    sanitize_text_field($sub_region->code)
-                                );
-                                if(absint($abbrev) !== 0){
-                                    $abbrev = sanitize_text_field($sub_region->code);
-                                }
-                                EEM_State::instance()->insert(
-                                    array(
-                                        // STA_ID	CNT_ISO	STA_abbrev	STA_name	STA_active
-                                        'CNT_ISO'    => $CNT_ISO,
-                                        'STA_abbrev' => $abbrev,
-                                        'STA_name'   => sanitize_text_field($sub_region->name),
-                                        'STA_active' => 1,
-                                    )
-                                );
-                            }
-                        }
-                    }
-                }
-            }
-        }
-    }
-
-
-    /**
-     * @param string $url
-     * @return array
-     */
-    private function getCountryJson($url='')
-    {
-        if (empty($url)) {
-            return array();
-        }
-        $request = wp_safe_remote_get($url);
-        if (is_wp_error($request)) {
-            return array();
-        }
-        $body = wp_remote_retrieve_body($request);
-        return json_decode($body);
     }
 
 
--- a/admin_pages/general_settings/OrganizationSettings.php
+++ b/admin_pages/general_settings/OrganizationSettings.php
@@ -7,12 +7,12 @@ use EE_Admin_File_Uploader_Input;
 use EE_Admin_Two_Column_Layout;
 use EE_Checkbox_Multi_Input;
 use EE_Core_Config;
+use EE_Country;
 use EE_Country_Select_Input;
 use EE_Currency_Config;
 use EE_Error;
 use EE_Form_Section_HTML;
 use EE_Form_Section_Proper;
-use EE_License_Key_Display_Strategy;
 use EE_Network_Core_Config;
 use EE_Organization_Config;
 use EE_Registry;
@@ -20,6 +20,8 @@ use EE_State_Select_Input;
 use EE_Text_Input;
 use EEH_HTML;
 use EEH_Template;
+use EEM_Country;
+use EEM_State;
 use EventEspresso\core\domain\services\pue\Stats;
 use EventEspresso\core\exceptions\InvalidDataTypeException;
 use EventEspresso\core\exceptions\InvalidFormSubmissionException;
@@ -28,6 +30,7 @@ use EventEspresso\core\libraries\form_sections\form_handlers\FormHandler;
 use EventEspresso\core\libraries\form_sections\strategies\filter\VsprintfFilter;
 use InvalidArgumentException;
 use LogicException;
+use ReflectionException;
 
 /**
  * OrganizationSettings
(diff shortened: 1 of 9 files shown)


admin_pages/general_settings/OrganizationSettings.php
```diff
@@ -467,10 +489,90 @@ public function process($form_data = array())
...
+        }
+        $json = $this->getCountryJson(
+            'https://raw.githubusercontent.com/astockwell/countries-and-provinces-states-regions/master/countries.json'
+        );
+        if (is_array($json)) {
+            foreach ($json as $country) {
+                if ($country->code === $CNT_ISO) {
+                    if($country->filename !== null) {
+                        $sub_regions = $this->getCountryJson(
+                            'https://raw.githubusercontent.com/astockwell/countries-and-provinces-states-regions/master/countries/'
```
nerrad: Are we sure this is a good resource to use?  It doesn't appear [to be maintained](https://github.com/astockwell/countries-and-provinces-states-regions/issues/15) and the last time it was updated was 4 years ago...  Other than that, nifty.

nerrad:
I missed that travis reports linting issues that'll need fixed.

admin_pages/general_settings/OrganizationSettings.php
nerrad: An alternative, would be for us to fork this and maintain it ourselves.  With it being a critical source for data in EE it might be useful for us to do that?

tn3rb: ya I thought about that too... what would be cool about that is that we could also add to the data as well, and include other things like cities, etc

tn3rb: forked to: https://github.com/eventespresso/countries-and-provinces-states-regions

admin_pages/general_settings/OrganizationSettings.php
```diff
@@ -494,85 +495,16 @@ public function process($form_data = array())
         if ($country instanceof EE_Country) {
             $country->set('CNT_active', 1);
             $country->save();
-            $this->getCountrySubRegions($country);
+            /** @var \EventEspresso\core\services\address\CountrySubRegionDao $countryDao */
+            $countryDao = LoaderFactory::getLoader()->getShared(
+                'EventEspresso\core\services\address\CountrySubRegionDao'
+            );
+            $countryDao->saveCountrySubRegions($country);
```
nerrad: are we not injecting CountrySubRegionDao for back compat reasons (which makes sense if so)?  I wonder if we should still inject `CountrySubRegionDao = null`  and only load if its not present? Would that be a good transition preserving back-compat but keeping dependencies in order?

core/services/address/CountrySubRegionDao.php
```diff
@@ -0,0 +1,122 @@
...
+     * @throws EE_Error
+     * @throws InvalidArgumentException
+     * @throws InvalidDataTypeException
+     * @throws InvalidInterfaceException
+     * @throws ReflectionException
+     */
+    public function saveCountrySubRegions(EE_Country $country_object)
+    {
+        $CNT_ISO = $country_object->ID();
+        $has_sub_regions = EEM_State::instance()->count(array(array('Country.CNT_ISO' => $CNT_ISO)));
```
nerrad: should models be brought in as a dependency in the constructor rather than used directly?

core/services/address/CountrySubRegionDao.php
```diff
@@ -0,0 +1,122 @@
...
+        if (empty($url)) {
+            return array();
+        }
+        $request = wp_safe_remote_get($url);
+        if (is_wp_error($request)) {
+            return array();
+        }
+        $body = wp_remote_retrieve_body($request);
+        return json_decode($body);
+    }
```
nerrad: How often will this data change?  Since we've forked it, I wonder if we should have the json file include a version property and then the initial request can check for version and compare with the version from any previous fetch?  So when a fetch is performed we save the json file version somewhere locally.  That way in future iterations if we want to expose a ui/ux for event managers to be notified that country/region data is updated and provide the opportunity for them to update locally we can do that?

admin_pages/general_settings/OrganizationSettings.php
tn3rb: it was more a case of not loading that unless it is needed which would only occur if a country is selected that does not have any states/provinces in the db. So it really only needs to load once in a blue moon. In most cases this may literally only ever need to load once, and that's only if the customer's set country does not have any states. Loading it by default for every other request to the page is totally unnecessary.

core/services/address/CountrySubRegionDao.php
tn3rb: now THAT should absolutely be the case... ug can't believe I missed that.. thnx

core/services/address/CountrySubRegionDao.php
tn3rb: hmmm... ya... was trying to just get a solution out the door... but if we don't account for that, then there isn't really any way to update this without going in and nuking the db... sigh...

admin_pages/general_settings/OrganizationSettings.php
nerrad: hmm sure, but all it is doing is constructing it which isn't much different from what we do elsewhere.  I'm actually kinda surprised you didn't inject this.

core/services/address/CountrySubRegionDao.php
tn3rb: kk added versioning to the JSON data: https://github.com/eventespresso/countries-and-subregions/blob/master/countries.json

core/services/validators/JsonValidator.php
```diff
@@ -0,0 +1,60 @@
...
+                break;
+            case JSON_ERROR_SYNTAX:
+                $error = ': Syntax error, malformed JSON';
+                break;
+            case JSON_ERROR_UTF8:
+                $error = ': Malformed UTF-8 characters, possibly incorrectly encoded';
+                break;
+            default:
+                $error = ': Unknown error';
+                break;
```
nerrad: I realize you didn't include other error types here because they depend on PHP 5.5+ but this seems like something we'll forget about.  Should we add detection for other types when PHP version > 5.5?

tn3rb: oh good catch... I totally missed that there were new errors added in 7.0

nerrad: It looks like they were added in 5.5...

admin_pages/general_settings/OrganizationSettings.php
```diff
@@ -75,11 +81,13 @@ public function __construct(
         EE_Registry $registry,
         EE_Organization_Config $organization_config,
         EE_Core_Config $core_config,
-        EE_Network_Core_Config $network_core_config
+        EE_Network_Core_Config $network_core_config,
+        CountrySubRegionDao $countrySubRegionDao
```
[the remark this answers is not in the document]
tn3rb: ya the `OrganizationSettings` class is only loaded in one place (`General_Settings_Admin_Page`) and there is zero reason for it to ever be loaded anywhere else. And in that one place where it IS being loaded, we are now using the Loader, so its dependencies will automatically get injected. There just simply isn't any way for this to break backwards compatibility unless someone has done some serious honey badgering in core, in which case they are on their own.

core/services/validators/JsonValidator.php
tn3rb: oops  ya 5.5 and a couple more in 7.0

tn3rb:
ok FINALLY fixed all of the linting issues... sry for bouncing this back at ya so many times

tn3rb:
oops.. i was wrong and the `OrganizationSettings` class is actually loaded in TWO places in `General_Settings_Admin_Page` (once for displaying the form and the second when processing) can I get another quick review please?

tn3rb:
## testing

- go to the **General Settings** > **Your Organization** admin page tab
- select a country from the "Country" select input that has never been active on your site (like "Turkmenistan")
- click **Save**
- [ ] confirm that the Country you selected is the displayed option in the above input
- [ ] confirm that the "State/Province" select input is empty and that some new help text has appeared beneath it
- [ ] confirm that the new help text is free of typos and grammatical errors
- click on the "State/Province" select input, find the optgroup for the Country you selected, and choose an option (like "Aşgabat")
- click **Save**
- [ ] confirm that the State/Province you selected is the displayed option in the above input and that the help text has disappeared
- now click on the **Countries** tab in the navigation
- [ ] confirm that the Country you selected is the displayed option in the "Select Country" input and that its data is displayed in the form, including the States/Provinces details.
- [ ] confirm that the "Appears in Dropdown Select Lists" options are set to "Yes" for  Country you selected as well as for each State/Province
- feel free to experiment with the **Your Organization** "Country" and "State/Province" inputs in this branch and attempt to find any potential issues

joshfeck:
One improvement to this would be to remove the pre-selected "Arizona" value, which will allow for the help text to appear on new installs.

while we're at it, we should really remove the pre-set values outlined in this screenshot:

screen shot 2018-10-24 at 2 40 52 pm

sethshoultes:
I agree we should remove the pre-set values from these fields.

sethshoultes:
Got an error when trying to save just now:
`Notice:  Undefined property: stdClass::$filename in  /Users/sethshoultes/Documents/Websites/espresso-4-9.dev.cc/wp-content/plugins/event-espress-core-github/core/services/address/CountrySubRegionDao.php  on line 145`

tn3rb:
@sethshoultes && @joshfeck 
want to try this again guys?

sethshoultes:
@joshfeck any other issues with this? 

@tn3rb would it be possible to have some magic happen that loads the states without having to save first?

joshfeck:
This all works very well.

tn3rb:
> would it be possible to have some magic happen that loads the states without having to save first?

absolutely, but that is going to be more complex because we have to make that query via AJAX (or REST API) and then update the DOM via JavaScript.
I think it would be better to just get this out as is and then make that additional enhancement vai another pull request since it will be more involved.

I initially jumped into doing this because I realized that it could be done pretty quickly if we don't get fancy.
It ended up taking way longer than I expected as so I wouldn't want to drag it out any longer.
As is, this is a pretty decent improvement over the existing UI.

sethshoultes:
I agree. Let's get it merged.

joshfeck:
It's a huge improvement.

garthkoyle:
Let's get this merged if we can. We don't want this falling too behind and it's a big improvement for a lot of users.

PR #753, opened 2018-10-24
Filter txn list table query params and allow for dynamic views

Pebblo:
This branch filters the `$query_params` var used by the get_transaction method which pulls all of the transactions for the EE Transactions list table.

I added this to allow me to add an 'Incomplete' transactions view - http://take.ms/78cmZ

Whilst the list table views are already filtered, I couldn't find a way to add the count for my new custom view, so I've also updated the method used to add the counts so that it loops over all views set and adds the counts for each, meaning we can add and remove views without too much fuss.

Adding this filter to an add-on, for add-ons to use, makes the add-on unable to be an add-on :|

## How has this been tested
Use this branch and confirm that all of the counts and views are correct when you go to Event Espresso -> Transactions.

Add this snippet to your site: https://gist.github.com/Pebblo/f76f4b2ab766072de44bb8f333db146f

Refresh and check again, click on the new 'Incomplete Transactions' view and confirm all of your counts are still correct and also that the list now only shows incomplete transactions.

## Checklist

* [ ] My code is tested.

### DIFF
--- a/admin_pages/transactions/EE_Admin_Transactions_List_Table.class.php
+++ b/admin_pages/transactions/EE_Admin_Transactions_List_Table.class.php
@@ -148,8 +148,8 @@ class EE_Admin_Transactions_List_Table extends EE_Admin_List_Table
      */
     protected function _add_view_counts()
     {
-        foreach( $this->_views as $view) {
-            $this->_views[$view['slug']]['count'] = $this->_admin_page->get_transactions($this->_per_page, true, $view['slug']);
+        foreach ($this->_views as $view) { 
+            $this->_views[$view[ 'slug' ]][ 'count' ] = $this->_admin_page->get_transactions($this->_per_page, true, $view[ 'slug' ]);
         }
     }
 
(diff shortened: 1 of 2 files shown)


admin_pages/transactions/EE_Admin_Transactions_List_Table.class.php
```diff
@@ -148,9 +148,9 @@ protected function _get_table_filters()
      */
     protected function _add_view_counts()
     {
-        $this->_views['all']['count'] = $this->_admin_page->get_transactions($this->_per_page, true, 'all');
-        $this->_views['abandoned']['count'] = $this->_admin_page->get_transactions($this->_per_page, true, 'abandoned');
-        $this->_views['failed']['count'] = $this->_admin_page->get_transactions($this->_per_page, true, 'failed');
+        foreach ($this->_views as $view) {
+            $this->_views[ $view['slug'] ]['count'] = $this->_admin_page->get_transactions($this->_per_page, true, $view['slug']);
+        }
```
tn3rb: should go through and make this change to ALL of our list tables (not as part of this PR though)

Pebblo:
Thanks, Brent :)

Sent for testing (see initial post)

joshfeck:
This works quite well. I tested this and it's ready for merging.

PR #756, opened 2018-10-29
Fix block asset registration for frontend scripts/styles

nerrad:

While testing the `EventAttendees` block as a part of #637, I noticed that when rendering the block on the frontend, all the backend scripts/styles were being loaded.  This shouldn't happen because there's a lot of unnecessary js that's specific to loading in the editor context.  Not only that, but it produces console errors due to the custom category for the blocks (which won't register on the frontend php side because its only intended for admin registration per the WP/gutenberg api).

This pull takes care of splitting out editor scripts/styles from frontend script/styles so that there is a clear distinction between the two. This pull also adds placeholder comments/files for demonstrating how to setup the build for frontend script generation once we actually start implementing blocks with scripts/styles for the frontend.  It also allows for reusability of script/style modules that are common between frontend and backend (the only difference being what's included in the final build).

## How has this been tested

Of primary concern with this pull for now is that:

- Existing behaviour of block registration for backend editor is unaffected.
- Scripts/styles currently being built are only used for editor context.

* [x] I've verified both the above via testing the changes merged into #637 

## Checklist

* [x] My code is tested.


assets/src/blocks/index-frontend.js
```diff
@@ -0,0 +1,5 @@
+/**
+ * @todo when there is javascript/css for blocks that needs to be on the
+ * frontend, this is where you import/export it from (final script/styles
+ * built from this file
+ */
```
nerrad: This is just  a placeholder for now.  But this is basically the entry script for the build process so it can be used to import any common script/style modules between admin and frontend and would then be built for consumption on the frontend.

core/services/assets/BlockAssetManager.php
```diff
@@ -212,7 +212,7 @@ public function addScript($handle, array $dependencies = array())
                 $this->domain->assetNamespace(),
                 $handle
             ),
-            $this->addDefaultBlockScriptDependencies($dependencies)
+            $dependencies
```
nerrad: Assuming that frontend context will need the same defaults as the editor context for dependencies is incorrect (especially when it comes to block editor api stuff).  We still may have a new helper method for frontend dependencies at some point (i.e. it's probably a safe bet `wp.i18n` would be one) but I'd like to iterate on that once we actually have a frontend script being built.

webpack.common.js
```diff
@@ -233,6 +233,10 @@ const config = [
 			'eventespresso-core-blocks': [
 				assets + 'blocks/index.js',
 			],
+			/** implement once there's javascript to output
+			'eventespresso-core-blocks-frontend': [
+				assets + 'blocks/index-frontend.js',
+			],*/
```
nerrad: Placeholder, we don't actually need to build this yet because there is nothing to build, but when we do we can just comment this out.  I debated whether I should do this or not because in general I dislike adding placeholders like this, but in this case I think its of benefit for our future selves to have.

core/services/assets/BlockAssetManager.php
tn3rb: should you not have to add the following dependencies for the editor blocks now?

```php
'wp-blocks',    // Provides useful functions and components for extending the editor
'wp-i18n',      // Provides localization functions
'wp-element',   // Provides React.Component
'wp-components', // Provides many prebuilt components and controls
CoreAssetManager::JS_HANDLE_EE_COMPONENTS
```

ie: removing these from the frontend context means you have also removed them from the admin context, but I don't see them being added back anywhere.

nerrad: This is just in the `addScript` method.  `addEditorScript` still has this call.

PR #757, opened 2018-10-30
Testing Event Attendees Block

nerrad:
The `EventAttendees` block is the first integration of EE with Gutenberg.  Along with adding this block to EE core, this pull also:

- expands the new javascript build process.
- adds a GB block framework in php for registering blocks and block assets.
- adds a new javascript framework for not only GB blocks, but also doing react work in general (components, model entities etc).

So while the focus of this ticket is testing the `EventAttendees` block, indirectly there's a number of new things being tested as well that will assist with building react based applications and future Gutenberg integrations (which will all get merged to GB).

### Environment setup

- Install this branch in a WordPress version environment prior to WordPress 5.0.
- Do _not_ have the Gutenberg plugin active.  I don't think the GB plugin is going to support earlier versions of WP once WP 5.0+ launches so I don't think we need to be testing against that.

### Tests

The following tests don't have to be exhaustive.  You are basically just making sure that there are no fatal errors related to gb stuff loading that shouldn't in earlier versions of WP.  Errors will pop up quite quickly if that happens.

* [x] Verify that creating events etc in the event editor work as expected 
* [x] Verify that doing a registration in the frontend works as expected.
* [x] Verify that doing a registration in the backend works as expected.
* [x] Verify that you can create a WordPress post without any issues.

### Environment setup

- We're primarily concerned with testing against the latest release of WordPress 5.0 (beta or release candidates), so install this in a WordPress environment that has WP 5.0+ installed.
- Optionally, repeat tests with the latest GB plugin release active.  This is optional because its mostly required for development purposes and people will be discouraged from having both GB plugin and WP 5.0+ active in production.

#### New block behaviour

The following tests are specific to the new `EventAttendees` block.

*General Testing Considerations:*

- watch for errors in the browser javascript console.
- try testing cross browsers. 
- try testing across various themes
- watch for any weird ui/ux behaviour and report it.  "Weird" is a admittedly a subjective thing and this is a vague consideration.  Basically this means if you see behaviour that seems glitchy or gives a bad experience, report it. 
- The testing notes focus mostly on just a single `EventAttendees` block being added to content, you will want to try adding multiple blocks as well.
- This block should work for any WordPress post type (post, pages, other cpts) that supports the Block Editor. So verify that it works in various post type editors as well.
- This block should work for the Event  Espresso CPT editors _once they support the block editor._

* [x] Verify that the Event Attendees block appears in the "Event Espresso" category of the Add block picker menu tool.
* Add an `EventAttendees` block to your content
* [x] Verify that you see a placeholder in your content with "To get started, select what event you want to show attendees from in the block settings"
* [x] Verify when you preview the post at this point, there is no rendered content for the block.
* [x] Go ahead and select the block and play with various settings to change the contents of the block.  Verify that the block content shows expected results.
* [x] for each variation you try in the previous step, ensure that the frontend for the post displays the content the same as how it appears in the editor.
* [x] In particular, ensure the "Display on Archives" feature works as expected for viewing block content on the frontend in an archive view.

#### Other behaviour

The following tests don't have to be exhaustive.  You are basically just making sure that there are no fatal errors related to new functionality in EE core.
* [x] Verify that creating events etc in the event editor work as expected 
* [x] Verify that doing a registration in the frontend works as expected.
* [x] Verify that doing a registration in the backend works as expected.


garthkoyle:
Looks like there is still development happening on this ticket. Is it ready for testing? If not, please let me know when it is ready.

nerrad:
The changes merged here were very minor changes (linting error, removing hard code id etc), nothing that would invalidate testing.  So it's still ready for testing.  If there's a significant change needing done that DOES invalidate testing I won't be pushing to this branch.

sethshoultes:
Tests without WordPress 5.0 are working for me. Moving on to testing with WP 5.0.

sethshoultes:
@garthkoyle @lorenzocaum @joshfeck @Pebblo what do you think about changing the default for "Display on Archives" feature to ON?

garthkoyle:
I agree with Tony. In addition, sometimes people can't figure out how to turn things off (or on), then we get blamed that we're "insecure" or "unable" to... So the defaults are an important piece to the perception of our applications.

nerrad:
> I don't think most users will want it enabled by default, we do get a few people asking how to add an attendee list to events but it's a relatively small number.

I just want to clarify that you are understanding the purpose of the "Display on Archives" option setting for a specific EventAttendees block.  Currently when someone adds an EventAttendees block to their content it will display attendees from the selected event/datetime/ticket as set in the block settings _for that specific block_ (they can have multiple blocks in the content) on the single post view for that post content.

The "Display on Archives" setting simply means that when the post containing the block is displayed in any archive view (home page showing list of posts, category archive showing list of posts etc), the block content is not shown.

You may have already understood it to behave that way and if so then this explanation wasn't needed but I wasn't sure so I thought I'd clarify.

garthkoyle:
Pre WP 5.0 (4.9.9-alpha-43656) seems to work as normal.

nerrad:
After discussion in slack, we've come up with the following decisions:

1. Toggles will continue to be used but I'll add some dynamic help text to provide supplemental information about what they do (text changes depending on state of toggle).
2. "Location Settings" changes to "Archive Settings"

PR #759, opened 2018-10-30
Split out vendor scripts from vendor bundle for things included in WP 5.0

nerrad:

WordPress 5.0 exposes new libraries as dependencies (script handles):

- react
- react-dom
- moment
- lodash

The problem is that in order to use those bundled scripts, WordPress 5.0+ must be running in the environment EE is loaded on.

The **easy** solution would be to just bump the WP version requirements when we merge in our gb stuff, but this increases maintenance burden on our customers because it means they'll have to get their website on the latest version of WP before upgrading to whatever version of WP this build lands in which is a big jump from the current WP requirements.  So currently this option has been given a no go (understandably) by our support team.

The solution implemented in this pull is:

* [ ] Copy bundled scripts from WP 5.0 core into our own `assets/vendor` directory.  the only one I didn't copy from the above list is `moment` because we use `moment-timezone` and `moment-duration` (which WP does not)... so we still have to use our own version (bundled in `ee-vendor`).
* [ ] Do wp version detection in our asset manager and register the above scripts on the same handle that WP uses if the environment has a lower version of WP. I created a new `VendorJavascriptAsset` for handling that asset registration.  This way if WP 5.0+ is registered, our script dependencies will just end up loading whatever is available.
* [ ] Remove the above scripts from being bundled with `ee-vendor`.
* [ ] Adjust the dependencies on existing `ee-vendor` so for the new vendor script handles.

## How has this been tested

* [ ] verify no unit test failures are introduced
* [ ] verify EventAttendees block still work as expected.
* [ ] Verify the plugin exit modal form works on the plugins page in a WP 5.0+ environment.
* [ ] Verify that the plugin exit modal form works on the plugins page in a WP <5.0 environment (and with GB plugin disabled as well)

## Checklist

* [ ] My code is tested.


nerrad:
Huh, the _only_ reason I did the separate Asset class for vendor scripts is because I didn't realize AssetManager had access to the `Domain` (for urls) - I must have been looking at the wrong constructor.  So ya, I agree no need for the extra class.

core/domain/services/assets/CoreAssetManager.php
```diff
@@ -335,7 +401,7 @@ private function getAccountingSettings() {
 
     /**
      * Returns configuration data for the js Currency VO.
-     * @sinc $VID:$
+     * @since $VID:$
```
tn3rb: should probably replace `$VID:$` with `4.9.62.p` here (or whatever other methods around it are using) since this obviously isn't being added in this branch and got skipped when setting versions

core/services/assets/AssetManager.php
```diff
@@ -123,6 +125,33 @@ public function addJavascript(
...
+     * @throws DomainException
+     * @since $VID:$
+     */
+    public function addVendorJavascript(
+        $handle,
+        array $dependencies = array(),
+        $load_in_footer = true
+    ) {
+        $dev_suffix = wp_scripts_get_suffix('dev');
+        $vendor_path = $this->domain->pluginUrl() . 'assets/vendor/';
```
tn3rb: should have mentioned this in my original comment, but if you think there is any chance that we will need to recreate the vendor path outside of this method, then this could be moved to its own method, ie:

```php
public function getVendorSourcePathForHandle($handle) {
    $dev_suffix = wp_scripts_get_suffix('dev');
    $vendor_path = $this->domain->pluginUrl() . 'assets/vendor/';
    return  "{$vendor_path}{$handle}{$dev_suffix}.js";
}
```

I can't immediately see how that would be necessary, so I'm fine with it as is, but if you think that could be useful, then better to do this now.

nerrad: ya I didn't create it outside of here (i.e. put it in the domain class) because I think its pretty specific to this need. I don't anticipate seeing more use-case for this and if we do it can be iterated on.

core/domain/services/assets/CoreAssetManager.php
nerrad: This was added in the GBM master branch, hence it's still needed as is so when it gets merged to master it gets the appropriate version.

nerrad: Your comment may have been because you thought `@since` is in the pattern matching for `$VID:$` replacement.  It's not.

PR #762, opened 2018-10-30
Remove usage of <ServerSideRender /> for EventAttendees block

nerrad:

For context see [this grinder post](http://p2.eventespresso.com/2018/10/22/gutenberg-side-effect-extensibility-of-ee/)

In this pull the `<EventAttendees />` block will be converted so that:

- it does not transform a shortcode to the block (we'll leave shortcodes as their own thing).
- it does not use SSR for rendering the content in the editor.  Instead the editor will have its own component for displaying the data.  Styles for the content in the editor will be also loaded on the frontend.  
- explore using react to lazy load the content on the frontend as well, however, in the interest of time we may still have the same template generated and loaded on the frontend as what is done in the client side in the editor.  So there still would be a consistent rendering.  IF there's time, we could offer the option to toggle between lazy loading and SSR on the frontend (which might be a good thing for events with a lot of attendees).

## Other things that are implemented in this pull:

* Organized panels for different options
* New Gravatar controls for controlling the size of the gravatar.
* Add more css classes to containers.  One change here is that every Event Espresso block receives the `event-espresso-blocks` css class and every *core* block receives the `ee-core-blocks` class.  One potential use-case for this is when troubleshooting client sites, the support agent can just use a user-agent style sheet that highlights all our blocks with a border in content (so they readily standout).  Another use-case is searching the page source when troubleshooting.  Another potential use-case will be for maintaining certain common style guides for all our blocks that can be applied using css specificity rules.
* Every request for attendees pulls gravatar images by default but only SHOWS the gravatar in the editor preview if `showGravatar` is true, this saves a round trip.
* Every attendees request pulls a limit of 100 attendees regardless of the set limit.  This resolves the "known issue" noted before.  The limit control simply affects how many attendees are displayed in the preview.
* Bunch of user context messaging added.  For example, on the initial block insertion, the user is instructed to select an event to get started.  Thus the _default_ behaviour of the block is to NOT show any attendees (differs from shortcode in that there is default query).

## How has this been tested

* [x] ensure existing and new unit tests pass.
* [x] Verify that `EventAttendees` block works as expected on the backend and frontend.

## Known issues

* [x] **Limit Range Control Slider**: Each time the range control slider is interacted with, a network request happens.  This has the potential to trigger a whole slew of requests for people sliding around on the control since with our store, each "limit" value is another cached query with results. **RESOLVED**

## Checklist

* [x] My code is tested.

Here's some screenshots showing the changes applied from this branch:

### Editor with first part of block panel:

screenshot-1

### Editor with second part of block panel:

screenshot-2

### Frontend view of post with blocks:

screenshot-3


assets/src/blocks/event-attendees/style.css
```diff
@@ -1,11 +0,0 @@
...
-    list-style-type: none;
-}
-
-.event-attendee-image-wrap-div {
-    display: inline-block;
-}
-
-.event-attendee-avatar-img {
-    margin: 0 1em -.5em 0;
-}
```
nerrad: I moved these into the `EventAttendeesList` component stylesheet and added specificity rules there that ensure the css is the same on the frontend as the editor backend.  Note, this file is still left in place and imported (even though its blank) because the block styles have the component styles as a dependency and this ensures that any styles related to the block are loaded in the correct context.  Right now the same styles are used backend and frontend so there's no splitting of those files yet.

assets/src/components/event-attendees/attendee-list-item.js
```diff
@@ -0,0 +1,60 @@
...
+import { isModelEntityOfModel } from '@eventespresso/validators';
+import warning from 'warning';
+import { Fragment } from '@wordpress/element';
+
+/**
+ * Internal imports
+ */
+import { AvatarImage } from '../ui/image';
+
+const AttendeeListItem = ( {
```
nerrad: This was added to the components folder (and thus exported with `@eventespresso/components`) because its not specific to a block.  It's feasible that it could be used elsewhere in a ui (and components are "context" agnostic).

assets/src/components/event-attendees/index.js
```diff
@@ -0,0 +1,70 @@
...
+		const reducerCallback = ( items, attendee ) => {
+			items.push( <AttendeeListItem
+				attendee={ attendee }
+				showGravatar={ showGravatar }
+				isLoading={ isLoading }
+				{ ...this.props }
+			/> );
+			return items;
+		};
+		const listItems = mapReducer( attendees, reducerCallback, [] );
```
nerrad: man this mapReducer is gonna come in handy 🎈

assets/src/components/query/limit/index.js
```diff
@@ -28,6 +29,7 @@ export const QueryLimit = ( {
 			min={ min }
 			max={ max }
 			onChange={ onLimitChange }
+			{ ...rest }
```
nerrad: `RangeControl` can receive more props so by putting this here it allows us to pass those props through our wrapping component - I used this to pass through the `help` prop for instance.

In general, I think whenever we build a **component**  that wraps a third party component, we should make sure we pass through extra props not explicitly overridden in our custom component unless there's very good reason to not pass them through.  Among other reasons, this allows for us automatically picking up any new prop availability/behaviour in the wrapped component without having to refactor our custom component.

assets/src/components/ui/image/avatar.js
```diff
@@ -42,13 +42,17 @@ export class AvatarImage extends Component {
...
+			width: avatarWidth,
+		};
 		return avatarUrl ? (
 			<div className={ avatarClass + '-image-wrap-div' }>
 				<img
 					className={ avatarClass + '-avatar-img avatar' }
 					src={ avatarUrl }
-					height={ avatarHeight }
-					width={ avatarWidth }
+					style={ style }
```
nerrad: Why style instead of height/width?  Unfortunately if any css selectors that match this img tag have `height` or `width` set they automatically override any height and width html attributes on the html tag (the default height/width is "cached" as the original size of the img). In order to have the img size changes seen immediately in the editor, we need to use style tags. 

For server rendering on the frontend I _do_ use the height and width html attributes because the size of the image isn't changing at all after it loads.  So that still allows clients to control the size via css if they want on the frontend.

assets/src/data/eventespresso/lists/entities.js
```diff
@@ -13,14 +15,19 @@ import { MODEL_NAMES } from '../../model';
...
 		) => source.getEntities( state, modelName, queryString );
 		selectors[ getMethodName( modelName, '', 'isRequesting', true ) ] = (
 			state,
 			queryString,
-		) => source.isRequestingEntities( state, modelName, queryString );
+		) => isResolving(
+			REDUCER_KEY,
+			methodNameForGet,
+			queryString
+		);
```
nerrad: So I discovered this bug while testing things on this branch when I decided to try the new dynamically generated model selectors (`getAttendees` and `isRequestingAttendees`).  I realized that in the original iteration `isRequestingAttendees` would end up checking the resolution state for `getEntities` because it's wrapping a call to `isRequestingEntities`.  However the resolution state would be indexed on the `getAttendees` selector (which also has a different set of arguments).  Making this change here (and for the schema registry) fixed things.

assets/src/data/eventespresso/lists/test/entities.js
```diff
@@ -87,12 +87,12 @@ describe( 'createEntityResolvers()', () => {
...
 			'getEvents',
 			'testQueryA',
-			{ path: '/wp-json/ee/v4.8.36/events?testQueryA' },
+			{ path: '/ee/v4.8.36/events?testQueryA' },
 		],
 		[
 			'getDatetimes',
 			'testQueryB',
-			{ path: '/wp-json/ee/v4.8.36/datetimes?testQueryB' },
+			{ path: '/ee/v4.8.36/datetimes?testQueryB' },
```
nerrad: So I realized while working on this branch that I was actually bundling our scripts with the `@wordpress/api-fetch` package instead of using the `wp.apiFetch` global exposed by WP. That global is automatically initialized to add the `wp-json` prefix to everything.  So, once I fixed that, I didn't need to add the prefix here.

assets/src/data/model/attendee/query.js
```diff
@@ -0,0 +1,129 @@
...
+		where.push( `where[Registration.Ticket.Datetime.DTT_ID]=${ forDatetimeId }` );
+	} else if ( forEventId !== 0 && ! isNaN( forEventId ) ) {
+		where.push( `where[Registration.EVT_ID]=${ forEventId }` );
+	}
+
+	if ( REGISTRATION_STATUS_IDS.includes( forStatusId ) ) {
+		where.push( `where[Registration.Status.STS_ID]=${ forStatusId }` );
+	}
+	if ( showGravatar === true ) {
+		where.push( 'calculate=user_avatar' );
```
nerrad: This used to be `calculate=userAvatar` and I was kinda shocked I needed to make this change.  It's possible you never caught that this wasn't working because you switched to using server side render.

tn3rb: it used to work but likely changed as a result of https://github.com/eventespresso/event-espresso-core/pull/681 which modified how calulated fields were retreived

core/domain/ui/blocks/event-attendees.php
```diff
@@ -0,0 +1,40 @@
...
+                    . ' width="' . $attributes['avatarSize'] . '"'
+                    . ' height="' . $attributes['avatarSize'] . '"'
+                    . ' src="' . $gravatar . '" alt="contact avatar">'
+                    . '</div>'
+                : '';
+            echo "<li>{$gravatar}<span>{$attendee->full_name()}</span>";
+        }
+        ?>
+    </ul>
+</div>
```
tn3rb: In DDD the UI layer should be a separate layer all on its own, and can even be split up to accommodate different kinds of UIs.
Ideally the Domain layer only contains all of your application's raw business logic, which can then be accessed via various kinds of UIs and/or request types such as web browser, RSS feed, REST API, CLI, AJAX, Cron, etc...

So based on the above, and the fact that we already have a `/ui/browser/` folder, something like the following would be better:

```php
- /domain/
- /ui/
- /ui/cli/
- /ui/browser/
- /ui/browser/blocks
```

we have not done a good job of this at all and have waaaaay too much logic everywhere that has presentation layer type stuff embedded in both domain and application logic, but it helps to remember to try and separate those things as much as possible because then your domain logic is far more reusable (and functional) when it is free of presentation logic.

core/services/assets/BlockAssetManager.php
```diff
@@ -136,8 +136,8 @@ public function addAssets()
     {
         $this->addEditorScript($this->getEditorScriptHandle());
         $this->addEditorStyle($this->getEditorStyleHandle());
-        $this->setScriptHandle($this->getScriptHandle());
-        $this->setStyleHandle($this->getStyleHandle());
```
tn3rb: uhhh... we still need to set the script and style handles else things will break!!!

for example:

- `BlockAssetManager::getScript()` references `$this->script_handle` which will no longer be set!
- `BlockAssetManager::getStyle()` references `$this->style_handle` which will no longer be set!
- `\EventEspresso\core\domain\entities\editor\Block::registerBlock()` references the getters for both of the above properties!!!
- you now have `addDefaultBlockScriptDependencies()` calling `getScriptHandle()` but `$this->script_handle` is no longer be set!
- you have added `addDefaultBlockStyleDependencies()` which calls `getStyleHandle()` but `$this->style_handle` is no longer be set!


Why exactly was this done?

and more importantly... how is this not blowing up for you when you try to use things? I can only guess that things have changed so blinking much with GB that all of this asset manager and block registration stuff is working completely differently now.

assets/src/data/model/attendee/query.js
nerrad: ahh

core/services/assets/BlockAssetManager.php
nerrad: 1. If that was the case then why does `addEditorScript` and `addEditorStyle` not have a `setEditorScriptHandle` and `setEditorStyleHandle` call?
2. `setScriptHandle` and `setScriptStyle` are actually called in the child class (`CoreBlocksAssetManager`)

```
/**
     * @since $VID:$
     */
    public function setAssetHandles()
    {
        $this->setEditorScriptHandle(self::JS_HANDLE_CORE_BLOCKS_EDITOR);
        $this->setEditorStyleHandle(self::CSS_HANDLE_CORE_BLOCKS_EDITOR);
        $this->setScriptHandle(self::JS_HANDLE_CORE_BLOCKS);
        $this->setStyleHandle(self::CSS_HANDLE_CORE_BLOCKS);
    }
```

This was done because nothing was actually *adding* the script and styles and thus things WERE broken.  The work I did is the first time the frontend script/styles were actually tested so...

tn3rb: right... sry forgot that asset handles were set there now... it wasn't like that initially but ended up changing.


so ya the change you made here was likely something I missed when handle setting was moved to the child classes.

PR #763, opened 2018-10-30
Fix: Don't add version string to asset source when hash name assets used

nerrad:

We just noticed that when "built" (built meaning through our new asset build process) assets are enqueued they are getting the version string `?ver=4.9.xx` added to them and they should _not_ be.  Having the version string basically makes the hash in the asset name pointless. 

Some preliminary investigation points to `BrowserAsset::version()` being the potential culprit except that it appears to have conditionals that should correctly detect when an asset source has `.dist.js` or `.dist.css` in the string.  Still, that's the first place to check for fixing this.

Also in this pull I'll make sure there's adequate phpunit test coverage for this process and either add tests or update tests if necessary.

## How has this been tested

* [x] Wrote unit test coverage
* [x] verified the javascript/css loaded on the WP plugins page (which is impacted by these changes) still works as expected.

## Checklist

* [x] My code is tested.

### DIFF
--- a/core/domain/values/assets/BrowserAsset.php
+++ b/core/domain/values/assets/BrowserAsset.php
@@ -2,6 +2,7 @@
 
 namespace EventEspresso\core\domain\values\assets;
 
+use DomainException;
 use EventEspresso\core\domain\DomainInterface;
 use EventEspresso\core\exceptions\InvalidDataTypeException;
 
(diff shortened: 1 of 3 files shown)


nerrad:
> You're going to need a new line at the end of JavascriptAssetTest.php to pass our linting tests however.

> (EDIT) or maybe not seeing as this already passed somehow ?!?!

The test folder is excluded from linting.  I don't care as much about code style in our test code.  It's already costly in time writing the dang things.

PR #769, opened 2018-11-02
Switch away from @eventespresso/i18n and use @wordpress/i18n directly (wip)

nerrad:

A while ago, I started using `@eventespresso/i18n` as a way to get around the fact that we started using wp.i18n functionality in environments that may not have `wp.i18n` so it was a way to ensure we could just use it anywhere (basically `eejs.i18n` is just a wrapper around the `@wordpress/i18n` package).

However, now that the release of WP5.0 is imminent, I think it'd be better for long term maintenance if we start using `@wordpress/i18n` everywhere instead and I can just give it the same treatment that I did for bundled vendor scripts in WordPress (such as `react`).  So essentially just copying the `wp.i18n` package into the `assets/vendors` directory and registering it on the `wpi18n` handle if its not already registered in WP.

**Note:** this is not an urgent requirement and can be done "whenever".  It's more of an optimization thing (and reduces maintenance overhead).  When I implement, I can just alias `@eventespresso/i18n` to the `wp.i18n` external instead of `eejs.i18n`.

## How has this been tested

* [ ] Verify no js unit-tests fail (indicating configuration problems)
* [ ] Verify files build successfully
* [ ] Verify existing GB blocks and react apps (plugins deactivation modal) still work as expected on the newly built files.

## Checklist

* [ ] My code is tested.

### DIFF
--- a/webpack.common.js
+++ b/webpack.common.js
@@ -5,6 +5,8 @@ const autoprefixer = require( 'autoprefixer' );
 const externals = {
 	jquery: 'jQuery',
 	'@eventespresso/eejs': 'eejs',
+	// @todo need to switch away from doing this and give it the same treatment
+	// as I did for other js libraries bundled with WP 5.0.
 	'@eventespresso/i18n': 'eejs.i18n',
 	'@wordpress/api-request': 'wp.apiRequest',
 	'@wordpress/data': 'wp.data',


nerrad:
I'm thinking it'd actually be good to put this ticket on the backburner until we bump minimum version required in EE to WP 5.0.  What we have now works fine across all WP versions we support and it reduces the need for additional boilerplate code to handle when EE is run in an environment that doesn't have wp.i18n.

`wp.i18n` itself is not a big script so while moving to just it will save some size in the js, I'm not sure it's enough of a deal to use it currently.

nerrad:
@joshfeck based on my previous comment, what pipeline do you think this would be best in?  I don't want to close it because it is worthwhile doing when its easy to do so (i.e. we bump minimum wp version required for EE).

tn3rb:
no longer applicable

PR #775, opened 2018-11-07
Make adjustments to tag list and update version numbers

joshfeck:

See the p2 at 
http://p2.eventespresso.com/2018/11/07/evaluating-decafs-tag-list/


joshfeck:
I did some further research and made a few changes to the tag lists. The p2 has more information about the research.

PR #776, opened 2018-11-09
Fix non-working production build of javascript.

nerrad:

After merging #331 into master I did a production build of the javascript (`npm run build`) to verify that things worked okay on that build.  The following issues surfaced when testing:

- attendees were being retrieved in requests but not displayed.
- intermittent issues where the selectors for tickets/datetimes would not show expected results.

What made these issues hard to debug are:

- development build had no problems
- there were no console errors pointing to any problems (however there was a console error related to component implementation).

The problem ended up being our validators making hardcoded assumptions about a `constructor.name` for an object when that object is minified and thus the name is minified as well.

This pull:

- fixes validators (`isModelEntity` and `isFactoryOfModel`) so they correctly account for minified scenarios. (which in turns fixes the attendee display and the inconsistent selector issues)
- has a minor fix for component structure to ensure components are properly keyed (using `<Fragment>` and nested.

## How has this been tested

* [x] Ran `npm run build` and tested all functionality of the `EventAttendees` block to make sure it worked as expected.
* [x] Ran the same tests on `npm run watch` to verify the same.
* [x] Ensure unit tests passed locally.

## Checklist

* [x] My code is tested.


assets/src/data/model/entity-factory/base-entity.js
```diff
@@ -95,6 +95,8 @@ class BaseEntity {
 	get isClean() {
 		return this.saveState === SAVE_STATE.CLEAN;
 	}
+
+	static name = 'BaseEntity'
```
nerrad: Javascript sets the `constructor.name` property on a class automatically from whatever the name is given in the code ( eg. `class BaseEntity { }`).  But you can override this by explicitly declaring the name property on the class constructor.  If we don't do this then when a minifier changes something like `class BaseEntity` to `class a`, there's no way to know accurately elsewhere in the code what the name of the class is.  However, the minifier will leave _explicitly_ defined `name` constructor properties alone.

nerrad:
@tn3rb, when you review this, I'd appreciate if you could also do a test with the `npm run build` production files to ensure the event attendees block works as expected for you too.  If all is good I'll merge it into master.

assets/src/data/model/entity-factory/base-entity.js
tn3rb: kk kewl.. .I may have to do the same in some of the REM code

PR #783, opened 2018-11-13
Add fail safe for loading of the network core config

tn3rb:
Josh  was dealing with a site that had the following error appear on their **General Settings** > **Your Organization** admin page

```php
Fatal error: Uncaught TypeError: 
Argument 4 passed to EventEspresso\admin_pages\general_settings\OrganizationSettings::__construct()
must be an instance of EE_Network_Core_Config,
instance of EventEspresso\core\services\address\CountrySubRegionDao given
```

The reason for the error was because `EE_Network_Core_Config` is stored as a property on `EE_Network_Config`, but for some reason had gotten converted to a `stdClass` object at some point prior to being saved to the database. This `stdClass` object still had the correct values saved to it, such as their site license key, but it was not being accepted as the correct dependency for the `OrganizationSettings` class because it type hints specifically for an instance of `EE_Network_Core_Config`.

The only code in `EE_Network_Config` that could explain how this happens would be some filters and actions that exist, however I was unable to find anything using those hook points in EE core or any of the EE addons I have installed on my local dev server. 

Other possible causes:

- the site owner or their dev(s) may have utilized one of the above hooks at some point and returned a `stdClass` object instead of an instance of `EE_Network_Core_Config`
- there were meta boxes from a couple of other plugins active on that site's **General Settings** > **Your Organization** admin page, so it's possible that one of them somehow interfered with the config saving, although I think this would be a huge stretch.
- the site had multiple backup/migration plugins active such as BackupBuddy, WP Migrate DB Pro, ManageWP, etc, and it's possible that one of these plugins unserialized and then re-serialized the database content in a request where EE was not running, which would mean that none of our classes would be available. This may have resulted in unknown objects being converted to a `stdClass` prior to re-serialization and saving. Our regular Config class has failsafes in place to _try_ and prevent properties from becoming corrupted (doesn't always work) but `EE_Network_Config` kinda does it's own things and did not have any fail safes in place.

Which leads to the purpose of this pull request. This branch adds a bit of logic to `EE_Network_Config` to ensure that its embedded `EE_Network_Core_Config` object does not become corrupted.

## How has this been tested
I pink sombrero'd the broken site and the fixes in this branch worked successfully there.
pink-sombrero-50x50 

## Checklist

* [x] My code is tested.


tn3rb:
## testing notes

- log in and visit the **General Settings** > **Your Organization** admin page
- [ ] confirm that no errors occur
- enter the super secret universal site license key `1234567890` into the **Support License Key** field and click **Save**
- [ ] confirm that no errors occur
- feel free to experiment with the **Your Organization** admin page in this branch

tn3rb:
@darren I pushed an additional fix for `EE_Registry` that basically ensures that a value is always set for all of a class's constructor arguments.
One of the issues with this broken site was that the error message was misleading because it implied that class signature for `OrganizationSettings` was somehow wrong or being interpreted incorrectly. 

What was really happening is that we were passing an array of arguments that looked like:

```php
array(4) {
  [0]=>
  string(11) "EE_Registry"
  [1]=>
  string(22) "EE_Organization_Config"
  [2]=>
  string(14) "EE_Core_Config"
  [4]=>
  string(55) "EventEspresso\core\services\address\CountrySubRegionDao"
}
```

to PHP when attempting to construct `OrganizationSettings`, but PHP must obviously ignore the indexes and was attempting to pass the fifth argument (`CountrySubRegionDao`) for the fourth constructor parameter (which should have been `EE_Network_Core_Config`).

The fix is to simply ensure that a default value is always set for each constructor parameter, even if that is null, ie:

```php
array(5) {
  [0]=>
  string(11) "EE_Registry"
  [1]=>
  string(22) "EE_Organization_Config"
  [2]=>
  string(14) "EE_Core_Config"
  [3]=>
  NULL
  [4]=>
  string(55) "EventEspresso\core\services\address\CountrySubRegionDao"
}
```

which will now result in the following (less confusing) error message:

```php
Fatal error: Uncaught TypeError: 
Argument 4 passed to EventEspresso\admin_pages\general_settings\OrganizationSettings::__construct()
must be an instance of EE_Network_Core_Config,
null given
```

core/EE_Registry.core.php
```diff
@@ -1354,7 +1354,8 @@ protected function _resolve_dependencies(ReflectionClass $reflector, $class_name
                     $arguments,
                     $index
                 );
-            } else {
+            }
+            if (empty($arguments[ $index ])) {
```
nerrad: Is this location the best location for this conditional check?  Is it something that could happen (and would be more efficient) earlier in the loop iteration?

tn3rb: no because it is represents the fallback behaviour when a dependency can not be resolved any other way and therefore has to happen after all other attempts have been made.

joshfeck:
I tested this, it's ready for merging.

PR #786, opened 2018-11-14
Fixed Datetime Filter Radio Input Values Fixes #785

### LINKED ISSUE #785: Datetime filter breaks multiple tickets if max tkt qty is set to 1
Reported here: https://eventespresso.com/topic/%EF%BB%BFan-error-has-occurred-you-need-to-select-a-ticket-quantity-before-you-can-proc/

If you set the max tkt qty for an event to 1, create enough datetimes tickets (I used 5 datetimes with 1 ticket per datetime) to trigger your 'show datetime selector' the ticket selector can't process any tickets other than the first.

This is my setup - http://take.ms/0ok4M

Frontend - http://take.ms/Q8jiV
(I selected ticket 3)

Ticket 1 works as expected.

Same happens with MER.

To be clear, your 'Show Date & Time Filter?' option in Event Espresso -> Events -> Templates needs to be set to Maybe and your threshold (for the example above) needs to be 5 or below.

If I deactivate that option and refresh the page I have no trouble registering on any ticket.

tn3rb:
plz see #785

## How has this been tested
Monkey based browser testing 

## Checklist

* [x] My code is tested.

### DIFF
--- a/modules/ticket_selector/assets/ticket_selector.js
+++ b/modules/ticket_selector/assets/ticket_selector.js
@@ -57,7 +57,7 @@ jQuery(document).ready(function ($) {
                         active_rows++;
                     } else {
 						$ticket_selector_row.addClass( 'ee-hidden-ticket-tr' );
-						$qty_input = $ticket_selector_row.find(
+						var $qty_input = $ticket_selector_row.find(
 							'.ticket-selector-tbl-qty-slct'
 						);
 						// set qty to zero for non-radio inputs
--- a/modules/ticket_selector/assets/ticket_selector.js
+++ b/modules/ticket_selector/assets/ticket_selector.js
@@ -56,7 +56,14 @@ jQuery(document).ready(function ($) {
                         $ticket_selector_row.removeClass('ee-hidden-ticket-tr');
                         active_rows++;
                     } else {
-                        $ticket_selector_row.addClass('ee-hidden-ticket-tr').find('.ticket-selector-tbl-qty-slct').val(0);
+						$ticket_selector_row.addClass( 'ee-hidden-ticket-tr' );
+						$qty_input = $ticket_selector_row.find(
+							'.ticket-selector-tbl-qty-slct'
+						);
+						// set qty to zero for non-radio inputs
+						if ( $qty_input.attr( 'type' ) !== 'radio' ) {
+							$qty_input.val( 0 );
+						}
                     }
                 });
             }


modules/ticket_selector/assets/ticket_selector.js
```diff
@@ -56,7 +56,14 @@ jQuery(document).ready(function ($) {
                         $ticket_selector_row.removeClass('ee-hidden-ticket-tr');
                         active_rows++;
                     } else {
-                        $ticket_selector_row.addClass('ee-hidden-ticket-tr').find('.ticket-selector-tbl-qty-slct').val(0);
+						$ticket_selector_row.addClass( 'ee-hidden-ticket-tr' );
+						$qty_input = $ticket_selector_row.find(
```
nerrad: It doesn't look like `$qty_input` has been declared anywhere.  In js, it's best practice to do so.

modules/ticket_selector/assets/ticket_selector.js
```diff
@@ -56,7 +56,14 @@ jQuery(document).ready(function ($) {
...
+							'.ticket-selector-tbl-qty-slct'
+						);
+						// set qty to zero for non-radio inputs
+						if ( $qty_input.attr( 'type' ) !== 'radio' ) {
```
nerrad: Is there the possibility of checkbox inputs?  If so, should this account for them as well? Should it be a whitelist check instead of a blacklist check?

tn3rb: no the Ticket Selector only uses select inputs or radio buttons.

tn3rb:
##testing 

- add/edit an event with the following:
    - 4 datetimes (D1, D2, D3, D4)
    - 4 tickets (T1, T2, T3, T4) all with their **Maximum Quantity** value set to `1` and different dates
- assign tickets to datetimes as follows:
    - T1 => D1
    - T2 => D2
    - T3 => D3
    - T4 => D4
- set **Maximum number of tickets allowed** in the Event Reg Options metabox to `1`
- click **Update**
- click on the **Events > Templates** tab and set the **Date & Time Filter** settings to the following values:
    - Show Date & Time Filter? => `maybe show date & time filter`
    - Date & Time Filter Threshold => `2`
    - Date & Time Filter Max Checked => `4` or more
- click **Save**
- navigate to the event on the frontend
- check that the Ticket Selector is using radio buttons and not dropdowns for selecting ticket quantities
- choose an option **other than the first radio button** and click **Register Now**
- cancel the registration
- repeat the ticket selection and submission process but choose different ticket options (including the first)
- cancel the registration
- repeat the ticket selection and submission process but for different events with different Ticket Selector configurations
- activate MER
- repeat the ticket selection and submission process
- [ ] confirm that tickets were processed and you were redirected to the registrations checkout
- feel free to experiment with the Ticket Selector in this branch and attempt to find situations where tickets are not processed

modules/ticket_selector/assets/ticket_selector.js
Pebblo: @tn3rb We had a request for using checkboxes for the ticket selector and I threw a snippet together that swapped out the select for a checkbox an then used the first value of the select if the checkbox was selected.

So would it be worth swapping this out? No objects if not, just noting that checkboxes have been requested/used previously.

Pebblo:
All good from me, thanks, Brent.

I added a comment above about checkboxes, although I released afterwards that it wouldn't matter anyway.

tn3rb:
@Pebblo 

where is this snippet you created?
thought it might be in here: https://github.com/eventespresso/ee-code-snippet-library/tree/master/templates
or  one of the other snippet folders but couldn't find it.

I would like to take a look because I may as well modify this to work with that snippet

Pebblo:
Sorry, I meant o include it, its here: https://gist.github.com/Pebblo/737310c7309436231008666ac27e9662

I don't think it'll be affected by this though as it hides the select and adds a checkbox using jQuery.

tn3rb:
thnx @Pebblo 
ya that gist won't be affected here because it is manipulating the hidden select input which will get picked up by the logic that processes the form

PR #792, opened 2018-11-21
fix potential for PHP fatal in PHP7.2+ boxes

nerrad:

We got a [report](https://eventespresso.com/topic/urgent-ee-4-9-71p-is-producing-the-following-php-errors/) of the following fatal error from a customer:

```
PHP Fatal error: Cannot use string offset as an array in /[redacted]/wp-content/plugins/event-espresso-core-reg/core/third_party_libs/pue/pue-client.php on line 1114
```

The fix in this pull intends to the address the potential for this kind of fatal surfacing.

## How has this been tested

* [x] Verified that pue-client functionality is not borked by this change.  Essentially what we're going is ensuring that what is retrieved from the db is an array and if its not then discarding it and saving as an array.

## Checklist

* [x] My code is tested.

### DIFF
--- a/core/third_party_libs/pue/pue-client.php
+++ b/core/third_party_libs/pue/pue-client.php
@@ -1083,6 +1083,7 @@ if (! class_exists('PluginUpdateEngineChecker')):
                 );
             }
 
+            $existing_notices = is_array( $existing_notices ) ? $existing_notices : array();
 
             //k make sure there are no existing notices matching the incoming notices and only append new notices
             //(unless overwrite is set to true).


nerrad:
@joshfeck I hope you don't mind me assigning to you.  I'll leave it up to you whether you want to try this on a customer's site if you get access before merging or just push it for merge now.  I'm confident it won't bork anything important.

nerrad:
> That looks like something else in WordPress.

Yup that's a WordPress core issue and they are already tracking a trac ticket for it: https://core.trac.wordpress.org/ticket/44979

joshfeck:
I've yet to hear back from the user that posted to the support forums. I suppose it'd be OK to merge this.

PR #793, opened 2018-11-23
Add shim for get_user_locale

nerrad:

Fatal errors are being reported for usage of this function in EE on WP versions earlier than 4.7.0.  Since we still support WP 4.5.0+ we should have a shim for this function.

## How has this been tested

* [x] I tested to ensure there is no unusual behaviour with WP version 5.0.RC1 (which would be cover any WP version > 4.7).  If there was a problem it would surface in some sort of error fairly quickly.
* [x] I have not tested personally, but should be verified that this fixes issues on WP versions < 4.7 (any wp version less would do).

## Checklist

* [x] My code is tested.

### DIFF
--- a/core/wordpress-shims.php
+++ b/core/wordpress-shims.php
@@ -93,3 +93,17 @@ if (! function_exists('wp_scripts_get_suffix')) {
         return $suffixes['suffix'];
     }
 }
+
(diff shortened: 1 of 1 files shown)


joshfeck:
I've verified the fix by testing on a site with WP 4.5.15.p

PR #802, opened 2018-11-29
move php docs for $query_params on EEM_Base::get_all to documentation

nerrad:
This allows for making more readable and easier to maintain documentation for this often used method.

This is a documentation change only.


mnelson4:
FYI
> Also, putting this documentation on EEM_Base::get_all() was really a little sloppy,
I meant when I dumped so much documentation into a single method's docblock. Not how you created the new markdown documentation page.

Kk so here's what I did:

* Renamed your new doc to "Model Query Params" and had it just focus on the structure of that special array (which again, in some ways, wish had just been an object)
* point to it from all the docs of model methods that use model query params
* updated model methods documentation to consistently document parameters, return type, and examples
* updated all the php doc-blocks which said something like "see $query_params on EEM_Base::get_all" to instead point to the new model query params documentation

nerrad:
Let's just publish it, I didn't really review it closely but docs are something we can iterate on as necessary so I'm not to keen on spending much more time on this.  I can't approve this using the code review tool because I created the pull.

PR #806, opened 2018-12-03
filter db engine when creating tables

mnelson4:

This should allow site owners to change what database engine a table uses. This should help with https://github.com/eventespresso/EE4-Promotions/issues/6, allowing folks having this problem to switch their table to `MyISAM` instead of `InnoDB`.

## How has this been tested

TBA

## Checklist

* [ ] My code is tested.

### DIFF
--- a/core/services/database/TableManager.php
+++ b/core/services/database/TableManager.php
@@ -184,6 +184,12 @@ class TableManager extends \EE_Base
      */
     public function createTable($table_name, $create_sql, $engine = 'MyISAM')
     {
+        $engine = apply_filters(
+            'FHEE__EventEspresso__core__services__database__TableManager__createTable__engine',
+            $engine,
+            $table_name,
+            $create_sql
+        );
         // does $sql contain valid column information? ( LPT: https://regex101.com/ is great for working out regex patterns )
         if (preg_match('((((.*?))(,\s))+)', $create_sql, $valid_column_data)) {
             $table_name = $this->getTableAnalysis()->ensureTableNameHasPrefix($table_name);


core/services/database/TableManager.php
```diff
@@ -184,6 +184,12 @@ public function dropIndex($table_name, $index_name)
      */
     public function createTable($table_name, $create_sql, $engine = 'MyISAM')
     {
+        $engine = apply_filters(
+            'FHEE__EventEspresso__core__services__database__TableManager__createTable__engine',
```
tn3rb: doesn't our filter naming schema only use single underscores to replace directory separators?

I think this should be: `FHEE__EventEspresso__core_services_database_TableManager__createTable__engine` no?

mnelson4:
Fixed the filter name, please review again. Thanks.

mnelson4:
@joshfeck can you test and review if you think this will actually help resolve folks' issues.

Pebblo:
This is based on the bug mentioned in this comment: https://github.com/eventespresso/EE4-Promotions/issues/6#issuecomment-441116649

However, reading over the MySQL bug report it sounds like we are adding a workaround for a bug with the ALTER command (which is also apparently fixed at the end of the ticket) so if I'm reading the thread correctly we are adding a fix using a bug that should have now also been fixed in MySQL, that doesn't seem right?

Unless I'm misunderstanding, the report in that MySQL ticket is that CREATE 'correctly' shows a warning that the tablespace already exists but then you can just bypass that using ALTER as it doesn't check for an `.idb` first and just overwrites it (if there was one).

This PR is just for a filter, so no big deal either way I guess, but seems a little flaky if we are going to advise people do the above. Any thoughts?

mnelson4:
Thanks for the thought and effort you're putting into this @Pebblo.

> we are adding a fix using a bug that should have now also been fixed in MySQL, that doesn't seem right?

:upside_down_face: zoiks I've been pretty confused by this issue.

So to clarify, the fallback solution is to just use MyISAM engine for the promotion and promotion_object tables. They don't currently have any strict foreign keys, so MyISAM works ok. Site owners can use this work and accompanying code snippet to do that.

The solution *is not* to first create the table using MyISAM, then alter it to InnoDb, which was the "[bug that should now have also been fixed in MySQL](https://bugs.mysql.com/bug.php?id=73225)".

So IMO, when folks get the error about a tablespace (idb file) already exists, the most correct fix is for them to remove that file and re-create the table. (Which is what you guys have been trying to suggest).

But in cases that doesn't work, a fallback would be to use this code snippet to just avoid using InnoDB altogether, and use MyIsam instead.

Does that make sense @Pebblo? Does it stop the car alarm in your head :wink:? 

(BTW I found that thread really MySQL bug report really confusing... I originally assumed the bug was that creating a table didn't automatically overwrite conflicting tablespaces, but altering a table did overwrite conflicting tablespaces was good. But it turned out the "fix" showed the opposite was true: the bug was that altering a table did overwrite conflicting tablespaces!)

Pebblo:
> Does that make sense @Pebblo? Does it stop the car alarm in your head wink?

Its does :)

> (BTW I found that thread really MySQL bug report really confusing... I originally assumed the bug was that creating a table didn't automatically overwrite conflicting tablespaces, but altering a table did overwrite conflicting tablespaces was good. But it turned out the "fix" showed the opposite was true: the bug was that altering a table did overwrite conflicting tablespaces!)

Exactly, that's why I was a bit :S with this fix, I thought we were trying to work around the file not being deleted by using MyISAM and then converting to InnoDB with ALTER.

Tested and works fine.

PR #807, opened 2018-12-04
Added logic to not adjust event subtotals for ticket scope promo line items

tn3rb:
plz see: https://github.com/eventespresso/EE4-Promotions/pull/15
In the above PR, I added some filters that facilitate the adding of ticket scope promotions, but that caused some issues with the line item display filters, so this tweak fixes things.

## How has this been tested

### DIFF
--- a/core/libraries/line_item_display/EE_SPCO_Line_Item_Display_Strategy.strategy.php
+++ b/core/libraries/line_item_display/EE_SPCO_Line_Item_Display_Strategy.strategy.php
@@ -337,9 +337,12 @@ class EE_SPCO_Line_Item_Display_Strategy implements EEI_Line_Item_Display
         $html .= EEH_HTML::td($line_item->quantity(), '', 'item_l jst-rght');
         // $total = $line_item->total() * $line_item->quantity();
         $total = $line_item->total();
-        if (
-            isset($options['event_id'], $this->_events[ $options['event_id'] ])
-            && ! ($line_item->OBJ_type() === 'Promotion' && $line_item->parent()->OBJ_ID() !== $options['event_id'])
+        if (isset($options['event_id'], $this->_events[ $options['event_id'] ])
+            && ! ($line_item->OBJ_type() === 'Promotion'
+                && ! ($line_item->parent()->OBJ_type() === 'Event'
+                    && $line_item->parent()->OBJ_ID() === $options['event_id']
+                )
+            )
         ) {
             $this->_events[ $options['event_id'] ] += $total;
         }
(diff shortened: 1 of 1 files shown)


tn3rb:
Sorry about not explaining things at all... so here's what happens normally when displaying line items somewhere:

 - line items get passed to a series of line item filter classes like:
     - `EE_Billable_Line_Item_Filter` removes line items that can not be billed for
     - `EE_Non_Zero_Line_Item_Filter` removes line items with totals or quantities of 0
 - because of this, we can not use the precalculated subtotals in the line items themselves, but need to recalculate things on the fly
 - line items for event scoped promotions currently get added as regular line items and do not affect individual ticket line item prices, so their values are considered when recalculating event subtotals
 - line items for ticket scoped promotions however directly affect the price of a ticket and therefore should not be considered when recalculating event subtotals because their effect is already applied in the form of a lower ticket price.

So now, looking at the original code in `EE_SPCO_Line_Item_Display_Strategy::_item_row()`:

```php
if (isset($options['event_id'], $this->_events[ $options['event_id'] ])) {
    $this->_events[ $options['event_id'] ] += $total;
}
```

you can see that for all regular line items, we would adjust the event subtotal

but with the new code:

```php
if (isset($options['event_id'], $this->_events[ $options['event_id'] ])
    && ! ($line_item->OBJ_type() === 'Promotion' && $line_item->parent()->OBJ_ID() !== $options['event_id'])
) {
    $this->_events[ $options['event_id'] ] += $total;
}
```

we have added a clause that does not adjust the event subtotal if the line item is for a promotion where the parent ID does not match the event ID for the corresponding event subtotal. This works because line items for ticket scoped promotions have a parent whose corresponding `OBJ_type` is equal to `Ticket` and not `Event` and again, we don't want to adjust the event subtotal because that line item has already affected the ticket price directly.

Hopefully that will all make sense, but if not don't hesitate to ask questions.

mnelson4:
hey @br3nt regarding https://github.com/eventespresso/event-espresso-core/pull/807 and https://github.com/eventespresso/EE4-Promotions/pull/15; according to my understanding, basically all line items under those with `LIN_type='line-item'` should be `LIN_type='sub-item'`- so ticket discounts, seeing how they're placed under the ticket they affect, should have `LIN_type='sub-item'`.

If we do that, https://github.com/eventespresso/event-espresso-core/pull/807 is unnecessary (including the part that makes EE core check for line items corresponding to models from an add-on, which is always nice to avoid). What do you think?

mnelson4:
I added to https://github.com/eventespresso/EE4-Promotions/pull/15 that, from what I can tell, makes this change unnecessary.

tn3rb:
killing this PR because the new filter added in https://github.com/eventespresso/EE4-Promotions/pull/15 works better

Issue #814, opened 2018-12-10
Investigate Divi conflict


joshfeck:
Please see HS thread:

https://secure.helpscout.net/conversation/727763885/596100?folderId=100570

joshfeck:
A quick recap on some troubleshooting I did:

Ee 4.9.69 + divi 3.18.3 + wp 5.0: Divi's front end editor works
EE current master + divi 3.18.3 + wp 5.0: Divi's front end editor is broken, JS errors
EE current master + divi 3.18.3 + wp 4.9.8 (no Gutenberg plugin activated) Divi's front end editor works

nerrad:
I spent a bit of time tonight looking into this.  The frustrating thing here is that without having access to the divi source, I can't really troubleshoot this effectively but here's what I've discovered:

- We extract webpack boilerplate into its own file that is then loaded as a dependency for our built files.  Since the boilerplate nearly never changes, this helps keep that cached long term in people's browsers and reduces (by about 1-2kb) the size of each of our bundles.  I discovered that divi throws up on this.  Could be because divi also builds their `bundle.js` file through webpack only its a different version of webpack maybe.  So if I switch our build process to not extract this built file divi doesn't blow up. However...
- Although divi loads, there's a whole bunch of errors in the console for things like `_` not being found (lodash symbol) and `wp` not being found (wp js global).  What's frustrating here is all these errors have divi's `bundle.js` as the source of the errors in the stack trace.  So on the surface it appears like divi is doing something in maybe reading the scripts on the page or mirroring behaviour or something but whatever its doing its clobbering other things loading on the page.  It's all conjecture though without seeing the source 😠 
- I was able to tweak our script so it JUST loads the base file and enqueues wp core's `lodash` as a dependency.  This confirms the conflict here because divi throws undefined `_` errors (which clearly does NOT come from our code, I ensured there was none of our own javascript loading).  This will be because WP core loads lodash in no conflict mode whenever its library is enqueued:

See `wp-includes/script-loader.php`
```
did_action( 'init' ) && $scripts->add_inline_script( 'lodash', 'window.lodash = _.noConflict();' );
```

So at a minimum, divi is not accounting for the fact that on the frontend the WordPress lodash library might be in use.

I suspect this might be a cause for some of the other console errors I was seeing as well.

## Next steps

* [ ] investigate namespacing the webpack boilerplate (runtime) built for our files OR (what I suspect we'll just have to do) is just have the webpack runtime in each file like normal.
[...]

PR #815, opened 2018-12-11
don’t load our blocks framework if in Divi front-end editor context

nerrad:

See #814 for background.  Essentially this pull kills any php side block registration (and thus any asset loading) when the divi frontend editor is loading.  Divi uses a request parameter (`et_fb`) when in that view so its trivial for us to just use that to kill loading.

Brent, I don't generally like putting compatibility things for other plugins directly in our code like this, but in this case I'm _hoping_ that Divi gets on the ball and fixes the conflict issues on their end.  Essentially they are not accounting for the very real possibility that other plugins/themes could start using the WP core block api for loading react, lodash etc on the frontend (along with other `wp` packages). For now, we don't really care because divi doesn't really fully support GB anyways, they currently just _bypass_ the whole thing.

## How has this been tested

* [x] tested against the issue reported in #814 with Divi loaded.
* [x] Ensured our assets still load on other non divi frontend editor routes.

## Checklist

* [ ] My code is tested.

### DIFF
--- a/core/EE_System.core.php
+++ b/core/EE_System.core.php
@@ -1327,7 +1327,8 @@ final class EE_System implements ResettableInterface
      */
     private function canLoadBlocks()
     {
-        return function_exists('register_block_type')
+        return apply_filters('FHEE__EE_System__canLoadBlocks', true)
+               && function_exists('register_block_type')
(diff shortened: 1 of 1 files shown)


joshfeck:
re-tested and it still works.

Issue #772, opened 2018-11-04
Customize Unauthorized message from PayPal


lorenzocaum:
**Summary**
With PayPal Express and PayPal Express Checkout Smart Payment Buttons, we check the credentials on save to make sure they work. If there is a problem such as entering them in the wrong order, then an event organizer will see something like this:

Error validating PayPal credentials. Unauthorized

![image 2018-11-02 at 7 40 26 am](https://user-images.githubusercontent.com/1079138/47913464-96a78880-de72-11e8-8ce9-06ef5b56cf05.png)

Could we tailor the "unauthorized" message so its more specific to the payment gateway?

That way we can provide some tips to get back on track.

Examples:

- [PayPal Express] PayPal did not accept your API credentials. Double-check your credentials and copy and paste them into the fields again and save changes. Alternatively, remove your existing credentials in PayPal.com and request new API credentials and try them.

- [PayPal Smart Buttons] PayPal did not accept your API credentials. Double-check your credentials and copy and paste them into the fields again and save changes. Alternatively, create a new app and try the new API credentials.

**Other notes**
Also see: https://github.com/eventespresso/eea-paypal-smart-buttons/issues/27

mnelson4:
PR merged

PR #820, opened 2018-12-13
improve paypal pro unauthenticatd message

mnelson4:

Added credentials check when saving PayPal Pro, and improved authenticated error message for PayPal Pr (it's now https://drive.google.com/a/eventespresso.com/file/d/1-Db0OwLUYC8ctOF7CcUjz6_9kvlmVP1i/view?usp=drivesdk). 

Addresses https://github.com/eventespresso/event-espresso-core/issues/772 

## Changes to Translated Strings

`Are you sure this is your API username, not your login username?%1$s`
changed to
`Are you sure this is your API username, not your login username? %1$s` (added a space)

`Are you sure this is your API password, not your login password.%1$s`
changed to
`Are you sure this is your API password, not your login password? %1$s` (changed a period to a question mark and added a space)

`Please verify your API signature is correct.%1$s`
changed to
`Please verify your API signature is correct. %1$s` (added a space)

I thought it might be more ok to change these translated strings because they're fairly new, only seen by event managers (not site visitors) and only seen when there is an error.

@joshfeck do you approve of these translated strings? or would you rather leave them as-is?

## How has this been tested

* [x] On master, purposefully set the wrong credentials for PayPal Pro. Then switch to this branch. Then go to make a payment. You should get a more helpful error message than before (saying "PayPal did not accept your API username, password, or signature. Please double-check these credentials and if debug mode is on.")
* [x] On this branch, try to save PayPal Pro with bad credentials. After the form submits, you should see a helpful error message saying the credentials were wrong, and some suggestions in the form about what could be wrong.
* [x] Try to save PayPal Pro with good credentials. It should work as normal. And you should be able to pay as usual.
* [x] On this branch, activate PayPal **Express** instead. Try to save it with invalid credentials. You should still get validation errors informing you of that.
* [x] Save PayPal **Express** with valid credentials and make a payment with it.

## Checklist

* [x] My code is tested.


caffeinated/payment_methods/Paypal_Pro/forms/SettingsForm.php
```diff
@@ -0,0 +1,51 @@
...
+    public function __construct(array $options_array = array(), $help_tab_link)
+    {
+        $options_array = array_replace_recursive(
+            array(
+                'extra_meta_inputs' => array(
+                    'credit_card_types' => new EE_Checkbox_Multi_Input(
+                        EE_PMT_Paypal_Pro::card_types_supported(),
+                        array(
+                            'html_label_text' => __('Card Types Supported', 'event_espresso'),
+                            'required' => true )
```
knazart: I'd wrap this parenthesis to a new line.

core/services/payment_methods/forms/PayPalSettingsForm.php
```diff
@@ -0,0 +1,182 @@
...
+                        'html_label_text' => sprintf(
+                            esc_html__('API Signature %s', 'event_espresso'),
+                            $help_tab_link
+                        ),
+                        'required'        => true,
+                    )
+                ),
+            )
+            )
+        );
```
knazart: Are we missing a second parameter (`$options_array`) for the `array_replace_recursive()` here ?

mnelson4: errrmm ya totally

mnelson4:
Thanks Nazar. I also remembered Brent asks that the classnames should be unique, even though the namespace makes that unnecessary (it's just nicer dealing with the classes in the code, eg `PayPalProSettingsForm` is more intuitive than just `SettingsForm`). That wasn't the case with the settings form used for PayPal Express, so we can't change that without breaking backward compatibility, but going forward we should do that.
Ok this is ready for another review.

knazart:
> I also remembered Brent asks that the classnames should be unique

Oh, kk, didn't hear about that. Although I was wondering about the "plain" `SettingsForm` naming in this spot myself.

Btw, I think you forgot to push 😃 I don't see the changes.

core/services/payment_methods/forms/PayPalSettingsForm.php
```diff
@@ -108,7 +115,10 @@ protected function checkForCredentialsErrors()
 
         if (empty($response_args['ACK'])) {
             EE_Error::add_error(
-                esc_html__('Your PayPal credentials could not be verified. Part of their response was missing.', 'event_espresso'),
+                esc_html__(
+                    'Your PayPal credentials could not be verified. Part of their response was missing.',
+                    'event_espresso'
+                ),
```
knazart: +1 for this. Should we do the same on lines `105`, `139` and `168` ?

mnelson4: yeah may as well tidy that up, thanks

mnelson4:
Thanks @knazart.
I'm realizing I should really also ask Brent to review, especially the new infrastructure class (`PayPalSettingsForm`)

core/services/payment_methods/forms/PayPalSettingsForm.php
```diff
@@ -0,0 +1,219 @@
...
+        );
+        if (is_wp_error($response) || empty($response['body'])) {
+            // If we got here then there was an error in this request.
+            // maybe is turned off. We don't know the credentials are invalid
+            EE_Error::add_error(
+                sprintf(
+                    // translators: %1$s Error message received from PayPal
+                    esc_html__(
+                        // @codingStandardsIgnoreStart
+                        'Your PayPal credentials could not be verified. There was an error communicating with PayPal, it was %1$s',
```
tn3rb: Depending on how the error is worded, this could end up being really confusing, so instead of:

> There was an error communicating with PayPal, it was %1$s

how about:

> The following error occurred while communicating with PayPal: %1$s

which does not have the possibility of resulting in weird looking grammar and/or sentence structure

core/services/payment_methods/forms/PayPalSettingsForm.php
```diff
@@ -0,0 +1,219 @@
...
+                        'event_espresso'
+                    ),
+                    $this->helpTabLink()
+                )
+            );
+            $this->get_input('api_password')->add_validation_error(
+                sprintf(
+                    // translators: $1$s HTML for a link to the help tab
+                    esc_html__(
+                        'Are you sure this is your API password, not your login password.%1$s',
```
tn3rb: needs a question mark and a space before the link

ie:

> 'Are you sure this is your API password and not your login password? %1$s',

core/services/payment_methods/forms/PayPalSettingsForm.php
```diff
@@ -0,0 +1,219 @@
...
+                    'If you are using PayPal Sandbox (test) credentials, Debug mode should be set to "Yes". Otherwise, if you are using live PayPal credentials, set this to "No".',
+                    // @codingStandardsIgnoreEnd
+                    'event_espresso'
+                )
+            );
+            $this->get_input('api_username')->add_validation_error(
+                sprintf(
+                    // translators: $1$s HTML for a link to the help tab
+                    esc_html__(
+                        'Are you sure this is your API username, not your login username?%1$s',
```
tn3rb: maybe:

>  'Are you sure this is your API username and not your login username? %1$s',

core/services/payment_methods/forms/PayPalSettingsForm.php
```diff
@@ -0,0 +1,219 @@
...
+                        'Are you sure this is your API password, not your login password.%1$s',
+                        'event_espresso'
+                    ),
+                    $this->helpTabLink()
+                )
+            );
+            $this->get_input('api_signature')->add_validation_error(
+                sprintf(
+                    // translators: $1$s HTML for a link to the help tab
+                    esc_html__('Please verify your API signature is correct.%1$s', 'event_espresso'),
```
tn3rb: needs a space before the link, ie:

> 'Please verify your API signature is correct. %1$s'

payment_methods/Paypal_Express/forms/SettingsForm.php
```diff
@@ -17,53 +14,22 @@
...
-
     /**
      * SettingsForm constructor.
      *
      * @param array $options_array
-     * @param string     $help_tab_link
+     * @param string $help_tab_link
+     * @throws \EventEspresso\core\exceptions\InvalidDataTypeException
+     * @throws \EventEspresso\core\exceptions\InvalidInterfaceException
+     * @throws \InvalidArgumentException
```
tn3rb: these classes should all be imported, so that you can see everything the class utilizes via the `use` statements at the top of the file

caffeinated/payment_methods/Paypal_Pro/forms/PayPalProSettingsForm.php
```diff
@@ -0,0 +1,46 @@
...
+ */
+class PayPalProSettingsForm extends PayPalSettingsForm
+{
+    /**
+     * SettingsForm constructor.
+     *
+     * @param array $options_array
+     * @param string $help_tab_link
+     */
+    public function __construct(array $options_array = array(), $help_tab_link)
```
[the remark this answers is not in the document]
mnelson4: the grandparent class, `EE_Payment_Method_Form`, has the options_array be optional, so I thought all child classes needed to likewise leave it optional, right?
See https://stackoverflow.com/questions/13423494/why-is-overriding-method-parameters-a-violation-of-strict-standards-in-php
Although when I tried it locally, I didn't get any errors. I would have expected to. Well, I'll try pushing it and see what travis says...

core/services/payment_methods/forms/PayPalSettingsForm.php
mnelson4: When I read that "and", it reads like "Are you sure this is your x and y?" (where x is "API username" and y is "not your login username"). Which reads a little strange.
What about "Are you sure this is your API username? (Not your login username)"?

core/services/payment_methods/forms/PayPalSettingsForm.php
mnelson4: ya space looks better. Same situation as before, but I don't think breaking this string is a big deal (for same reason as https://github.com/eventespresso/event-espresso-core/pull/820#discussion_r242644962).

payment_methods/Paypal_Express/forms/SettingsForm.php
mnelson4: oups, missed those. Thanks.

mnelson4:
Thanks Brent. I implemented most of your suggestions but had a reply to a few others.

caffeinated/payment_methods/Paypal_Pro/forms/PayPalProSettingsForm.php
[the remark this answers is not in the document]
tn3rb: oh sry I missed that that parameter got passed up to yet another parent, so ya keep it as is since that class is already out in the wild.

core/services/payment_methods/forms/PayPalSettingsForm.php
tn3rb: hmmm... well it would only break the existing translation if we also change that string that's in another file right?
I would think that adding the question mark here would simply add another string to the PO file, which is not necessarily desirable.

I'm okay with changing it because I favour enhancement over backwards compatibility but I'm also okay with deferring the decision to @joshfeck / the support tEEm since they're the ones that have to deal with the angry mobs.

core/services/payment_methods/forms/PayPalSettingsForm.php
tn3rb: > it reads like "Are you sure this is your x and y?"

just throw the "not" in there and it becomes:

> "Are you sure this is your x and **not** y?"

or if you really want to think like a programmer:L

> "Are you sure this is your x and **!** y?"

or what about:

```
if ( $you === 'sure' && isApiCredential($username) && ! isLoginCredential($username) ) {
```

core/services/payment_methods/forms/PayPalSettingsForm.php
tn3rb: I agree with you but if you want to let @joshfeck decide, then that's fine with me

mnelson4:
@joshfeck this ticket currently makes some changes to translated strings that you should either approve or reject. I outlined them on https://github.com/eventespresso/event-espresso-core/pull/820#issue-238525919 under "Changes to translated strings". 
If you don't approve I'm happy to put them back.

kristinoc:
@joshfeck what PayPal test credentials can i use to test the proper credentials -- after I use the "bad" credentials?

joshfeck:
@kristinoc you can ping me in slack and I'll show you where we keep that information.

kristinoc:
I entered a bad API password on master then switched. - and when I tested  a transaction I got this message 
ticket820_2

joshfeck:
I checked through all the checklist steps. This works well and is ready for merging.

PR #828, opened 2018-12-18
Removed bot trap timestamp Fixes #670

### LINKED ISSUE #670: Remove timing component of bot trap
We see quite a few support requests where they get a false positive "server timing error", like this one:
image

Since there could be reasons why a ticket selector form could be over 1 hour old (interruptions, caching, or even someone really taking their time to mull over whether they want to attend the event) I think the timing component of the bot trap should be removed. Further, the ticket selector can be protected with reCAPTCHA now as well. 

The hidden input part could probably remain.

tn3rb:
Fixes #670 

## testing

- log into the admin and go to the **Event Espresso - Registration Form** > **Reg Form Settings** tab and find the **Bot Trap Settings** section
- [x] confirm that the "Encrypt Bot Trap Data" question no longer exists and no errors appear on the screen
- change the value of "Enable Bot Trap" question and click "Save"
- [x] confirm that the settings saved and that the value "sticks" (navigate away from the page and back again - value should be the same)
- now change the value for the "Enable Bot Trap" question to "Yes" (if not set to that already) and click "Save"
- log out, switch to the frontend of the site, and register for an event 
- [x] confirm that registration process went as expected
- register for event again, but this time, prior to submitting ticket selections, use your browser's dev tools, to examine the ticket selector's html and find the email input named `tkt-slctr-request-processor-email` (it is not visible when viewing the form, assuming you are not a bot)
- enter an email address into the above input and then submit the ticket selector form
- [ ] confirm that you were not advanced to the registration form but instead receive the "Thank you so much. Your ticket selections have been received for consideration." message.
- feel free to experiment with the bot trap in this branch and attempt to find any other issues


joshfeck:
I get this error when saving the Reg Form Settings page on this branch:

```
An error has occurred:
Invalid or missing Bot Trap settings. Please refresh the form and try again.
EED_Bot_Trap - update_bot_trap_settings_form - 302
```

joshfeck:
All working! This is ready to be merged.

PR #834, opened 2018-12-19
Fix block-level elements breaking the payment method description output

joshfeck:
…ion field

Originally reported in https://events.codebasehq.com/projects/event-espresso/tickets/11094

> Reported here: https://eventespresso.com/topic/payment-method-description-breaks-when-using-tags/
> 
> The user added a h1 tag to a payment method description and it output like this - https://monosnap.com/file/yikin8KgvqKjx3ZBzQ7cIoWvUVduRa
> 
> The description is output within a p tag so none of the block level elements are valid within the payment method description although its shows they are.

## How has this been tested

This can be tested by adding a h1 tag or similar block element to the payment method description field, then go to register for a paid ticket, then select that payment method. With master, the description breaks out of the paragraph, with this branch the contents stay within the div.

## Checklist

* [x] My code is tested.


Pebblo:
Thumbs up from me!

Mering in.

PR #850, opened 2018-12-20
Fix issue where espresso_event_status() always echos even if $echo = false

joshfeck:
See https://events.codebasehq.com/projects/event-espresso/tickets/11370
for backstory.

The only place in core I could find that uses this template tag is the RSS feed template. So for testing you could check the RSS feed and make sure the event status displays the same.


joshfeck:
This is ready for testing.

For testing you could check the event RSS feed and make sure the event status displays the same as it does on master.

sethshoultes:
I switched between master and this branch and the RSS feed seems to be working as expected. I tested with a regular event, password protected event, and a private event. As you can see, with this branch selected, only the normal event and the password-protected event are available:

espresso-4-9 dev cceventsfeed 2019-02-04 08-27-14

PR #852, opened 2018-12-21
Bug/rest api taxonomy relationships regression, Fixes #845

### LINKED ISSUE #845: Fix regression of REST API Taxonomy-related routes
Originally fixed in https://events.codebasehq.com/projects/event-espresso/tickets/10083 and these appear to be broken again:

    wp-json/ee/v4.8.36/events?include=Term_Taxonomy
    wp-json/ee/v4.8.36/term_relationships
    wp-json/ee/v4.8.36/term_taxonomies?include=Term_Relationship

mnelson4:

Fixes the usage of term_relationships in REST API requests, which was broken on https://github.com/eventespresso/event-espresso-core/pull/804. 
Adds a test to avoid another regression.

## How has this been tested

Unit tested but it would be good to double-check:
* [x] send a request to `{yoursite}/wp-json/ee/v4.8.36/term_relationships`
* [x] `{your_site}`/wp-json/ee/v4.8.36/events?include=Term_Taxonomy`
* [x] and `{your_site}`/wp-json/ee/v4.8.36/term_taxonomies?include=Term_Relationship`
* [x] double-check the mobile apps load events as usual

## Checklist

* [x] My code is tested.


joshfeck:
This fixes #845

tests/testcases/core/libraries/rest_api/controllers/Read_Test.php
```diff
@@ -1072,6 +1072,42 @@ public function testHandleRequestGetAllVenues()
...
+     * regression.
+     * @since $VID:$
+     */
+    public function testHandleRequestGeAllTermRelationships()
+    {
+        $e = $this->new_model_obj_with_dependencies(
+            'Event',
+            array(
+                'status' => EEM_Event::post_status_publish
+            )
```
tn3rb: just as an FYI,
because our min php version is now at 5.4 you can use the short array syntax **if you prefer**,

ie:

```php
       $e = $this->new_model_obj_with_dependencies(
            'Event',
            ['status' => EEM_Event::post_status_publish]
        );
```

mnelson4: yes, yes I would. Good point thanks.

mnelson4:
Adjustment was only made to the test code, so I don't think this needs another review (plus the adjustment was what Brent suggested).
So I'm just going to wait for the tests to finish before passing off for testing. (Although I think Josh has already tested, still I think I should get a confirmation)

mnelson4:
Also FYI @tn3rb, the issue existed because the code incorrectly assumed each model would have a primary key, which `EEM_Term_Relationship` of course doesn't. The methods we're now using are able to use the combined keys instead of a single primary key.

joshfeck:
This is tested and ready for merging.

PR #861, opened 2018-12-26
Bug/11315/fix long titles for events breaking registraiton due to wpdb insert error

joshfeck:
For backstory, plz see:
https://events.codebasehq.com/projects/event-espresso/tickets/11351

@mnelson4 can you update this branch if need be, verify the checklist, then request a review?

## How this has been tested
* [ ] Create an event with a name longer than 245 characters. Register for the event. It should work fine (on EE master, you'd get an error right after submitting the ticket selector form).
* [ ] Disabled multibyte string support. I went into my php.ini file, changed the line that said `extension=php_mbstring.dll` to `;extension=php_mbstring.dll` (commented it out). restarted the webserver. The repeated the above test. It should still work.

## Note
We may want to instead add similar logic to `EE_Line_Item::set()` to avoid making the column's length too big, or adding it to the field classes. But doing this doesn't preclude that (and that might get bogged down in more elaborate fixes, so let's get this patched first).

## Checklist

* [ ] My code is tested.


mnelson4:
I've gone through the testing steps I added to the original PR description, but it'd probably good for someone else to also (at least the first step).

PR #864, opened 2018-12-27
Added button to Transaction admin page for recalculating line items. Closes #726

### LINKED ISSUE #726: Add button to recalculate line items within a transaction
See here: https://eventespresso.com/topic/need-to-send-revised-invoice-after-payment/

We've had a couple of requests where users have either forgotten to set a ticket as taxable or their tax rate has changed and they need to update a transaction to include/update the tax.

Brent mentioned in slack that we may be able to add a button in the edit transaction page used to recalculate the line items for the above.

**Is your feature request related to a problem? Please describe.**
If you create a paid ticket on an event and do not set the ticket to be taxable, register and pay on said ticket.

Now update the ticket in the event editor to make it taxable, other than re-registering the previous attendee onto the same ticket there's no way to update the previous registration to request the missing tax.

**Describe the solution you'd like**
Possibly a button to recalculate line items?

This seems similar to the issue mentioned here: https://github.com/eventespresso/eea-attendee-mover/issues/4

tn3rb:
plz see #726
After opening up the appropriate files and looking into the code a bit I realized that all that was needed was to copy one of the existing buttons on the transaction admin details page, then add a new route and callback for calling `EEH_Line_Item::apply_taxes()` on the Transactions total line item. This was literally only about 15 minutes of work on top of the 5-10 minutes it took to open the files and search through all of the code to determine what needed to be done. I concluded it would be more efficient to simply add the few lines of code required to do this while I already had everything open.

## testing

- log into the admin, go to the Transactions admin page, and click on any Transaction to view its details page
- [x] confirm that a new button labeled "Recalculate Taxes and Total" appears below the **Transaction Items** table
- take note of the Transaction's tax line items (or lack thereof) and grand total
- in the first column of the **Transaction Items** table, click on the Event link for one of the tickets so that the Event editor opens in another tab or window
- find the ticket in the editor and click to view its advanced settings
- click on the "This ticket is taxable" checkbox to change whatever it is currently set at and click **Save**
- return to the Transaction admin details page in the other tab or window and click on the "Recalculate Taxes and Total" button
- [x] confirm that the taxes and transaction total have been updated as expected
- experiment with taxes and transactions in this branch and attempt to find any potential issues
- plz compare any issues against master branch to determine if they are caused by the work in this branch


tn3rb:
I think that's a great idea... here it is:

recalculate taxes and txn total button

Pebblo:
Running through the steps nothing seems to happen when I click the button.

The transaction page reloads but everything is still the same.

I registered onto a ticket that wasn't set as taxable.
Paid using Stripe ($10)
When complete edited the ticket and set it as taxable, saved.
Opened up the transaction for the above and clicked the recalculate taxes button, page reloads but the transaction is the same - https://monosnap.com/file/XhucBQwpf6uGzFJId7hTEOll1Mfqkl

I thought it 'may' have been an issue is the transaction was complete or the reg approved (your screenshot shows a payment reminder button, so that transaction obviously wasn't complete) so retested on another incomplete transaction, same thing.

tn3rb:
hmmm... kinda baffled why my testing notes obviously don't work...
could have sworn that I had done things exactly as outlined and got positive results...

anyways... had to reset the "is taxable" field on the line items as well.

core/db_models/EEM_Line_Item.model.php
```diff
@@ -99,59 +93,159 @@ class EEM_Line_Item extends EEM_Base
...
+    // various line item object types
+    const OBJ_TYPE_EVENT = 'Event';
+
+    const OBJ_TYPE_PRICE = 'Price';
+
+    const OBJ_TYPE_PROMOTION = 'Promotion';
+
+    const OBJ_TYPE_TICKET = 'Ticket';
+
+    const OBJ_TYPE_TRANSACTION = 'Transaction';
```
tn3rb: here's the new constants I added

core/helpers/EEH_Line_Item.helper.php
```diff
@@ -1637,72 +1909,156 @@ public static function non_empty_line_item(EE_Line_Item $line_item)
...
+                $ticket = $ticket_line_item->ticket();
+                if ($ticket instanceof EE_Ticket && $ticket->taxable() !== $ticket_line_item->is_taxable()) {
+                    $ticket_line_item->set_is_taxable($ticket->taxable());
+                    $ticket_line_item->save();
+                }
+            }
+        }
+    }
+
+
```
tn3rb: here's the new method

admin_pages/transactions/Transactions_Admin_Page.core.php
```diff
@@ -726,14 +714,14 @@ protected function _transaction_details()
         $this->_template_args['txn_datetime']['value'] = $this->_transaction->get_i18n_datetime('TXN_timestamp');
         $this->_template_args['txn_datetime']['label'] = esc_html__('Date', 'event_espresso');
 
-        $this->_template_args['txn_status']['value'] = self::$_txn_status[ $this->_transaction->get('STS_ID') ];
+        $this->_template_args['txn_status']['value'] = self::$_txn_status[ $this->_transaction->status_ID() ];
```
tn3rb: we should not be bypassing the dedicated methods on the model objects in case they get overridden

admin_pages/transactions/Transactions_Admin_Page.core.php
```diff
@@ -1635,28 +1650,26 @@ public function apply_payments_or_refunds()
...
-                EE_Error::add_error(
-                    esc_html__(
-                        'You do not have access to apply payments or refunds to a registration.',
-                        'event_espresso'
-                    ),
-                    __FILE__,
-                    __FUNCTION__,
-                    __LINE__
-                );
-            }
```
tn3rb: unnecessary if-else block

admin_pages/transactions/Transactions_Admin_Page.core.php
```diff
@@ -2222,28 +2248,26 @@ public function delete_payment()
...
-                EE_Error::add_error(
-                    esc_html__(
-                        'You do not have access to delete a payment.',
-                        'event_espresso'
-                    ),
-                    __FILE__,
-                    __FUNCTION__,
-                    __LINE__
-                );
-            }
```
tn3rb: another unnecessary if-else block

admin_pages/transactions/Transactions_Admin_Page.core.php
```diff
@@ -2513,4 +2537,44 @@ public function get_transactions($perpage, $count = false, $view = '')
...
+     */
+    public function recalculateLineItems()
+    {
+        $TXN_ID = ! empty($this->_req_data['TXN_ID']) ? absint($this->_req_data['TXN_ID']) : false;
+        /** @var EE_Transaction $transaction */
+        $transaction = EEM_Transaction::instance()->get_one_by_ID($TXN_ID);
+        $total_line_item = $transaction->total_line_item(false);
+        $success = false;
+        if ($total_line_item instanceof EE_Line_Item) {
+            EEH_Line_Item::resetIsTaxableForTickets($total_line_item);
```
tn3rb: added call to new method here

admin_pages/transactions/Transactions_Admin_Page.core.php
nerrad: All good reasons to change.  Unfortunately, even if the dedicated methods are refactored, any refactor would still have to account for the public `get` because its public (a mistake we made in this design).  Regardless, I think changing it where possible is still worthwhile.

admin_pages/transactions/Transactions_Admin_Page.core.php
```diff
@@ -752,23 +734,21 @@ protected function _transaction_details()
...
-        } elseif ($this->_transaction->get('TXN_total') == 0) {
+        } elseif ((float) $this->_transaction->total() > (float) 0) {
+            if ((float) $this->_transaction->paid() > (float) 0) {
+                // monies owing
+                $this->_template_args['amount_due_class'] = 'txn-overview-part-payment-spn';
+            } elseif ((float) $this->_transaction->paid() === (float) 0) {
+                // no payments made yet
+                $this->_template_args['amount_due_class'] = 'txn-overview-no-payment-spn';
+            }
+        } elseif ((float) $this->_transaction->total() === (float) 0) {
```
nerrad: `EE_Transaction::total` already returns a float... why cast again?

admin_pages/transactions/Transactions_Admin_Page.core.php
```diff
@@ -752,23 +734,21 @@ protected function _transaction_details()
...
-                  && $this->_transaction->get('TXN_paid') == 0
-        ) {
-            // no payments made yet
-            $this->_template_args['amount_due_class'] = 'txn-overview-no-payment-spn';
-        } elseif ($this->_transaction->get('TXN_total') == 0) {
+        } elseif ((float) $this->_transaction->total() > (float) 0) {
+            if ((float) $this->_transaction->paid() > (float) 0) {
+                // monies owing
+                $this->_template_args['amount_due_class'] = 'txn-overview-part-payment-spn';
+            } elseif ((float) $this->_transaction->paid() === (float) 0) {
```
nerrad: `EE_Transaction::paid` already returns a float, why cast again?

admin_pages/transactions/Transactions_Admin_Page.core.php
```diff
@@ -1062,26 +1043,22 @@ public function txn_details_meta_box()
         $this->_template_args['line_item_table'] = $Line_Item_Display->display_line_item(
             $this->_transaction->total_line_item()
         );
-        $this->_template_args['REG_code'] = $this->_transaction->get_first_related('Registration')
-                                                               ->get('REG_code');
+        $this->_template_args['REG_code'] = $this->_transaction->primary_registration()->reg_code();
```
nerrad: Are you certain that the `primary_registration()` will always return the same registration as what `get_first_related` would?  On the surface it looks like you're changing the logic here (`EE_Transaction:primary_registration()` does more work under the hood).

admin_pages/transactions/Transactions_Admin_Page.core.php
```diff
@@ -1105,8 +1082,8 @@ public function txn_details_meta_box()
         }
 
         $this->_template_args['txn_details']['registration_session']['value'] = $this->_transaction
-            ->get_first_related('Registration')
-            ->get('REG_session');
+            ->primary_registration()
```
nerrad: I have the same observation here as made earlier, `primary_registration` does more logic than `get_first_related( 'Registration' )` does. That may be intentional, but I just want to make sure.  I get that this likely will be an improvement because `primary_registration()` _ensures_ we're only getting the primary registration (no matter what its order was in saving).

core/db_classes/EE_Line_Item.class.php
```diff
@@ -1414,12 +1704,13 @@ public function timestamp($raw = false)
...
      * @param string $type one of the constants on EEM_Line_Item
      * @return EE_Line_Item[]
+     * @throws EE_Error
      */
     protected function _get_descendants_of_type($type)
     {
         EE_Error::doing_it_wrong(
             'EE_Line_Item::_get_descendants_of_type()',
-            __('Method replaced with EEH_Line_Item::get_descendants_of_type()', 'event_espresso'),
+            esc_html__('Method replaced with EEH_Line_Item::get_descendants_of_type()', 'event_espresso'),
```
nerrad: The class and method name probably should not be localized.

core/db_classes/EE_Line_Item.class.php
```diff
@@ -1430,12 +1721,17 @@ protected function _get_descendants_of_type($type)
...
+     * @throws InvalidDataTypeException
+     * @throws InvalidInterfaceException
+     * @throws ReflectionException
      */
     public function get_nearest_descendant_of_type($type)
     {
         EE_Error::doing_it_wrong(
             'EE_Line_Item::get_nearest_descendant_of_type()',
-            __('Method replaced with EEH_Line_Item::get_nearest_descendant_of_type()', 'event_espresso'),
+            esc_html__('Method replaced with EEH_Line_Item::get_nearest_descendant_of_type()', 'event_espresso'),
```
nerrad: The class::method_name probably should not be localized here.

core/helpers/EEH_Line_Item.helper.php
```diff
@@ -807,7 +949,12 @@ public static function set_event_subtotal_details(
      * any old taxes are removed
      *
      * @param EE_Line_Item $total_line_item of type EEM_Line_Item::type_total
-     * @throws \EE_Error
+     * @return float|bool
```
nerrad: Is this correct? It looks to me like this only returns boolean?

core/helpers/EEH_Line_Item.helper.php
```diff
@@ -1367,35 +1601,48 @@ public static function visualize(EE_Line_Item $line_item, $indentation = 0)
...
-    {
+    public static function calculate_reg_final_prices_per_line_item(
+        EE_Line_Item $line_item,
+        $billable_ticket_quantities = array()
+    ) {
+        $running_totals = [];
         // init running grand total if not already
-        if (!isset($running_totals['total'])) {
+        if (! isset($running_totals['total'])) {
             $running_totals['total'] = 0;
```
nerrad: Not sure what the need for the `isset` checks is here.  What about instead of initializing an empty array, you do (and thus remove the `isset` checks as well:

```php
$running_totals = [
	'total' => 0,
	'taxable' => [ 'total' => 0 ]
];
```

core/helpers/EEH_Line_Item.helper.php
```diff
@@ -1620,12 +1891,16 @@ public static function non_empty_line_items(EE_Line_Item $line_item)
...
+     * @throws InvalidArgumentException
+     * @throws InvalidDataTypeException
+     * @throws InvalidInterfaceException
+     * @throws ReflectionException
      */
     public static function non_empty_line_item(EE_Line_Item $line_item)
     {
         if ($line_item->type() === EEM_Line_Item::type_line_item &&
-            $line_item->OBJ_type() === 'Ticket' &&
+            $line_item->OBJ_type() === EEM_Line_Item::OBJ_TYPE_TICKET &&
```
nerrad: heh, javascript standards for multiline conditionals have crept in here :)  For php the `&&` are supposed to go at the beginning of the line right?

core/helpers/EEH_Line_Item.helper.php
```diff
@@ -1637,72 +1912,156 @@ public static function non_empty_line_item(EE_Line_Item $line_item)
...
+     * @throws InvalidInterfaceException
+     * @throws ReflectionException
      */
     public static function get_items_subtotal(EE_Line_Item $total_line_item)
     {
-        EE_Error::doing_it_wrong('EEH_Line_Item::get_items_subtotal()', __('Method replaced with EEH_Line_Item::get_pre_tax_subtotal()', 'event_espresso'), '4.6.0');
+        EE_Error::doing_it_wrong(
+            'EEH_Line_Item::get_items_subtotal()',
+            esc_html__(
+                'Method replaced with EEH_Line_Item::get_pre_tax_subtotal()',
```
nerrad: In all these `doing_it_wrong` strings, we probably should not be translating the class and method name.  However, I understand if you decide not to make the change because it would break existing strings.

admin_pages/transactions/Transactions_Admin_Page.core.php
tn3rb: missed that, will fix

admin_pages/transactions/Transactions_Admin_Page.core.php
tn3rb: missed that, will fix

admin_pages/transactions/Transactions_Admin_Page.core.php
tn3rb: > Are you certain that the primary_registration() will always return the same registration as what get_first_related would?

no I'm not certain at all, which is why I changed it, because what we want IS the primary registrant, not just whatever one gets returned first. The `REG_code` template argument is used as a default value for the `Notes / Extra Accounting` input when you apply a payment to the transaction. Here's the help text from that input:
```
An extra field you may use for accounting purposes or simple notes. Defaults to the primary registrant's registration code.
```

admin_pages/transactions/Transactions_Admin_Page.core.php
tn3rb: Well both the primary and additional registrations can be changed, and a transaction can have registrations from multiple sessions, so it could be debated what session ID to display for the transaction, but it will still likely always make the most sense to use the session ID for the primary registrant. ie: why would you use the session ID from any other registrant?

core/db_classes/EE_Line_Item.class.php
tn3rb: missed that, although doing so will break any existing translations, which I doubt exist for this text, so will change

core/db_classes/EE_Line_Item.class.php
tn3rb: will change

core/helpers/EEH_Line_Item.helper.php
tn3rb: me likey! will change

core/helpers/EEH_Line_Item.helper.php
tn3rb: DOH!!! 🤦‍♂️

core/helpers/EEH_Line_Item.helper.php
tn3rb: no I think fixing is more important, plus, this allows eon translation to rule them all: `Method replaced with %1$s`

admin_pages/transactions/Transactions_Admin_Page.core.php
nerrad: > ie: why would you use the session ID from any other registrant?

I agree, but we're not introducing new code, we're changing old code (as flawed as it is).  So along with considering the _improvement_ this brings, I think it is also important to consider what effects this might have on code that was written against this flawed code (and might have made its own adjustments).  Again, just a sanity check.

admin_pages/transactions/Transactions_Admin_Page.core.php
tn3rb: ya we unfortunately can't get rid of the public base class method and can only correct any uses we find, and you're right that an override needs to also override the parent `get()` (ex: `EE_Registration::get()`)

Maybe one day `EE_Base_Class::get()` can be deprecated during a refactor.

admin_pages/transactions/Transactions_Admin_Page.core.php
tn3rb: As stated, this is only being used to populate a **template arg** which is only being used in one place.
This is not the same thing as a regular model object getter where "there may be code working with the existing behaviour"

admin_pages/transactions/Transactions_Admin_Page.core.php
tn3rb: there's no side effects... this is for a template arg, and there's a infinitesimally small chance at best that someone could be using this in some other way that would be affected.

Here is where the session ID is being used:
transaction session details

which is normally hidden from view and probably only ever looked at by support staff

Pebblo:
Thanks, Brent.

So this time it works in that the line items update, however, the transaction status does not change to incomplete if you've already made a payment, see: https://monosnap.com/file/FIWJi99m7jsnYtRmyduGGzOU0BRvCb

Is that expected?

tn3rb:
nope... will fix

tn3rb:
here ya go @Pebblo... sry for the back n forth

Pebblo:
Hey @tn3rb,

I'm still getting the same WRT the transaction status:

https://monosnap.com/file/OCrQFOnsaLZkdoHVZJewrhKESnrIlP

I'm on Version 4.9.78.rc.006 and I can see `$success = $transaction->update_status_based_on_total_paid();` in `Transactions_Admin_Page.core.php`
(Just to confirm I have the latest changes)

Ticket with $20 value (no tax), register and pay with Stripe.
Edit ticket and set as taxable.
Open up the transaction and hit the recalc button, tax shows but transaction remains Complete.

tn3rb:
I missed that `EE_Line_Item::recalculate_total_including_taxes()` was also recalculating the total on the transaction, which meant that the conditional in `Transactions_Admin_Page::recalculateLineItems()` that compared the transaction total before updating the status would fail because the totals already matched. This actually simplifies things quite a bit because we can now simply instruct `EE_Line_Item::recalculate_total_including_taxes()` to also update the transaction status after updating the transaction total.  I did this via a new parameter because this method is called from many many places and I did not want to affect any existing behaviour.

tn3rb:
here ya go @Pebblo... sry for the back n forth **AGAIN** 🙄

Pebblo:
All good now! thanks @tn3rb :)

PR #868, opened 2018-12-29
Update node and npm version requirements and fix travis builds

nerrad:

Travis is failing because it's installing version `6.5.0.next.0` for the builds (that's included with the latest version of node) and this fails the npm version check that happens pre-build. To fix this, I just needed to specify the npm version to install.  As a part of this, I've switched to following what WP (via gutenberg config) is setting for minimums (currently node v10, npm v 6.0.0) as this is updated by simply updating the `@wordpress/wp-scripts` package (which also exposes the check-engines script to use).

As a part of this pull I also bumped up the required npm and node version in our package.json.  There's been a number of updates since that was originally set 

## How has this been tested

* [ ] Verified locally and via travis

## Checklist

* [ ] My code is tested.


nerrad:
Ooops, `wp-scripts check-engines` is in an unreleased (next) version of the `@wordpress/scripts` package so I've reverted back to using `check-node-version` for now.  We can switch to `@wordpress/scripts check-engines` once that is released (see #869)

PR #870, opened 2018-12-29
update travis build to include php 7.3 (fixes #794)

### LINKED ISSUE #794: Prepare EE for PHP 7.3
**Describe the bug**

Our automated tests on travis are showing a number of fails in a PHP 7.3 environment. We don't currently have those fails block "success" for the entire build but with PHP 7.3 dropping soon (December) we need to get working on figuring out why our tests are failing and discover if there's any fixes we need to add to EE for php 7.3.

As a part of this work, we should probably update our travis configuration so that PHP7.3 failing builds are not skipped (for success measurement) and add a build for the next version of PHP (that is allowed to fail).

Might be a good time to do another PHP version stat measurement for our users and see if we can plan to bump our minimum version requirements again.

nerrad:

PHP 7.3 stable was released at the beginning of December.  It's time to update our travis build to include that as a non-failing build so that we're ensuring EE runs well on PHP 7.3

## How has this been tested

* [x] Verify things pass in travis
* [x] This does impact the single transaction details view in the admin.  Specifically the Attendees metabox in that view.  Ensure that all the attendees (and the relevant details loads the same as master on this branch).

## Checklist

* [x] My code is tested.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,17 +29,13 @@ before_script:
     # Install the specified version of PHPUnit depending on the PHP version:
     if [[ -n "$TRAVIS_PHP_VERSION" ]]; then
       case "$TRAVIS_PHP_VERSION" in
-        7.2|7.1|7.0|nightly)
+        7.3|7.2|7.1|7.0|nightly)
           echo "Using PHPUnit 6.1"
           composer global require "phpunit/phpunit=6.1.*"
           ;;
-        5.6|5.5|5.4|5.3)
+        5.6|5.5)
           echo "Using PHPUnit 4.8"
           composer global require "phpunit/phpunit=4.8.*"
-          ;;
-        5.2)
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -29,18 +29,19 @@ before_script:
...
+        7.3|nightly)
+          # when wp tests support phpunit 7 we'll update this.
+          echo "Using PHPUnit 6"
+          composer global require "phpunit/phpunit=6.5.*"
           ;;
-        5.6|5.5|5.4|5.3)
+        7.2|7.1|7.0)
+          echo "Using PHPUnit 6"
+          composer global require "phpunit/phpunit=6.5.*"
+          ;;
```
tn3rb: is this right? I see it in other places, but gosh it sure done look wrong

nerrad: yes its bash/shell syntax

Issue #873, opened 2019-01-03
Attendee 1 form is missing, on ee4 sandbox, when multi-attendee registration 


joshfeck:
**Describe the bug**

Attendee 1 form is missing when "Personal Information" Group is required for additional attendees & multiple tickets are registered for. 

**To Reproduce**

1. Go to https://ee4m.eventespresso.com/happenings/2019-event/ **or** set up an event that requires the Personal Information Group for Additional Attendees
2. Select 2 or more tickets, and submit TS form
3. You'll be redirected to SPCO step 1
4. See missing attendee 1 form

**Expected behavior**

Attendee 1 form included
**Screenshots**

![screen shot 2019-01-02 at 6 09 54 pm](https://user-images.githubusercontent.com/891156/50620065-4b8a1000-0ecb-11e9-8230-fce0eed50c47.png)

**Environment:**
 - Version of Event Espresso: current master
 - Version of WordPress: 5.0
 - Browser: Firefox, Chrome
 - PHP Version 7.1.25

**Additional context**

**Not** an issue on this other site, also running EE4 master:
https://josh.eventespresso.com/events/wordpress-5/

joshfeck:
Update: This is only an issue on master, not in the release (4.9.75.p)

tn3rb:
`4.9.76.rc.006` appears to still be good, but issue starts in `4.9.76.rc.007`

mnelson4:
caused by https://github.com/eventespresso/event-espresso-core/pull/798/files#diff-7ec6cba948e6194ffe136de7715e376bR186

joshfeck:
I guess #875 wasn't properly linked to this issue so it did not auto-close.

PR #875, opened 2019-01-03
instead of only using FKs to identify a duplicate, use all the columns

mnelson4:

Fixes an unreleased issue from https://github.com/eventespresso/event-espresso-core/pull/798 which caused primary attendee question groups to be disassociated with their event.
See https://github.com/eventespresso/event-espresso-core/issues/873
This fixes it by, instead of only considering a join table row a duplicate based solely on its foreign keys, also take into account other columns (in the case of event-question-group, its `EQG_primary` value). This causes the primary question group relation to not get overwritten.

**Note**: if you have an event on which you were reproducing https://github.com/eventespresso/event-espresso-core/issues/873, just switching to this branch and updating the event **won't** fix it. This just prevents the issue from occurring in the future for newly saved events. (IMO that's ok because this hasn't been released yet, so it's just us EE team who are inconvenienced- had it been released we'd want to retroactively fix those events)

## How has this been tested

* [ ] 1. Set up an event that requires the Personal Information Group for Additional Attendees
2. Select 2 or more tickets, and submit TS form
3. You'll be redirected to SPCO step 1
4. See missing attendee 1 form

## Checklist

* [ ] My code is tested.

### DIFF
--- a/core/db_models/relations/EE_HABTM_Relation.php
+++ b/core/db_models/relations/EE_HABTM_Relation.php
@@ -183,11 +183,6 @@ class EE_HABTM_Relation extends EE_Model_Relation_Base
         // the other tables) use the joining model directly!
         if (! $existing_entry_in_join_table) {
             $this->get_join_model()->insert($all_fields);
-        } else {
-            $this->get_join_model()->update(
-                $all_fields,
-                array($foreign_keys)
-            );
         }
         return $other_model_obj;
     }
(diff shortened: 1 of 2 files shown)


joshfeck:
Original issue is fixed, but now we're seeing duplicated entries in the esp_event_question_group table added each time the event is saved.

screen shot 2019-01-03 at 5 30 01 pm

PR #884, opened 2019-01-10
tweaked batch job error message to mention they can refresh the page …

mnelson4:
…to retry

Adds helpful text in case there was an error during a batch job. It says to just refresh the page. Created as a quick response to https://github.com/eventespresso/eventsmart.com-website/issues/361

This issue can be experienced by all EE users, not just ES users.

## How has this been tested

* [ ] open `core/libraries/batch/JobHandlers/RegistrationsReport.php` and edit `continue_job` by adding a bunch of random characters to the start of it, after the opening `{`. 
* Then create a registration report. 
* You should get an error once the report starts. It should now instruct you to refresh the page to try again. 
* Undo you local changes.
* Refresh the page, the report should generate as normal now.

## Checklist

* [ ] My code is tested.


mnelson4:
Seeing how this is just a text change, asking for @lorenzocaum's feedback.

lorenzocaum:
Hi Mike, where do I add the extra characters? I tried in this spot and saw a fatal:

image 2019-01-15 at 7 34 51 am

mnelson4:
Ya that’s thd right spot, but I think I was wrong in saying to add “whatever”. Try this `$foo->bar()`.
Also, I’m realizing now you should only save this change AFTER you started the registrations report but BEFORE it finishes. This is easiest if you have a report with hundreds of registrations.
With that, the migration script should start but have an error during it, and give you the new error message that says you can refresh the page to continue.

lorenzocaum:
Hi Mike, I tried those steps and I see undefined for all registrations report and filtered registrations report.

I started the registration report and once I saw it get two batches of 50, then I applied the change and then it showed undefined:

image 2019-01-15 at 1 52 55 pm

image 2019-01-15 at 1 53 04 pm

mnelson4:
Huh I'm not sure how you're getting that, I'm just retrying it and still getting the expected result (I thought perhaps WP_DEBUG was a player, but I get the same behaviour regardless of its value).

Are you testing on a web-accessible server @lorenzocaum ?

lorenzocaum:
Hey Mike, yes it is a live site. I reached out through Slack about this pull request.

mnelson4:
Aha thanks. I see there is a registration session expiry notice that's overriding the EE batch job message.
I'd suggest creating a separate issue for this, because you'd experience this issue on master branch too, right? This might be a general issue with WPEngine...

But back to this ticket, I'd suggest testing somewhere where on master you see the old batch job error text (instead of "null" like you see on this WPEngine site).

lorenzocaum:
Hi Mike, I tried on the playground site and that worked!

image 2019-01-15 at 4 16 23 pm

On the registration session expiry notice that's overriding the EE batch job message, I can create a new issue. 

Could you explain for me how that happens so I can include those notes in the issue?

mnelson4:
hey @lorenzocaum thanks for being willing to create the ticket, but I realized I should probably make it instead directly instead of requiring you to be a middleman. See https://github.com/eventespresso/event-espresso-core/issues/895

mnelson4:
@joshfeck are you ok with this change to a translated string? Or should we put the new text into a translated string? Or leave as-is (although Garth, iirc, requested somehow improving the UX when there is an error when the batch system has an error. https://github.com/eventespresso/eventsmart.com-website/issues/361 was related.)

joshfeck:
I'm OK with changing the string because it's an admin-only facing text string.

PR #886, opened 2019-01-11
Fix npm version installed in travis environment and switch to wp-scripts check-engines

nerrad:

Travis builds were failing again because the `next` version was getting installed which resulted in fails for the npm version checks.  So:

- switched to `npm install npm --latest-npm` which should make sure we're just installing the latest production release (not the "next" version)
- updated `@wordpress/scripts` to the latest version so we can take advantage of `wp-scripts check-engines` for node and npm version checks which in turn means we now inherit whatever the minimum requirements are set for the WP project.

## How has this been tested

* [x] verify that travis builds for javascript pass.

## Checklist

* [x] My code is tested.


nerrad:
Force merging even without a complete travis build because the job that mattered (js tests) passed.  No review needed because this only affected development tests.

PR #896, opened 2019-01-16
Loosen up restrictions on error reporting when invisible reCAPTCHA fails

tn3rb:
plz see: https://github.com/eventespresso/eventsmart.com-website/issues/303

We're getting sporadic reCAPTCHA fails on specific sites but there is not enough information to know what's going on. 
This branch adds an additional error message and loosens restrictions on those returned from Google so that we can get a better idea of what's actually happening. Right now it's impossible to know exactly where the error is occuring and all we do know is that Google is returning false for the reCAPTCHA token verification.

I've also updated our reCAPTCHA module logic to use the new `Request` class instead of the old `EE_Request` object.


tn3rb:
### testing

- go to the **EE - Registration Form** > **Reg Form Settings** admin page tab
- set **Use reCAPTCHA** to `Yes`, enter valid keys, check off `Ticket Selector` for Protection,  and then click Save
- take note of the option set for **Invisible reCAPTCHA Badge Position**
- log out of the admin and go to an event with available tickets
- [ ] confirm that the reCAPTCHA badge appears on the screen in its designated position
- select some tickets and attempt to proceed to registration
- if a reCAPTCHA modal appears then attempt to solve it fellow human being who is totally not a robot
- [ ] confirm that you were able to proceed to registration
- experiment with reCAPTCHA in this branch and attempt to find any other potential issues
- plz compare any issues against master branch to determine if they are caused by the work in this branch

joshfeck:
seems fine after testing. This can be merged.

PR #897, opened 2019-01-16
imrprove recipient (context) links in message template list table (fixes #725)

### LINKED ISSUE #725: Review links on the Message templates screen
**Summary**
This is for another pain point that I see and thats editing a message template and then not seeing the changes come through. This is likely because another context was edited. For example, they may have wanted to edit the message to the attendee but they actually edited the message to the event admin.

image 2018-10-11 at 4 08 14 pm

This needs more discussion but I think dropping the links on "Email" and "Edit" would help because that would leave just a couple of options in most cases:

- Edit the event admin
- Edit the registrant

**Other notes**
Also see #723 and #724

nerrad:

See #725 for background.  This pull implements the suggested changes there to the context links in the Messenger column of the Message Template List Table in the admin.

Result:

screenshot-of-changes

## How has this been tested

* [ ] Verified the links display as expected on the Default Message Templates tab of the messages admin (verified links worked)
* [ ] Same as above for the Custom Message Templates tab.

## Checklist

* [ ] My code is tested.

### DIFF
--- a/admin_pages/messages/Messages_Template_List_Table.class.php
+++ b/admin_pages/messages/Messages_Template_List_Table.class.php
@@ -171,15 +171,13 @@ class Messages_Template_List_Table extends EE_Admin_List_Table
     {
         // Return the name contents
         return sprintf(
-            '%1$s <span style="color:silver">(id:%2$s)</span><br />%3$s%4$s',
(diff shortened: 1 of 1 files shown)


mnelson4:
well I gave it a whirl, looked over the code, and checked there weren't any warnings, so this looks good.

PR #914, opened 2019-01-24
Document default where conditions for REST API

mnelson4:

Just adds documentation regarding `default_where_conditions` for the REST API, and expands a little on its value `full_this_minimum_others` when used directly from the models.
Also fixed some spots `$VID:$` was left in the docs. I thought these would be automatically changed to the released version as part of the build process, but that wasn't the case.

## How has this been tested

No testing necessary as this only affects documentation files which aren't released.

## Checklist


docs/C--REST-API/ee4-rest-api-GET-filtering-results.md
```diff
@@ -239,8 +239,82 @@ https://demoee.org/wp-json/ee/v4.8.36/events?caps=read_admin
...
+
+```php
+https://demoee.org/wp-json/ee/v4.8.36/datetimes?default_where_conditions=minimum&caps=read_admin
+```
+will *not* have default datetime where conditions added (so both deleted and non-deleted datetimes will be returned).
+
+```php
+https://demoee.org/wp-json/ee/v4.8.36/datetimes?where[Ticket.TKT_price][<]=10&caps=read_admin
+```
+will have default datetime and ticket where conditions added (so only non-deleted tickets for non-deleted tickets will be returned).
```
nerrad: something seems off about the grammar here: "non-deleted tickets for non-deleted tickets"

mnelson4: ah thanks- ya I meant to say "non-deleted **datetimes** for non-deleted tickets"

PR #919, opened 2019-01-25
Bug/add instance checks reg admin pages to avoid fatal errors due to orphaned data

joshfeck:

I tested out what happens if I delete a user account where the user account was the author of an event. Which deletes the event from the _posts table, leaving orphaned data in need of an event. I found a few fatal errors when viewing the registrations for the event. 

This PR adds a few additional checks to avoid fatal errors in this circumstance.

## How has this been tested

- [x] Create a temporary user account and set an event with **multi-attendee** registrations to be the author of that event 
- [x] Take note of the event ID
- [x] Go to Event Espresso > Registrations > Event Check-in and select the event
- [x] In another browser window or tab Delete the user in WP > Users, do not attribute content to that user
- [x] Refresh the Event Check In page, there should be no fatal errors
- [x] Click through to one of the multi-attendee transactions for that event to navigate to the single reg page. 
- [x] There should be no fatal errors in the "Other Registrations in this Transaction" box.

## Checklist

* [x] My code is tested.


caffeinated/admin/extend/registrations/Extend_Registrations_Admin_Page.core.php
```diff
@@ -1156,16 +1156,16 @@ protected function _event_registrations_list_table()
         );
         $this->_template_args['after_list_table'] = $this->_display_legend($legend_items);
         $event_id = isset($this->_req_data['event_id']) ? $this->_req_data['event_id'] : null;
-        $this->_template_args['before_list_table'] = ! empty($event_id)
+        /** @var EE_Event $event */
+        $event = EEM_Event::instance()->get_one_by_ID($event_id);
+        $this->_template_args['before_list_table'] = ! empty($event_id) && $event instanceof EE_Event
```
nerrad: You can simplify this here:

```suggestion
        $this->_template_args['before_list_table'] = $event instanceof EE_Event
```

Pebblo:
Works fine! Merging.

PR #928, opened 2019-01-31
Improve a11y for row actions on EE list tables (fixes #925)

### LINKED ISSUE #925: Improve a11y for actions in list tables.
## Issue Overview

As a part of #922, @joshfeck requested taking a look at what sort of a11y improvements we could make to the actions in list tables.  As I noted there:

* [ ]  WP removes the `title` attribute and adds `aria-label="Trash"` etc in its list tables for the actions.  This would require updating our `EE_Admin_List` parent to do the same which means _all_ our list tables will be affected (could be a good thing)?
- we're not overriding any of the css for the `.row-actions` provided by WP here so any accessiblity things.  For screen-readers the actions are actually accessible because the element is in the dom (just positioned off screen for visibility).  There's some js that makes it "visible" by changing the positioning when hovered.

We decided to break this work into its own issue/pull because it increases the scope of work and testing needed.

This will be good work to do because it will improve the accessibility of these actions.

nerrad:

This pull request uses `aria-label` instead of `title` for all the EE list table actions.

## How has this been tested

* [ ] Verify the actions for list tables still work as expected.

## Checklist

* [ ] My code is tested.


nerrad:
@joshfeck it turns out that each EE List table would have to be converted over to use aria-label instead of title.  I pushed up an example for the `Events_Admin_List_Table` that can be used as a guide for how the conversion works.  I wonder if it'd be more efficient to have another team member work on this so I can focus on some of the more difficult coding tickets?  I'm totally fine with continuing on with this if desired, just looking to optimize time here.

So the steps to completing the work would involve:

- open one of the List Table classes in EE core (check the folders in `admin_pages` and `caffeinated`.  List tables will have `List_Table` in the file name.
- Search for a call to `row_actions(` and that will give an indicator of where to go for the edits.
- Once you've found the links that are being used for actions, modify them similarly to how I did for the Event List table: change `title` to `aria-label`; add the name of the entity along with the action (choose wording that would make sense for a screen reader picking out); make sure that the value is wrapped in `esc_attr`; make sure you use a `/* Translators: info for translators*/` comment if needed.

I'll assign to you for now. Just to be clear, I'm not trying to push off work I don't want to do - I just am trying to be aware of where my time could be best spent.

joshfeck:
- [x] Events_Admin_List_Table.class.php
- [x] Venues_Admin_List_Table.class.php
- [ ] Event_Categories_Admin_List_Table.class.php
- [ ] Venue_Categories_Admin_List_Table.class.php
- [ ] EE_Admin_Transactions_List_Table.class
- [ ] EE_Registrations_List_Table.class
- [ ] Payment_Log_Admin_List_Table.class

joshfeck:
This fixes #925

joshfeck:
The admin tables for the Default Prices are getting the a11y treatment in #1039.

PR #929, opened 2019-02-01
Update EventAttendees block with new default for limit and number input control (Fixes #799)

### LINKED ISSUE #799: Attendee Block's "Number of Attendees to Display:" slider control could be perceived to only allow displaying a max of 100 attendees
Events will have more than 100 attendees, and the slider control stops at 100.

You **can** type in a higher limit if you ignore the slider control and input numbers directly into the text field. I wonder if the option would be easier to use if the slider was removed or replaced? A slider that stops at 100 would seem to be best used for controlling percentage based option values (like setting a width or scaling an image).

Also, I think the default value of 10 is a bit low (the shortcode also defaulted to 10 and this led to more than a few "My attendee list is missing names!" support requests). Could the default limit be 100?

nerrad:

See #799 for details on what prompted this work.  This pull:

- Makes the slider in the `QueryLimit` control optional via a new `withSlider` prop (defaults to current behaviour).
- Ups the default for the limit on the event attendees block to `100` from `10`.
- Only shows number input for limit control (via the modifications done for `QueryLimit`.
- removes the max of 100 on the number input (which means the default of 1000 set by the `QueryLimit` component is added - I think that's sufficient).
- add some optimization so that the attendees component remembers the highest limit set for the set of query parameters and doesn't do an additional query if a lower limit is selected for those selected parameters.  This prevents unnecessary network requests (and makes things feel snappier).
- Updates the test snapshots for expected change on tested render.
- Updates docs for the `<QueryLimit/>` component.

## How has this been tested

* [x] Ensure existing unit-tests pass
* [x] Verify adding a new event attendees block works as expected.
* [x] Verify existing event attendees blocks (created before this branch) are still working as expected.

## Checklist

* [x] My code is tested.


assets/src/blocks/event-attendees/edit.js
```diff
@@ -504,8 +506,18 @@ export default withSelect( ( select, ownProps ) => {
 		status = defaultProps.status,
 		orderBy = defaultProps.orderBy,
 		order = defaultProps.order,
+		limit = defaultProps.limit,
 	} = ownProps.attributes;
 
+	// This ensures that we don't query unnecessarily since if the limit is
+	// lower than a query we've already done, then we already have cached data
+	// for this limit (and cache is still busted by any other query changes)
```
tn3rb: clever 👍

PR #937, opened 2019-02-04
Include seconds in timestamp for check-in list table view (fixes #885)

### LINKED ISSUE #885: Include hour, minute, and second in check-in timestamps displayed
Requested by Garth:

Will make it easier to calculate the period of time between scans/check (in/outs). Plus we can more easily understand if a check in/out event is truly happening at a different time.
pasted_image_at_2016_04_15_10_18_am

nerrad:

See #885 for original report.  This pull results in this:

screenshot

Note: The date format will not change from what is set as the default format for the site.  However, the time format is now `H:m:s a`.  Ultimately, the time format was always simply what was the format set for the site so if the site had seconds in it's time format then that's what would get used here.  This _overrides_ the site time format for this view.

## How has this been tested

* [x] Verify the individual check-in records view in the check-in list table now has seconds included with the timestamp.
* [x] Verify the date format (not time format) is still following what is set for the site.

## Checklist

* [x] My code is tested.


joshfeck:
verified

PR #938, opened 2019-02-04
Add filter for swapping GRP_ID in on the fly for messages being generated (fixed #927)

### LINKED ISSUE #927: Filter the GRP_ID used when generating messages
See: https://eventespresso.com/topic/email-template-location/

Looks like this could be useful for various features, I mentioned this in Slack and Darren checked to see if it was possible:

> no, there’s _probably_ a way but if I haven’t found anything yet it’d probably end up being a very hacky way. So @tony this may be something that is relatively easy to add a filter for (but would require the implementor has a bit of php knowledge to know how to use the filter correctly).  In `EE_Messages_Generator::create_and_add_message_to_queue`, we could add a filter for setting the `GRP_ID` on the message.  Presumably the implementor would be able to get what they need from the request to know whether to serve the french template or not. We could expose to the filter callback all the data available there for what kind of message etc is being queued for generation.
If you want to create an issue for that it can be added to the queue for doing.  I estimate an effort of `1` to do it.

nerrad:
See #927 for background.

This pull adds a filter that allows for swapping in the `GRP_ID` of a message template on the fly when messages are being _generated_.  As with most filters, it needs to be used with care because it's technically possible for people to bork their messages being generated if they aren't using this properly.  However it is a powerful option for people who _do_ know what they are doing.

I've exposed as much data as I reasonably can on this filter.  In the hands of a skilled developer they should be able to accomplish some nifty things with it (including language specific templates for the message being generated).

## How has this been tested

I added the following filter to a site-specific plugin to verify that filter works as expected. You'd have to substitute `86` for whatever the GRP_ID is for the custom template you want to swap in.  Keep in mind you need to swap in a correct template for handling that message type and messenger:

```php
add_filter(
    'FHEE__EE_Messages_Generator__create_and_add_message_to_queue_GRP_ID',
    function ($GRP_ID, EE_Message $message) {
        return $message->message_type() === 'registration' && $message->messenger() === 'email' ? 86 : $GRP_ID;
    },
    10,
    2
);
```

Basically the filter above simply returns a specific Custom Message Template GRP_ID (86) whenever the message template being generated is for a template belonging to the `registration` message type and `email` messenger.

> **Note:** In the context of message template generation (it varies), the `GRP_ID` for the template to generate with is not know this early in the process as it's not derived until later.  For now, this means its a bit more complicated to swap out templates matching _specific_ message template groups but the refactoring needed to expose the GROUP ID here would involve potentially significantly more work.  The filter signature is sufficient though that if we do refactor this down the road, the filter signature should remain the same.
> 
> I chose this route as well because that way the filtered group id is actually saved with the message queued for generation in the db.

For testing I recommend using some variation of the filter implementation above and then try a test registration (or whatever triggers a message generation) and verify the filter works as expected.

## Checklist

* [x] My code is tested.


core/libraries/messages/EE_Messages_Generator.lib.php
```diff
@@ -961,8 +960,17 @@ public function create_and_add_message_to_queue(EE_Message_To_Generate $message_
...
-        if ($GRP_ID = EE_Registry::instance()->REQ->get('GRP_ID')) {
+        $GRP_ID = EE_Registry::instance()->REQ->get('GRP_ID', 0) > 0
+            ? EE_Registry::instance()->REQ->get('GRP_ID')
+            : apply_filters(
+                'FHEE__EE_Messages_Generator__create_and_add_message_to_queue_GRP_ID',
+                0,
+                $message,
+                $message_to_generate,
+                $test_send
+            );
```
tn3rb: couple things:

 - this should be converted over to use the new Request class
 - you're potentially calling that `get()` method twice
 - you're **only** filtering the `GRP_ID` if one isn't already set??!?! is that correct? ie: why not just filter the `GRP_ID` regardless of its value and let the callback decide?

imho this would be better if it were simply:

```php
/** @var \EventEspresso\core\services\request\RequestInterface $request */
$request = LoaderFactory::getLoader()->getShared('EventEspresso\core\services\request\RequestInterface');
$GRP_ID = apply_filters(
    'FHEE__EE_Messages_Generator__create_and_add_message_to_queue_GRP_ID',
    $request->getRequestParam('GRP_ID', 0),
    $message,
    $message_to_generate,
    $test_send
);
```

then any callbacks hooking into that can decide whether to filter the `GRP_ID` even if it isn't `0`, but even if you don't want to do that, at least change this to be:


```php
/** @var \EventEspresso\core\services\request\RequestInterface $request */
$request = LoaderFactory::getLoader()->getShared('EventEspresso\core\services\request\RequestInterface');
$GRP_ID = $request->getRequestParam('GRP_ID', 0);
$GRP_ID = $GRP_ID > 0
    ? $GRP_ID
    : apply_filters(
        'FHEE__EE_Messages_Generator__create_and_add_message_to_queue_GRP_ID',
        0,
        $message,
        $message_to_generate,
        $test_send
    );
```

core/libraries/messages/EE_Messages_Generator.lib.php
```diff
@@ -252,7 +254,7 @@ protected function _reset_current_properties()
      * @throws EE_Error
      * @throws InvalidArgumentException
      * @throws \EventEspresso\core\exceptions\InvalidDataTypeException
```
tn3rb: why not import this class too?

core/libraries/messages/EE_Messages_Generator.lib.php
nerrad: The original intent was to have any GRP_ID set via the request be explicit because that generally is already an explicit template request.  The filter would only be for when there is no GRP_ID in the request to allow _setting_ an explicit GRP_ID.  Normally, at this point in the stack trace the only time there is a GRP_ID is when it's set via the request.

With that said, after thinking about it, I think it'll be more flexible if we actually filter everything because that way consuming code _can_ override the `GRP_ID` from the request if desired.  It'd still be possible for the client code to derive whether the GRP_ID comes from the request and leave it alone if desired.

I also set the default to be the group id from the `EE_Message` instance on the off chance in the future we start setting `EE_Message::GRP_ID` earlier.  Currently it'll just return `0` if its not set so effectively the same as what I had before.

core/libraries/messages/EE_Messages_Generator.lib.php
nerrad: Ya I think this was automatically added by my IDE and I didn't notice it.  I decided to do a review of the entire file for lint (warning) suggestions.

joshfeck:
Tested and verified, this is ready for merging.

PR #939, opened 2019-02-05
Disable the file exists check for image URLs in the admin

Pebblo:
We get so many issues with this file exists check that I've just completely disabled it for now.

If this needs further investigation for a better/proper fix can we please just put this pull request through to disable the check until we get time to do so as we keep running into the same old validation check error in General Settings and/or Payment Methods.

If someone wants to break there images by adding URL's that aren't valid lets roll with that for now.

## Testing Checklist
* [x] Do a fresh install of EE (or just reset your current EE via the maintenance page). The payment method's button images should be set properly as usual.
* [x] Purposefully set a payment method's image to something that doesn't exist. It won't be automatically fixed. (Because we can't rely on the server to correctly identify when a URL is valid or not, and so we usually shouldn't automatically fix ostensibly "broken" image URLs)
* [x] Purposefully set a payment method's image to a blank string. It should get reset to the default button URL again.
* [x] Go to general settings and purposefully set the logo image to something non-existent. It should be saved without a problem.
* [x] Delete all EE4 data. Install EE4 **decaf**. Notice the payment method button URLs point to files inside `event-espresso-decaf`. Then deactivate decaff and delete it. Activate caffeinated. Visit the payment methods page. The button URLs should have automatically been updated to the EE caff folder (it, not be broken)

## Checklist

* [ ] My code is tested.

### DIFF
--- a/admin_pages/payments/Payments_Admin_Page.core.php
+++ b/admin_pages/payments/Payments_Admin_Page.core.php
@@ -258,7 +258,6 @@ class Payments_Admin_Page extends EE_Admin_Page
          * recheck here.
          */
         EE_Registry::instance()->load_lib('Payment_Method_Manager');
-        EEM_Payment_Method::instance()->verify_button_urls();
         // setup tabs, one for each payment method type
         $tabs = array();
         $payment_methods = array();
--- a/admin_pages/payments/Payments_Admin_Page.core.php
+++ b/admin_pages/payments/Payments_Admin_Page.core.php
@@ -258,6 +258,7 @@ class Payments_Admin_Page extends EE_Admin_Page
          * recheck here.
          */
         EE_Registry::instance()->load_lib('Payment_Method_Manager');
+        EEM_Payment_Method::instance()->verify_button_urls();
         // setup tabs, one for each payment method type
         $tabs = array();
         $payment_methods = array();
--- a/core/db_models/EEM_Payment_Method.model.php
+++ b/core/db_models/EEM_Payment_Method.model.php
@@ -264,9 +264,15 @@ class EEM_Payment_Method extends EEM_Base
         $payment_methods = is_array($payment_methods) ? $payment_methods : $this->get_all_active(EEM_Payment_Method::scope_cart);
         foreach ($payment_methods as $payment_method) {
             try {
-                // If there is really no button URL at all, reset it to the default.
+                // If there is really no button URL at all, or if the button URLs still point to decaf folder even
+                // though this is a caffeinated install, reset it to the default.
                 $current_button_url = $payment_method->button_url();
-                if (empty($current_button_url)) {
+                if (empty($current_button_url)
+                || (
+                        strpos($current_button_url, 'decaf') !== false
+                        && strpos($payment_method->type_obj()->default_button_url(), 'decaf') === false
+                    )
+                ) {
                     $payment_method->save(
                         [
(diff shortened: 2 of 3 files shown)


joshfeck:
Hey Mike, 

This probably needs further review since your changes haven't been reviewed, right?

mnelson4:
Well usually the PR creator would review the changes from the reviewer. But if you'd like we can ask Brent or Darren to review

mnelson4:
> There was one feature that worked quite well for those upgrading from decaf to premium, and that feature was still there with Tony's commit. It's gone with the additional commits that Mike added.

Sorry @joshfeck I'm not sure what feature you're referring to.

My additional commits make it so if you enter an invalid payment method button URL (or a valid one which the server for some reason incorrectly said was invalid), it stays as-is. Whereas before they would get reset to the payment method's default. To my knowledge that should be the extent of what I added. Is there something else my commits affected?

joshfeck:
@mnelson4 with just Tony's commit, the validation that prevented the form from saving was removed.

Your commits removed the feature that was useful for those upgrading from decaf to caffeinated. Can we revert back to Tony's commit only?

mnelson4:
> That feature being the decaf image path would be corrected to premium image path for payment icons

ooooh I see. So on decaf the image path would be `https://mysite.com/wp-content/plugins/event-espresso-decaf/...image.png` and then after upgrading to caf it became `https://mysite.com/wp-content/plugins/event-espresso/...image.png` right?

Ya great catch, my changes would make it so the image path *would not* get updated.

But again, the problem with just Tony's commit was that if someone entered a valid image URL for their payment method, but the server wasn't letting the HTTP ping to verify that image was valid, it would *automatically* reset the image back to the default, whereas we want it to stay how the admin entered it. (Eg https://drive.google.com/file/d/1zIqOsWZMy-Tb8V7jVypGTbw0--4_S6P8/view shows that when I purposefully enter a URL that it doesn't think is valid, it gets automatically changed. That's a problem if the server only *thought* the URL was invalid, which where people are having these problems). Does that problem make sense and seem valid?

So it sounds like on upgrading to caf, we'd like to proactively fix broken image URLs (like before) but when saving an individual payment method, we want to leave them exactly how the admin entered them. Is that right @joshfeck?

joshfeck:
Yes. If possible we want the image URLs for payment method buttons to get proactively fixed when updating from decaf to caffeinated. Maybe check the path for 'decaf' and make the correction if decaf isn't activated?

mnelson4:
> Maybe check the path for 'decaf' and make the correction if decaf isn't activated?

The above commit checks if the button URLs are for decaf, whereas the current defaults are for caff. In that case, assume they've updated from decaf to caff and so the payment method button URLs need to be updated.

Doing this instead of just reverting to Tony's original commit because that commit by itself fails on the second test in the testing checklist (I also most recently explained the issue on comments https://github.com/eventespresso/event-espresso-core/pull/939#issuecomment-461117559 and https://github.com/eventespresso/event-espresso-core/pull/939#issuecomment-460894192)

Pebblo:
@mnelson4 Nice! I like it.

Catches current decaf installs with the default or custom images and updates from decaf.

mnelson4:
passing off for testing again. I believe this checks all the boxes (you can purposefully set a payment method button image to be something considered invalid; but when you upgrade from decaf to caf the default button images don't get broken)

joshfeck:
Tested. Ready for merging. Thanks for making this happen everyone!

PR #941, opened 2019-02-05
Event Editor Dates and Tickets Refactor (WIP)

tn3rb:
- edit an event
- scroll down to see the new "Available Tickets" metabox
- click on the Price Calculator for one of the tickets (you may need to adjust filters to see a ticket)
- code can be found in `/assets/src/editor/events/tickets/editor-ticket/price-calculator/ticket-price-calculator-form.js` and `/assets/src/editor/events/tickets/editor-ticket/price-calculator/ticket-price-calculator-form-modal.js` 
- `TicketPriceCalculatorForm` is wrapped with `withFormHandler` (`/assets/src/components/form/base/with-form-handler.js`)
- `TicketPriceCalculatorFormModal`  is wrapped with `withEditorModal` (`/assets/src/higher-order-components/editor-modal/index.js`) and now also by the `withSelect` HOC which is retrieving the ticket and related prices
- as stated in chat, eventually I will be moving by data "subscriptions" (connections to the store) to the top level of my app (`/assets/src/admin-refactor/index.js`)
- there's a lot of commented out garbage in `TicketPriceCalculatorFormModal` from my attempts to retrieve data asynchronously, but basically, it has a `loadHandler()` callback that is used by `withFormHandler` to provide form data
- I want to work things so that that function only returns data when it has been correctly resolved because otherwise it doesn't work well with the existing form framework, primarily because `withFormHandler` is unable to set its loading state properly
- it doesn't really make sense to me why something like `select( 'eventespresso/core' ).getTicketById( ownProps.ticketId )` appears to repeatedly return `null` or `undefined`, and would make more sense imho if it just did nothing until data was either resolved or an error was generated, although that is likely beyond the scope of what you have done and a design decision made in WP core


nerrad:
As a general observation, the changeset on this branch is quite a bit concerning.  The _only_ changes that should show up in the changes are things specific to work done in this branch but I'm seeing a number of changes that are in files that don't look like work specific to this branch (but potentially regress work on master), or new files that are overwriting deleted files on master (i.e `blocks/widgets/event-attendees`... that was renamed in the GB master branch at one point).

I'm not quite sure how this branch got in this state but I'd be _very_ hesitant about merging a variation of this branch with all these suspect changes.  I realize this pull isn't intended to be in a merge-able state (it's just for the purpose of deciding a direction for the work you're doing), but something you may want to consider is branching a new branch off master and just cherrypicking in work from this branch so you have a cleaner/more accurate history and can have more confidence for eventual merge to master in the future.

tn3rb:
> do you think that is an indicator that components are being re-rendered for every time you see `null` or `undefined` returned?

ya it appears that way but I don't know for certain that it is because of the `withSelect` component or whether it is with some other code.
normally I only see things render once or twice when loading, but `TicketPriceCalculatorForm` is rendering multiple multiple times, mostly with the ticket and prices props showing as `undefined`.

> I'm not quite sure how this branch got in this state but I'd be _very_ hesitant about merging a variation of this branch with all these suspect changes. I realize this pull isn't intended to be in a merge-able state (it's just for the purpose of deciding a direction for the work you're doing), but something you may want to consider is branching a new branch off master and just cherrypicking in work from this branch so you have a cleaner/more accurate history and can have more confidence for eventual merge to master in the future.

ya there's no way I'll merge this into master as is, but instead of cherry picking, I'll likely cut a new branch off master and then just copy over the new files and manually edit the other few files that need changing. I don't need the commit history for a new feature.

What's potentially more disturbing, as I wrote in chat, is that the changes were brought into this branch (at least some of them) on a commit labeled "merge from master", so it's possible that those changes **were** in master at some point but have since been overwritten by a bad merge.

need to do some more investigating

nerrad:
> normally I only see things render once or twice when loading, but `TicketPriceCalculatorForm` is rendering multiple multiple times, mostly with the ticket and prices props showing as `undefined`.

Just to be clear, when you say "is rendering multiple multiple times" are you determining that from console.log in the `withSelect` or via some other means?  Keep in mind that seeing the response from the selects _does not automatically mean_ that the components are re-rendered.  It's expected that the selects will repeatedly get called but not every response will result in a re-render if there's no change in the result.

> I'll likely cut a new branch off master and then just copy over the new files and manually edit the other few files that need changing.

Ya that's what I was thinking of when I was using the term cherrypick (which admittedly is typically used for actually bringing in commits).  Glad we're on the same page there :)

> What's potentially more disturbing, as I wrote in chat, is that the changes were brought into this branch (at least some of them) on a commit labeled "merge from master", so it's possible that those changes **were** in master at some point but have since been overwritten by a bad merge.

Ya there's a number of possible causes, two offhand that I can think of:

- the commits were brought in, but on the merge commit the file in this branch was "newer" so it's changes were considered ahead of the changes merged from master so they overwrote the master changes.  Keep in mind that seeing a commithash as a part of a _merge_ is just a historical record.  The actual _merge_ might itself overwrite changes from master if git determines work in the merged to branch is newer.
- At some point you did a merge from the old Github/master branch after it had been _squash merged_ to master. This can cause wierd behaviour when you change the base branch and start merging from master in because the commits that were in the GB master  branch are considered "newer" because they don't exist in the master history (due to the squash merge).  And actually, this may have nothing to do with when you merged from GB master - doing the squash merge of GB/master might have messed things up because the history in this branch diverges from master.

I think ultimately this branch being originally based off of GB/master which was squash-merged into master is the likely culprit.  I seem to recall having to restart a couple branches I had based off of GB/master when I saw something similar happening in them when I switched their pull to be based off master.

tn3rb:
when I say "rendering multiple multiple times" I mean the component's render function is firing multiple times.

nerrad:
> when I say "rendering multiple multiple times" I mean the component's render function is firing multiple times.

Hmm ya that doesn't sound good.  We'll have to determine why that's happening.

assets/src/editor/events/tickets/editor-ticket/price-calculator/ticket-price-calculator-form-modal.js
```diff
@@ -0,0 +1,210 @@
...
+			.getTicketById( ownProps.ticketId );
+		const prices = isModelEntityOfModel( ticket, 'ticket' ) ?
+			select( 'eventespresso/core' )
+				.getRelatedEntities( ticket, 'prices' ) :
+			[];
+		return {
+			ticket,
+			prices,
+		};
+	} )( TicketPriceCalculatorFormModal )
```
nerrad: To get the number of renders down, I changed the withSelect to this:

```js
/**
 * Enhanced TicketPriceCalculatorForm with Modal
 */
export default withEditorModal( {
	title: __( 'Ticket Price Calculator', 'event_espresso' ),
	customClass: 'ee-ticket-price-calculator-modal',
	closeButtonLabel: __( 'close ticket price calculator', 'event_espresso' ),
} )(
		const ticket = select( 'eventespresso/core' )
			.getTicketById( ownProps.ticketId );
		const prices = isModelEntityOfModel( ticket, 'ticket' ) ?
			select( 'eventespresso/core' )
				.getRelatedEntities( ticket, 'prices' ) :
			[];
		return {
			ticket,
			prices: isEmpty( prices ) ? DEFAULT_EMPTY_ARRAY : prices,
		};
	} )( TicketPriceCalculatorFormModal )
);
```

You'd need to `import { isEmpty } from 'lodash'` and define a `DEFAULT_EMPTY_ARRAY` somewhere in the upper scope.

I don't think similar treatment needs done to `ticket` because it should be always returning either `null` or the same ticket `BaseEntity` instance.

PR #942, opened 2019-02-06
Improve `Registry::pushData` (fixes #855)

### LINKED ISSUE #855: Refactor Registry::pushData() to merge supplied data into existing instead of adding under new numerical key
Originally reported by @tn3rb in https://events.codebasehq.com/projects/event-espresso/tickets/11479

Currently `Registry::pushData()` will append new data onto the existing data array using a numerical key, so something like:

```
$registry->pushData(
    'myAddonDataNeededForXFactorPackage',
    array(
        'totalPrice' => 15
    )
);
```

would result in the final data object looking like:

```
 [
     'apiKey' => 'qq7-1-039844',
     'xFactorSignupLabel' => 'Git me signed up!',
     0 => [
         'totalPrice' => 15
     ]
 ]
```

This is problematic because JS code relying on that total price data would need to access it using something like :

```
myAddonDataNeededForXFactorPackage[0]['totalPrice']
```

the problem is that any other data being pushed onto that same top level key, **before** the above, will increment that numerical key, ie:

```
// some other code that loads earlier
$registry->pushData(
    'myAddonDataNeededForXFactorPackage',
    array(
        'foo' => 'bar'
    )
);
// the data we were expecting to be first
$registry->pushData(
    'myAddonDataNeededForXFactorPackage',
    array(
        'totalPrice' => 15
    )
);
```

would result in the final data object looking like:

```
 [
     'apiKey' => 'qq7-1-039844',
     'xFactorSignupLabel' => 'Git me signed up!',
     0 => [
         'foo' => 'bar'
     ],
     1 => [
         'totalPrice' => 15
     ]
 ]
```

which would of course break any code relying on `myAddonDataNeededForXFactorPackage[0]['totalPrice']`

An easy fix would be to simply merge any incoming data with the existing data.

To handle collisions I suggest implementing some class constants that define different ways of proceeding, similar to how PHP's `extract()` method handles duplicate variable names: http://php.net/manual/en/function.extract.php

so maybe some constants like:

```
COLLISION_ERROR = 0; // throws invalidArgumentException
COLLISION_OVERWRITE = 1; // replaces existing data
COLLISION_PREFIX = 2; // prefixes new var name with supplied prefix
```

####COLLISION_ERROR 
this option would detect if any keys for the incoming data match existing keys and throw an error

####COLLISION_OVERWRITE 
this option would simply replace any existing keys with the new incoming data

####COLLISION_PREFIX 
this option would allow client code to recover from their data being moved to a new key, ie:

```
// some other code that loads earlier
$registry->pushData(
    'myAddonDataNeededForXFactorPackage',
    array(
        'totalPrice' => 10
    )
);
// the data we were expecting to be first
$registry->pushData(
    'myAddonDataNeededForXFactorPackage',
    array(
        'totalPrice' => 15
    ),
    Registry::COLLISION_PREFIX,
    'foo_'
);
```

would result in the final data object looking like:

```
 [
     'apiKey' => 'qq7-1-039844',
     'xFactorSignupLabel' => 'Git me signed up!',
     'totalPrice' => 10,
     'foo_totalPrice' => 15
 ]
```

which would allow client code to check for the existence of its expected data in both locations before entering into an error state.

This can be assigned to Darren after triaging

nerrad:

See #855 for original issue.  While I agree the current behaviour of `pushData` is less than ideal, I think some of the suggestions for improvement are a bit overkill.  I don't think we should have variable merge strategies because it's an unknown on whether they will even be needed.  I'd prefer to have `pushData` have a specific known opinionated behaviour and there are alternatives if consumers don't want risk that behaviour.  If we discover that having different merge strategies are necessary, then we can add specific interfaces for each of them.  For now, this pull implements the following:

- `pushData` will always result in the provided value being added to an array on the given key.  So if you do `EE_Registry::pushData( 'test', 'foo' );` when `test` doesn't exist in the registry data, then `EE_Registry::getData( 'test' )` will return `[ 'foo' ]`.
- When an array is fed as the value using `pushData`, it will get _merged_ with the existing value on that key (if it exists).  The merge strategy is what `COLLISION_OVERWRITE` is described as in the issue.  `array_merge` is used, so that means arrays with numeric keys are appended to the existing array.
- If the consumer tries to submit a value for a key that already has a scalar value in the registry data, then an exception is thrown because `pushData` will not overwrite or convert existing key/value pairs cached that are not arrays.
- There are php unit tests added that describe and assert the above behaviour.

## How has this been tested

* [x] Verify new unit tests pass.

I also verified that `pushData` isn't used anywhere in our add-ons or EE core currently.  I'm fairly confident that it's usage in third party add-ons is close to nil as well.

## Checklist

* [x] My code is tested.

### DIFF
--- a/core/services/assets/Registry.php
+++ b/core/services/assets/Registry.php
@@ -245,12 +245,16 @@ class Registry
     /**
      * Similar to addData except this allows for users to push values to an existing key where the values on key are
      * elements in an array.
-     * When you use this method, the value you include will be appended to the end of an array on $key.
-     * So if the $key was 'test' and you added a value of 'my_data' then it would be represented in the javascript
+     *
+     * When you use this method, the value you include will be merged with the array on $key.
+     * So if the $key was 'test' and you added a value of ['my_data'] then it would be represented in the javascript
      * object like this, eejs.data.test = [ my_data,
      * ]
-     * If there has already been a scalar value attached to the data object given key, then
+     * If there has already been a scalar value attached to the data object given key (via addData for instance), then
      * this will throw an exception.
+     *
+     * Caution: Only add data using this method if you are okay with the potential for additional data added on the same
+     * key potentially overriding the existing data on merge (specifically with associative arrays).
      *
      * @param string       $key   Key to attach data to.
      * @param string|array $value Value being registered.
@@ -276,7 +280,11 @@ class Registry
                 )
             );
         }
-        $this->jsdata[ $key ][] = $value;
+        if ( ! isset( $this->jsdata[ $key ] ) ) {
+            $this->jsdata[ $key ] = is_array($value) ? $value : [$value];
+        } else {
+            $this->jsdata[ $key ] = array_merge( $this->jsdata[$key], (array) $value);
+        }
     }
 
 
--- a/tests/testcases/core/services/assets/RegistryTest.php
+++ b/tests/testcases/core/services/assets/RegistryTest.php
@@ -114,6 +114,38 @@ class RegistryTest extends EE_UnitTestCase
     }
 
 
+    public function pushDataWithArrayProvider() {
+        return [
+            'initial creation of dataset when it does not exist' => [ 'test', 'foo', [ 'foo' ] ],
+            'adding a string to existing dataset' => [ 'test', 'bar', [ 'foo', 'bar' ] ],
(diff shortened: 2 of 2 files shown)


nerrad:
> I wonder if we should type hint the incoming value as an array though so that it is absolutely clear that the data will end up as an array?

I intentionally didn't hint as an array because that makes it easier for client code to add additional scalar items to the existing array (see tests for examples).

PR #947, opened 2019-02-07
Ensure QSG_identifier is unique when saving new question groups (fixes #720)

### LINKED ISSUE #720: Avoid Fatal Error that results from inputting a duplicate Question Group ID
**Describe the bug**

A fatal error is thrown when you attempt to save a Question Group that has a Group Identifier that matches the Group Identifier of another Question Group. 
**To Reproduce**

Add a question group and give it a unique Group identifier, copy that Group Identifier to your clipboard, then save and close
Add another question group and give it the same Group Identifier (paste from clipboard)
A fatal error is thrown:
`Fatal error: Uncaught Error: Call to a member function questions() on null in .../caffeinated/admin/extend/registration_form/Extend_Registration_Form_Admin_Page.core.php:701`

**Expected behavior**

This could use some user proofing to avoid throwing a fatal error. Possible solutions might include adding some server side validation to catch this, and show a friendly error message to tell the user to change the Group ID. Or another idea would be to append a -2 (like what WP does when a post slug matches another post's slug)

nerrad:

The question group saving process already generates a unique value for the `QSG_identifier` field if its left blank when creating the question group.  However because `QSG_identifier` is set as a UNIQUE index in the db schema, if a user provided string matches something already saved in the db, then a sql error is triggered and the question group is not created.  This in turn causes the fatal. In this pull:

- Prevent the conditions causing the fatal by checking to see if there is already a Question Group in the db matching that `QSG_identifier`.  If there is then append a unique id to the existing string to make it unique before inserting.
- Future proof against possible future fatals by bailing early if saving a new Question Group fails (and show an error message to the user).

## How has this been tested

* [x] Follow the reproduction steps in the original issue (#720) and verify no more fatals.

## Checklist

* [x] My code is tested.


joshfeck:
tests yielded expected results

PR #948, opened 2019-02-08
stop build process if only doing a doc change

nerrad:

This pull implements a travis configuration change to stop travis builds when there are only `md` or docs changes in the pull.  This should save having unnecessary builds run.

## How has this been tested

* [ ] verify normal builds are unaffected.
* [ ] Do a pull based off of this one as a test that only has a doc change.


nerrad:
Doesn't work reliably so closing.

PR #951, opened 2019-02-08
Optimize usage of react key for AttendeeListItem component

nerrad:

Keys in a react list element should be something that identifies that element as unique across re-renders.  This means in the case of list elements based off of entities its best to use the ID of that entity because it ensures the list item is not re-rendered if the id doesn't change.  In the case where an array of attendees might have an item that does not have an id, `cuid` is used instead.

## How has this been tested

* [x] Verify existing unit tests pass.
* [x] Load a post that had the event attendee block _prior_ to this branch and verifies it still loads as expected.
* [x] Create a new post and add an event-attendees block and verify it works as expected.

## Checklist

* [x] My code is tested.


assets/src/components/event-attendees/index.js
```diff
@@ -35,9 +36,10 @@ export default class EventAttendeeList extends Component {
 			return '';
 		}
 		const listItems = attendees.map(
-			( attendee, index ) => {
+			( attendee ) => {
+				const id = attendee.id || cuid();
```
tn3rb: so this is safe from collisions, but will also generate a new cuid on every component rerender, which will in turn trigger a rerender for each item using a cuid for its key, since those items would now have a new cuid.

I wonder if there is a quick and easy way to either:

- generate a hash from some of the primary entity properties and use that (assuming those properties don't change)
- save a temp ID to each entity, such as a cuid, and then the entity id can be used and will stay with that entity for the entire time it exists in the store until it is persisted.

nerrad: > generate a new cuid on every component rerender

To be clear, a cuid is ONLY generated if there is no attendee id. I'm actually going to change this because `AttendeeListItem` EXPECTS an instance of `BaseEntity` so if there's no id it's going to bail anyways.  So what I'll do is just skip the list item if it doesn't have an id.

PR #955, opened 2019-02-13
Fix bugs in Money and Currency Value objects

nerrad:

While working on REM, @tn3rb noticed that the money vo's in the `BaseEntity` instances he was working with were not using his site's set currency values for the internal `Money.formatter` object as expected.  This pull fixes that along with another bug I discovered in the Currency vo.  The following is what is implemented in this pull:

- Make sure we're not sharing a reference of `Accounting` across all constructed Money vo's (this wasn't the cause of the bug but could surface other bugs)
- The accounting npm package we use differs from the standalone browser library version in that the `settings` property on the accounting object is a flat object.  The browser version `settings` property is a nested object (with `currency` and `number` properties).  The flat object is sufficient, so the solution was just to make sure we're passing through the expected object. _This is ultimately what fixed the original reported issue_.
- In debugging things I discovered the Currency vo was using the wrong property for setting the default `subunits` property value.

## How has this been tested

* [x] Added unit tests that verified the discovered bugs were present and then they demonstrated the fixes worked.  Verified no other unit tests were broken.
* [x] monkey testing in the browser to ensure the money and currency vos worked as expected.

## Checklist

* [x] My code is tested.


assets/src/vo/currency.js
```diff
@@ -84,7 +84,7 @@ export class Currency {
 		this.decimalMark = currencyConfig.decimalMark || this.decimalMark;
 		this.thousandsSeparator = currencyConfig.thousandsSeparator || this.thousandsSeparator;
 		this.subunits = currencyConfig.subunits ||
-			Math.pow( 10, this.decimalMark );
```
tn3rb: oof ouch owie!

tn3rb:
failed unit test was an off by 1 second error... restarted test... should pass

nerrad:
merging this to master as changes here don't impact existing published functionality and further testing of the value-objects will be done in the upstream work implementing it.

PR #959, opened 2019-02-14
Don’t autoload these, some can have a lot of data Fixes #346

### LINKED ISSUE #346: Don't autoload ee_dms_map_Core_4.1.0_events options
See original Codebase ticket: https://events.codebasehq.com/projects/event-espresso/tickets/11227

Per Josh:
I was working on a support token where they were having issues with the site going down (largely due to a runaway ee_transient_schedule option and ee_ssn_ transient options). While looking into this I saw that their options table has a number of fairly large autoloading options that store what appears to be information about what was migrated over from EE3. Are these options necessary to autoload on every request? Can they be safely deleted to help speed up a site?

joshfeck:

See #346 
## How has this been tested

Do a migration from EE3, then check the options table for rows with option name that start with `ee_dms_map_`. They should not be set to autoload.

## Checklist

* [x] My code is tested.


sethshoultes:
Looks like this is working:
image

Issue #278, opened 2018-02-27
Private events lose their private status when sold out!


jonappleton:
> Before submitting your issue, please ensure that what you are reporting is specific to Event Espresso 4.
Yes EE Version 4.9.46.p

When submitting an issue, please follow this rough guideline for your report.

### Is this a bug report?

* [ X] yes
* [ ] no

### What branch did this occur in?
Version 4.9.46.p

### What version of WordPress is on the site this issue happened?
4.8.5

### If you were directed to create an issue here by the EE support team, what is the link to your original EE support forum thread?
https://eventespresso.com/topic/private-events-lose-their-private-status-when-sold-out/#post-263807

### Outline the steps to reproduce
Create and event
Set it to private (which shows in EE admin as Inactive)
Set it to have 1 ticket qty for sale
Register someone for that ticket so qty left becomes zero
EE then changes the post_status of that event from "private" to "sold_out"
So now the event shows as being public and no longer private

We have an upcoming schedule table that shows events where post_status = published or sold_out as we still like to show the sold out events as those booked on to them can click to see location details etc.

We’ve just noticed that when the last ticket sells on a private event then EE changes the post_status from private to sold_out, causing our private events to show in our public website schedule.

I did some digging and found that EE stores the old post_status in the postmeta table with a key of “_previous_event_status” so I have now added an extra check to my php so when is iterates through the results it says

if($previous_event_status == “private” && $currentEventStatus == “sold_out”) {

//dont show table row } else {
//yes display this event
}

This seems to work for me but just wondering if I’m making extra work for myself and is there a simpler way to ignore/hide private events when pulling events from SQL

This feels like a bit of a bug that EE overrides the private status like this?

Thanks
Jonathan

mnelson4:
hey @jonappleton ya I'm reproducing the issue. When events get sold out, they're given a new status of "sold_out"... which is public! D'oh! It seems we should have a "sold_out_private" and "sold_out_protected" too.
I'm going to see if we already have an internal ticket for this...

jonappleton:
Thanks for checking

mnelson4:
Hey @jonappleton believe it or not, this is the first report of this issue. (I guess not a lot of people use protected/private events that get sold out). We're going to work on a quickfix solution and a more long-term solution.
[...]

PR #960, opened 2019-02-15
dont change private events to sold out beceause then they are no longer private

mnelson4:
…

Fixes private events becoming public when sold out. Instead, they retain the status "private" and are just marked "inactive". See https://github.com/eventespresso/event-espresso-core/issues/278

## How has this been tested

* [x] Create a private event, set the ticket quantity to 1, register for that ticket. Verify the event is still not visible to unauthenticated users. In the WordPress dashboard, the event's "active status" will still appear as "upcoming". Oh well.
* [ ] Possible bug in master: change the registration to "declined". The event's "active status" will still be "upcoming".
* [x] Repeat the above, but with a regular event. It should be marked sold out as normal.

## Checklist

* [ ] My code is tested.


mnelson4:
Kk, although in the course of playing with this, I noticed once a private event's tickets were all sold, its active status was stuck as "inactive" (even when I declined the registrations). I noticed a spot in the code where we were solely looking for published events, but I think we should also be looking for private events. Now, when you declined the registration that sold out the private event, the event's "active status" will be "upcoming". Also, when a private event sells out, its active status just remains "upcoming"... This is just an improvement (not complete fix), so I'm not too fussy about that personally...

mnelson4:
oh and regarding the rant: thanks for bringing that up, although I'd need to immerse myself more in the topic to have anything useful to say. :no_mouth:

mnelson4:
thanks @tn3rb, passing for testing

joshfeck:
This is tested and ready for merging.

PR #965, opened 2019-02-15
Add checkin action handling to `eventespresso/core` store

nerrad:

While working on https://github.com/eventespresso/eea-barcode-scanner/pull/7 I realized that there is need for handling the toggling of checkins.  Checkins are a unique data model in that whenever a checkin is toggled for a given registration and datetime, a new record is created.  This allows for an audit trail of when a registration checked in and out of an event.

For the purpose of the barcode scanner, the ui presents a button for both indicating the current toggle action available (requiring the latest checkin record) and initiating toggling to the relevant checkin state (i.e. checked in or out).

The complexity of the check-in model is represented by the creation of an [rpc endpoint](https://github.com/eventespresso/event-espresso-core/blob/master/docs/C--REST-API/ee4-rest-api-rpc-checkin.md) specifically for handling checkins. 

While the current selectors and actions available in our stores _could_ be used for handling things (with the exception of the toggle checkin endpoint), consuming code would have to do more work to correctly handle checkin changes since when a checkin is "toggled" via the rpc endpoint, a NEW record is returned (not an update to an existing record) - I'll add more details at another time just how much more code this creates for consumer.

For the purpose of simplifying things this pull seeks to:

- Add a new selector (and resolver) to `eventespresso/core` that is used for retrieving the latest checkin record `getLatestCheckin( registrationId, datetimeId );`
- Add a new action (generator) to `eventespresso/core` that is used for `receiving` the latest checkin: `receiveLatestCheckin( checkinEntity )`.
- Add a new action (generator) to `eventespresso/core` that is used for toggling a checkin (and in turn updates necessary store state via further dispatches): `toggleCheckin( checkinEntity )`

Under the hood these will use both new code, and existing actions for recording the checkin record in the state.

The shape of the new state tree (using Immutable!) will be something like this:
```
{
	modelSpecific: {
		getLatestCheckin: {
			[JSON.stringify([ registration.id, datetime.id] )]: checkInEntity.id,
		}
	}
}
```

So to be clear it simply stores the checkin id and the selector would retrieve the actual entity from the `entities` tree in the state.

These serve the goal of simplifying the api for consumers subscribing to the latest checkin because they would simply have to do something like this.

```js
export default compose( [
		const { registration, datetime } = ownProps;
		return {
			checkin: select( 'eventespresso/core' ).getLatestCheckin(
				registration.id,
				datetime.id
			),
		}
	} ),
	withDispatch( ( dispatch ) => {
		return {
			onChange: ( checkin ) => dispatch( 'eventespresso/core' )
				.toggleCheckin( checkin.regId, checkin.dttId ),
		}
	} ),
] )( WrappedComponent );
```

This pull introduces the following:

- new "model specific" interface for adding new specific type model queries.  This includes a new action (`receiveSelectorValue`), a new selector (`getSelectorValue`), and new reducer.  These are interfaces to the model-specific state that can be used by more concrete model specific queries.
- The first model specific item to use this new interface is the interaction with latest checkin.  This includes new actions (`receiveLatestCheckin`, `toggleCheckin`), a new selector and resolver (`getLatestCheckin`).  Reducer is unneeded because that's handled via the model-specific interface.
- added a new control to base-controls for handling selects that have a resolver.  This is a better implementation than calling a generator from `base-resolvers`.  After this pull is merged, I'll do another pull to implement this wherever else we are using `base-resolvers` (which impacts more existing things so will need more testing).
- added unit test coverage.
- added documentation.

## How has this been tested

* [x] Added new unit tests, ensure existing tests don't fail.
* [x] Monkey testing in the browser to ensure new selects/actions work as expected.

## Checklist

* [x] My code is tested.


nerrad:
One thing we _might_ be able to do to abstract this more for other uses is dynamically introduce state for this kind of _specific_ type of requests. In other words, the shape of the state would always be something like:

```
const modelSpecificState = {
	[ selectorName ]: {
		[ hashedArgs ]: value,
	}
}
```

`hashedArgs` would be a hash of the args coming into the selector, and the value would be the whatever is used for the results (in the case of normalization could be an array of ids or an id).

That way future  model specific stuff could simply use a common action for storing and retrieving state and then the specific stuff would be done on the value.

assets/src/data/eventespresso/core/model/checkin/resolvers.js
```diff
@@ -0,0 +1,59 @@
...
+		// this is the cause for the error?
+		return null;
+	}
+	const factory = yield resolveSelect(
+		SCHEMA_REDUCER_KEY,
+		'getFactoryForModel',
+		'checkin'
+	);
+	if ( ! isModelEntityFactoryOfModel( factory, 'checkin' ) ) {
+		return null;
```
tn3rb: same point as earlier... if the model is invalid we just do... nothing?
maybe there should be some kind of `InvalidModel` exception/error we can throw/return in these cases

docs/AA--Javascript-in-EE/data/eventespresso/core/actions/model-specific/persist.md
```diff
@@ -0,0 +1,21 @@
...
+
+## Arguments
+
+| Argument       | Type                  | Description                                                                                                                                                            |
+| -------------- | --------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| registrationId | `number    `          | The registration id for the registration having its checkin state changed.                                                                                             |
+| datetimeId     | `number`              | The datetime id for the datetime the registration is having the checkin state changed for.                                                                             |
+| force          | `boolean`             | This is used to flag whether consideration should be given to whether the registration can be checked-in for the given datetime.  If true, then no restrictions apply. |
```
tn3rb: > This is used to flag whether consideration should be given to whether the registration can be checked-in for the given datetime. 

kinda awkward and unclear, maybe something like:

> Flag to indicate whether registration checked-in restrictions for the given datetime should be ignored.

or
 
> Flag to indicate whether restrictions on registration checked-in for the given datetime should be ignored.

docs/AA--Javascript-in-EE/data/eventespresso/core/selectors/model-specific.md
```diff
@@ -0,0 +1,12 @@
+The following selectors for model specific queries that are available on the object returned by `wp.data.select( 'eventespresso/core' )`.
+
+## `getLatestCheckin( registrationId, dateTimeId )`
+
+This retrieves the latest checkin entity from the state for the given value.  There is a corresponding resolver so if this selector has not been resolved yet for the given arguments, the resolver will retrieve the latest checkin record from the server and assign it to the state as an instance of `BaseEntity`.
```
tn3rb: > for the given values.

plural since more than one value is supplied, or maybe:

> for the given arguments. 

since that is the phrasing used in the sentence that follows the above

tests/javascript-config/unit/fixtures/checkin-schema.js
```diff
@@ -0,0 +1,256 @@
...
+				'default': 0,
+				'primary_key': true,
+			},
+			'REG_ID': {
+				'description': 'Registration Id',
+				'type': 'integer',
+				'readonly': false,
+				'default': 0,
+				'foreign_key': {
+					'description': 'This is a foreign key the points to the given models.',
```
tn3rb: ???

maybe:

> This is a foreign key that points to the given models

or better yet (imho)

> This is a foreign key that points to the Registration model

tests/javascript-config/unit/fixtures/checkin-schema.js
```diff
@@ -0,0 +1,256 @@
...
+					],
+				},
+			},
+			'DTT_ID': {
+				'description': 'Datetime Id',
+				'type': 'integer',
+				'readonly': false,
+				'default': 0,
+				'foreign_key': {
+					'description': 'This is a foreign key the points to the given models.',
```
tn3rb: same typo as above

assets/src/data/eventespresso/base-controls.js
```diff
@@ -63,6 +80,27 @@ const controls = {
...
+		return new Promise( ( resolve ) => {
+			const hasFinished = () => selectData( 'core/data' )
+				.hasFinishedResolution( reducerKey, selectorName, args );
+			const getResult = () => selectData( reducerKey )[ selectorName ]
+				.apply( null, args );
+
+			// trigger the selector (to trigger the resolver)
+			const result = getResult();
+			if ( hasFinished() ) {
+				return resolve( result );
```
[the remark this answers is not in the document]
nerrad: Because we do an initial trigger of the selector and only _use_ the value if it has resolved.

If I use `getResult()` directly, then I have to still call it outside of the `hasFinished()` conditional.  So this actually reduces the number of times the select has to be invoked.

I _might_ be able to get away with re-using the `result` value in the subscribe callback.

assets/src/data/eventespresso/core/model/checkin/actions.js
```diff
@@ -0,0 +1,114 @@
...
+		);
+		return null;
+	}
+	const factory = yield resolveSelect(
+		SCHEMA_REDUCER_KEY,
+		'getFactoryForModel',
+		'checkin'
+	);
+	if ( ! isModelEntityFactoryOfModel( factory, 'checkin' ) ) {
+		return null;
```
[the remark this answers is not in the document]
nerrad: Ya I'm not certain what to put here at this point and whether it should be a user facing error or a console.log.  I initially went with this thinking that this could be left up to the consuming code to handle (by dealing with a "null" value returned - which is doable).

Since I'm still uncertain _what_ should be done I've left it as is for now until I start implementing and can discover what might work best.

assets/src/data/eventespresso/core/model/checkin/resolvers.js
```diff
@@ -0,0 +1,59 @@
...
+import { REDUCER_KEY as SCHEMA_REDUCER_KEY } from '../../../schema/constants';
+
+/**
+ * Resolver for the getLatestCheckin action.
+ *
+ * @param {number} registrationId
+ * @param {number} dateTimeId
+ * @return {BaseEntity|null} Null if there is an error or no checkin record.
+ */
+export function* getLatestCheckin( registrationId, dateTimeId ) {
```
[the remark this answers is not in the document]
nerrad: hmm ya wouldn't hurt.

assets/src/data/eventespresso/core/model/checkin/resolvers.js
nerrad: Just to be clear, this is not a model check.  It is a check for the factory that builds the models.  It'd have to be a fairly significant critical error if the factory isn't available for the given model (i.e. the REST API is not accessible on a site - which in turn would cause problems elsewhere).

If anything, I'm thinking of just removing this check, because if there is no factory available its a significant critical exception and thus the error from the usage of factory in the next line would be sufficient to indicate something wrong to the developer.

I'll think on this though, I might just add a warning after all (as it would mostly be a development issue).

tests/javascript-config/unit/fixtures/checkin-schema.js
nerrad: This will require a separate issue/pull request as it came from a direct REST response (that I copied and pasted into this file for a fixture).  In other words, this is a grammar issue in the REST API code in core.

tests/javascript-config/unit/fixtures/checkin-schema.js
nerrad: same response as above.

assets/src/data/eventespresso/base-controls.js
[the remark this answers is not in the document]
nerrad: Okay so the reason why things are the way they are is this:

- The initial `const result = getResult()` invokes the selector and if the selector has not been resolved, this triggers the resolution.  If it HAS resolved then it returns the resolved value which can then just be returned on line 93.  So `result` only gets returned if resolution is finished.  As mentioned earlier, assigning to the const allows for skipping another unnecessary call to the selector.
- the call to `resolve( getResult() )` in line 99 is inside the subscription callback.  We need to invoke the selector every time the subscription fires because that it will fire at various resolution states for that selector.  That's why we can't simply use the value on the `result` const because it will get stale.

assets/src/data/eventespresso/core/model/checkin/actions.js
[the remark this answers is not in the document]
nerrad: Implemented a warning

assets/src/data/eventespresso/core/model/checkin/resolvers.js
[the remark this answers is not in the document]
nerrad: added a warning

assets/src/data/eventespresso/core/model/checkin/resolvers.js
nerrad: added a warning

Issue #924, opened 2019-01-30
 Change the color of the view transaction button


joshfeck:
See https://events.codebasehq.com/projects/event-espresso/tickets/8746

PR #967, opened 2019-02-19
Relocate "Action buttons" on Registration Details page to be consistent with location of buttons on Transaction Details page

joshfeck:

See #924 and https://events.codebasehq.com/projects/event-espresso/tickets/11459

## How has this been tested

- [x] Load up a single registration page to see the button's new location. You may need to clear browser cache for CSS changes to take effect. 

Here are screenshots showing the differences, along with a screenshot of the txn page for comparison:
**before**
screen shot 2019-02-19 at 12 55 12 pm
**after**
screen shot 2019-02-19 at 12 55 22 pm
**txn page button locations**
screen shot 2019-02-19 at 12 55 37 pm

## Checklist

* [x] My code is tested.


joshfeck:
Hey Seth I went ahead and also applied the change to the "View Transaction" button that you  requested in [this codebase ticket](https://events.codebasehq.com/projects/event-espresso/tickets/11459). To be fair, you probably don't remember what prompted you to request that change, right? I think it looks OK, but I after reading Darren's questions, I do wonder about why that button color should be changed.

Usually a button is given the "primary" color if it's the main action in its container. In the context of a "Registration Details" box, I don't think there should be more weight given to "View Transaction" vs. "Resend Registration" because the two aren't related. 

Anyway, we can roll with this change or if you'd prefer to keep the "View Details" button as it was before (gray background with black text) I can switch it back.

sethshoultes:
I like the changes and am okay with the blue "View Details" button. I'm thinking we should make the "Send Payment Reminder Button" blue as well:
image

nerrad:
Seth can you expand on your rationale for why you think buttons should be made a primary action?  I think it'd be good for us to have a consistent design pattern we use with rationale for what makes a button "primary" button.  There's convention that could be used as examples as well (wp itself, github etc).

Rationale has to be more than just "it looks good to me" or "it seems better to me" because that is highly subjective.

joshfeck:
@sethshoultes To be clear, the color change for the View Transaction button was something you requested here:
https://events.codebasehq.com/projects/event-espresso/tickets/8746
The unfortunate thing about that ticket was there was no reason given for the change. Going forward, we really need have some kind of stated purpose for why make a change to an existing UI element. Changes to existing UIs should be done with care, so there should be a good reason for these. 

With regards to your follow up request: The transactions page is not part of this PR, this PR is about making a change to the registrations page. Since you brought this idea up here though, we'll address it here:

The "Send Payment Reminder" button should *not* be changed to be a primary button (or blue, depending on the set WP admin color scheme setting). The reason being there is already one primary button within the box. Having two primary buttons breaks standard UI patterns. So we'll avoid doing that.

sethshoultes:
Oh, I see now. Ya, I just thought it looked better. We can leave off the blue in those buttons. However, I do think the relocation of the buttons looks better and makes more sense below the tables.

PR #968, opened 2019-02-19
Experiment: use wp.data store for filter state

nerrad:
This is based off of #961 to demonstrate how the `wp.data` can be used for managing filter state.

Some things to note:

- This was done rapidly as a proof of concept.  Thus there's no inline jsdocs or tests
- I did the new `filter-state` data store as a generic implementation.  This is required so you can have multiple filter-states managed (and delimited via the `context` property).
- The initial implementation is done in the `withTicketListFilterState` hoc.
- Future iterations could abstract things further with there being a generic hoc that is fed what to use for context.  So feasibly it has the potential to eliminate a lot of code.
- I didn't remove any of the existing filter-state code, but I think this demonstrates the potential for less code overall.
- I tested this locally and the filters worked as expected.  There are a couple console errors, but it looks like they may be unrelated to this work (it looks like prop-types warning)


assets/src/data/eventespresso/filter-state/index.js
```diff
@@ -0,0 +1,53 @@
...
+		state[ context ];
+};
+
+const filterStateReducer = ( state = DEFAULT_STATE, action ) => {
+	const { type, context, filter, value } = action;
+	if ( type === SET_FILTER ) {
+		return {
+			...state,
+			[ context ]: getState( state, context ).set( filter, value ),
+		};
```
nerrad: This will return a new state object every time this specific action is dispatched even if there are no changes. This will result in unnecessary renders.

I realize that what I originally had in this pull was wrong too.  So probably what we need to do is something like:

```js
const filterStateReducer = ( state = DEFAULT_STATE, action ) => {
	const { type, context, filter, value } = action;
	if ( type === SET_FILTER ) {
		const newValue = getState( state, context ).set( filter, value );
		if ( state[ context ] !== newValue ) {
			return {
				...state,
				[ context ]: newValue,
			};
		}
	}
	return state;
};
```

As I did this, I realized that there's probably places in the existing store reducers where I'm  doing things wrong (mutating the original state) so I'll watch for that as a part of my review pull in #970. It's likely that there's a few places where I'm doing things incorrectly and that could be the cause of the wonky `withSelect` behaviour you were seeing.

assets/src/data/eventespresso/filter-state/index.js
```diff
@@ -0,0 +1,53 @@
...
+	}
+	return state;
+};
+
+const setFilter = ( context, filter, value ) => {
+	return {
+		type: SET_FILTER,
+		context,
+		filter,
+		value,
```
nerrad: You need to be wary about what is set as the value here. If it's an object or array, then that means that there will be a new immutable map created every time it's set even if there is a value equivalency. For instance, if this is called with something like:

```js
dispatch( 'eventespresso/filter-actions' ).setFilter( 'some-context', 'filter-name', [ 10, 20 ] );
```

And there is already a value for that filter equalling `[ 10, 20 ]` in the state, then when the reducer is called it will result in a new immutable map.  You can see a demo of this if you go to the [immutable.js](https://immutable-js.github.io/immutable-js/) site, open up your browser console, and you can do the following in the console:

```js
valueTest = Immutable.Map().set( 'foo', [ 10, 20 ] );
valueTestB = valueTest.set( 'foo', [ 10, 20 ] );
valueTest === valueTestB
> false
valueTest = ImmutableMap().set( 'foo', 10 );
valueTestB = valueTest.set( 'foo', 10 );
valueTest === valueTestB
> true
```

Of course if a reference is passed around, it wouldn't be an issue (and likely in most use-cases a reference _would_ be passed around).  However, it's still a potential "gotcha".

I'm not sure we should trust that consuming code will understand that limitation when setting in the store.  One way of handling this would be to only allow scalar values to be set via the action (so that ensures there won't be any unexpected side-effects for incoming objects/arrays). It appears that will always be the case from your implementation in this pull so we'll probably want to enforce that.

assets/src/data/eventespresso/filter-state/index.js
```diff
@@ -0,0 +1,53 @@
...
+	};
+};
+
+const getFilter = createSelector(
+	( state, context, filter, defaultValue ) => {
+		return getState( state, context ).get( filter, defaultValue );
+	},
+	( state, context, filter, defaultValue ) => [
+		getState( state, context ).get( filter, defaultValue ),
```
nerrad: Hmm, I don't think `defaultValue` should be passed through on this argument as this argument is used by rememo to determine whether there is a state change and if there is to "recalculate" by calling the original selector (which is the first argument on `createSelector`).  And actually (again, I know you were going by my original example), if we're only dealing with scalar values here we don't need to use rememo at all.  So you could simply write this selector as:

```js
const getFilter = ( state, context, filter, defaultValue ) => {
	return getState( state, context ).get( filter, defaultValue );
};
```

PR #970, opened 2019-02-20
Review existing data stores and ensure action-generators and resolvers are only using actions

nerrad:

In #965 I introduced a new control, `ResolveSelect` that handles waiting for the result of a selector with an attached resolver.  This removes the need for things like `getSchemaByModel` found in `base-resolvers`.

This pull request reviews all the existing code in the registered stores and ensures all action-generators and resolvers are yielding actions (which means using the controls where necessary.  

Benefits of doing this:

- Tests are more explicit.
- logic is clearer
- in some cases I think this may prevent race conditions with resolution starting for the same select in multiple places.
- debugging is more easier.

This pull does the following: 

- adds new `resolveDispatch`, `resolveGetEntityByIds`, and `resolveGetRelatedEntities` controls and implements everywhere along with the `resolveSelect` control.
- review all data stores to look for cases of defaults that could cause unnecessary renders or unintended mutations (there were a few).
- update tests to match new expectation.
- removes now obsolete `base-resolvers.js` and other obsolete files after the refactor.

## How has this been tested

* [x] Unit tests pass
* [x] Monkey testing of the stores in the browser console to make sure nothing is broken.  I did not do _exhaustive_ tests of everything, but focused on a few critical selectors and dispatches that would highlight issues.
* [x] This impacts the `EventAttendees` block so it needs tested to verify that there is no impact from the changes in this pull.

## Checklist

* [x] My code is tested.


assets/src/data/eventespresso/core/actions/persist-relations-generators.js
```diff
@@ -37,9 +49,11 @@ function* persistAddRelationsForModel( modelName ) {
  * names with values of relation Ids persisted.
  */
 function* persistDeleteRelationsForModel( modelName ) {
-	return yield persistRelationsForModel(
+	return yield resolveDispatch(
+		CORE_REDUCER_KEY,
+		'persistRelationsForModel',
 		modelName,
-		false
+		false,
```
tn3rb: trailing comma in a function call? typo?

assets/src/data/eventespresso/core/actions/persist-relations-generators.js
```diff
@@ -55,15 +69,25 @@ function* persistDeleteRelationsForModel( modelName ) {
...
+		addRelation
+	);
 	if ( isEmpty( relationState ) ) {
-		return {};
+		return DEFAULT_EMPTY_OBJECT;
 	}
-	const relationsPersisted = {};
 	const entityIds = keys( relationState );
+	const relationsPersisted = entityIds.length > 0 ?
+		{ ...DEFAULT_EMPTY_OBJECT } :
```
tn3rb: ????

assets/src/data/eventespresso/core/actions/persist-relations-generators.js
```diff
@@ -94,27 +118,31 @@ function* persistRelationsForEntityId(
...
 	const relationNames = relationState[ entityId ] ?
 		keys( relationState[ entityId ] ) :
-		[];
+		DEFAULT_EMPTY_ARRAY;
 	if ( isEmpty( relationNames ) ) {
-		return {};
+		return DEFAULT_EMPTY_OBJECT;
 	}
-	const persistedRelations = {};
+	const persistedRelations = { ...DEFAULT_EMPTY_OBJECT };
```
tn3rb: ???

joshfeck:
I did some testing on the Attendees block and it all works as expected.

PR #972, opened 2019-02-20
Fix fatal error in ticket list shortcode parser (fixes #971)

### LINKED ISSUE #971: Prevent Fatal error: Uncaught TypeError: Argument 2 passed to EEH_Parse_Shortcodes::parse_ticket_list_template() must be an instance of EE_Ticket, null given, called in /event-espresso-core-reg/core/libraries/shortcodes/EE_Ticket_List_Shortcodes.lib.php on line 99 with cancelled registrations
A few users have reported this but we couldn't really narrow down the problem, usually, the user had used the Attendee Mover to move a registration and they'd get this error when resending the registration emails or even on the Invoice:

```
Notice: Undefined index: ticket in /event-espresso-core-reg/core/libraries/shortcodes/EE_Ticket_List_Shortcodes.lib.php on line 97

Fatal error: Uncaught TypeError: Argument 2 passed to EEH_Parse_Shortcodes::parse_ticket_list_template() must be an instance of EE_Ticket, null given, called in /event-espresso-core-reg/core/libraries/shortcodes/EE_Ticket_List_Shortcodes.lib.php on line 99 and defined in /event-espresso-core-reg/core/helpers/EEH_Parse_Shortcodes.helper.php:97 
Stack trace: 
#0 /event-espresso-core-reg/core/libraries/shortcodes/EE_Ticket_List_Shortcodes.lib.php(99): EEH_Parse_Shortcodes->parse_ticket_list_template(Array, NULL, Array, Array) 
#1 /event-espresso-core-reg/core/libraries/shortcodes/EE_Ticket_List_Shortcodes.lib.php(61): EE_Ticket_List_Shortcodes->_get_ticket_list_for_main() 
#2 /event-espresso-core-reg/core/libraries/shortcodes/EE_Ticket_List_Shortcodes.lib.php(44): EE_Ticket_List_Shortcode in /event-espresso-core-reg/core/helpers/EEH_Parse_Shortcodes.helper.php on line 97
```

Turns out the problem is with cancelled registrations if you have a registration on that same event and use [TICKET_LIST] in some locations, for example in the main content of the Registration Approved template.

To reproduce create add `[TICKET_LIST]` (or `[RECIPIENT_TICKET_LIST]`) to the main content of the registration approved message template (registrant context).

Register onto a single event with more than 1 ticket type on a single datetime and register onto both (or more) tickets in a single group. E.g - https://monosnap.com/file/XxR9aINwnoEU91BHfapAPffh5ZV0Qw

Pay and then cancel just 1 registration within the admin.

Now resend the registration email on any of the registration except the one you just cancelled.

**Additional context**
Just to let know you know what I found, I've traced this back to the lines items here: https://github.com/eventespresso/event-espresso-core/blob/master/core/libraries/messages/data_class/EE_Messages_incoming_data.core.php#L422-L430

The reg_obj loop [HERE](https://github.com/eventespresso/event-espresso-core/blob/master/core/libraries/messages/data_class/EE_Messages_incoming_data.core.php#L337-L341) works correctly and has all of the registrations (including the cancelled one) but skips the cancelled registration when setting up `$tickets`

The line items code I linked to above pulls the line items for ALL of the tickets on that event, so when it loops over the cancelled ticket it adds another element to `$tickets` that contains just a `line_item` and `sub_line_itmes`.

nerrad:

This (hopefully) fixes #971.  Thanks to the excellent troubleshooting done by @Pebblo, there is a flaw where line items were being assigned to the `$tickets` array in `EE_Messages_incoming_data::_assemble_data` even though the $ticket itself may not be assigned to that array because it belonged to a cancelled registration.

The fix involves simply making sure we _only_ setup ticket line items if there is a ticket already in the prepared array.  This _should_ fix the reported issue.

## How has this been tested

Testing should involve the reproduction steps outlined in #971.  However, I was unable to reproduce the fatal error using those steps so the fix here is a bit of a shot in the dark.  I have thus not verified it actually fixes the issue myself.

Should be noted that this change _does_ impact:

* [x] Any message types using line item data (so invoices/receipts)
* [x] Any usage of the line item shortcodes (see `EE_Line_Item_List_Shortcodes`)

A wide variety of test conditions (free tickets/paid ticket, multiple tickets across multiple datetimes, very simple registrations, MER etc) should be covered in user testing to expose any potential issues with this fix.

## Checklist

* [x] My code is tested.


Pebblo:
I spent a while just going over various event setups and confirming that messages worked even with the TICKET_LIST in place to reproduce the error condition and then confirmed that the Invoice & Receipt loaded as expected.

No issues found.

Issue #954, opened 2019-02-11
Bug in basic auth implementation causes WordFence to flag normal users as brute force attempts


360Zen:
I have Event Espresso running on a staging site with HTTP authentication (nginx basic auth) set up to protect it from public viewing. For example purposes, we'll say the username for this authentication is "staging". This staging site also uses WordFence to provide brute force protection from repeated login attempts under the same username. Unfortunately, Event Espresso appears to be breaking this brute force protection. 

I believe it is because of some code in this file: https://github.com/eventespresso/event-espresso-core/blob/master/core/third_party_libs/wp-api-basic-auth/basic-auth.php. On lines 45-47, the HTTP credentials are retrieved in place of standard login credentials. Then, on line 87, there is an attempt to authenticate with these credentials using wp_authenticate(). So, with all of these things combined, when using the staging site, any non-logged-in page views are flagged by WordFence as login attempts with the username "staging".

I noticed that this substitution is not done in the original library being used here: https://github.com/WP-API/Basic-Auth/blob/master/basic-auth.php. Is there any reason to try to use HTTP credentials as a substitute for proper WP login credentials?

joshfeck:
I believe so, yes, the comments in Event Espresso's version of basic auth explain:

//account for issue where some servers remove the PHP auth headers
	//so instead look for auth info in a custom environment variable set by rewrite rules
	//probably in .htaccess
    //and, as a last resort, look in the querystring

What we had found is many servers would not allow basic auth to authenticate with the EE4 apps. So the workarounds that you noticed were added in later, starting with some code that was offered in https://github.com/WP-API/Basic-Auth/issues/1 and iterated on later.

What you could do is install the original basic-auth plugin and the `function_exists()` checks in Event Espresso 4 core will make sure EE's version doesn't load.

360Zen:
I understand that this was done for a reason, but I think it was a mistake to close this bug. This issue causes WordPress to attempt to authenticate a server-level user over, and over, and over again on every page load. That's abusive behavior in terms of server resources, and has potential security implications as well, since a user could authenticate simply using HTTP authentication, while bypassing WP's normal cookie + POST data system. What if my HTTP auth credentials happened to be the same as one of my actual WP users? Then we'd be in real trouble.

I'd request that this issue be re-opened and treated as a security complaint.

joshfeck:
Hi,

There's already discussion/planning elsewhere to:
1) Disable the Auth mechanism by default
[...]

PR #974, opened 2019-02-21
dont enable basic auth on non-rest-routes

mnelson4:

This makes the brute force attack vector smaller by only enabling WP basic auth on REST routes, not all routes. Related: https://github.com/eventespresso/event-espresso-core/issues/954. This way there are fewer routes on which HTTP basic auth can be used for brute force attacking a site.

## How has this been tested

* [x] Verify the mobile apps still work with this branch
* [x] Setup a site with basic auth (see https://docs.nginx.com/nginx/admin-guide/security-controls/configuring-http-basic-authentication/). On master, if the nginx username+password is the same as WP's, when you visit the frontend, the admin topbar will appear, even if you didn't log in via the normal WP dashboard login. On this branch though, the admin topbar won't appear.
* [x] Verify no conflicts with standard basic auth plugin, WooCommerce, Application Passwords
* [x] Test a authenticated request with Postman
## Checklist

* [ ] My code is tested.


core/EE_System.core.php
```diff
@@ -394,6 +394,7 @@ public function load_espresso_addons()
                 array('activate', 'activate-selected'),
                 true
             )
+            && ($this->request->isWordPressApi() || $this->request->isApi())
```
tn3rb: this is great but would be slightly better if this conditional block was the first one in the entire set, since it is the easiest and therefore fastest to evaluate, ie:

```php
        if (($this->request->isWordPressApi() || $this->request->isApi())
            && $this->request->getRequestParam('activate') !== 'true'
            && ! function_exists('json_basic_auth_handler')
            && ! function_exists('json_basic_auth_error')
            && ! in_array(
                $this->request->getRequestParam('action'),
                array('activate', 'activate-selected'),
                true
            )
        ) {
            include_once EE_THIRD_PARTY . 'wp-api-basic-auth' . DS . 'basic-auth.php';
        }
```

mnelson4: y kk!

mnelson4:
Updated with your recommendation @tn3rb.

joshfeck:
Tested, ready for merging.

joshfeck:
@mnelson4 I think we might need to revert this for now because I went to test the app for another PR and the app would not authenticate. If I remove what was added here, the app works again.

I'm a bit puzzled why this all worked a month ago. I remember testing the apps and they worked OK then.

joshfeck:
Update: I found why the app wouldn't connect to my test site: Permalinks settings were set to Default (instead of post name). What's strange is I thought the REST API didn't work when Permalinks Settings were set to Default, but apparently they do now. So the change in this PR only breaks the apps if the site's Permalinks settings are set to Default. 

I guess we don't need to revert this as it's not common for Permalinks to be set to default. Strange eh?

mnelson4:
Yeah that's strange, I'd like to see if I can reproduce and find the cause...

mnelson4:
I found the issue and made PR https://github.com/eventespresso/event-espresso-core/pull/1069

Issue #811, opened 2018-12-06
REST API cant filter datetimes by deleted field


mnelson4:
## Issue Overview

Queries like this always return an empty set: `/datetimes?where[DTT_deleted]=false` returns an empty set (even when there are definetely non-trashed datetimes) and so does `/datetimes?where[DTT_deleted]=true`.

## Bug report or feature request?

* [x] Bug
* [ ] Feature
* [ ] Neither

## Environment Data:

Version of EE: 
4.9.71

Version of WordPress:
4.9.7
PHP Version:  
7.1
Browser used:  
Firefox/postman

## Steps to Reproduce (for bugs)

1. Create an event with two datetimes, and trash one.
2. Send a request to `wp-json/ee/v4.8.36/datetimes`, you should see the one untrashed one. Then add `?where[DTT_deleted]=false` and somehow you'll get an empty set. Then change it to `?where[DTT_deleted]=true` and it's still empty. Weirdness.

## Expected Behaviour

`/wp-json/ee/v4.8.36/datetimes?where[DTT_deleted]=false` should return undeleted datetimes. And instead using `where[DTT_deleted]=true` should... probably tell you you're not allowed to see deleted datetimes, as they're not visible on the frontend; or maybe show them. It's up for debate, but probably an error would be better.

## Current Behaviour

See steps to reproduce.

joshfeck:
A request to:
`wp-json/ee/v4.8.36/datetimes/?where[DTT_deleted]=0&include=DTT_id`
will get the expected results (use 0 instead of false)

However, it looks like there are some other related bugs with the models or the client code for the UI. What happens is after I trash a datetime in the UI, then look in the database, the DTT_deleted field is still 0 (should be 1) **and** the EVT_ID field is change to 0. 

Along with the above, there's a confusing error message that's shown when you trash a datetime:
> You cannot add this datetime to this ticket because the ticket has a sold amount that is greater than the amount of spots remaining on the datetime.
Which makes no sense in the context of trashing a datetime.

So it appears there's an issue where `false` isn't interpreted as 0, and then some other issues when a ticket is trashed in the event editor UI. 

Maybe we need a few more issues for these other items?

sethshoultes:
Ya, let's create issues for the other items.

mnelson4:
Huh, ya some nutty behaviour there when trashing datetimes. I think that admin code is more Darren's baby. So I'm actually not sure how to create a trashed datetime. 
[...]

PR #977, opened 2019-02-21
Bug/boolean input interpreted as string in rest api

mnelson4:

The strings "false" and "true", when provided for REST API parameters for boolean model fields, are interpreted as booleans instead of strings. This addresses the mistake I made on https://github.com/eventespresso/event-espresso-core/issues/811, so I figured others were likely to make it too.

## How has this been tested

Unit tested.
* [x] Create an event, register for it, change the ticket's price, which will archive the ticket and create a new one. Then send an authenticated request using the REST API to `http://yoursite.com/wp-json/ee/v4.8.36/tickets?where[TKT_deleted]=true`. Your archived ticket should get returned. 

## Checklist

* [x] My code is tested.


joshfeck:
Tested and the results were the expected results.

PR #987, opened 2019-02-27
Clean up a few things in the skeleton add-ons

joshfeck:
Originally reported in https://events.codebasehq.com/projects/event-espresso/tickets/8223. The part about the main add-on class autoloading itself was fixed somewhere along the way. 

This PR cleans up the unit test class names & makes the example responses for the Onsite gateway sound slightly more professional.


mnelson4:
Oh also, I don't think this needs to be tested seeing how it doesn't affect any released code. It can be merged as-is.

mnelson4:
I've put this in the milestone, although I'm not 100% sure that's the right thing to do.

joshfeck:
Yeah since this isn't packaged up with the release it's not going into the milestone

PR #988, opened 2019-02-27
Fet/demo copy date functionality

nerrad:
App for demonstrating using the new [relation type awareness in the stores](https://github.com/eventespresso/event-espresso-core/pull/982).  

- Go to WP Data Demo tab in the Event Admin
- Enter an Event Id.

Very raw, very proof-of-concept.


assets/src/demo/app.js
```diff
@@ -0,0 +1,188 @@
...
+		);
+		createRelations(
+			'event',
+			ownProps.eventId,
+			'datetimes',
+			[ newDateTime ]
+		);
+	}
+	return { onDateClone };
+} )( DatesList );
```
nerrad: Note how DatesList (which is responsible for outputting the list of datetimes and the clone button) has the `withDispatch` HOC wrapping it.  This HOC primarily takes care of the clone button behaviour which when triggered:

* Retrieves related tickets on the datetime clicked.
* clones the new Datetime entity from the one clicked and creates the relation to tickets on that new entity
* creates the relation to the event id on that new entity (I _could_ have also just retrieved the event from the store state if I wanted to).

Note, this does not feed the `datetimes` array to the component as that will automatically get updated further upstream.

assets/src/demo/app.js
```diff
@@ -0,0 +1,188 @@
...
+	function getContent() {
+		return isInitializing ?
+			<Spinner /> :
+			<Fragment>
+				<DatetimesList eventId={ eventId } datetimes={ datetimes } />
+				<TicketsList tickets={ tickets } />
+			</Fragment>;
+	}
+	return <div>{ getContent() }</div>;
+};
```
nerrad: Note the implementation of the `isInitializing` property (which is actually a state via the `withState` HOC) that allows for hiding the datetime and ticket lists until the main query is done.

assets/src/demo/app.js
```diff
@@ -0,0 +1,188 @@
...
+		const { isInitializing } = ownProps;
+		const { getEntitiesForModel } = select( 'eventespresso/core' );
+		if ( isInitializing ) {
+			return ownProps;
+		}
+		return {
+			datetimes: getEntitiesForModel( 'datetime' ),
+			tickets: getEntitiesForModel( 'ticket' ),
+		};
+	} ),
```
nerrad: This `withSelect` HOC takes care of retrieving the datetimes and tickets used for the display but only when `isInitializing` is false.  That way it can confidently use the general selectors once the state is populated by the main query.  From that point on any changes to the datetimes or tickets state will immediately get reflected in the rendered component.

assets/src/demo/app.js
```diff
@@ -0,0 +1,188 @@
...
+		}
+		const datetimeIds = Datetimes.map( ( datetime ) => datetime.id );
+		getRelatedEntitiesForIds( 'datetime', datetimeIds, 'tickets' );
+		const relatedEntitiesIsResolved = hasFinishedResolution(
+			'eventespresso/core',
+			'getRelatedEntitiesForIds',
+			[ 'datetime', datetimeIds, 'tickets' ]
+		);
+		setState( { isInitializing: false } );
+	} ),
```
nerrad: This `withSelect` HOC is responsible for populating the state with all the datetimes and tickets related to the selected EventId.  The state can be populated using the new efficient queries (getRelatedEntitiesForIds).  

The sole purpose of this HOC is to populate the state and set the `isInitializing` local state.  `isInitializing` is then used by the other hocs as a signal for the state being populated and thus other component rendering can occur from whats in the store state.

Once this HOC has populated the state, it effectively is "done", hence the early bail on line 131.  Note, this currently does not handle a change in eventId via the input (again just a raw demo).  If I was to complete this, I'd have the `onEventClick` action reset the local `isInitializing` state and clear the `eventespresso/core` state of all records from the original event.

tn3rb:
closing this as it is no longer needed

PR #989, opened 2019-02-28
Bug/fix to persist not returning expected values

nerrad:

While working on REM, @tn3rb discovered a bug with the the store persisting entity changes to the db.  In troubleshooting, we discovered that `toPersist` was working as expected for the `BaseEntity` instances, but remained complicated to use for the use-case it was in use for.  In the original implementation, the design was that `forPersist` would only return fields and values from the entity for fields the entity was instantiated with.  The intent being, that when the entity is persisted, any _defaults_ would be set by the database vs client side (non included fields in an insert request would get have default values determined by the server on saving to the db).  The problem with this approach is that it's reasonable to expect consuming code to create a "blank" entity and then use the setters to set the values at a later point.  When `forPersist` is called later however, those new set values would not get persisted.

The solution in this pull involved:

- Creating a new property for `BaseEntity` labelled `fieldsToPersistOnInsert`.  This property automatically captures the initial fieldNames passed in on instantiating the `BaseInstance` as well as any additional fieldNames that have a value set to their property through the life of the instance.  The value of `fieldsToPersistOnInsert` is a javascript [Set](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Set) object which ensures there's only ever unique field names in the iterator.
- When `forInsert` is called (either directly or via `forPersist`), it will iterate through `fieldsToPersistOnInsert` for determining what fields and values are exported from the instance.
- When `forUpdate` is called (either directly or via `forPersist`), it will use `Object.keys` on the `BaseEntity` instance as the iterator thus included all enumerable properties (and their values) on the export (which currently are just field names for the model).

## How has this been tested

* [x] Ensure existing unit-tests pass.
* [x] I wrote new tests covering the flaw that was exposed.  Ensure these tests pass.

While `BaseEntity` instances _are_ used in released code (EventAttendees block), I'm confident the existing tests cover usage there because:

a. `EventAttendees` block is read only, so the affected functions are not used.
b. Any impact due to instantiation of `BaseEntity` instances would surface immediately in the existing unit test coverage.

## Checklist

* [x] My code is tested.

### DIFF
--- a/assets/src/data/model/entity-factory/base-entity.js
+++ b/assets/src/data/model/entity-factory/base-entity.js
@@ -52,6 +52,11 @@ class BaseEntity {
 		);
 		createGetter( this, 'modelName', modelName );
 		createGetter( this, 'originalFieldsAndValues', entityFieldsAndValues );
+		createGetter(
+			this,
+			'fieldsToPersistOnInsert',
+			new Set( Object.keys( entityFieldsAndValues ) )
+		);
 		createEntityGettersAndSetters( this );
 		createPersistingGettersAndSetters( this );
 	}
--- a/assets/src/data/model/entity-factory/create.js
+++ b/assets/src/data/model/entity-factory/create.js
@@ -91,7 +91,7 @@ export const createGetterAndSetter = (
 	instance,
 	fieldName,
 	initialFieldValue,
-	opts = {}
+	opts = {},
 ) => {
 	let propertyValue = initialFieldValue;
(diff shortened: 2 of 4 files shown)


nerrad:
I merged this without complete test runs because they were just restarted due to a rebase from master.  The javascript tests passed successfully.  There was a fail in php tests but that was due to a time of day date fail that happens from time to time.

PR #990, opened 2019-02-28
Improve dependency tree for value objects (fixes #956)

### LINKED ISSUE #956: Review package dependencies for value-objects and make sure it's independent.
## Issue Overview

While working on #955 I discovered that rebuilding the vo touched more packages than I expected.  It would be preferable to have any changes to `@eventespresso/value-objects` be self contained.

This issue is created as a reminder to myself to investigate this and:

- Work out what needs to be done to ensure `@eventespresso/value-objects` can standalone (has no other `@eventespresso/*` dependencies (except for `@eventespresso/eejs`.
- Make sure all other packages are importing `@eventespresso/value-objects` for using vo's rather than importing directly (so its used as an external).
- Make sure the webpack config references the correct external for `@eventespresso/value-objects`
- Update the `CoreAssetManager` php class so that the `vo` package has the correct dependencies and also is listed as the correct dependency for other packages.

nerrad:

See #956 for original report.  This improves the build files fairly significantly (remember the below numbers are for dev builds, prod builds are smaller because they are minified.)

Size files:

| File | Before | After |
| --- | ------ | ------ |
| ee-model | 3.2mb | 1mb |
| ee-data-stores | 4.6mb | 2.4mb |

By changing the direct import of files to be referencing externals this resulted in significant size savings for a couple of the built files.  This also means that when `@eventespresso/value-objects` has changes it won't cause a rebuild of the above files.

## How has this been tested

* [x] ensure unit-tests pass.
* [x] This impacts the `EventAttendees`  block, so verify it works as expected.

## Checklist

* [x] My code is tested.


nerrad:
See testing notes.  The EventAttendees block should be tested to verify there's no regressions with this change.

PR #995, opened 2019-03-01
Added `$_FILES` global to Request

tn3rb:
The Attendee Importer add-on that @mnelson4 is working on requires CSV files to be uploaded, and therefore needs to have the `$_FILES` global data available. It was decided to add this to the existing `Request` class instead of accessing it directly.

## How has this been tested
unit tests

## Checklist

* [x all publicly displayed strings are internationalized (usually using `esc_html__()`, see https://codex.wordpress.org/I18n_for_WordPress_Developers)
* [x] My code is tested.


mnelson4:
FYI I double-checked and could still create an event and register for it on this branch. I'm not sure if there is any other manual testing you'd suggest @tn3rb?

tn3rb:
ya I don't think there is anything else necessary since the Request is created right at the start of the bootstrap process so if something was going to go wrong it would blow up on basically any request. On top of that, the new parameter is optional, so even if some third party code is trying to build a Request object on their own for some crazy reason, it should still not cause any troubles.

PR #998, opened 2019-03-04
Add reset state actions to our stores.

nerrad:

While working on the demo app (#988) and the Barcode Scanner app (https://github.com/eventespresso/eea-barcode-scanner/pull/7) I realized that there will be need for some sort of reset state action for our stores that allows dispatches to restore state to its default shape.  For instance with the barcode scanner app the user might scan a new registration for checkin that we'll want to reset all the tracked related registrations etc in the state.

Note: a [pull I did for GB](https://github.com/WordPress/gutenberg/pull/14225) recently that adds more sledgehammer like control over the resolution cache in the `core/data` store.  It will make it easier for resetting that state in this pull.  The downside is it will be tied to a version of WordPress so I can only use it IF the functions exist.

## How has this been tested

* [x] unit test coverage
* [x] no impact to customer facing code.

## Checklist

* [x] My code is tested.


assets/src/data/eventespresso/core/reducers/relations.js
```diff
@@ -320,6 +320,96 @@ const removeRelatedEntitiesForEntity = ( state, action ) => {
...
+};
+
+/**
+ * Helper for clearing the entries in the dirty relations state for the given
+ * modelName.
+ *
+ * This clears entries in both the `index` and `entityMap` maps.
+ *
+ * "Main state" is used in this function to represent the top level state the
+ * model's entries is being deleted from (i.e. entityMap or index).
```
tn3rb: should be:

> to represent the top level state the model's entries **are** being deleted from

assets/src/data/eventespresso/core/actions/reset.js
```diff
@@ -0,0 +1,238 @@
...
+import { ACTION_TYPES } from './action-types';
+import { REDUCER_KEY } from '../constants';
+import * as modelSpecificSelectors from '../model/model-selectors-index';
+
+const { resets: types } = ACTION_TYPES;
+
+/**
+ * Resets the entire state to its default for the store.
+ */
+export function* resetEntireState() {
```
tn3rb: all of the other function names for the reset actions mirror the constant names
and I think that would be a good precedent to follow here for consistency's sake.

so either:

- rename this function to `resetAllState()` 
- rename the constant to `RESET_ENTIRE_STATE`

depending on which one you think better describes the intent of the action,
I think my preference is maybe for the latter, but I'll leave that up to you

nerrad:
@tn3rb I implemented your suggestions but then decided to go ahead and merge this (after verifying js tests passed) because the changes don't really need reviewed again.

PR #1006, opened 2019-03-07
Fix checkins migration fatal error (fixes #983)

### LINKED ISSUE #983: Avoid Fatal error during Database Migration related to migrating the _events_attendee.checked_in column
**Describe the bug**

This error is logged when doing the migration from EE3 to EE4 if unexpected data is in the `_events_attendee` table's `checked_in` column:
`PHP Fatal error:  Uncaught Error: Cannot use object of type stdClass as array in //wp-content/plugins/event-espresso-core/core/data_migration_scripts/4_1_0_stages/EE_DMS_4_1_0_checkins.dmsstage.php:73`

**To Reproduce**

**Expected behavior**

The invalid data could be worked around and the migration allowed to continue. Worse case scenario the check in data will no be accurate.
**Screenshots**

screen shot 2019-02-26 at 11 08 50 am

screen shot 2019-02-26 at 11 09 12 am

**Environment:**
 - Version of Event Espresso:   4.9.78
 - Version of WordPress:  5.1
 - PHP Version  7.1.25

joshfeck:
See #983 

@mnelson4 there may be a better way to address this, like telling $wpdb->get_row() to output an array instead. This PR should at least get the ball rolling. 

## How has this been tested

- [x] Do a migration from EE3 for a site that has checkins, on a server that has PHP 7+

## Checklist

* [x] My code is tested.


mnelson4:
Thanks @joshfeck for getting this going. Ya your fix was going to work fine, but in the course of reviewing I noticed the variable `$is_checked_in` (which we were setting when the fatal error occurred) isn't actually used! Not sure how that happened. So modified the code to reflect what I believe was the purpose: don't add another checkin for in EE4 if one already existed for this EE4 registration.

I'd be happy to pass back to you for review, but I believe the process is to pass to Brent or Darren.

core/data_migration_scripts/4_1_0_stages/EE_DMS_4_1_0_checkins.dmsstage.php
```diff
@@ -97,7 +112,7 @@ private function _try_to_find_datetime($old_attendee)
 
         $new_event_id = $this->get_migration_script()->get_mapping_new_pk($wpdb->prefix."events_detail", $old_attendee['event_id'], $wpdb->posts);
         if (! $new_event_id) {
-            $this->add_error(sprintf(__("Could nto find new event ID with old event id '%d', on attendee row %s; and because of that couldnt find the correct datetime for Check-in", "event_espresso"), $old_attendee['event_id'], $this->_json_encode($old_attendee)));
+            $this->add_error(sprintf(esc_html__('Could not find new event ID with old event ID %d, on attendee row %s; and because of that couldn\'t find the correct datetime for Check-in', 'event_espresso'), $old_attendee['event_id'], $this->_json_encode($old_attendee)));
```
tn3rb: since this translation string will break due to the typo fix anyways, let's also switch it over to using numbered placeholders (and format it), ie: 

```php
$this->add_error(
    sprintf(
        esc_html__(
            'Could not find new event ID with old event ID %1$d, on attendee row %2$s; and because of that couldn\'t find the correct datetime for Check-in', 
            'event_espresso'
        ),
        $old_attendee['event_id'],
        $this->_json_encode($old_attendee)
    )
);
```

joshfeck:
Verified Mike's fixes, this is ready to be merged.

PR #1010, opened 2019-03-08
Fet/attendee importer

mnelson4:
Work needed for attendee importer. It's tricky committing in isolated chunks because I don't want to commit something until it's shown to be helpful. Once it's shown to be useful we can break into useful chunks if needed.

## How has this been tested
* [x] On master branch, create a ticket whose price is crazy complex (like with lots of price modifiers, taxes, etc; and maybe toss some promotion discounts in there), then register for it (maybe even register multiple registations, and initially choose Invoice, then approve one or two of them, and pay for just those few). Then switch to this branch and re-register, and verify the behaviour remains the same.
* [x] Install and activate the EE4 files add-on plugin (3rd party add-on on wp.org). Verify this branch doesn't break it.
* [x] Verify you can still create registration CSV reports (using the standard batch system; not the legacy system you can swap back-in with a filter)

## Checklist

* [ ] My code is tested.


core/services/request/files/FileSubmission.php
```diff
@@ -0,0 +1,182 @@
+<?php
```
mnelson4: this is part of https://github.com/eventespresso/event-espresso-core/pull/999 which was already reviewed (just haven't gotten around to merging it)

modules/batch/EED_Batch.module.php
```diff
@@ -93,6 +103,21 @@ public static function set_hooks_admin()
...
     }
 
+    /**
+     * @since $VID:$
+     * @return LoaderInterface
+     * @throws InvalidArgumentException
+     * @throws InvalidDataTypeException
+     * @throws InvalidInterfaceException
+     */
+    protected function getLoader()
```
mnelson4: Can't override `EED_Module::__construct()` (because it's `final`) to inject this, so just using this getter instead.

core/services/locators/FqcnLocator.php
```diff
@@ -148,16 +148,27 @@ protected function getDirectoryFromPartialNamespace($partial_namespace)
...
+            $prefix = $psr4_loader->prefixes($namespace_to_try . Psr4Autoloader::NS);
+            // nope? then the incoming namespace is invalid
+            if (! empty($prefix) && ! empty($prefix[0])) {
+                break;
+            }
+            array_unshift(
+                $removed_namespace_segments,
+                array_pop($namespace_segments_to_try)
+            );
+        }// nope? then the incoming namespace is invalid
```
tn3rb: My original code only took the Vendor portion of the namespace into consideration,
but we should really only ever need to test the first two elements of the namespace segments against registered prefixes,
so i think this could be simplified for both readability and efficiency...

how about:

```php
    protected function getDirectoryFromPartialNamespace($partial_namespace)
    {
        if (empty($partial_namespace)) {
            throw new InvalidClassException($partial_namespace);
        }
        // load our PSR-4 Autoloader so we can get the list of registered namespaces from it
        $psr4_loader = \EE_Psr4AutoloaderInit::psr4_loader();
        // breakup the incoming namespace into segments so we can loop thru them
        $namespace_segments = explode(Psr4Autoloader::NS, trim($partial_namespace, Psr4Autoloader::NS));
        // we're only interested in the Vendor and secondary base, so pull those from the array
        $vendor_base = array_slice($namespace_segments, 0, 2);
        $namespace = $prefix = null;
        while (! empty($vendor_base)) {
            $namespace = implode(Psr4Autoloader::NS, $vendor_base);
            // check if there's a base directory registered for that namespace
            $prefix = $psr4_loader->prefixes($namespace . Psr4Autoloader::NS);
            if (! empty($prefix) && ! empty($prefix[0])) {
                // found one!
                break;
            }
            // remove base and try vendor only portion of namespace
            array_pop($vendor_base);
        }
        // nope? then the incoming namespace is invalid
        if (empty($prefix) || empty($prefix[0])) {
            throw new InvalidClassException($partial_namespace);
        }
        $this->setNamespace($namespace, $prefix[0]);
        // but if it's good, add that base directory to the rest of the path, and return it
        return $prefix[0] . implode(DS, array_diff($namespace_segments, $vendor_base)) . DS;
    }
```

modules/batch/EED_Batch.module.php
tn3rb: ya that was my bad

core/helpers/EEH_Line_Item.helper.php
```diff
@@ -1743,14 +1743,14 @@ public static function calculate_final_price_for_ticket_line_item(
...
+        if (empty($final_prices_per_ticket_line_item) || empty($final_prices_per_ticket_line_item[$total_line_item->ID()])) {
+            $final_prices_per_ticket_line_item[$total_line_item->ID()] = EEH_Line_Item::calculate_reg_final_prices_per_line_item(
                 $total_line_item
             );
         }
         // ok now find this new registration's final price
-        if (isset($final_prices_per_ticket_line_item[ $ticket_line_item->ID() ])) {
-            return $final_prices_per_ticket_line_item[ $ticket_line_item->ID() ];
+        if (isset($final_prices_per_ticket_line_item[$total_line_item->ID() ][ $ticket_line_item->ID() ])) {
+            return $final_prices_per_ticket_line_item[$total_line_item ][ $ticket_line_item->ID() ];
```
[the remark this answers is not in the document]
mnelson4: you're right! I forwarded the physical aggression by giving my phpstorm settings a slap on the wrist for this :fist_left:

core/libraries/form_sections/inputs/EE_File_Input.input.php
```diff
@@ -0,0 +1,115 @@
...
+
+    /**
+     * @param array $options
+     * @throws InvalidArgumentException
+     */
+    public function __construct($options = array())
+    {
+        if (isset($options['allowed_file_extensions'])) {
+            if (!is_array($options['allowed_file_extensions'])) {
+                throw new InvalidArgumentException(esc_html__('A valid allowed_file_extensions array was not provided to EE_File_Input', 'event_espresso'));
```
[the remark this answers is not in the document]
mnelson4: sure. I think I can just cast to an array right? eg `$this->allowed_file_extensions = (array) $options['allowed_file_extensions'];`

core/libraries/form_sections/inputs/EE_File_Input.input.php
```diff
@@ -0,0 +1,115 @@
...
+            }
+            $this->allowed_file_extensions = $options['allowed_file_extensions'];
+        } else {
+            $this->allowed_file_extensions = ['csv'];
+        }
+        if (isset($options['allowed_mime_types'])) {
+            if (!is_array($options['allowed_mime_types'])) {
+                throw new InvalidArgumentException(esc_html__('A valid allowed_mime_types array was not provided to EE_File_Input', 'event_espresso'));
```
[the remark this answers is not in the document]
mnelson4: :eyes:

core/libraries/form_sections/inputs/EE_File_Input.input.php
```diff
@@ -0,0 +1,115 @@
...
+            . ' extension="'
+            . implode(
+                ',',
+                array_map(
+                    function ($file_extension) {
+                        return  $file_extension;
+                    },
+                    $this->allowed_file_extensions
+                )
+            )
```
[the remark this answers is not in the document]
mnelson4: Well, you see... :door: :eyes: what in the world could that be?! :running_man: 

(I have no idea why this happened, I assume it needed some more complicated logic at one point)

core/libraries/form_sections/strategies/normalization/EE_File_Normalization.strategy.php
```diff
@@ -0,0 +1,52 @@
...
+     * @param string $value_to_normalize
+     * @return FileSubmissionInterface
+     */
+    public function normalize($value_to_normalize)
+    {
+        if ($value_to_normalize instanceof FileSubmissionInterface || is_null($value_to_normalize)) {
+            return $value_to_normalize;
+        } else {
+            throw new EE_Validation_Error(
+                esc_html__('The file input has an inavlid format.', 'event_espresso')
```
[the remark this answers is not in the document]
mnelson4: thnaks! (typo intended, that time)

core/services/locators/FqcnLocator.php
mnelson4: works like a charm. Good point!

modules/batch/EED_Batch.module.php
mnelson4: ...although thinking about it, I suppose we could change that and it would be backward compatible, right? WDYT?

tn3rb: ya nothing would break if the `final` declaration was removed, but modifying the constructor can be tricky

core/libraries/form_sections/inputs/EE_File_Input.input.php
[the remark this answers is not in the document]
tn3rb: i spent way too long looking at that trying to figure out if it did something else that I was just missing (wouldn't be the first time). I too assumed that you had either had different logic at one point that then got changed or are just suffering from that goold old newborn sleep deprivation.

core/services/request/files/FileSubmissionInterface.php
```diff
@@ -5,7 +5,11 @@
 /**
  * Class FileSubmissionInterface
  *
+<<<<<<< HEAD
  * Data concerning a file that was submitted via an HTML form.
+=======
+ * Description
+>>>>>>> master
```
tn3rb: oops! unresolved conflict here

mnelson4: :dizzy_face: oups!! Good catch.

mnelson4:
Thanks for the catch @tn3rb ! But what do you think about my comment on https://github.com/eventespresso/event-espresso-core/pull/1010#discussion_r270574501 ?

tn3rb:
> But what do you think about my comment

I already responded a few days ago... GitHUb might be hiding it...

mnelson4:
> I already responded a few days ago... GitHUb might be hiding it...

I think maybe you replied to that BEFORE doing the review so it didn't appear as part of the review (so I missed it). But thanks, I'll pass this off for review as-is.

mnelson4:
I noticed a warning while registering using this branch, which I fixed in the above commit, but it wouldn't be bad to re-review.

core/helpers/EEH_Line_Item.helper.php
```diff
@@ -1749,7 +1749,10 @@ public static function calculate_final_price_for_ticket_line_item(
             );
         }
         // ok now find this new registration's final price
-        if (isset($final_prices_per_ticket_line_item[ $total_line_item->ID() ][ $ticket_line_item->ID() ])) {
+        if (isset(
+            $final_prices_per_ticket_line_item[ $total_line_item->ID() ],
+            $final_prices_per_ticket_line_item[ $total_line_item->ID() ][ $ticket_line_item->ID() ]
```
tn3rb: this shouldn't be necessary as PHP checks each index sequentially, so the previous single line of code should do what you want.

from the PHP docs:

```php
$a = array (
    'test' => 1,
    'hello' => NULL,
    'pie' => array(
        'a' => 'apple'
    )
);

var_dump(isset($a['test']));            // TRUE
var_dump(isset($a['foo']));             // FALSE
var_dump(isset($a['hello']));           // FALSE

// The key 'hello' equals NULL so is considered unset
// If you want to check for NULL key values then try: 
var_dump(array_key_exists('hello', $a)); // TRUE

// Checking deeper array values
var_dump(isset($a['pie']['a']));        // TRUE
var_dump(isset($a['pie']['b']));        // FALSE
var_dump(isset($a['cake']['a']['b']));  // FALSE
```

notice in the last block of code that they did not have to do:

```php
isset($a['pie'], $a['pie']['a']);
```

(EDIT)
Just saw your commit note now (wasn't visible from the code review screen) 

> check both the top-level item and the lower item are set. Because isset has a bad habit of giving the wrong result when you ask for both at the same time

really? how so? I have not experienced this (not saying it is not the case).

I looked through the comments on https://www.php.net/manual/en/function.isset.php
and found the following scenarios that may be the kind of cases you are referring to:

 - https://www.php.net/manual/en/function.isset.php#112046
 - https://www.php.net/manual/en/function.isset.php#47396

the first case deals with objects that are using `__isset()` 
and the second case deals with string values getting converted to arrays and returning unexpected results, which I'm guess is from before PHP 5.4 considering the comment is 14 years old.

Are either of these what you are referring to? Or are you seeing something else?

mnelson4: oh, ya I got a warning and incorrectly assumed it was because we were checking both the top-level array key and the lower one simultaneously. But like you said, that's actually not the problem. The error only coincidentally went away when I added this change.

(And ya, I was vaguely recalling php.net/manual/en/function.isset.php#112046, but got it mixed up.)

Looking at it more closely, the warning was

> Warning: Illegal offset type in E:\Software\laragon\www\laragonee\wp-content\plugins\event-espresso-core\core\helpers\EEH_Line_Item.helper.php on line 1753

Which wasn't even that exact line... all sorts of oups.

The latest commit fixes the actual problem. For realz this time.

mnelson4:
Reverted previous commit (see https://github.com/eventespresso/event-espresso-core/pull/1010#discussion_r272268079) and added another one which does the trick better (I was trying to use an object as an array key :sob: )

mnelson4:
Passing off for testing. I added some testing notes on the original PR description eh.

mnelson4:
Merged in some changes Brent did for the attendee importer.

joshfeck:
Verified the checklist, all is good.

PR #1015, opened 2019-03-11
Remove Google + option and shortcode in default templates Fixes #931

### LINKED ISSUE #931: Remove Google + options and default templates usage
There's an option in Event Espresso > General Settings for Google plus, and the default message templates has some boilerplate that use the Google plus shortcode. 

I think we'll need to keep the message shortcode around for awhile, but we can remove its usage from the default templates.

joshfeck:

See #931 and https://www.blog.google/technology/safety-security/project-strobe/

## How has this been tested

- [ ] Go to the General Settings page and verify the Google + input is removed

- [ ] Reset message templates (or start with EE plugin activation on new site) to see no usage of `[CO_GOOGLE_URL]` shortcode in the default message templates 
## Checklist

* [x] My code is tested.


kristinoc:
tested and Google + input is not showing in Gen Settings social profile area

Message template options still showing

Steps: 
I reset all data from the Maintenance tab and reactivated EE.
I also used the Reset Templates option when viewing a default message. 

[CO_GOOGLE_URL] shortcode in the default message templates is still showing

ticket931

joshfeck:
It's actually expected that the [CO_GOOGLE_URL] shortcode would still be available to use. We're not removing it as a shortcode because that would break existing templates. 

The actual change that was made was removing any _usage_ of the shortcode from the default templates. In other words, sites can still use the [CO_GOOGLE_URL] shortcode in messages, the shortcode just won't be included in the contents of default templates.

PR #1023, opened 2019-03-14
Improve help text for Default Registration Status option within event editor

joshfeck:

The current help text for this option, within the event editor, isn't helpful. This Pull Request adds more detailed help text, and a link next to the option's label. The same help text already exists in the default settings page.
Screen Shot 2019-03-14 at 3 34 38 PM

## How has this been tested

View the event editor, click the new link next to the label, and verify the new help text that's available to view after clicking the link.

* [x] My code is tested.


joshfeck:
Still to do: Take this opportunity to re-work the information for "Active Status" section of the same help tab. The current list is kind of long so the actual setting information gets buried. Also, the Active Status isn't really an option, it's really an indicator. So I'm thinking of changing the help text to say it's an indicator based on if the event is past, present, future, or if a special custom status is set in the "update event" box"

sethshoultes:
The changes look good and make sense

PR #1024, opened 2019-03-15
Fix bugs in js data stores exposed by work on REM

nerrad:
While working on REM, Brent discovered another significant bug surfacing in causing a callstack error in the `@wordpress/data` module.  After debugging this I discovered the cause of the bug was BaseEntity instances getting mutated and having a circular reference added to them.  Where this was happening was hard to track down but eventually I discovered it by sealing BaseEntity instances after they are created (which I should have had in place to begin with).

In the process of fixing this bug I also cleaned up a couple other things I discovered in the store.

## How has this been tested

* [x] Ensure existing unit tests don't fail (any issues with adding `Object.seal` would surface in existing tests.
* [x] Ensure bug Brent reported was fixed in a cherrypicked branch with these changes applied.
* [x] Ensure existing `EventAttendees` block works as expected.

## Checklist

* [x] My code is tested.


assets/src/data/eventespresso/core/resolvers/relations.js
```diff
@@ -127,9 +127,11 @@ export function* getRelatedEntities( entity, relationModelName ) {
...
 	if ( ! isEmpty( existingEntities ) ) {
 		fullEntities = keepExistingEntitiesInObject(
-			existingEntities.map(
-				( entitiesObject, entityObj ) =>
-					entitiesObject[ entityObj.id ] = entity,
+			existingEntities.reduce(
+				( entitiesObject, entityObj ) => {
+					entitiesObject[ entityObj.id ] = entity;
+					return entitiesObject;
+				},
```
nerrad: This is the actual fix for the issue you discovered @tn3rb.   Crazy stuff and not at all apparent when simply manually reviewing code.  The debugging was showing a new property getting added to TicketEntities of `undefined` and the value was a _DateTimeEntity_ instance!  So I was looking for where that would explicitly be getting set.  Completely missed this in my manual reviews and only surfaced once I sealed the BaseEntity instances.

assets/src/data/model/entity-factory/base-entity.js
```diff
@@ -59,6 +59,7 @@ class BaseEntity {
 		);
 		createEntityGettersAndSetters( this );
 		createPersistingGettersAndSetters( this );
+		Object.seal( this );
```
nerrad: This has the added benefit of making sure there won't be a repeat of the snafu in the store.

PR #1025, opened 2019-03-15
Make action constants for admin-routes simpler (js-library)

nerrad:

I'm currently implementing `eejs.routes` constants in the barcode scanner work and I've discovered that the action constants are more verbose than they need to be.

Here's an example of before/after to get the checkin action url for the registrations admin pages:

**Before**

```js
const adminRoute = getAdminUrl( 
    ADMIN_ROUTES.REGISTRATION,
    ADMIN_ROUTE_ACTIONS[ ADMIN_ROUTES.REGISTRATIONS].EVENT_CHECKIN
);
```

**After**

```js
const adminRoute = getAdminUrl( 
    ADMIN_ROUTES.REGISTRATIONS,
    ADMIN_ROUTE_ACTIONS.REGISTRATIONS.EVENT_CHECKIN
);
```

## How has this been tested

* [x] verify no existing unit tests fail
* [x] verify works as expected via monkey testing in the browser
* [x] Doesn't impact `EventAttendees` block but I've verified that anyways.

## Checklist

* [x] My code is tested.


nerrad:
merged before all tests were run because its just the js tests that matter.

PR #1028, opened 2019-03-20
Fix getAdminUrl helper for `eejs.routes` returning incorrect string

nerrad:

I noticed while implementing this method in the barcode scanner app work I'm doing that the url is returning the incorrect value.

## How has this been tested

* [x] in the Barcode scanner app
* [x] (fixed) unit test.

## Checklist

* [x] My code is tested.


nerrad:
merging to master with just js tests passing because this only affects js (not php)

PR #1029, opened 2019-03-21
Make select2 inputs wider by default

mnelson4:

Makes the EE inputs using select2 wider by default. Previously they defaulted to being only 100px wide (although the attendee mover, the main add-on that comes to my mind that used this input, overrode this).
This is done by adding a CSS class ("ee-select2") onto inputs using the display strategy `EE_Select2_Display_Strategy`, and adding some styling for it, and removing the default width of 100 set in `EE_Select_Ajax_Model_Rest_Input`.
Josh noted this on https://github.com/eventespresso/eea-attendee-importer/issues/16, under "Step 3". And I set about fixing this generally rather than just fixing it in the two usages of select2 inputs in that add-on.

It's probably best to have a general fix than having every plugin that uses these inputs need to override them.

## How has this been tested

* [ ] Use this branch and the attendee mover. The input select2 inputs (to select the event, or the ticket) should be the same width as before.
* [ ] Can you think of anywhere else in EE that we have select2 inputs? If so they should work fine too (Mike is currently using for the attendee importer)
* [ ] Select inputs (like those created in SPCO from questions) should have the same HTML as before (the only difference being that if the class or style are empty, they don't get outputted, eg instead of `<select style="" class="">` we now have `<select>`)

Also, I temporarily removed the attendee mover's styles that fixed the select2 inputs (the CSS style `.select2{min-width:75% !important;}` in `espresso_attendee_mover.css`) and the select2 input, as expected, [looked really small with EE core master](https://drive.google.com/a/eventespresso.com/file/d/1NNIuYSJDIzDoaAgqvI63mDUUCFjVrkNM/view?usp=drivesdk). But with this branch of EE core, [that same input looks good again](https://drive.google.com/a/eventespresso.com/file/d/1CO53zaqYoRfqzHJ1_wamtrbqXSqbiVT5/view?usp=drivesdk) (and again, without needing the attendee mover's custom styling). So, by default, select2 inputs look better with these changes.

## Checklist

* [x] My code is tested.


joshfeck:
Looks OK to merge.

PR #1030, opened 2019-03-21
default to having select2s using the REST API to use descending order

mnelson4:

Makes `EE_Select2_Ajax_Model_Rest_Input`s default to showing items in descending order, so most recently created items appear first by default. (As Josh suggested on https://github.com/eventespresso/eea-attendee-importer/issues/16, under "Step 3")

Valid question here. I just thought this would be a better default (it's what our admin list tables all do). I could be wrong though.

## How has this been tested

* [ ] Fire up the attendee mover. When you start typing in the name of an event, the MOST RECENTLY CREATED events should appear first, instead of the oldest first.

## Checklist

* [x] My code is tested.

### DIFF
--- a/core/libraries/form_sections/inputs/EE_Select_Ajax_Model_Rest_Input.input.php
+++ b/core/libraries/form_sections/inputs/EE_Select_Ajax_Model_Rest_Input.input.php
@@ -72,11 +72,22 @@ class EE_Select_Ajax_Model_Rest_Input extends EE_Form_Input_With_Options_Base
             'query_params',
             array()
         );
-        // make sure limit and caps are always set
+        // Make sure limit and caps are always set.
         $query_params = array_merge(
-            array( 'limit' => 10, 'caps' => EEM_Base::caps_read_admin ),
+            [
+                'limit' => 10,
+                'caps' => EEM_Base::caps_read_admin
+            ],
             $query_params
         );
+
+        // And default to showing the most recently created items first.
+        if ($model->has_primary_key_field()) {
+            $query_params['order_by'] = [
+                $model->primary_key_name() => 'DESC'
+            ];
+        }
+        
         $this->_value_field_name = EEH_Array::is_set(
             $input_settings,
             'value_field_name',


joshfeck:
I think at the very least, this needs further discussion before we can move ahead with a change like this. There's already similar discussion in this PR for the attendee mover add-on:
https://github.com/eventespresso/eea-attendee-mover/pull/9

A few thoughts on showing events in a particular order:

In the context of adding registrations via an import tool, the workflow would go something like this:
1) Create an event
2) Import attendees into that event

So in this type of workflow, it makes sense to show the most recently added events first because you're likely going to be adding attendees to one of events you've just added.

The context of moving an attendee from one event to another is a different workflow. I'm really uncertain which way to order the events in this context. I can imagine that if an event manager has just added many events going into the next year, then the changes in this PR would likely be an unwelcome one because now they may need to scroll through a bunch of events to get to the end of the list where the events happening sooner. An example: An attendee phones in and wants to change their registration for tomorrow's event to instead attend next week's event. In this case it'd make sense to show events that starting closer to today's date. 

In any case, even if we settle on a default for this input, are other add-ons or other uses of this input allowed to change what's used to order by & sort by?

mnelson4:
> In any case, even if we settle on a default for this input, are other add-ons or other uses of this input allowed to change what's used to order by & sort by?

Yes, client code can override this default ordering.

>  In that scenario, the changes in this PR would likely be an unwelcome one because now they may need to scroll through a bunch of events to get to the end of the list that has the next upcoming events...

Oh that's a good point. IMO in that case showing the nearest upcoming event ordered by *date* would probably be best (not sure if start or end date would be better). But failing that, ordering by ID in ascending order would by my second choice.

Other situation to consider: if we had a select2 for selecting questions, message templates, or prices, I suppose we'd prefer ASCENDING order, rather than this branch's default of descending order (because that's the order we show those things in the admin list tables actually; in contrast to events, registrations, contacts, and transactions, which was what I was originally thinking about).

So I'm fine with closing this, and just changing the ordering for the attendee importer. Thoughts @joshfeck ?

joshfeck:
Yeah let's close this and just change the ordering for the attendee importer.

PR #1032, opened 2019-03-21
Fix getRelatedEntities in eventespresso/core store not accounting for different response types

nerrad:

`getRelatedEntities` in the `eventespresso/core` store is expected _all_ responses from the relation endpoint (eg. `event/10/datetimes`) to return an array for a successful response.  However, it is possible that a single entity would be returned if the relation is not a one to many or many to many relation (eg. `registration/123/attendee`).  This was an oversight and rectified by using the schema to determine what type to expect in the response.

In this pull:

- add new selector & resolve `getRelationResponseType` to the `eventespresso/schema` store.
- implement the new selector in the `getRelatedEntities` resolver in `eventespresso/core`.

I decided to normalize things so `getRelatedEntities` always still returns an array, so client code can expect a consistent response and deal with it accordingly (i.e. an array), but internally handles the variant response types from the rest api.

## How has this been tested

This does not impact any existing public facing customer features so internal testing only.

* [x] Unit tests (both new, updated and existing)
* [x] Verified this fixes the issue first apparent in the Barcode Scanner work I'm doing.

## Checklist

* [x] My code is tested.


assets/src/data/eventespresso/core/resolvers/relations.js
```diff
@@ -87,9 +87,26 @@ export function* getRelatedEntities( entity, relationModelName ) {
...
+		relationResponseType;
+
+	let relationEntities = yield fetch( {
 		path: relationEndpoint,
 	} );
+
+	relationEntities = relationResponseType === 'object' &&
+		! isEmpty( relationEntities ) ?
+		[ relationEntities ] :
+		relationEntities;
```
tn3rb: What else can be returned when fetching `relationEntities`?
Can `null` be returned?
Because if so, then it would remain as `null` instead of being `[ null ]`...
which might be exactly what you want, but I'm just checking to be sure.

Would it possibly be better to simply check whether or not `relationEntities` is an `array`, and if not, then put it in one?
ie:

```js
relationEntities = relationResponseType === 'array' &&
		! Array.isArray( relationEntities ) ?
		[ relationEntities ] :
		relationEntities;
```

or even just:

```js
relationEntities = ! Array.isArray( relationEntities ) ?
		[ relationEntities ] :
		relationEntities;
```

assets/src/data/eventespresso/schema/selectors.js
```diff
@@ -152,6 +152,24 @@ export function isRequestingRelationEndpointForEntityId(
...
+ * @return {string} The type for the relation returned for the given model and
+ * relation.
+ */
+export const getRelationResponseType = ( state, modelName, relationName ) => {
+	modelName = singularModelName( modelName );
+	relationName = pluralModelName( relationName );
+	const relationSchema = getRelationSchema( state, modelName, relationName );
+	return relationSchema !== null ?
+		relationSchema.type :
+		'';
```
tn3rb: Why not just return `array` here as the default, instead of having to convert elsewhere?
I'm assuming it's better to know the real type and then be at liberty to change it if need be, even if that ends up being repeated in several places.
Another option might be to add a `default` parameter to this function (and any others that call it),
so that the default can be set ahead of time, and save additional processing elsewhere. 

ie:
```js
export const getRelationResponseType = ( state, modelName, relationName, default = '' ) => {
	modelName = singularModelName( modelName );
	relationName = pluralModelName( relationName );
	const relationSchema = getRelationSchema( state, modelName, relationName );
	return relationSchema !== null ?
		relationSchema.type :
		default ;
};
```

assets/src/data/eventespresso/core/resolvers/relations.js
nerrad: Ya this needs improved a bit.  Going to rework.

assets/src/data/eventespresso/schema/selectors.js
nerrad: I don't think this should have a default as its schematic information returning the expected return type.  How would client know what the server is going to return in order to set a default?  I think it's better that this function has a return value that is consistent when there is no response type available for the given arguments.  Caller code then can decide what it's going to do when this returns that value.  In this case, I'm returning an empty string so it shares the same data type it would be if there _was_ a non-empty value.

nerrad:
I reworked the `getRelatedEntities` resolver so that it actually removes the need for doing a call to get the expected response type and instead does some type coercion as needed.  I think it now better handles various responses.

I've left the new `eventespresso/schema` selector/resolvers in place though because they could be useful in other contexts in the future.

assets/src/data/eventespresso/core/resolvers/relations.js
tn3rb: ya that looks pretty bullet proof to me now

PR #1035, opened 2019-03-26
Fix getRelatedEntitiesForIds using improper primary key string

nerrad:

Currently the `getRelatedEntitiesForIds` selector & resolver in `eventespresso/core` prepares query strings that only work for `EE_Belongs_To_Relation` type relations (i.e. the event relation to datetime).  However, for any other relation type (for relations that don't have a join table) it will fail because the query string for the primary key will be incorrect.  For example if you call it with the following signature:

```js
const datetimes = getRelatedEntitiesForIds( 'datetime', [ 10, 20 ], 'event' );
```

That will produce a querystring like:

```js
const queryString = 'https://ee.test/wp-json/ee/v4.8.36/events/?where[DTT_ID][IN]=10,20';
```

The problem here is that the query is incorrect.  It _should_ be `Datetime.DTT_ID` not `DTT_ID` because the datetime model is a `EE_Has_Many_Relation` to the event model (and thus the datetime id does not exist in the event model).  

In order to fix this, I've introduced a new selector and resolver to the `eventespresso/schema` store: `getRelationPrimaryKeyString( modelName, relationName)`.  This spits out a string that can be used for entity queries filtered by the relation primary key id and correctly accounting for the join type.  This then is implemented in the `getRelatedEntitiesForIds` resolver so that the fetch path is correctly generated for relations that don't involve a join table.

## How has this been tested

* [x] Added new unit tests
* [x] Tested against the work done for the new Barcode Scanner app and verified it fixes the issue there (first spotted this in that work).

**Note:** This does not impact any current customer facing features so no impact there.

## Checklist

* [x] My code is tested.


nerrad:
@tn3rb heads up that while the selector/resolver I added in the `eventespresso/schema` store is still useful, there's still problems that need worked out in the `getRelatedEntitiesForIds` that need iterated on in a different pull.

assets/src/data/eventespresso/schema/selectors.js
```diff
@@ -152,6 +158,53 @@ export function isRequestingRelationEndpointForEntityId(
...
+	(
+		state,
+		modelName,
+		relationName
+	) => {
+		modelName = singularModelName( modelName );
+		relationName = pluralModelName( relationName );
+		const singularRelationName = singularModelName( relationName );
+		const relationType = getRelationType( state, modelName, relationName );
+		if ( relationType === '' ) {
```
tn3rb: since `singularRelationName` isn't needed yet, maybe bump its creation down below this return clause. 
ie:

```js
		const relationType = getRelationType( state, modelName, relationName );
		if ( relationType === '' ) {
			return '';
		}
		const singularRelationName = singularModelName( relationName );
```

assets/src/data/eventespresso/schema/resolvers.js
```diff
@@ -134,6 +136,37 @@ export function* getRelationEndpointForEntityId(
...
+	const relationType = yield resolveSelect(
+		SCHEMA_REDUCER_KEY,
+		'getRelationType',
+		modelName,
+		relationName
+	);
+	relationName = singularModelName( relationName );
+	if ( relationType === '' ) {
+		return '';
+	}
```
tn3rb: since `relationName` is not needed yet, maybe bump its creation below this return clause.
ie:

```js
	if ( relationType === '' ) {
		return '';
	}
	relationName = singularModelName( relationName );
```

PR #1045, opened 2019-03-31
CHORE: Update @wordpress/scripts and @wordpress/jest-preset-default to latest

nerrad:
This pull

- updates `@wordpress/scripts` to the latest version which includes a number of changes (including new linting rules, updating jest, etc).
- removes no longer needed `@wordpress/es-lint` dependency (because that's a part of the updated `@wordpress/scripts` package.
- updates `@wordpress/jest-preset-default` package to latest (for parity with jest updates).

Currently this is a WIP and on hold because [there's a bug](https://github.com/facebook/jest/issues/7981) in the Jest version included with this update (Jest 24.1) that causes a couple of our tests to fail because of our dependency on moduleMapping for our internal package like things.  The bug is fixed [here](https://github.com/facebook/jest/pull/8210).

I'll have to monitor when the new version of Jest is released with that fix and then do a pull for `@wordpress/scripts` to get that change.

This isn't a critical update but it does bring some helpful new lint rules/tools so it'd be nice to update to it when possible.


nerrad:
I'm going to close this pull in favor of the dependabot auto updates that I've been handling.  More granular updates mean for a bit easier resolution of issues.

PR #1046, opened 2019-04-01
Add ability to filter start and end dates for default EE_Datetime entity (Fixes #950)

### LINKED ISSUE #950: Add filter hooks to allow changing the default date time start and end date times
In `EEM_Datetime::create_new_blank_datetime()` there's a place where it adds 30 days to today's date for the event start and end dates, and also sets the start time to be 8:00 AM and end time to be 5:00 PM. In some cases it would be very useful to change these defaults (like if all events started at 11:00AM, for example). 

Another example: In this support topic:
https://eventespresso.com/topic/30-days-added-to-event-start-date/
the end user would prefer the default datetime start date to be today's date.

nerrad:

See #950 for background.  This pull:

- Adds four new filters: `FHEE__EEM_Datetime__create_new_blank_datetime__start_date`, `FHEE__EEM_Datetime__create_new_blank_datetime__end_date`, `FHEE__EEM_Datetime__create_new_blank_datetime__start_time`, `FHEE__EEM_Datetime__create_new_blank_datetime__end_time` for modifying the parameters on a generated default `EE_Datetime` entity.
- Due to start and end time filters expected an array response for the time and time format, there is some minimal validation done to ensure the returned filter value is an array and has the correct count of arguments.  If not, the appropriate exception is thrown.  This provides some immediate fail indication if code is no implementing the filters correctly.
- Adds tests for new filters to verify they work.

## How has this been tested

* [x] Ensure existing unit tests pass
* [x] Ensure new unit test passes.
* [x] Filters can be manually checked via a site-specific plugin to ensure they provide the expected flexibility for modifying the Date used (I have not done this myself).
* [x] Ensure otherwise without any usage of the filters that the event editor works as expected when creating a new Event (existing behaviour is preserved).  I have not tested this myself.

## Checklist

* [x] My code is tested.


core/db_models/EEM_Datetime.model.php
```diff
@@ -150,40 +154,97 @@ protected function __construct($timezone)
...
-                'DTT_EVT_start' => $this->current_time_for_query('DTT_EVT_start', true) + MONTH_IN_SECONDS,
-                'DTT_EVT_end'   => $this->current_time_for_query('DTT_EVT_end', true) + MONTH_IN_SECONDS,
+                'DTT_EVT_start' => $start_date,
+                'DTT_EVT_end'   => $end_date,
                 'DTT_order'     => 1,
                 'DTT_reg_limit' => EE_INF,
             ),
             $timezone_string
         );
+        $start_time = apply_filters(
```
mnelson4: It would probably be good to add a comment here defining the format of the value being filtered (eg it took me a while to figure out what the "ga" meant).

core/db_models/EEM_Datetime.model.php
```diff
@@ -150,40 +154,97 @@ protected function __construct($timezone)
...
+     * @throws InvalidArgumentException
+     * @throws InvalidDataTypeException
+     * @throws ReflectionException
+     * @throws InvalidInterfaceException
      */
     public function create_new_blank_datetime()
     {
         // makes sure timezone is always set.
         $timezone_string = $this->get_timezone();
+        $start_date = apply_filters(
```
mnelson4: Maybe a comment here too would be good, explaining the type of the value being filtered.

joshfeck:
This was tested and it works quite well. Thanks!

PR #1047, opened 2019-04-02
Fix missing reset action type handling in reducers.

nerrad:

While continuing testing of the Barcode Scanner App rewrite, I noticed that the recently added reset state actions (`eventespresso/core` store) have some missing reducer handler.  Specifically:

- The entities reducer is missing handling for `RESET_ALL_STATE` and `RESET_STATE_FOR_MODEL`.
- The model specific selector reducer is missing handling for `RESET_ALL_STATE`.

This pull adds the missing logic and updates test coverage for that logic. 

## How has this been tested

* [ ] verified existing and updated unit tests pass
* [ ] verified via testing in the Barcode Scanner App that the reset actions work as expected in updating the store.

## Checklist

* [ ] My code is tested.


assets/src/data/eventespresso/core/reducers/test/entities.js
```diff
@@ -2,11 +2,13 @@
...
 import { EventFactory } from '@test/fixtures';
+import { fromJS, Map } from 'immutable';
+import { DEFAULT_CORE_STATE } from '@eventespresso/model';
 
 /**
  * Internal dependencies
  */
-import {
+import entitiesHandler, {
 	receiveEntity,
```
tn3rb: is this correct?
is that var supposed to be before the brace?

nerrad: yes, its the default export.

Issue #1034, opened 2019-03-25
Fix issues with EE3 to EE4 attendee data migrations


joshfeck:
While testing #881 I found some issues with EE3 to EE4 database migrations where EE3 attendee data went missing after migrating to EE4. The same issues happened regardless of whether I had master checked out or the branch for #881. It didn't matter which branch was activated while doing the migration, the same thing happened.

**To Reproduce**

In this case I set up one event, with two extra question groups, and set 1 custom question to each extra question group. I also set 1 custom question to the Personal Information group. Then I did 3 registrations for the event. Then I migrated to EE4.

**Expected behavior**

Expected each of the 3 attendees to have the correct information migrated

**Screenshots**

### _events_attendee table
<img width="365" alt="Screen Shot 2019-03-25 at 5 32 23 PM" src="https://user-images.githubusercontent.com/891156/54955540-00a8f300-4f24-11e9-94b3-88df1a46bdbb.png">

### _events_answer table
<img width="453" alt="Screen Shot 2019-03-25 at 5 33 36 PM" src="https://user-images.githubusercontent.com/891156/54955612-27672980-4f24-11e9-95ba-3557f8c73108.png">

### _esp_registration table
<img width="316" alt="Screen Shot 2019-03-25 at 5 34 46 PM" src="https://user-images.githubusercontent.com/891156/54955660-4cf43300-4f24-11e9-8c66-0a7bdce33a08.png">

### _esp_attendee_meta table
<img width="774" alt="Screen Shot 2019-03-25 at 5 35 11 PM" src="https://user-images.githubusercontent.com/891156/54955712-701ee280-4f24-11e9-8d42-da2023ad7bea.png">

**note in the above, no address information**

### _esp_answer table
<img width="372" alt="Screen Shot 2019-03-25 at 5 36 43 PM" src="https://user-images.githubusercontent.com/891156/54955758-9a70a000-4f24-11e9-8923-047e3789fc96.png">

**note in the above, the answers are related to REG ID 3 only, and the answer values are actually for REG ID 1**

**Environment:**
 - Version of Event Espresso:  4.9.80.rc (current master)
 - Version of WordPress: 5.1.1
 - Browser Firefox
 - PHP Version 7.3

mnelson4:
Ugh ya that's pretty borked.
When I ran the migration from EE3 with the above setup, I got a bunch of warnings after the first migration, including

```
[...]

PR #1049, opened 2019-04-02
Fix DMS weirdness because event table selections overwriting attendee table selections

mnelson4:

Fixed several errors during the EE 4.1 migration (like missing address info on migrated attendees, incorrect registration counts, answers all for the same attendee, etc). 
These errors arose because some EE3 event table values (eg address, city, phone) were being used instead of the attendee table values  (ever since we started avoiding migrating attendees for deleted events). The fix involved just making sure we ignore the event table's values.

## How has this been tested

* [x] In EE3, create multiple question groups, an event that uses all question groups for primary and additional attendees, and allows group registrations and requires their full registration info. 
* Register a few times for that event.
* Also, create and register for another event, then trash that event.
* Deactivate EE3 and activate this branch of EE4 and run the migrations.
* Verify the new registrations all have the correct answers to the custom questions, the database column's `REG_count` is correct, their address info is populated correctly on the attendee, and they're for the correct transactions.
* Also verify that the attendees for the deleted event were NOT imported into EE4.

## Checklist

* [x] My code is tested.


mnelson4:
Thanks @tn3rb, passing for testing.

joshfeck:
Related issue #1034

joshfeck:
Tested so far and I can see that the issues reported in #1034 are fixed with this branch. Will continue testing through the checklist.

joshfeck:
Finished testing the checklist

PR #1058, opened 2019-04-09
Automatically check command types

mnelson4:

`EventEspresso\core\services\commands\CommandHandler` has a new method, `invokeHandle`, which is a wrapper of `handle` that also checks the command provided is of the correct type. Just saves having to check the command's type in all implementations of `handle`.

It seems DRYer to me to do this in the parent, simplifies each child class, and overall saves trees.

## How has this been tested

* [x] Verify code that executes commands works properly. Eg attendee mover, attendee importer, etc.

## Checklist

* [ ] My code is tested.


core/exceptions/InvalidEntityException.php
```diff
@@ -37,7 +40,7 @@ public function __construct($actual, $expected, $message = '', $code = 0, Except
                     ? get_class($actual)
                     : gettype($actual),
                 $expected,
-                var_export($actual, true)
+                $object
```
mnelson4: Wouldn't `var_export` work more often than hoping the object can be turned into a string? This might have warning sometimes right?

tn3rb: `$object` is the output buffer contents after var dumping the object...
this code is just before the message code above:

```php
ob_start();
var_dump($actual);
$object = ob_get_clean();
```

mnelson4: oh I see, ya I missed that. kk!

joshfeck:
Verified code that executes commands works properly, including the attendee mover and the attendee importer. I'm not sure what should be tested for "etc.", but if it's sufficient to test only the attendee mover & importer, then this is ready for merge.

mnelson4:
Thanks Josh!
(Sorry, what I meant by "etc" was any other code that's known to use the commands system; I didn't specify them because I couldn't think of any others besides the two mentioned, and thought the tester might know of others.)
Anyways, really just those two is sufficient.

Issue #654, opened 2018-08-18
Form system labels use the form input's name instead of ID


mnelson4:
## Issue Overview

Form system labels's `for` attribute uses the input's name, not its ID.

I was playing with the HTML Code Sniffer for Accessibility (see https://squizlabs.github.io/HTML_CodeSniffer/), and saw a warning saying our reg form inputs don't have labels. I thought "sure they do", but then I noticed clicking on the labels didn't select the input. So something was wrong. It's because the `for` attribute on the labels has the form input's name, not its ID. It should be its ID.

This affects usability for both regular users and for users on screenreaders etc.

## Bug report or feature request?

* [x] Bug
* [ ] Feature
* [ ] Neither

## Environment Data:

Version of EE: 4.9.66

Version of WordPress:4.9.8

PHP Version:  7.1

Browser used:  Firefox

## Steps to Reproduce (for bugs)

Add the HTML code sniffer bookmarklet to your browser (should work for any browser). Start to register for an event and get to SPCO step 1 (attendee info). Click on the bookmarklet and view the errors. Among them is one saying 

> This textinput element does not have a name available to an accessibility API. Valid names are: label element, title undefined, aria-label undefined, aria-labelledby undefined. 

Eg https://drive.google.com/file/d/1pseY2oOLm34H4jhwVOyPB1BLqRlVjQTs/view

## Expected Behaviour

The HTML code sniffer shouldn't complain at this, and clicking on the form input's label should select the corresponding input.

## Current Behaviour

Also, clicking on the label does NOT select the input.

sethshoultes:
@mnelson4 is this something you can fix?

mnelson4:
I think so

mnelson4:
I would still like to investigate this- it's a bit annoying when clicking labels don't select the right form input.

sethshoultes:
I agree. Do you have an estimate of time this will take?

mnelson4:
An hour for initial investigation (according to http://p2.eventespresso.com/2018/12/27/highlights-from-it-doesnt-have-to-be-crazy-at-work/#comment-19972). After that I'll have a better idea, but it might just be an easy switch (like 30 minutes of development); if not we would re-evaluate it.

PR #1061, opened 2019-04-09
html labels should use IDs not names

mnelson4:

Fixed EE HTML label tags so their "for" attribute uses the input's ID, not its name, which makes it more accessible and so clicking on the label selects the input.
In SPCO and anywhere using EE forms, if you click on the label of an input, the input is not selected (with the exception of checkbox inputs and radio inputs, they've been working ok).
This was way back when, on https://github.com/eventespresso/event-espresso-core/commit/c4d911fb8b7e5c93e4 (codebase ticket https://events.codebasehq.com/projects/event-espresso/tickets/6081) we accidentally switched the labels to use the HTML names of inputs instead of their IDs.

## How has this been tested

* [x] Register for an event, it should work fine as usual.
* [x] When registering, click on any of the question labels (eg "First Name"). The input should become selected.
* [x] Register for an event with all sorts of questions (eg checkboxes, selects, radio buttons). They should all work as expected.

## Checklist

* [x] My code is tested.


joshfeck:
This is ready for merging.

joshfeck:
This fixes #654

Issue #1062, opened 2019-04-10
REST API _links are erroneously for join-models when querying across a relation


mnelson4:
## Issue Overview

Eg, `mysite.com/wp-json/ee/v4.8.36/datetimes/1/tickets` returns entities with `_links` that appear to actually be for `datetime_tickets` entities, not `tickets`.
However `mysite.com/wp-json/ee/v4.8.36/tickets`'s `_links` are fine.

This was causing Brent and Darren some grief in their Javascript code that consumes the REST API.

## Bug report or feature request?

* [x] Bug
* [ ] Feature
* [ ] Neither

## Environment Data:

Version of EE: 4.9.80

Version of WordPress:any

PHP Version:  any

Browser used:  any

## Expected Behaviour

If the results are `tickets`, the `_links` should only be for them.

## Current Behaviour

When querying for related data across a has-and-belongs-to-many-relation, `_links` is for the join model/entitiy/table.

joshfeck:
#1064 was merged

PR #1064, opened 2019-04-10
Fix REST API _links when querying across HABTM relation

mnelson4:

Fix REST API _links when querying across HABTM relation. 
This was happening because when we query across a has-and-belongs-to-many relation, we add the fields from the join table onto the response. (TBH I don't remember why but that's what we've been doing and there was some reason for it.) The only trouble was that we were also adding the join table's `_links`, and overwriting the queried model's `_links`. 
This PR instead keeps the queried model's `_links` and ignores the join table's.

See https://github.com/eventespresso/event-espresso-core/issues/1062

## How has this been tested

* [x] Send a request to `mysite.com/wp-json/ee/v4.8.36/datetimes/{id}/tickets`. Look at the results' `_links` property. There should NOT be one for `ticket` because that's what the entity is- a ticket.
* [ ] Double-check this doesn't break anything using the REST API, like the mobile apps or attendee mover.

## Checklist

* [ ] My code is tested.


tn3rb:
I have good news and bad news....
tested this fix on my site where I first encountered the issue, and it fixed that issue,
but then a second issue appeared.
flipping the fix back to the original code, made the new issue go away, but the original issue come back (of course).
will need to discuss this with Darren first as it looks like the new issue is a problem in the JS models.
basically i'm getting results for a query, but then a few seconds later they get overwritten with nothing

joshfeck:
This was chatted about in Slack and it got the go ahead for testing:

https://eventespresso.slack.com/archives/C080DDVLM/p1554927556768200

joshfeck:
I've tested this but I'm having some trouble with the mobile apps on master. So it looks like before we can verify the apps aren't broken we'll need to fix whatever is going on in master that's affecting the apps.

PR #1066, opened 2019-04-11
Add StepBubbleMenu Component

nerrad:

The `<StepBubbleMenu />` component is a accessible, configurable, ui element used to describe steps to the end user.  Heres an example of its usage in an early iteration of the new Barcode Scanner App

Example usage

Customization options include:

- number of steps
- which steps can be interacted with (clickable) by the user.
- label for each step
- slug for each step
- value for each step.

More information about the component can be found in the [attached documentation](https://github.com/eventespresso/event-espresso-core/blob/7507065b0ea1f099b6e8a417368a498741228995/docs/AA--Javascript-in-EE/components/ui/menu/step-bubble/README.md) for this pull.

## How has this been tested

* [x] Covered by unit tests
* [x] Functionality was tested as a part of the work for the Barcode Scanner Addon
* [x] Accessibility was tested/verified using the `VoiceOver` macosx screenreader application as well as via the [Axe Accessibility Analysis Browser Extension](https://www.deque.com/axe/) (chrome).

## Checklist

* [x] My code is tested.


assets/src/components/ui/menu/step-bubble/index.js
```diff
@@ -0,0 +1,58 @@
...
+	clickable,
+	activeBubble
+} ) => {
+	return (
+		<div className="ee-step-bubble-menu-container">
+			<div className="step-bubbles-container" role="menu" tabIndex={ 1 }>
+			{
+				bubbleData.map( ( bubble ) => {
+					let slug = bubble.slug || bubble.label;
+					let canClick = clickable.indexOf( bubble.slug ) > -1;
```
tn3rb: imho `clickable` should just be another property of the objects passed in `bubbleData`.
Going only by what I can see in this PR, it makes no sense to pass an array of objects with details about each bubble... 

**except** for whether they are "clickable"...

and **then** have to pass **another** array to inform the system about this extra detail about each bubble. Why is that not just part of each bubble step's data?

It could be argued that `activeBubble` could also be a property of `bubbleData`, but that may be external due to how/when it gets toggled.

but this seems way more logical to me:

```js
<StepBubbleMenu
    bubbleClick={ () => null }
    bubbleData: [
		{
            label: 'foo',
            active: true,
            clickable: true,
            value: 42,
        },
		{
            label: 'bar',
            active: false,
            clickable: true,
            value: 39,
        },
		{
            label: 'baz',
            active: false,
            clickable: false,
            value: 0,
        },
    ]
/>
```

Even if you needed to toggle the active state afterwards you could still find the specific bubble data object in the array and update it.
That of course could be made even easier by using an object instead of an array and then using the slugs as keys:

```js
<StepBubbleMenu
    bubbleClick={ () => null }
    bubbleData: {
		foo: {
            label: 'Foo',
            active: true,
            clickable: true,
            value: 42,
        },
		bar: {
            label: 'Bar',
            active: false,
            clickable: true,
            value: 39,
        },
		baz: {
            label: 'Baz',
            active: false,
            clickable: false,
            value: 0,
        },
    }
/>
```

which would allow you to directly toggle the active status via something like:

```js
bubbleData.foo.active = false;
```

which would be much easier to work with.

Another benefit of this approach is that the `slug` property is guaranteed to be usable as a key, or URL param, or whatever, since it literally IS a key already.
With the existing code, the `slug` property is not required as part of the supplied bubble data and will be copied from the `label` if missing. The `label` however may not be usable in things such as URLs without having to convert it first (ie: "I'm a Label, Not a Key"), whereas the slug does not require conversion and is therefore more stable.

To make this even more versatile, you could also include `bubbleClick` as yet another property of `bubbleData` which would allow each bubble to have it's own callback if so desired.

```js
<StepBubbleMenu
    bubbleData: {
		foo: {
            label: 'Foo',
            active: true,
            clickable: true,
            value: 42,
			callback: doFoo,
        },
		bar: {
            label: 'Bar',
            active: false,
            clickable: true,
            value: 39,
			callback: doBar,
        },
		baz: {
            label: 'Baz',
            active: false,
            clickable: false,
            value: 0,
			callback: doBaz,
        },
    }
/>
```

There is of course nothing stopping one from using the same callback for every step, but at least this way you have the option to also have one or more of them different than the others.
It also appears from the logic in the `StepBubble` component, that the only purpose of the `clickable` property is to determine when to apply the one provided callback to a step's `onClick` and `onKeyPress` actions. By passing the callbacks directly as part of each bubble's data object, this also becomes completely unnecessary, because the mere presence of a callback in the bubble step data would be indication enough that it is clickable.

So now we are down to:

```js
<StepBubbleMenu
    bubbleData: {
		foo: {
            label: 'Foo',
            active: true,
            value: 42,
			callback: doFoo,
        },
		bar: {
            label: 'Bar',
            active: false,
            value: 39,
			callback: doBar,
        },
		baz: {
            label: 'Baz',
            active: false,
            value: 0,
        },
        // baz is not clickable because no callback was provided
    }
/>
```

Although, the `clickable` property **could** also serve an alternate purpose and actually prevent a particular step from being clicked until some other criteria has been met (such as completing previous steps) despite that step having a valid callback

assets/src/components/ui/menu/step-bubble/style.css
```diff
@@ -0,0 +1,74 @@
...
+    border: none;
+    background-color: transparent;
+}
+
+.step-bubbles-container .ee-step-bubble-item:focus {
+    outline: unset;
+    border: none;
+}
+
+.step-bubbles-container .ee-clickable {
```
tn3rb: there is an existing `clickable` class in WP core but that likely only applies to the admin, so this might still be good, but should eventually be moved into a global ee stylesheet (which you are probably already planning to do)

assets/src/components/ui/menu/step-bubble/step-bubble.js
```diff
@@ -0,0 +1,68 @@
+/**
+ * Internal imports
+ */
+import StepText from './step-text';
+
+/**
+ * External imports
+ */
+import classNames from 'classnames';
```
tn3rb: oh cool... didn't know this was available to us. I had meant to ask about this.

tn3rb:
Have more than a few suggestions for improving how the bubble data is structured.
Basically it comes down to applying properties to the objects they actually affect as opposed to having to infer an object's properties from something else. Out of the four properties passed to `StepBubbleMenu`, three of them actually apply to the bubble steps themselves and you have had to add extra logic in order to handle passing those properties from the parent component to the child, when they could have simply been part of the child's data from the start.

Another question I have that is not directly related to the code is how come there is no button for progressing to the next step when the current step has been completed? It appears that you can click on any step you want at any time. I would think this would cause unnecessary complications and require additional logic in order to prevent errors occuring due to information missing that should have been provided in a previous step.

For example, in the gif you have above, I would think that choosing an Event would be a requirement before choosing a Datetime would even be possible, no? And the "Scan" step is dependent on both of the previous steps being completed. So how come you can click anywhere you want?

The UI would be much better user-proofed if instead of clicking any bubble you want to progress, there was another button labelled something like "Next", or "Proceed to FooBar", or whatever, that lead the user to the next step and **only** became clickable after they have successfully completed the current step. I know that **YOU** know you can click those step bubbles to navigate the menu, but to users with low tech skills that may not be apparent at all and they may not figure that out. A dedicated "Next" button would be far more intuitive.

assets/src/components/ui/menu/step-bubble/style.css
nerrad: Ya just keeping styles in here until the new style system is put in place.

nerrad:
First of all, thank you for the feedback.  On the initial read, everything you wrote made sense.  I did try implementing some of what you suggested but in the end I reverted to mostly what I had before.  Please understand, I didn't arbitrarily arrive at the structure for this component the way it is and I hope my explanation helps.

Before you read the explanation, I'm hoping it would be helpful if you review the implementation of this component in the Barcode Scanner App (see [this file](https://github.com/eventespresso/eea-barcode-scanner/blob/12d055af25cf4747ae361ca580746ef49db5269d/assets/src/barcode-scanner-app/index.js)).  This component was crafted in concert with an actual implementation while thinking about future uses of the component as well. 

> imho clickable should just be another property of the objects passed in bubbleData.
Going only by what I can see in this PR, it makes no sense to pass an array of objects with details about each bubble

The `bubbleData` configuration object is meant to contain things that are distinct between various steps (so `label`, `slug`, and `value`) and are also intended to be things that generally do not change for the life of the view.  At most, I could foresee an entire step being removed, or an entire step being added, but in general once the menu is setup with the configuration objects, those values won't change.  The reason why `clickable` is kept as an array of slugs is because it's much more likely that what _does_ change will be what menu item is available for selection during the life of an app.  

Let's look at the case of the barcode scanner app.  Currently, on initialization, it receives an array of ALL the slugs.  However, that's probably not the way we'll want to keep it.  Instead, what I'll likely do is just include an empty array on initialization because nothing should be clickable (we want step bubble active state to be controlled entirely by the app).  When an event is selected, the active bubble is changed to the second step and at that point it would be far easier to just add the selected slug to the `clickable` state vs traversing the configuration array (or object if it was changed like you suggested). 
I think having this as a simple array makes it easier to add/remove slugs as needed without touching the configuration object which generally doesn't change at all during the life of the view.

> It could be argued that activeBubble could also be a property of bubbleData, but that may be external due to how/when it gets toggled.

This is not a property of the `bubbleData` configuration for similar reasons as `clickable`.  `activeBubble` is something that is expected to change multiple times in the course of a view and it's more efficient (although arguable whether the performance gain _that_ significant) and easier to simply set that one value vs traversing the configuration to set the previous step as inactive and the current step as active.  

> With the existing code, the slug property is not required as part of the supplied bubble data and will be copied from the label if missing. The label however may not be usable in things such as URLs without having to convert it first (ie: "I'm a Label, Not a Key"), whereas the slug does not require conversion and is therefore more stable.

You make a good point about the label/slug and on review I didn't like the optional nature of the slug, especially when it is referenced via the `clickable` and `activeBubble` properties.  So I decided to make it required and have done so by asserting incoming bubble configuration objects have the required properties and they are of the correct type (see d8d3b829).

> To make this even more versatile, you could also include bubbleClick as yet another property of bubbleData which would allow each bubble to have it's own callback if so desired.

I'm not sure there's much value in each bubble configuration to have its own onClick callback over the current implementation because the provided callback _receives_ the slug of the bubble that was clicked.  You can see the example implementation of this in the barcode scanner implementation here: 

https://github.com/eventespresso/eea-barcode-scanner/blob/12d055af25cf4747ae361ca580746ef49db5269d/assets/src/barcode-scanner-app/index.js#L23-L25

So effectively you can accomplish executing different behaviour via what slug was clicked.

> Another question I have that is not directly related to the code is how come there is no button for progressing to the next step when the current step has been completed? It appears that you can click on any step you want at any time. I would think this would cause unnecessary complications and require additional logic in order to prevent errors occuring due to information missing that should have been provided in a previous step.

The barcode scanner app is still not controlling the behaviour of what's clickable or not in its current implementation.  The gif I used as an example was simply to show the step bubble component in action visually but you're right, for that particular use-case it wouldn't make much sense to have the user be able to click to step two without having selected an event in step one.  It won't be that way when the barcode scanner app is complete.

The reason why there is no button for progressing to the next step is because I feel that's something that should be left up to the consuming code to control.  There's a lot of variables affecting not only how the "next step" button might appear in the ui, but also what might trigger the next step button being available or even whether it even is visible.  In the case of the barcode scanner example, it will controls toggling to the next step.  For example, when an event is selected in step one, it toggles to step two by setting the `activeBubble` property to the slug for step two.  You can see the related code here:

https://github.com/eventespresso/eea-barcode-scanner/blob/12d055af25cf4747ae361ca580746ef49db5269d/assets/src/barcode-scanner-app/index.js#L38-L45

Again, I _really do_ appreciate the feedback and I hope some of the explanation I've given helps outline the reasoning behind the choices I made.

tn3rb:
I took a look at the scanner app and how it integrates with the bubble step menus and found it to be overly complex when it absolutely does not need to be. So I cut a branch and took a stab at simplifying things. I'm really at a total loss and can not see any reasons for some of the things you were doing. If there are good reasons then, I would love to see the React documentation that promotes doing things those ways.

What kinds of things? 

- The biggest offender I saw was the complete duplication of state between the `BarcodeApp` and the `Selectors` components. You were literally passing the state from the `BarcodeApp` into the `Selectors` component and then copying it directly. I could understand if maybe some kind of transformation or processing was required but that was not the case. It was an exact copy. Everything I have read about state management cautions heavily against duplicating state and most articles promote having as minimal reliance on state as possible. It's one of, if not **THE** foundation of functional programming which is the preferential way to approach React.

- using `componentDidUpdate()` to control whether to re-render. This is another thing  that is heavily cautioned against. Even you yourself have warned me about not doing this because of the possibility that it will cause errors and skip a necessary render. I can't even figure out why you did this because I'm not seeing anything but minimal numbers of renders. In my refactoring that got rid of the duplicated state and is passing everything as props I am only seeing one render for each component per step.

- functions calling functions calling functions calling functions... This is another thing that I just can't even begin to understand why you do it. Let's look at the callback chain that fires when the event selector value changes:

    - the `EventSelect` component has its `onSelect` property set to use `Selectors.onEventSelect()` for it's callback
    - `Selectors.onEventSelect()` calls `Selectors.props.onDataUpdate()`
    - `Selectors` is getting ALL of its props from `BarcodeApp.state` meaning `BarcodeApp.state.onDataUpdate` is (a copy? or reference?) to `BarcodeApp.onDataUpdate()`
    - `BarcodeApp.onDataUpdate()` runs through a bunch of `if` statements and ultimately calls `BarcodeApp.updateEventState()` and `BarcodeApp.updateDatetimeState()` and may also call `setState()` if the event id has changed.
    - then `BarcodeApp.updateEventState()` calls `setState()`
    - and `BarcodeApp.updateDatetimeState()` calls `setState()`

that's 6 steps to go through whenever the event selector value changes. And the only real logic that happens is spread between `onDataUpdate()`, `updateEventState()`, and `updateDatetimeState()` but is really actually pretty minimal. It just seems like a lot of code because you have to chase all over the place to find out what's actually happening. This kind of thing is not necessary. You may think that you are abstracting things, but it is not the kind of abstraction that is useful. Useful abstraction is where you isolate some portion of your code that changes and allow that to be swapped out with other blocks of code that operate the same way but give different results. That's not the case in the above code. None of that can be swapped out with an alternate block of logic. There would be absolutely no way to substitute the bubble click handler for one of those steps with some other logic (not that we would want to, or need to, but I'm just making a point). Good abstraction though would allow that or have some other really obvious benefit.

In the refactor that I did, I consolidated all of the logic that occurs when the event selector changes into one callback on `BarcodeApp` called `onEventChange()`, and simplified how that is passed through the `Selectors` component to `EventSelect`. Now please note that in your existing code, you were usually referencing state and/or component props directly via `this.state` or `this.props` whereas I am actually deconstructing `onEventChange` in `Selectors.render()` and then passing it along as a constant as opposed to directly referencing `this.props.onEventChange()`. But to make this comparison fair, I'm going to pretend that I referenced things directly using `this.props`.

So here's my refactored function chain:

    - the `EventSelect` component has its `onSelect` property set to use `Selectors.props.onEventChange()` for it's callback
    - `BarcodeApp` sets the `Selectors.onEventChange()` property to use `BarcodeApp.onEventChange()`

...that's it!!!

That's all that is needed. Again,  in my refactor `Selectors.props.onEventChange()` is actually being deconstructed in `Selectors.render()` and then passed to `Selectors.getEventSelector()` where it gets passed to the `EventSelect` component. So really there are 3 steps in my path, but to be fair, if your app had to do the same, then it would add another 2 steps in that function chain that someone would have to follow to get to the end of it all. That's 8 steps vs 3, or 6 vs 2 depending on if you want to reference `this.props` directly or not. I'm not concerned about this in terms of performance, as any hit taken from hopping function to function is going to be negligible. But people other than yourself may have to deal with this code at some point, and it is just not easy to follow when everything requires tracing a half dozen or more function calls and the actual logic is chopped up and spread between more than one of them. Look at this:

```js
	onDataUpdate = ( stepSlug, selectedValue ) => {
		if ( selectedValue === null ) {
			if ( stepSlug === slugs.MENU_CHOOSE_EVENT ) {
				this.updateEventState();
				this.updateDatetimeState();
			}
			if ( stepSlug === slugs.MENU_CHOOSE_DATETIME ) {
				this.updateDatetimeState();
			}
			return;
		}
		if ( stepSlug === slugs.MENU_CHOOSE_EVENT &&
			selectedValue.value &&
			selectedValue.value !== this.state.selectedEventId
		) {
			this.updateEventState( selectedValue );
			this.updateDatetimeState();
			this.setState( { currentStep: slugs.MENU_CHOOSE_DATETIME } );
		}
		if ( stepSlug === slugs.MENU_CHOOSE_DATETIME &&
			selectedValue.value &&
			selectedValue.value !== this.state.selectedDatetimeId
		) {
			this.updateDatetimeState( selectedValue );
			this.setState( { currentStep: slugs.MENU_SCAN } );
		}
	};
```

this is what handles the change events for the event and datetime selectors. There's a bunch of `if` statements, a couple calls to `updateEventState()` in different places, **FOUR** different calls to `updateDatetimeState()` in different places, and then a couple of calls to `setState()`. Then `updateEventState()` and `updateDatetimeState()` each make a call to `setState()`. This is really not easy to figure out quickly what is going on when the event selector changes. Anybody in the future that needs to try and debug this is not going to have a good time!

The refactor I did has a single `onEventChange()` callback with a nearly direct path between it and the selector that calls it, with one call to `setState()`. Same thing with `onDatetimeChange()`. And all the logic for each event handler is in one place. 

Here's a new rule of thumb that I want you to try and follow from now on:

#### When you have logic that needs to fire whenever some other event is triggered (like a click handler), try to put ALL of that logic in one place and try to link the event source with the event handler as directly as possible. If there is a LOT of logic in the handler, or there are BIG blocks that can be shared with other functions, then it is ok to split that logic up. Otherwise keep it together so others don't have to go searching for it.

Now let's look at some of your response from above...

> First of all, thank you for the feedback. On the initial read, everything you wrote made sense. 

and on subsequent reads? is there anything I suggested that doesn't make sense now?

> I did try implementing some of what you suggested but in the end I reverted to mostly what I had before. Please understand, I didn't arbitrarily arrive at the structure for this component the way it is and I hope my explanation helps.

nobody is saying that you "arbitrarily" arrived at anything. I would hope that that was never ever the case. Do you do things "arbitrarily" when you are writing code? I hope not. Do you think that I think you don't put any thought into what you are doing? Because that's totally not the case. I think you're an incredibly smart guy and I know that you take a lot of pride in your work and I would never think that you are just slapping stuff together arbitrarily. But you hope your explanation helps me... what? ...understand that... "OH... you do things for REASONS!!!". It would help much better in the future if you just posted those reasons directly and then we could actually discuss them. And please understand, that last sentence is very condescending. I mean... you _hope_ your explanation helps me understand... but it doesn't sound like you think I'm actually capable of understanding things.

> The bubbleData configuration object is meant to contain things that are distinct between various steps (so label, slug, and value) and are also intended to be things that generally do not change for the life of the view.

That doesn't address my points at all which were about **where** properties rightfully belong and not about whether they change or not. And intending that a particular set of attributes for an object should not change does not negate whether some other properties belong on that object or not. Let's look at the `StepBubbleMenu`:

Out of the four properties assigned to it:

- bubbleClick
- bubbleData
- clickable
- activeBubble

**ALL** of them are passed along and become properties of `StepBubble`, either directly or indirectly as a property of `bubbleData`.

Look:

```js
bubbleData.map( ( bubble ) => {
	assertStepConfiguration( bubble );
	const { slug, label, value } = bubble;
	const canClick = clickable.indexOf( bubble.slug ) > -1;
	return (
		<StepBubble
			key={ slug }
			label={ label }
			slug={ slug }
			stepValue={ value }
			isActive={ activeBubble === slug }
			canClick={ canClick }
			bubbleClick={ bubbleClick }
		/>
	);
} )
```

`StepBubble.isActive` is derived from `StepBubbleMenu.activeBubble` but is ultimately a property of `StepBubble` because `StepBubbleMenu` does nothing else with that property.
`StepBubble.canClick` is derived from `StepBubbleMenu.clickable` but is also a property of `StepBubble` because again, `StepBubbleMenu` does nothing else with that property except use it to configure `StepBubble`.
Here's the best one, `bubbleClick ` is just passed **DIRECTLY** to `StepBubble` !?!?! `StepBubbleMenu` literally does nothing with this property so there is just no way you can argue that this should be a property of `StepBubbleMenu`. If it belongs on `StepBubbleMenu`... then how come `StepBubbleMenu` isn't doing anything with it?

It doesn't really matter what you say these are intended to be when it is clear that in reality they are properties of `StepBubble`.
I'm not making these criticisms arbitrarily okay? I'm looking at the code and I'm seeing you play hot potato with properties and variables and functions and it makes things hard to follow with little to no benefit.

So here's another new rule of thumb for you to follow:

#### properties that describe object A do not belong on object B, or object C, or any other object. Put them on object A.

let's move on...

> Let's look at the case of the barcode scanner app. Currently, on initialization, it receives an array of ALL the slugs. However, that's probably not the way we'll want to keep it. 

so why is it the way it is now then?

> Instead, what I'll likely do is just include an empty array on initialization because nothing should be clickable (we want step bubble active state to be controlled entirely by the app). 

**we want step bubble active state to be controlled entirely by the app** ??? REALLY?? Because that's NOT what is happening now. Active state is being controlled by `StepBubbleMenu`. Look at the code I posted above. You are passing `activeBubble` to `StepBubbleMenu` and it is comparing that against the current bubble step slug when mapping the array and then using the results to set `StepBubble.isActive`.

HOWEVER... if we instead passed `isActive` as one of the properties for each step in the `bubbleData` array (or object) then the barcode scanner app would have to set that property as part of the logic in its event handlers that fire when the selector values change, and therefore the step bubble active state WOULD be controlled entirely by the app. The arguments you are using to dismiss my code review actually support the suggestions I have made. 

My refactor currently still passes `activeBubble` to `StepBubbleMenu` which is still comparing that against the current bubble step slug when mapping the array, because that is still a really easy way to set that property. That said, I see now that it can also be set easily by the event handlers in my refactor, and since "we want step bubble active state to be controlled entirely by the app", I'm going to go and make that change right now. (done!)

> The barcode scanner app is still not controlling the behaviour of what's clickable or not in its current implementation.

it is now in my refactor so you can no longer advance to a step until the previous steps have been completed.

> `With the existing code, the slug property is not required as part of the supplied bubble data and will be copied from the label if missing. The label however may not be usable in things such as URLs without having to convert it first (ie: "I'm a Label, Not a Key"), whereas the slug does not require conversion and is therefore more stable.`
>
> You make a good point about the label/slug and on review I didn't like the optional nature of the slug, especially when it is referenced via the clickable and activeBubble properties. So I decided to make it required and have done so by asserting incoming bubble configuration objects have the required properties and they are of the correct type (see d8d3b82).

except `typeof slug !== 'string'` is not enough to guarantee that a slug can be used as a key or URL param or whatever, so it appears you didn't really get my point. The perfectly valid string `"I'm a Label, Not a Key"` would still pass your verification, but can not be used as a key or URL param. My refactor drops `slug` as a `bubbleData` property and instead uses them as keys for that **object** (converted from array) which now guarantees they are valid keys.

> I'm not sure there's much value in each bubble configuration to have its own onClick callback over the current implementation because the provided callback receives the slug of the bubble that was clicked.

`BarcodeApp.onBubbleClick()` still receives the slug in my refactor too because the bubble steps can still share the same click handler. The benefit is that there is also the option to use a different click handler if need be, which... ya never know... could happen at some point in the future for some reason?!?!? The logic you use in your dismissal can easily be turned around and used against the way you have done things too eh? Where is the benefit in NOT having each bubble step having *the potential* to have its own click handler? It certainly wasn't any more complicated to add that functionality. So why not do it? As already stated above, `StepBubbleMenu` is passing `bubbleClick ` directly to `StepBubble`, so it's hard to argue that that's NOT where it belongs. It really blows me away that you would even attempt to argue this point. 

> The reason why there is no button for progressing to the next step is because I feel that's something that should be left up to the consuming code to control. There's a lot of variables affecting not only how the "next step" button might appear in the ui, ...

I totally agree with every point you made regarding having a "next" button not be part of the bubble step component.
We can bridge that discussion again when it comes time to review the barcode scanner app.

I really am sorry for how I reacted yesterday, because I should have given you the opportunity to respond to my code review before jumping to the conclusion that you would basically dismiss my valid constructive suggestions. Yesterday I truly felt bad for being a jerk and being nasty to you. Today I'm feeling extra sorry because now I don't know if you have dismissed my code review out of spite for my behaviour yesterday or whether this is how you would have responded anyways.

Code reviews are not a joke. If I see some wonky code, I'm going to comment on it and the code author should have the consideration to take my comments seriously. That applies to all of us, myself included. I'm not accusing you of thinking that code reviews are a joke, I'm just pointing out that they take time and effort on behalf of the reviewer, and the reviewer is attempting to help you and Event Espresso to be better.

I pinged you in chat yesterday and commented that I had **just** passed my code review back to you for "some rework" (which was my attempt at politely telling you that there were some "code smells" that needed to be cleaned up) and commented that we could discuss things in chat.  By that I was meaning that if you needed further clarification or had questions, then we could talk about how to fix things.
 
You responded in a matter of minutes with the comment that "there’s some things I agree can be changed and others that I don’t think can be changed."

This is how I know that you are way way way too eager to dismiss people's code reviews and why I got upset so quickly. In the 2-3 minutes between my passing the code review back to you and you're giving me that response, I'm pretty darn certain that you did NOT have enough time to fully consider my suggestions and/or have a look at your code to see how my suggestions might affect and/or change things, for better OR worse. Even if my code suggestions were really bad, that was NOT enough time to give things the consideration they deserve. Now granted... I sure didn't give you any extra time to do that because of my blowing up, and although I'm certain your comments were totally premature, it's always possible that you may have seen things differently had I given you time to look at the code some more. But we have gone down this road too many times before Darren. I blew up because I jumped to the conclusion that you were basically going to give me the kind of response that you ended up giving me above. Most of your comments are that I just need to look at some other code you've done and hopefully I'll understand why you are doing things, and some of your other arguments better support my position than your own, but overall you really haven't addressed my points adequately (imho), and I don't know if that's because they weren't clear enough (despite the code examples), or because you don't understand them (i doubt this actually), or that you haven't given them adequate consideration (i think it's this), or whether you just don't think your code can't get any better. 

But anyways... this isn't a competition, this isn't about who's a better programmer, because we all have different strengths and weaknesses, and hopefully if we all work together we can be greater than the sum of the parts. I absolutely **hate** having to tear into you like this. I do not enjoy this one bit and would love to never have to do this again. But I am the lead dev for Event Espresso and ultimately the code quality of this project is my responsibility and IMHO you are too dismissive of code reviews, and that needs to change. You can not just skim over people's suggestions and jump right to planning what to say to dismiss them (sorry but that's what it looks like).

bottom line:

**You need to start taking code reviews more seriously and give other's contributions more consideration.**

plz look at the two branches I pushed:

 - `eea-barcode-scanner/Gutenberg/convert-scanner-shortcode-to-block-brent`
 - `event-espresso-core/FET/add-step-bubble-nav-component-brent`

you will see that there are no additional renders, nothing overly complex, but things are much simpler, cleaner, and more direct, which will hopefully make things easier to debug for anyone in the future. That said, I'm also not saying that what I did is the be all end all of good coding. That's not the point. I want you to take my suggestions seriously though (especially those two rules of thumb I suggested) and use them as a guide to be an even more amazing programmer than you already are. I'm serious, I have tons of respect for you and everything you do.

nerrad:
# Code Changes

Thanks for taking the time to do the branches illustrating the approach you want to take for this component (and that also completes the step clickability control and improves the selectors event handling logic).

There is a couple of bugs in the `BarcodeApp` app branch though:

### `BarcodeApp` is mutating state.

In the event handler for `onEventChange` and `onDatetimeChange` the code has this:

```js
const bubbleData = this.state.bubbleData;
```

This simply sets bubbleData as a reference to `this.state.bubbleData` so when you mutate `bubbleData` you're also mutating state.

The [react docs](https://reactjs.org/docs/react-component.html#state) declare:

> Never mutate this.state directly, as calling setState() afterwards may replace the mutation you made. Treat this.state as if it were immutable.

There are multiple articles online that dive into the dangers of mutating state directly so I'm not going to go into that here.

### Users can click bubbles to go back to earlier steps, but the active state for the bubble does not update.

In your `BarcodeApp` branch I fixed the above two issues.

In my branch, I've updated the barcode scanner app so:

- it implements your suggested changes for selector event handlers (discrete `onEventChange` and `onDatetimeChange` callbacks).  I think it works great and does make the code a bit cleaner/clearer.
- it works similar to how your branches now work after the fixes but uses the approach I'm in favour of.

A reason why I updated my branch is because I think it helps highlight some reasoning behind why I chose my approach over the "everything in the configuration object" approach. Namely, in my branch, the `bubbleData` configuration object doesn't need modified at all thus there's no need to traverse through the configuration option tree to update properties on it (necessary because state is to be treated as immutable in react). Since the number of steps are dynamic, I think think it helps the api be a bit more easier to work as the the number of steps vary across applications.

With that said, let me be clear that I'm not saying my way is right and your way is wrong.  I think both approaches are *valid* approaches to the component.  I just have the opinion that the logic and api is a bit easier to work with in my approach.  With that said, if you disagree with me, that's okay, just let me know what you want to go with (now that both approaches are functional) and I'll update the related pulls.

## Responses

### Using `componentDidUpdate` to control whether to re-render.

Yes this is heavily cautioned against because it's easy to use improperly. In this case I used it properly because I didn't update the state for the active selector unless specific conditions were met.  I was making this even more restrictive than a pure component (Pure components implement `shouldComponent` that does a shallow compare of prop and state changes, whereas regular components re-render on _any_ prop/state change).  This is a moot point though because I like your proposed changes that remove the need to use it.  I just think it's worth clarifying that `componentDidUpdate` does have a purpose and can be used.

### "functions calling functions calling functions"

Welcome to `react` (or even javascript!).  I don't think its terribly abnormal to do what I did, but I agree that it _is good_ to reduce unnecessary calls and your proposed changes do that.  I often do not have things working optimally the first pass (or even sometimes the second pass!), that's why we iterate.  I should note I never indicated that the Barcode App was "complete".  Nevertheless, point taken (and I implemented your changes to reduce the unnecessary functions).

### "But you hope your explanation helps me... what? ...understand that... "OH... you do things for REASONS!!!""

When I wrote "I hope my explanation helps" I acknowledge that you perceived that as condescending and I'm sorry for that.  I do believe you're capable of understanding things. I chose these words with the goal that you would acknowledge the explanation for why I chose to implement things the way I did. I wasn't confident that you had an opportunity to receive my point of view on this. That does not mean that I expect you to read my explanation and automatically agree with me or that all of a sudden convert to my way of thinking.  I was merely taking an opportunity to flesh out the decision making behind my choices.  I also did not mean (in case that's the way you took it) that I think my way is right and your way is wrong.  I think both approaches are valid and was wanting make sure that there was documented reasons for why I chose my approach to counter-balance the documented reasons you had for implementing your approach.

Nevertheless, point taken. I have to continue to be careful of the words I write and how they might be perceived in the future.  It's something I'm always striving to work on but still fail terribly at times.

> It would help much better in the future if you just posted those reasons directly and then we could actually discuss them.

I thought I did, but based on your words, I didn't.  I'm struggling with how to resolve this in the future because if I think I'm giving an explanation, but you don't think I am, then there's a bit of an impasse.  So all I can say is I'll keep trying harder.

### Here's the best one, bubbleClick is just passed DIRECTLY to StepBubble !?!?! StepBubbleMenu literally does nothing with this property so there is just no way you can argue that this should be a property of StepBubbleMenu. If it belongs on StepBubbleMenu... then how come StepBubbleMenu isn't doing anything with it?

> Here's the best one, ...

What do you mean by that?  Are you saying that's the best example of a poor argument or "REASON" I gave in any of my answers?  Are you being sarcastic?  It's phrases and wording like this that lead me to making statements like "Please understand, I didn't arbitrarily arrive at the structure for this component" because it appears to me that is exactly what you are thinking.  I'm willing to consider how I write things, but I'd appreciate the same effort from you.  

To the specific example you are making here, this is a common pattern in react in which props are passed through components in the tree (especially for things like onClick events).  What's puzzling to me is that you are making this argument when your example is effectively doing the same thing except you've just moved the `<StepBubble>` generation into a `getStepBubbles` function along with `StepBubbles` still passing down a onClick handler via an `action` prop vs the `bubbleClick` prop name I was using. The only difference in your case that the click handler came in as part of the bubbleData object, whereas in my case it's a distinct prop passed through.  The behaviour of passing props through is one of the things that led to the rise of using global state (via redux) for complicated apps or the new React hooks api so that props do not _have_ to be passed through.  We _could_ use global state here (which I think would be overkill) or explore using React.Hooks but I think that's unnecessary in this case. 

I understand that you are making the point that the onClick handler is a property of `StepBubble` and thus should be a part of the configuration object `StepBubbleMenu` receives to _build_ the `StepBubble` components. For reasons (yes "reasons!"  what else would I call it?) I'm attempting to explain, I felt that _in this case - i.e. I don't think its a universal principle_ I thought it would better coming into `StepBubbleMenu` as a part of the `props` object (which is also an object describing the `StepBubbleMenu` as a unit?) for it constructing the individual `stepBubbles`.  In other words, from my perspective `bubbleClick` is a property of `stepBubbleMenu` as it doesn't change for any of the `stepBubbles` that might trigger the click event. In other words, the `<StepBubbleMenu>`  requires a click handler that will be fired when _any_ menu item is clicked (if it's clickable). Let me be clear, I'm _not_ discarding your approach, or failing to consider it, I'm simply articulating why I prefer my approach and the reason for why I implemented it.

### "...so why is it the way it is now then?"

Because as I mentioned in the [pull description](https://github.com/eventespresso/eea-barcode-scanner/pull/7):

> Priority wise, the goal of this pull is to get to the app to a functional state replacing the feature set of the current shortcode powered app. This means that everything should work, but it won't be in a releasable state until the ui/ux receives some polish.

In other words, its not something I consider "complete" (and I have not put it in the Code Review pipeline or requested review).  UI/UX polish includes improving behaviour around menu reactivity and selection (which is also one of the unchecked todo items in the pull request description).  Also I have not gone through and added comment blocks for all the code or added tests yet because it is not ready for review.  The pull request description includes a todo list of things left to do before it's ready for review/release.

### we want step bubble active state to be controlled entirely by the app ???

> we want step bubble active state to be controlled entirely by the app ??? REALLY?? Because that's NOT what is happening now. Active state is being controlled by StepBubbleMenu. Look at the code I posted above. You are passing activeBubble to StepBubbleMenu and it is comparing that against the current bubble step slug when mapping the array and then using the results to set StepBubble.isActive.

From my perspective, I consider what the code is doing in my branch as being controlled by the Barcode app.  `activeSlug` is a string matching a slug passed into `StepBubbleMenu` by the `BarcodeApp` component.  The slug for each step is also passed through as a prop from `BarcodeApp` component.  The props passed to `StepBubbleMenu` are the configuration for the `StepBubbleMenu` component as a unit. So effectively the `BarcodeApp` component is giving the instruction to `StepBubbleMenu` that when it declares the `StepBubble` components, it is to mark the one with the same slug as `activeSlug` as `active`.  Thus, the `BarcodeApp` still has control. 

I'm guessing why you don't consider it as having control is because the equality check happens in `StepBubbleMenu` and thus the logic deriving the actual active state is contained in this component?  I'm looking it from the the perspective of an api provided to components using `StepBubbleMenu` that declare whatever slug you provide `StepBubbleMenu` as the `activeBubble` prop, that will be the step marked active from the bubble components built using the provided configuration objects. So effectively, the passed in props _are_ controlling the active state.

### "except typeof slug !== 'string' is not enough to guarantee that a slug can be used as a key or URL param or whatever, so it appears you didn't really get my point."

I'm sorry you thought I didn't get your point.  I was trying to avoid writing a book and stick to explaining why I made the choices I did to give you some insight into my thought process.  So to address your point about `slug` not being usable as a key or URL param.

- re URL param.  The step menu is not a href linking mechanism thus urls are not involved.  If the controlling component (i.e. what is _implementing_ the `StepBubbleMenu` component) wants to react to the `onClick` event on a menu so that it triggers a redirect to a url then it's up to that component to validate its slugs accordingly.  From the perspective of the `StepBubbleMenu` why should it care? It's not a _requirement_ for anything it does.
- re `key`.  I'm not entirely clear on what you mean by that but I'm assuming you mean the `key` special prop on the `StepBubble` component. There's no clear documentation on what a `key` in a react component can/cannot be but I know for sure it can be any string or number so I'm not sure why a non spaced slug would have to be used.
- Further, are you aware that properties in javascript objects don't have to be `some_slug` or `some-slug`?

```js
const a = {
    [ 'this has a space in it' ]: 'foo',
    [ 'mix of spaces and-dashes' ]: 'bar',
};
// console.log( a );
// {
//    this has a space in it: 'foo',
//    mix of spaces and-dashed: 'bar',
// }
```

Thus its entirely possible for a property in an object to be something that is "unsafe" for url or html attribute use.

So with that said, I'm kind of puzzled why you think using keyed objects by slugs will resolve the issues you named.

I only tested for slug and label being strings because that satisfies the requirements `StepBubbleMenu` has for those properties.

### "BarcodeApp.onBubbleClick() still receives the slug in my refactor too because the bubble steps can still share the same click handler."

> BarcodeApp.onBubbleClick() still receives the slug in my refactor too because the bubble steps can still share the same click handler. The benefit is that there is also the option to use a different click handler if need be, which... ya never know... could happen at some point in the future for some reason?!?!? The logic you use in your dismissal can easily be turned around and used against the way you have done things too eh? Where is the benefit in NOT having each bubble step having the potential to have its own click handler? It certainly wasn't any more complicated to add that functionality. So why not do it? As already stated above, StepBubbleMenu is passing bubbleClick directly to StepBubble, so it's hard to argue that that's NOT where it belongs. It really blows me away that you would even attempt to argue this point.

To be clear, I got your point that it would allow for distinct click handlers for each `stepBubble`  to allow for some future hypothetical requirement.  However,

- what I implemented _does not prevent that hypothetical future use from being implemented_ as the slug provided to the callback still helps derive what specific step triggered the event. So really the only reason for adding the callback to the step bubble configuration object is because you consider it as a property that describes the configuration object and thus belongs in that object.
- I did provide an explanation for why I did not do as you suggested and that was because it already receives the slug of the callback and thus a single callback accomplishes the required needs. To me, it is a property of the `StepBubbleMenu` component - i.e. When something in the menu is clicked, this handler is called.  To expand on that here, I think doing it the way you are suggesting will be in the majority of use-cases, superfluous code.  The api for the `StepBubbleMenu` is opinionated. An `onClick` event is triggered when something in the Menu component is clicked (when it is clickable or if triggered via accessibility tools). That's it.  So why complicate the configuration object for each bubble more than it has to be? That was my reasoning _and why it made sense to me to implement it the way I did_.  I understand you disagree and your reasons for why you disagree, I'm just giving you the reasons for _why I did what I did_.

To _reiterate_, none of what I'm saying is an attempt to justify the position that my approach is right and your approach is wrong.  It's an attempt to give insight into why I prefer my approach.  As mentioned in my original response, I had something similar to what you are proposing in an earlier iteration of this component when doing the barcode scanner app and then tried this as an alternative.  So I definitely _have_ considered your approach.

## Regarding further non-code related issues.

Some of your response in this pull request isn't really dealing with the pull specifically but your frustration with what you perceive to be my dismissal of your code reviews and other things you've named.  I don't think this is the appropriate venue for this kind of discussion so I'll address those things in a slack post in our team chat.

tn3rb:
pushed some more changes to:

- `event-espresso-core/FET/add-step-bubble-nav-component-brent`
- `eea-barcode-scanner/Gutenberg/convert-scanner-shortcode-to-block-brent`

I've reworked the css a bit for the bubble steps so that they match the ones we use in core for PHP code in:
```/core/services/progress_steps/display_strategies/number_bubbles/```
and also so that they work with mobile:

320 x 480
BubbleStepMenuMobile

480 x 320
BubbleStepMenuMobile2

I've pushed some other changes to the barcode scanner to 
```eea-barcode-scanner/Gutenberg/convert-scanner-shortcode-to-block-brent-plus-css```
which were just the result of me playing around with the scanner yesterday while tweaking the media queries for the bubble step menu. I realize the barcode scanner isn't done, so feel free to use or discard the changes in this branch.
If you want to pass off all of the final UI tweaking to me, that's fine. I can do that while you tweak the js model query logic in the admin refactor stuff.

One final point, I don't know if this is also the result of the barcode scanner just not being finished yet, but I wanted to mention it. 

I noticed if I used the "Lookup Attendee" option on the "Scan" step and entered a Reg Code from a registration in my system, it would return the registration and attendee details including whether the registration was approved or not and/or required payment or not. That sounds fine but the issue is that the registration did not belong to the event and datetime that had been selected in the previous steps. Since the returned attendee details make no mention of the event or datetime, it simply appears that the information is valid for the current selections. However, the registration could belong to **any** event, even one that occurred years ago. This could easily allow people to print bogus tickets using barcodes / QR codes from a previously paid for past event to gain fraudulent entrance to a current event they have not paid for.

Despite the above issue, my overall impression is that this is going to be so much quicker and nicer to work with than the PHP version (which will likely be true for all of our React based rebuilds).

nerrad:
Okay @tn3rb here's what has been done:

### Barcode Scanner App repo and branches.

- force pushed your "css" branch to the branch attached to https://github.com/eventespresso/eea-barcode-scanner/pull/7.  So going forward you'll want to create a fresh branch from the repo for any future changes.
- I deleted your two extra branches from the repository so there will be no confusion.
- Any more discussion about that work should probably happen in the pull so we're not making this one any more cluttered than it needs to be :)
- I like the changes you did there (style is great).  I don't consider it done or ready for (more) ui/ux work yet because as you pointed out (which I've added as a checklist item) there's still some bugs.  I also want to get the css changes in EE core done so they can be built on top of in the barcode scanner app.  I also still have a few things to extract to ee core from the work in the app yet.  I'll let you know when its ready for your ui/ux work :)

> Despite the above issue, my overall impression is that this is going to be so much quicker and nicer to work with than the PHP version (which will likely be true for all of our React based rebuilds).

Most definitely!  The only thing I'm not sure of yet is extensibility.  But we can cross those bridges as we encounter them.

### This pull

- I pushed the changes from your branch into the branch associated with this pull.  I force pushed, but it was a clean merge, so the existing branch on github can be pulled into the local branch for this pull without issue.
- I deleted your branch on the repo so there's no confusion, future work should be pushed to the branch for this pull.
- I have a couple minor tweaks to your latest code that could use a review.
- I also pushed fixes to the tests due to the changes.

nerrad:
cancelled all the php tests in the latest test build because this is a js_only pull.  If those pass I'll merge.

PR #1069, opened 2019-04-11
account for permalinks disabled when deciding when to enable REST API Basic Auth

mnelson4:

Fix mobile app (and all REST API requests using basic auth) authentication when permalinks disabled.

While testing https://github.com/eventespresso/event-espresso-core/pull/1064#issuecomment-481932322, Josh had trouble using the mobile apps on master.
He noted on https://github.com/eventespresso/event-espresso-core/pull/1064#issuecomment-481932322 that he discovered the problem only happened with a site using default permalinks. That issue is where the bug was introduced.

I did some digging and found the culprit was `EventEspresso\core\domain\services\contexts\RequestTypeContextDetector::isEspressoRestApiRequest()`, which wasn't aware a request like `mysite.com/?rest_route=ee/v4.8.36/registrations` was a valid REST request.

Ideally we would be using the same logic `rest_api_loaded()` uses to detect a REST request: `isset($GLOBALS['wp']->query_vars['rest_route'])`, but the global `$wp` hasn't been created yet by this stage of the request. So the workaround is to just handle the two cases: permalinks enabled (in which case it will have the REST URL prefix in the path, the case we handled before) or permalinks disabled (in which case it will have the GET parameter `rest_route` in there).

## How has this been tested

* [x] setup a site with permalinks disabled. Try to use the mobile apps with it. It should work fine.
* [x] Setup a  site with permalinks enabled and repeat. It should also work.
* [x] If reasonable, verify no regression from tests on https://github.com/eventespresso/event-espresso-core/pull/974#issue-255168265

## Checklist

* [x] My code is tested.


joshfeck:
ready for merging

PR #1070, opened 2019-04-12
Add Event Category column to events overview list table (fixes #426)

### LINKED ISSUE #426: Add a category column to the events overview
Codebase https://events.codebasehq.com/projects/event-espresso/tickets/8506

Per Seth:
Please add an option in the screen options menu to display a category column to the event overview in the admin. It can default to being turned off, but it would be very valuable to me to be able to see the categories a particular event is in.

Per Darren:
What about other list tables for things that have categories (Venues, People)?

Per Seth:
Ya, that would be great. Right now I am setting up multiples of the same event, but in different categories and it would be REALLY helpful to see what category each event is assigned to.

nerrad:

See #426 for original request.  In this pull,

- Event Category column is added after the venue column if there are event categories existing on the site.
- For *new installs* it is hidden by default.  Existing installs will see the column if the above conditions are met.
- It is not a sortable column, nor are the terms linked.  

**Note:** The issue was sparse on details for what was desired so I intentionally kept the initial iteration very limited in scope.  What was implemented resulted from [some feedback in slack](https://eventespresso.slack.com/archives/C080L8G68/p1555081032036500).

## How has this been tested

Testing Steps (checked items are things I've tested):

* [x] Verify on new installs that the Event Categories column is hidden
* [x] Verify that the Event Categories column shows up as an option (or in the list table) when there are _any_ event categories.
* [x] Verify the column does not show up in the list table (or as an option in the screen options) if there are no event categories on the site.
* [x] Verify event categories show up in the row for an event that has categories and they are the correct categories.

## Checklist

* [x] My code is tested.


admin_pages/events/Events_Admin_List_Table.class.php
```diff
@@ -515,4 +537,31 @@ public function column_actions(EE_Event $item)
...
+     *
+     * @throws EE_Error
+     * @throws InvalidArgumentException
+     * @throws InvalidDataTypeException
+     * @throws InvalidInterfaceException
+     */
+    private function addConditionalColumns()
+    {
+        $event_category_count = EEM_Term::instance()->count(
+            [['Term_Taxonomy.taxonomy' => 'espresso_event_categories']]
```
mnelson4: There is a constant `EEM_CPT_Base::EVENT_CATEGORY_TAXONOMY` that could be used eh

nerrad: ooo ya good catch, I'll use the constant.

joshfeck:
I found a strange issue related to displaying the columns when the People add-on is activated. What happens is the People column for the event is squished into the Actions column, when it's a filtered view.

Screenshot of the initial view where everything looks good:
Screen Shot 2019-04-12 at 6 38 31 PM

Then after clicking the Filter button (no filter settings changed, just clicked Filter)
Screen Shot 2019-04-12 at 6 38 51 PM

Then after clicking the Reset button the columns are OK again:
Screen Shot 2019-04-12 at 6 39 12 PM


I don't believe adding the column in this PR causes this, but having one more default column seems to expose the issue where it otherwise would have gone unnoticed.

joshfeck:
Apparently cpt-to-people.css isn't loaded when the event list table is filtered.

nerrad:
> Apparently cpt-to-people.css isn't loaded when the event list table is filtered.

Good sleuthing!  It sounds like something that we'd need to fix in the people add-on?

joshfeck:
opened https://github.com/eventespresso/eea-people-addon/issues/16

PR #1071, opened 2019-04-12
Change default for showing Event Espresso link to false (fixes #932)

### LINKED ISSUE #932: Set credit link option to default to "No"
The "Link to Event Espresso in your Registration Page?" option needs to be opt-in for decaf. See also [this P2](http://p2.eventespresso.com/?p=4204#comment-16654).

nerrad:
See #932 for details.

This simply changes the default for the `EE_Admin_Config::show_reg_footer` option to false for all new installs.

**Note:** this _only_ applies to new installs of Event Espresso Core and applies to both decaf and caf installs.

## How has this been tested

* [x] Verify that on new installs of EE decaf that the option defaults to false.
* [x] Verify that on new installs of EE caffeinated that the option defaults to true.

I had to change the loading order of the `brewing_regular.php` file in this pull which affects caffeinated payment method registration. So we need to verify that no caffeinated payment methods are affected

* [x] Verify that caffeinated payment methods on an existing site prior to this branch still work as expected (if you test one, then all of them should be fine).
* [x] Verify that caffeinated payment methods on a brand new site activated on this branch still work as expected (if you test one, then all of them should be fine).

## Checklist

* [x] My code is tested.


nerrad:
> If that doesn't jump out as a "yes, absolutely, let's do that now!" idea, we can just sit on it (and it would be a separate ticket anyway.)

Whoever tests this can provide feedback on this comment and if they agree, then ya a new issue can be created.

nerrad:
For the latest changes:

- in my initial fix to the test I was just calling `set_current_screen` and then I noticed fails happening in various modules and shortcodes because they were defining constants in their `set_admin_hooks` callbacks.  So I started fixing those but then I realized if those functions were getting called, then there's things loading for tests that shouldn't be loading (i.e. every test ended up being in the admin context).  So I left those changes in because I think its good to make sure we're only ever setting constants once anyways.  I'm sure there's more modules/shortcodes doing the same thing but that could be done separately.
- Instead, what I did was reset the `$current_screen` global to `null` which is what it is normally for test runs.
- I also ended up adding a long running test to the `big_rest_tests` group because I noticed it was taking a large amount of time to run (relatively).  It could have been done in a separate pull but just as easy to put in here.

joshfeck:
I disagree that the option should be moved or made more prominent. 

The request in #932 and the related P2 was to make this change for decaf only. Can some logic be added so the option is set to No only if the version at time of installation is decaf, otherwise it's set to Yes?

nerrad:
> Can some logic be added so the option is set to No only if the version at time of installation is decaf, otherwise it's set to Yes?

Oops sorry, I forgot it was supposed to be decaf only.  I'll put this back to my in progress tree and will try to adjust that.  I'm uncertain whether this is straightforward to do because some of the defaults are set really early in the request so I don't know if the functionality we have for detecting decaf only happens early enough.

nerrad:
@mnelson4 latest changes are in ba807ac where the default is now conditionally dependent on whether the environment is decaf or not.

admin_pages/general_settings/AdminOptionsSettings.php
```diff
@@ -149,7 +157,7 @@ public function generate()
                             ),
                             'default'         => isset($this->registry->CFG->admin->show_reg_footer)
                                 ? filter_var($this->registry->CFG->admin->show_reg_footer, FILTER_VALIDATE_BOOLEAN)
-                                : true,
+                                : DomainFactory::getEventEspressoCoreDomain()->isCaffeinated(),
```
mnelson4: I'd prefer we do `apply_filters(...)` here, and add a filter from within caff. That seems like the more normal way to do things, rather than asking "is this a caff install?" (having code that checks if it's a premium version of the plugin sounds bad, and if we were doing it to outright disable some functionality it would be in violation of the WP.org plugin guidelines, right?)

If that's not possible, this is fine then.

nerrad: oh ya! that's a good point :)  I agree.  I'm not sure if we can do the same for the `EE_Config` default because of loading order for the config but I'll try it  there.

nerrad: > having code that checks if it's a premium version of the plugin sounds bad, and if we were doing it to outright disable some functionality it would be in violation of the WP.org plugin guidelines, right?

Ya on the surface it might look bad, but in this case its actually changing the default for an option (so not disabling the ability to toggle on).  In this case it is an option that wp.org coding guidelines would frown on for being "on" by default.

nerrad:
Who knew changing a default would take so long :)

@mnelson4 in the latest changes:

- I filter the default in the EE_Admin_Config constructor (option defaults to false).
- No need to also filter in the AdminOptionsSettings form handler as the ternary there could be removed (`filter_var` assists with the isset check).
- `brewing_regular.php` hooks into the filter and changest the default to true.
- I had to change the loading order of `brewing_regular.php` so that it is loaded before the EE_Config is constructed.  I looked over all that is loaded in brewing regular and it was all hooked into filters that loaded later in the execution path _except_ for the payment registration api.  So I modified that call so it is executed when configuration is loaded (which is pretty much the same loading order as it would have been before).
- `brewing_regular` has a `AHEE__EE_Brewing_Regular__construct__complete` action that is invoked in it's constructor, I did a search through the EE codebase and I don't see any instance of that being called, so impact there should not be affected as well.

admin_pages/general_settings/AdminOptionsSettings.php
mnelson4: Ya it mostly looks bad at first, but is actually ok I think.

mnelson4:
Yeah, in slack Darren said @tn3rb should review too.
So Brent, I'd suggest you especially review the changes on https://github.com/eventespresso/event-espresso-core/pull/1071/files/acfa66e4e3f110d38030eb4c282ed2ea545b361f, where we now initialize caff before the config (instead of after, which is what's in master right now).
Darren explained the changes here: https://github.com/eventespresso/event-espresso-core/pull/1071#issuecomment-483416608

tn3rb:
ya I think these changes are an improvement in terms of how things load and/or get filtered instead of manipulated directly.

The one thing I find odd is the addition of the `EE_Brewing_Regular.defaultFilters()` whose single filter is not really doing anything different than the other filters in `EE_Brewing_Regular.__construct()`. So not sure why it is in its own method by itself. I would either do one of the following:

 - just move the single filter in `defaultFilters()` into the constructor so that everything is together and then delete that new method

or

 - move all of the other `add_filter()` and `add_action()` calls in the constructor (but not the `do_action()`) into `defaultFilters()` and maybe rename that new method to something like `set_hooks()`. This way everything is together.

nerrad:
> So not sure why it is in its own method by itself.

Ya I didn't make that really clear, I gave it that name because I thought it would be good to have a specific method for any filters that modify "default" values for things.  So it is a function that _only_ sets filters for modifying defaults (as opposed to action and filters that do anything else but modify defaults).  I could be even more targeted here and name the function `defaultFiltersForConfig`?

Does that help with my reasoning behind the method and name?  I'm happy to just make it more generic have a `setHooks` method (or move back into the constructor with a doc block labelling it as a filter setting defaults.

nerrad:
Hmm, although I _did_ add this comment block:

```
/**
     * Various filters for affecting default configuration values in the caffeinated
     * context.
     */
```

nerrad:
Discussed in Slack with @tn3rb and he pointed out that these are currently in the constructor:

```
add_filter('FHEE_EE_Html_messenger__add_powered_by_credit_link_to_receipt_and_invoice', '__return_false');
add_filter('FHEE__ee_show_affiliate_links', '__return_false');
```

So per my reasoning, they should be in this new method as well.  I agree.  I'm also going to take the opportunity to do a bit of reorganizing so all the things are moved out of the constructor into more specific methods.

joshfeck:
This was tested and is ready for merging.

PR #1094, opened 2019-04-16
Bug/avoid unnecessary alter table queries

mnelson4:
Note: This PR now also contains changes from 

* https://github.com/eventespresso/event-espresso-core/pull/1100
* https://github.com/eventespresso/event-espresso-core/pull/1094

Their milestone should be updated also when this is merged.

## Problem this Pull Request solves

Corrects various problems in the MySQL table definitions in the DMSes from 4.1-4.7 which caused WorPress' `dbdelta` to issue unnecessary `alter table` MySQL commands, which took extra time.
Mostly, this meant making the DB column types lowercase, and defining their size.
See https://github.com/eventespresso/event-espresso-core/issues/978 where this was requested.

## How has this been tested

* [x] Superficially, you should delete EE4's data (or start wit a new install), 
* then install EE3 and create some data, including
* * set a question's required text to something bigger than 100 characters.
* * a paid event, register for it, and add a payment from the admin. set the "Transaction Type" field to be "123456789a1234567890b123467890c1234567890".
* deactivate EE3 and activate EE4
* run the first (4.1.0) migration script, you should see NO errors from it.
* continue all the migrations. There should be no new errors and the tables should be created a-ok. Also
* * the question with the long required text should have the same required text (see https://github.com/eventespresso/event-espresso-core/pull/1100)
* * Notice the payment that had long "Transaction type" should still exists in EE4 (whereas on master, it would have failed to migrate; see https://github.com/eventespresso/event-espresso-core/pull/1101).
* [ ] The alter DB queries are ran during AJAX calls, which unfortunately makes finding out exactly what they are tricky. I don't believe the Query Monitor plugin will be of much help. I had to set my site up with xdebug, and add a breakpoint to `wp-admin/includes/upgrade.php` line 2493, which is inside `dbdelta`, and contains `foreach ($cfields as $fieldname => $fielddef) {`. And look at the values of `$cfields`, which means "changed database columns," to see if it thinks there are any tables that got changed when they really didn't. (The easiest way to check if a column "should" be altered during a DMS is to compare the current DMS, eg `EE_DMS_Core_4_3_0` and its definition for the table containing that database column, to the previous DMS, eg `EE_DMS_Core_4_2_0`, and its definition for the table containing that database column. If they're the same, that column shouldn't be mentioned in `$cfields`. Really only new columns should be in there. Anyways, that's all quite a bit of work, but I've gone through it and am satisifed with it, so just the first test is the only test that's really necessary.

## Checklist

[...]

PR #1100, opened 2019-04-17
make QST_required_text a text column instead of varchar. let the peop…

mnelson4:
# Note: this is a PR into another bug branch (to clarify what's being changed here) NOT master

Fixes migrations from EE3 when the question's required text was larger than 100 characters. (In EE3 there was no limit to its size, so this restores that behaviour).

## How has this been tested

* [ ] Nuke EE4 or create a new site.
* Install EE3 and set a question's required text to something bigger than 100 characters.
* Deactivate EE3 and activate EE4 and run the migrations. The question's required text should remain the same.
* In EE4, try to set a new question's require text to something larger than 100 characters. It should work fine now.

## Checklist

* [x] My code is tested.

### DIFF
--- a/core/data_migration_scripts/EE_DMS_Core_4_9_0.dms.php
+++ b/core/data_migration_scripts/EE_DMS_Core_4_9_0.dms.php
@@ -412,7 +412,7 @@ class EE_DMS_Core_4_9_0 extends EE_Data_Migration_Script_Base
 				QST_deleted tinyint(2) unsigned NOT NULL DEFAULT 0,
 				PRIMARY KEY  (QST_ID),
 				KEY QST_order (QST_order)';
-        $this->_table_is_changed_in_this_version($table_name, $sql, 'ENGINE=InnoDB');
+        $this->_table_has_not_changed_since_previous($table_name, $sql, 'ENGINE=InnoDB');
         $table_name = 'esp_question_group_question';
         $sql = "QGQ_ID int(10) unsigned NOT NULL AUTO_INCREMENT,
 				QSG_ID int(10) unsigned NOT NULL,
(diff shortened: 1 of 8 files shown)


core/data_migration_scripts/EE_DMS_Core_4_9_0.dms.php
```diff
@@ -404,15 +404,15 @@ public function schema_changes_before_migration()
...
+				QST_required_text text NULL,
 				QST_order tinyint(2) unsigned NOT NULL DEFAULT 0,
 				QST_admin_only tinyint(1) NOT NULL DEFAULT 0,
 				QST_max smallint(5) NOT NULL DEFAULT -1,
 				QST_wp_user bigint(20) unsigned NULL,
 				QST_deleted tinyint(2) unsigned NOT NULL DEFAULT 0,
 				PRIMARY KEY  (QST_ID),
 				KEY QST_order (QST_order)';
-        $this->_table_has_not_changed_since_previous($table_name, $sql, 'ENGINE=InnoDB');
+        $this->_table_is_changed_in_this_version($table_name, $sql, 'ENGINE=InnoDB');
```
nerrad: Is this so that _existing_ 4.9.0 sites when updated to 4.9.x will get the table updated?

mnelson4: Hmmmm thinking about it it's not necessary. I originally thought it would help when someone updates from 4.9.x to this version, because we're telling it the DB changed. But on an update like that we're going to check all the database tables anyway. The methods `_table_has_not_changed_since_previous` and `_table_is_changed_in_this_version` really just affect migrations. And in this case, if someone's going through all the migrations, there's actually nothing changed in this table.

(The only case this could help is someone has been on 4.8.x, then upgrades to this version of 4.9.x; they'll get the changes to this database table. But I doubt there are very many users hanging onto a 4.8.x version; they're basically all on 3.1.x, and those ones are going to run the full gamet of migrations, and so won't benefit from this.)

mnelson4:
merged into base branch. Updating testing notes on the other pr https://github.com/eventespresso/event-espresso-core/pull/1094

PR #1101, opened 2019-04-17
Bug/bigger pay gateway column

mnelson4:
# Note: this PR is NOT off master!

Fixes migrations from EE3 where old payments "txn_type" had a value between 32 and 45 characters. The column used to store that data in EE4 only allowed 32 characters, (whereas in EE3 it allowed up to 45), so there would be a database error during migration under those conditions.
This avoids the problem by allowing the column to have 45 characters. (It's removed later anyway, so it's not actually permanently taking up more space anyway!)

## How has this been tested

* [ ] Nuke EE4 and activate EE3. Create a paid event, register for it, and add a payment from the admin. set the "Transaction Type" field to be "123456789a1234567890b123467890c1234567890". 
* Deactivate EE3 and migrate to EE4. Only do the first migration script though. You should see NO errors from it. You can continue the migration. Notice the payment still exists in EE4 (whereas on master, it would have failed to migrate).

## Checklist

* [x] My code is tested.


mnelson4:
Merged into https://github.com/eventespresso/event-espresso-core/pull/1094, and updated its testing notes.

PR #1102, opened 2019-04-17
Migrate partial payments, txn_type to extra accounting, and avoid column name collisions during migrations

mnelson4:

Fixed migrating payments for attendees that were only partially paid (before their payments would not get migrated into EE4, now they do).
Fixed migrating line items (only the line items for the first transaction would get migrated).
Fixed so EE3's payment "Transaction Type" (events_attendee.txn_type) gets migrated to EE4's payment "Extra Accounting) (esp_payment.PAY_extra_accntng). 
It was originally only migrated to esp_payment.PAY_gateway, but that column was removed in 4.6.
Also avoided some of the gotchas noticed in this ticket for other migration stages (by avoiding database column name collisions).

## How has this been tested

* [ ] Nuke EE4 and activate EE3. 
* create a custom question
* create a paid event using that custom question,
* register for it (but do NOT create the registrations from the admin, those are known to not migrate correctly) and pay for it with something like PayPal.
* register for it again, but don't pay at all (use invoice)
* register a 3rd time, use invoice, then add a full payment for it from the admin, and set the "Transaction Type" to "Star Wars". 
* register a 4th time, use invoice, and add a PARTIAL payment for in in the admin.y for it but don't pay.
* Deactivate EE3 and activate EE4. Run the migrations
* The 1st registraiton's payment should have migrated over fine.
* The 2nd registration should have migrated over fine, with no payment. You should see the registration from the transaction's details page.  If you refresh the transaction admin page, it should appear fine.
* The 3rd registration's payment should be migrated fine; AND it now has "Extra Accounting" set to "Star Wars" (on master it wouldn't).
* The 4th registration's payment should be migrated fine; the transaction won't be complete, but they at least get credit for what was paid previously.

## Checklist

* [x] My code is tested.


joshfeck:
I followed the testing steps, but I'm afraid the migrated transaction doesn't show a payment in the Event Espresso > Transaction page. I have a few screenshots.

On the transactions list table, the payment is shown ($8.00 of a $10.00 ticket)
Screen Shot 2019-04-19 at 4 26 26 PM

Then on the transaction page, it says no payment was made:
Screen Shot 2019-04-19 at 4 26 56 PM

Also you'll notice that no registrations are listed. I think this is a problem in master too. I checked 4.9.79.p tag and the registration is listed in the transaction.

mnelson4:
Hmmm yes I'm reproducing for partial payments, but complete payments seem to be migrated over ok. The line items are only getting created for fully-paid transactions. (Also strange: if you visit the transaction page, initially it says a payment of x dollars is owing, but then when you refresh the page, it drops to 0 dollars owing, because the transaction total gets updated by the non-existent line items.)

> I think this is a problem in master too. I checked 4.9.79.p tag and the registration is listed in the transaction.

Are you saying that after you've migrated, if you switch to 4.9.79.p, you can see the registrations on the transaction? If so, I'm not seeing that (this is what I did: create some partial payments in EE3, migrate to EE4 using master, confirm the registrations don't appear on the transaction page as you reported, then switch to 4.9.79.p and refresh the transaction page, and don't see any difference.)

joshfeck:
No I'm saying if I do the migration with the 4.9.79.p tag activated, migrating the same data from EE3, the registration is listed in the transaction. If I do the migration with master, the registration is not listed in the transaction.

joshfeck:
Works for me!

PR #1108, opened 2019-04-18
Bug/migration touchups

mnelson4:

* Don't migrate Invoice Address field from Event Espresso 3
More information in https://events.codebasehq.com/projects/event-espresso/tickets/11490

* Fix conditional in EE_Data_Migration_Manager PHP5.6+ environments
More information in https://events.codebasehq.com/projects/event-espresso/tickets/10595

* Fix Undefined index: invoice_css migrating from 4.4 -> 4.5
https://events.codebasehq.com/projects/event-espresso/tickets/7975

Lumped together onto one ticket because they're each so small IMO. But if one holds up the others, yeah, let's break them into separate tickets...

## How has this been tested

* [ ] Delete EE4's data and activate EE3...
* Activate the invoice payment method and give the "Payee Address" a really silly name
* in wp-config, set `define('EE_MIGRATION_STEP_SIZE', 4);`
* Deactivate EE3 and activate EE4. 
* Run the first migration. Note that, with a couple exceptions (because the migrations aren't too strict about doing exactly the number of records requested, they're revels that way), it should only migrate 4 items at a time, instead of the default 50.
* Run the migrations up until 4.5. After that one, there should be no warning like mentioned on https://events.codebasehq.com/projects/event-espresso/tickets/7975
* Finish the migrations. 
* View the invoice payment method. The address should NOT have been migrated from EE3 (for the reasons Josh mentioned on https://events.codebasehq.com/projects/event-espresso/tickets/11490)

## Checklist

* [x] My code is tested.


core/EE_Data_Migration_Manager.core.php
```diff
@@ -655,7 +655,7 @@ public function migration_step($step_size = 0)
         // ok so we definitely have a data migration script
         try {
             // how big of a bite do we want to take? Allow users to easily override via their wp-config
-            if (! absint($step_size) > 0) {
+            if (absint($step_size) < 1) {
```
nerrad: much clearer conditional :)

core/data_migration_scripts/4_5_0_stages/EE_DMS_4_5_0_invoice_settings.dmsstage.php
```diff
@@ -75,7 +75,8 @@ protected function _migration_step($num_items = 1)
             // message templates are only initialized AFTER migrations and those two are new in 4.5. So if we wanted to
             // update them from a DMS, we'd need to have the DMS create the message templates which is quite a lot of code;
             // also we don't want to build a dependency on the messages code because it is likely to change soon
-            if (! in_array($invoice_settings['invoice_css'], array( '', 'simple.css' ))) {
+            if (isset($invoice_settings['invoice_css'])
+                && ! in_array($invoice_settings['invoice_css'], array( '', 'simple.css' ))) {
```
nerrad: We can use short array syntax now, so you could do something like

```suggestion
                && ! in_array($invoice_settings['invoice_css'], ['', 'simple.css']))) {
```

PR #1116, opened 2019-04-20
New CSS style build system

nerrad:
(This pull addresses #1116)

This pull introduces a new global CSS build system to Event Espresso core.  What do I mean by "global CSS"?  I mean various css stylesheets that form the basis for common css elements used throughout the EE application (and it's add-ons).  The goals of this build system are:

- A set of common styles and selectors for the EE app to use throughout the app.
- Expose selectors that help with accessibility (meeting WCAG AAA standards).
- Fairly simple to extend/modify.
- Ability to add theming eventually.
- Appropriate `media` css query blocks for various viewport breakpoints (for mobile etc).

In this initial iteration the following was implemented:

### CSS Styles

The following css style categories were implemented:

| Category | Description |
|--- | ---- |
| buttons | Various base button styles |
| colors | Various foreground, background, and transparency styles that can be used as modifiers on other elements.
| entity-status | A set of entity status color/background-color selectors.  Eg. `ee-status-color-TCM` for the transaction approved status |
| shadows | Various styles related to shadows |
| sizes | Various styles related to sizing |

Note, this is the _initial_ iteration so it's expected that there's still some work to expand on the styles available.  We can add to this as needed over time but with the goal of having a central place to work from rather than having global styles spread around the app.

### Structure

Currently the layout of these stylesheets follows this rough schema:

- All global styles are located in `assets/src/components/ui/styles` and this is _excluded_ from the component entry point.
- `styles/root` contains all base styles that implement [css variables](https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties) and defaults.  The entry point for each built stylesheet includes these.  These stylesheets are automatically built using the custom `css-builder` script.
- `styles/themes` contains directories for the themes to build (currently only one theme: `default`)
- `styles/themes/default` (and each theme subdirectory contents) contains the `config.js` file which is the file defining the configuration for that theme.  This configuration is used by the `css-builder` to build all the `*-variables.css` files in the theme folder.  You _can_ override "root" styles by adding custom `{any name}.css` stylesheets in the theme folder and to it add the selectors you want to override.
- Each theme folder also includes an auto-generated `index.js` file which is used by Webpack as the entrypoint for building the final stylesheet.
- `styles/themes/default/demo` Each theme subdirectory will have a generated `demo` folder that exposes an html file that can be used to visually review the available styles for that theme.

Built stylesheets follow this order:

- All variable declarations from the theme `*-variables.css` files.
- styles from the `*.css`  files in `styles/root`.
- Any remaining styles found in the theme folder that don't have `-variables` in the file name (which then can override the root styles).

Demo screenshot:

demo page screenshot

### CSS-builder script

To simplify the build process and to reduce human error, I've created a css-builder script (lives in `bin/css-builder`) that will read a configuration file from the theme folder (eg. `styles/themes/default/config.js` and build the following:

- all styles in the `styles/root` folder.
- all `*-variables.css` files in the `styles/themes/default` folder.
- the `demo` folder and files for the css demo view.  See it [here](http://htmlpreview.github.io/?https://github.com/eventespresso/event-espresso-core/blob/553e892dcef5ecda867761105f7d2f84f55aca56/assets/src/components/ui/styles/themes/default/demo/index.html) (note: This link in this pull comment is specific to a commit,  in the doc linking to this demo file, it will be based on the master branch).  The index.html can be opened in any browser and will give a demo of the css selectors.
- the `index.js` file for the Webpack entry point in the final css file build process.

**Currently** this script _only_ builds the default theme and root stylesheets.  However, with a little more iteration it could be updated to generate a scaffold for new themes and also build the final files for new themes.  So in later iterations this will greatly aid with building various themed global styles.

You can execute this script manually by running this in the same directory as the `event-espresso` package.json file: 

```bash
npm run build:css
```

[asciicast](https://asciinema.org/a/0FEDv4xlKD9K5UJPOrHBFFzor)

However, this is also automatically setup to run when you do `npm run watch` and will automatically rebuild whenever there are any changes to either the build-script itself (new templates/logic etc) or files in the `assets/src/components/ui/styles` folder.

### Documentation

I've started documentation for the global css process in `docs/AAA-CSS-in-EE/README.md`.  This too can be iterated on over time.

### Other:

- stylesheets are built to `ee-core-default-theme.{hash}.dist.css` and exposed on the `CoreAssetsManager::CSS_HANDLE_CORE_CSS_DEFAULT`.  In the future themed global sheets could follow the same naming schema.
- currently the ee components stylesheet lists this global stylesheet as a default.
- There is some overlap with legacy css selector names in the new selectors but I haven't brought over 1:1 sameness between legacy styles and new as this is a fresh opportunity to rethink how we do styles in EE.  Eventually we can start dropping some of the legacy styles and move our existing codebase over to use the new styles as the new global stylesheets are iterated on.  However I did none of that in this pull.
- I expect we may tweak some of the initial styles I've got in this pull as a part of the review process.  But please be aware that the colors in particular have been carefully selected to be WCAG AAA compliant for the `with-foreground-color` and `with-background-color` modifiers so any changes should also be verified compliant.
- Once this lands in master, we should start using it with any NEW styling we write.  This means whenever global styles need tweaked, they should be done in a separate pull iterating on this stylesheet and merged into master so it can be used in other work.
- The styles have not been tested against the WordPress core GB component styles (example the `button-component`) and I think we may want to account for that in a special `wp-components-overrides.css` stylesheet in `styles/root` as we encounter them.  We may also need to do the same for general WP styles as well.
- I wonder if we should use completely new class names that aren't in legacy stylesheets (i.e. `ee-btn` instead of `ee-button`) so that we can do a clean break from the legacy styles?

## How has this been tested

Currently the new stylesheet just loads anywhere we enqueue `CSS_HANDLE_COMPONENTS`.  So I checked the following to make sure things are not broken:

* [x] Event Attendees Block
* [x] new Barcode Scanner App (which implements the WP button component).

With the latter (barcode scanner app), there's still some issues with the active state of button components and the new styles (plus the legacy stylesheet loads that has some of the button classes).  I'd like to fix this in later iterations after this pull.

## Checklist

* [x] My code is tested.


nerrad:
I just whitelisted a package used by npm-watch which was added in this pull (this is what called the failed build in travis).  The whitelisted package is not distributed so the license incompatibility here does not apply for our purposes.

tn3rb:
There's some really cool stuff in this branch but I think we should rework things so that it will be easier to implement themes in a really semantic manner. I've been thinking about this stuff for a long, long, **long** time now while working on REM and the admin refactor, and I started to do some of the same things that you have done in this branch (in terms of defining all kinds of color variations). When it came time to use them though, I realized that they were not as great as I thought they would be, and would also make it more difficult to implement theming down the road.

That said, we still want to be able to dynamically define a bunch of colors and have stylesheets generated for us, just like you are doing in this branch (the style demos are the bomb too). But to make the transition to theming easier, instead of defining color specific css classes (like `.ee-green-background-color.with-linear-gradient`) and then using those directly in the stylesheets, we should instead have themes define a set of color variables and then use those variables in our stylesheets. 

So instead of:

```css
.ee-light-blue-background-color {
    background-color: var(--ee-color-light-blue, #00b1ca);
}
.ee-light-blue-background-color.with-accessible-foreground {
    color: var( --ee-color-light-blue-foreground-contrast, #041525 );
}
.ee-light-blue-background-color.with-accessible-foreground.with-hover:hover,
.ee-light-blue-background-color.with-linear-gradient.with-accessible-foreground.with-hover:hover,
.ee-light-blue-background-color.with-linear-gradient.with-hover:hover,
.ee-light-blue-background-color.with-hover:hover {
    background-image: none;
    background-color: var( --ee-color-light-blue-hover-contrast, #00e1ff );
}
.ee-light-blue-color {
    color: var(--ee-color-light-blue, #00b1ca);
}
```

a theme would define the following color variables
- `--primary`
- `--secondary`
- `--tertiary` (or accent)

and maybe
- `--greyscale-#` (where #'s go from 1 to 4 or 5 or whatever, ie: `--greyscale-1`, `--greyscale-2`, `--greyscale-3`, etc)

as well as a bunch of other stuff that we'll discover we need as we go,
so a theme's color config would simply define these variables:

```js
const myTheme = {
    colors: {
        primary: #2D3E50;
        secondary: #1BBC9B;
        tertiary (or accent): #E64C66;
        border: #AEAEAE;
        background: #FFFFFF;
    },
```

and then our stylesheets would use these variables to style their elements:

```css
.ee-info-box {
    color: var(--ee-primary);
}
.ee-info-box h4 {
    color: var(--ee-secondary);
}
```

so an element could be green in one theme, and red in another, and blue in another, or whatever color because none of the colors are set directly. This will be far more versatile in the long run and will likely make it easier for us to share components much easier because styles will automatically work better with other components, because again, no colors are explicitly set anywhere other than the theme. 

Of course there will still be some things that we will want to consistently style with the same relative color from theme to theme. For example, the status colors for transactions where we will always want the TCM (completed) status to be some shade of green and the TFL (failed) status some shade of red. And that will work exactly like what you have done in this branch:

```css
.ee-status-color-TCM {
    color: var(--ee-color-green, #70cc50);
}
.ee-status-color-TFL {
    color: var(--ee-color-pink, #ff55c0);
}
```

Because themes can be constructed using any set of colors but can generally be described as being dark or light (most themes at least), it doesn't make sense to use terms like "darker" or "lighter" or specific alpha transparency references like `25-transparency` (although transparency can work for some things like shadows, but it's not good to use for setting background colors for hover effects because it will allow background elements to show through which is very likely undesirable) so variations in color are better referred to in terms of "contrast" (or maybe some other term), meaning how much greater in difference they are from the theme's main background color. 

- `-high-contrast` darker foreground against light background (or lighter foreground against dark background for dark themes)
- `-low-contrast` lighter foreground against light background (or darker foreground against dark background for dark themes)

> plz note that "low contrast" can still be WCAG compliant, ie: "low contrast" could mean a 4.75:1 contrast ratio whereas "high contrast" could mean a 10:1 or greater ratio.

color modifiers can then be applied to the above to create varying shades:
- `--primary-high-contrast` (farther from bg color)
- `--primary`
- `--primary-low-contrast` (closer to bg color)

This is similar to what you have done in this branch with things like `.ee-pink-color.with-accessible-background` but is more versatile because it can be used with virtually any color because `--primary-high-contrast` is completely abstracted from any specific color. This also simplifies a lot of things and reduces the amount of code we have to add because something like `.with-linear-gradient` can simply be something like:

```css
.with-primary-gradient {
    background-image: linear-gradient( var(--primary-low-contrast), var(--primary-high-contrast) );
}
```

which now works for any color that is used for `--primary`

These variables can also be used to set the colors used for things like hover states.

```css
// regular color to high contrast on hover ie: starts out normal, gets **extra** saturated on hover
.with-primary-hover { background: var(--primary); }
.with-primary-hover:hover { background: var(--primary-high-contrast); }

// low contrast to regular color on hover ie: starts out less saturated, becomes normal on hover
.with-low-primary-hover { background: var(--primary-low-contrast); }
.with-low-primary-hover:hover { background: var(--primary); }
```

this code **NEVER** has to change but can represent any number of colors and support **ANY** theme that defines those variables properly.
So the above works equally well for a super dark theme using blacks and dark greys, or a super light theme using pastels and off-whites, or a theme where everything is a shade of blue, or... anything!

That said, I personally think that creating a slew of css classes (like hover classes) just doesn't work well and really makes a mess of the markup because everything has 500 different css classes applied to it to define how it is supposed to look. This also goes against the concept of having semantic markup. So I personally find it easier to just use a specific css class that describes what a thing is and then apply styles to that using variables instead of hard-coded values:

```css
.specific-button {
    background: var(--primary);
    border: var(--primary-border);
    border-radius: var(--button-radius);
    box-shadow: var(--button-shadow);
    color: var(--primary-text);
}
.specific-button:hover { background: var(--primary-high-contrast); }
```

again, the above can look totally different depending on the theme, but this also keeps the markup much cleaner:

```html
<button class="specific-button">click me</button>
```

instead of 

```html
<button class="specific-button ee-green-background-color with-accessible-foreground with-hover with-border-radius-small with-small-dark-grey-box-shadow">click me</button>
```

which is cluttered, unsemantic, and not very versatile because a lot of the appearance has been hard coded into the markup.
Again, I have made the mistake of going down this path many times and it just does not work well. It just creates a big mess. One of the primary reasons that css was created was to be able to extract all of the styling from your markup and put it into a different file. Using a gazillion css classes to define all of the individual stylings for each component goes against that and is not much different than having inline styles.


Some other variables that we could have themes define include:

- `--add` applied to buttons, containers, etc to indicate addition ie: "click to add" or "thing is queued for adding"
- `--remove` applied to buttons, containers, etc to indicate subtraction ie: "click to trash" or "thing is queued for deletion"
- `--has` applied to buttons, containers, etc to indicate existence of something ie: "thing A has got it"
- `--none` applied to buttons, containers, etc to indicate lack of something ie: "thing A does not have it"

these could be used to create specific css classes like `.ee-add-bg-color` and `.ee-remove-bg-color` but again, I think it is far better to just use the variables directly in our stylesheets:

```css
.ee-add-ticket { background: var(--add, var(--ee-color-green) ); }
.ee-remove-ticket { background: var(--remove, var(--ee-color-red) ); }
```


With regards to font sizes we don't need 24 different levels and we certainly don't need anything below 8px. Even 9px is almost intolerable as a font size and that's throwing accessibility out the window. I can't even fathom why you would create a class for setting the font size to 1px. How would that even work? A 1px high character would just be... a dot?

Anyways, I think we can get away with the following for font size modifiers, although these (or something similar) _could_ maybe also be applied to containers or other things as well:

```css
--tiny: calc( var(--ee-font-size) * .5 )
--smaller: calc( var(--ee-font-size) * .667 )
--small: calc( var(--ee-font-size) * .833 )
--medium: calc( var(--ee-font-size) * 1 )
--big: calc( var(--ee-font-size) * 1.167 )
--bigger: calc( var(--ee-font-size) * 1.334 )
--huge: calc( var(--ee-font-size) * 1.5 )
```

This is similar to what you did with `with-font-size-small` (but would create a greater variation between the biggest and smallest), and works better than setting strict hardcoded font sizes like 11px or having classes like `.with-font-size-11` applied all over the place.
If needed we could also have a generic `--scale` modifier that could be applied to all kinds of things like padding, and margins, and widths, to help scale EVERYTHING up (for accessibility purposes).

These can be associated directly with css classes (ie: `.tiny-text`) but it is better to use selectors that are more semantic. Semantic markup describes what the content **is** or the purpose of the content as opposed to describing what the content should look like (something I am repeatedly guilty of doing cuz it's far too easy to fall into). For example, applying a css class like `.huge-text` to something as a way to make it standout from other content may not work in a theme that wishes to use color, or big margins, or some other variation to give things more importance. A more semantic approach would be to use class names that describe the importance of the thing they are applied to:

ie:
```css
.ee-high-priority { font-size: var(--huge); }
```
vs
```css
.ee-high-priority { color: var(--foreground-high-contrast); }
```

This also works well in conjunction with accessibility because it can help to indicate things that should also implement other accessibility features:

```
<div id="warning" class="ee-high-priority" aria-live="assertive"></div>
```

Some other modifiers that we will likely want themes to define may include:

- border-radius
- padding
- margin

which could be combined with the size modifiers from above to end up with things like:

```css
--tiny-padding: calc( var(--ee-padding-size) * .5 )
--smaller-padding: calc( var(--ee-padding-size) * .667 )
--small-padding: calc( var(--ee-padding-size) * .833 )
--medium-padding: calc( var(--ee-padding-size) * 1 )
--big-padding: calc( var(--ee-padding-size) * 1.167 )
--bigger-padding: calc( var(--ee-padding-size) * 1.334 )
--huge-padding: calc( var(--ee-padding-size) * 1.5 )
```

although we may want to simplify and not have so many levels of variation (especially for border-radius).

Anyways, I've rambled on enough, here's the bottom line:

 - gazillions of css classes for micromanaging appearance is not a good path to go down. I've tried it. It's bad.
 - let's simplify things and start with less variations and get rid of things we'll likely never use (like pink shadows)
 - let's focus on generating really useful variables and then just use those variables in our stylesheets instead of css classes
 - let's maybe try to get theming working sooner rather than later because that can help with accessibility

nerrad:
> gazillions of css classes for micromanaging appearance is not a good path to go down. I've tried it. It's bad.

Just to be clear, the stylesheets in `root/styles` are all auto-generated so managing them per theme is not an issue.  The _main_ reason I started off with what is there is because it exposes some flexible **accessible** theming to elements without the implementor having to think about whether it will be WCAG AAA compliant.

For instance you gave your example as where things look verbose (I modified it for how it's implemented in the stylesheets in this branch:

```html
<button class="ee-button ee-green-background-color with-accessible-foreground with-hover with-border-radius-small with-small-dark-grey-box-shadow">click me</button>
```

Now if the implementor wanted to change the color to the pink shade and the button remains accessible then the implementor just has to change `ee-green-background-color` to `ee-pink-background-color`.  

The main point of these modular and modifier selectors is to expose some accessible styles for specific things apps might build that aren't defined in any of the global styles and thus provide a way for those apps to build things that will automatically adjust with changes in themes down the road.

Another thing I had in mind is that many of the things we might define in global stylesheets using the old legacy method of ui/ux will be better in specific react components.  For example `.ee-info-box` would be better living in an `InfoBox` component that has all the styling specific to that component _within_ that components exports.

With that said, I don't disagree with your desire to define some aggregate selectors that can be used and I definitely think we should add them. But, for the intended use case - exposing selectors that allow for implementing accessible companion modifiers that can be reliably implemented - I think what I have works well.

_With that said, I'm totally okay if you disagree and want to remove them_.  I hope what we end up with has accessibility in mind.

------

> let's simplify things and start with less variations and get rid of things we'll likely never use (like pink shadows)

All for this, and I particularly like some of the examples you give for using `calc` to add the variations.

------

> let's focus on generating really useful variables and then just use those variables in our stylesheets instead of css classes

I'm not sure what you mean by `just use those variables in our stylesheets instead of css classes` because its written like that's not being done in this branch.  I'm assuming you have read the [initial description for this pull](https://github.com/eventespresso/event-espresso-core/pull/1116#issue-272121269) in which I stated this:

> `styles/root` contains all base styles that implement css variables and defaults. The entry point for each built stylesheet includes these. These stylesheets are automatically built using the custom css-builder script.

In other words, the only thing the theme defines is `config.js` in the theme directory and everything else is autogenerated.  What may not be clear is that the `styles/root` is the base styles for ALL themes.  Because variables are used, themes should not need to define any custom stylesheets (but facility is there to do so in case for a particular theme we want to override a specific selector).

Also when you use the phrase "our stylesheets"  are you referring to the global stylesheets that are being created or are you referring to stylesheets that may be defined with specific components?

-------

> let's maybe try to get theming working sooner rather than later because that can help with accessibility

I'm wary about getting theming working sooner because we haven't even worked out the styles for the default theme yet.  I think we should be focusing on working out those styles and then start using them to see where there may be flaws in the approach and what improvements can be made.  If we implement theming then we've got to _maintain_ all those different themes and any changes we make to the process will have to be adjusted in all the various theme configs as well.  We've got too much other stuff on our plate to add to it right now. Let's get one solid theme working well first ('default'). 

The code that has been implemented in this branch is not a blocker to doing any theming when we eventually _do_ do it.  But I'm wary of going all the way with it until we have more solid styles in place for the default theme.

Also, I'm not sure how getting theming working sooner can help with accessibility? Can you expand on what you mean by that? Shouldn't _all_ themes be accessible? 

-------

Some next steps:

I understand that styles are something we want to try and get "right" because it provides much of the base for the ui/ux throughout the app, but I'm wary of falling for the trap of getting things perfect in our first go.  Hence, I'm wary of this branch getting all the styles added that we think we'll need (or removing what we think we wont' need) before it gets merged.  As I was doing this branch I was thinking of places I could use what's implemented now (and tried it in a branch based off of the barcode scanner app as noted in the description) to get a feel for how workable it was.  With that said, here's some suggested next steps to keep things moving:

* I'd like it if you can go through and do the second pass of making any changes to variables/styles that are implemented in this branch to get it the way you think it should be.  Nothing beats having something concrete (vs pseudo examples) code to help illustrate what you intend to be used.
* For the above it'd probably be good to implement what you need for REM that you can implement.
* I'd prefer if we hold off on getting "theming" working for the reasons I noted in addressing your point about it and focus on getting this branch merged with some basic styles being used right now.  We can iterate on this work as we encounter more "global" styles we need to expose in things we build and then once we're confident the default global theme styles work well, we can implement theming (if/when that is something prioritized).

tn3rb:
> Just to be clear, the stylesheets in root/styles are all auto-generated so I don't think managing the definitions per theme is not as much of an issue.

yes the auto-generation is great and I love that aspect of things but "managing the definitions per theme" is not the issue, the issue is ending up with what I have called "gazillions of css classes for micromanaging appearance" which results in markup like:

```html
<button class="ee-button ee-green-background-color with-accessible-foreground with-hover with-border-radius-small with-small-dark-grey-box-shadow">click me</button>
```

that's just css classes on that element. Add in an id and a bunch of aria and other attributes and that's going to be insanely cluttered. It's also unsemantic and inflexible because it is hardcoding the element's appearance into the markup itself.

> The main reason I started off with what is there is because it exposes some flexible accessible theming to elements without the implementor having to think about whether it will be WCAG AAA compliant.

This is great and we want to keep that intent but implement it in a slightly different manner through the use of css variables as opposed to css classes. 

With regards to that button markup above you said:

> Now if the implementor wanted to change the color to the pink shade and the button remains accessible then the implementor just has to change ee-green-background-color to ee-pink-background-color.

Sure, an implementer can define the appearance for an element any way they want **when** they are in the process of developing their component, but a user can NOT change that later on by switching themes. Look at the "Admin Colour Scheme" options in the WP admin that we have both agreed would be good to integrate with what we are doing. Those admin styles change the button colors, but if an element has a hardcoded `ee-green-background-color` class applied to it, then at best, the user can only get a different shade of green when they switch themes. But what if they are using the "Blue" admin theme where everything is blue? They're still stuck with a green button because that style decision is hardcoded right into the markup for that element.

The direction I want us to go is not to utilize css classes that define the appearance of things, but instead utilize classes and/or markup that is more semantic and then allow themes to control the style. So instead of hardcoding a `ee-green-background-color` class on an element, or any other classes that define appearance, we only use classes that define purpose and/or functionality. So we end up with markup like:

```html
<button class="ee-submit-edit-ticket-form ee-primary-button">click me</button>
```

and then use css variables to apply our WCAG compliant styles to it instead of css classes:

```css
 /* in some global stylesheet */
.ee-primary-button {
    background: var(--primary);
    border: var(--primary-border);
    border-radius: var(--button-radius);
    box-shadow: var(--button-shadow);
    color: var(--primary-text);
    font-size: var(--medium);
    padding: var(--small-padding);
}
.ee-primary-button:hover {
    background: var(--primary-high-contrast);
}

/* in stylesheet for edit ticket form to override defaults */
.ee-submit-edit-ticket-form {
    font-size: var(--huge);
    padding: var(--big-padding);
}
```

so now that button is completely free to change its appearance simply by changing the applied theme because none of its appearance has been hardcoded directly in its markup. Both of the css classes applied to it (`ee-submit-edit-ticket-form` and `ee-primary-button`) can and **do** define the button's appearance, but they do so in a semantic and totally flexible way. Please note that there is nothing stopping the above from being WCAG compliant. In fact, I believe that it will be **easier** for us to implement WCAG compliance across the board if that is baked right into the themes as opposed to relying on having to add specific css classes like `.with-accessible-foreground` to an element. What happens to elements that do not have that class applied to them? I'm assuming it means they are not accessible (or not _as_ accessible?).

If instead we use a variable like `--primary` (or `--primary-color`, or `--primary-bgColor`, or whatever descriptive name) to define an element's background color and then another variable like `--primary-text` (or `--primary-text-color`, or `--text-color-on-primary`, or whatever descriptive name) to define it's font color, then that can be made WCAG compliant as part of the theme and will apply to **ALL** elements that use those color variables without having to do anything else. And as long as our themes are all WCAG compliant, then changing themes will maintain that compliance. I realize that this last point is also the case with the existing code in this branch if you apply the appropriate css classes, but I just don't want to rely on specific css classes in order to be compliant. It should just apply with no extra effort.

> The main point of these modular and modifier selectors is to expose some accessible styles for specific things apps might build that aren't defined in any of the global styles and thus provide a way for those apps to build things that will automatically adjust with changes in themes down the road.

right. I want to achieve the same thing but via the use of css variables as opposed to css classes. An implementer can utilize the variable variants like `--small-text`, `--big-text`, `--huge-text`, etc, to control the appearance of their components and know that their selections will (hopefully) blend harmoniously with everything else on the page because those text variables will all scale accordingly as different themes are applied. For text though, I think we should create some semantically named classes that combine multiple properties together to help maintain WCAG compliance, since font size, text color, and background color, all interact together to affect accessibility. I'm thinking something like:

 - `.priority-1` huge font size + high contrast font color
 - `.priority-2` large font size + super high contrast font color
 - `.priority-3` large font size + high contrast font color
 - `.priority-4` large font size + default font color
 - `.priority-5` large font size + low contrast font color (low contrast is ok because of large font size)
 - `.priority-6` default font size + default font color
 - `.priority-7` small font size + high contrast font color
 - `.priority-8` smaller font size + super high contrast font color

but that naming schema is open to better suggestions. They don't have to be named so linearly and could be more descriptive and have more variation like `.super-important-text` or `.fine-print`, which are still both descriptive in terms of purpose and/or functionality without actually explicitly controlling the appearance.

> Another thing I had in mind is that many of the things we might define in global stylesheets using the old legacy method of ui/ux will be better in specific react components. For example .ee-info-box would be better living in an InfoBox component that has all the styling specific to that component within that components exports.

totally 100% agree with this. That was just some code I grabbed to use as an example. That said, I would still want the css for that component to utilize the css variables from our themes for styling its elements, as opposed to defining its appearance via css classes.

> exposing selectors that allow for implementing accessible companion modifiers that can be reliably implemented - I think what I have works well.
> 
> With that said, I'm totally okay if you disagree and want to remove them. I hope what we end up with has accessibility in mind.

I don't disagree that what is in this branch can work well in some ways, but again, I don't want to rely on specific css classes like `.with-accessible-foreground` to be compliant. We should try to be compliant without having to add extra markup to do so. It should just happen without any extra effort or thought. To aid with this and to make design decisions easier, similar to my comments above regarding text, we may want to have some semantically named css classes that group together several properties that can be applied to container elements, like:

```css
.ee-primary,
.ee-secondary,
.ee-accent {
    border-radius: var(--default-borderRadius);
    font-size: var(--default-fontSize);
    padding: var(--default-padding);
}

.ee-primary {
    background: var(--primary-bgColor);
    border: var(--primary-borderColor);
    color: var(--primary-textColor);
}

.ee-secondary {
    background: var(--secondary-bgColor);
    border: var(--secondary-borderColor);
    color: var(--secondary-textColor);
}

.ee-accent {
    background: var(--accent-bgColor);
    border: var(--accent-borderColor);
    color: var(--accent-textColor);
}
```

so if someone wants a section of their component to be the primary element then they can just add the `.ee-primary` class to it, which helps to define its purpose in a semantic way, and also applies a preconfigured set of WCAG compliant styles regardless of what theme is applied.

> I'm not sure what you mean by just use those variables in our stylesheets instead of css classes because its written like that's not being done in this branch. 

didn't mean to imply that at all. What is happening in this branch that I don't want to do is have tons of css classes that explicitly define the appearance for an element. Here's all the classes that are being generated just for the color green (two shades of green):

```css
.ee-green-background-color {
    background-color: var(--ee-color-green, #70cc50);
}
.ee-green-background-color.with-accessible-foreground {
    color: var( --ee-color-green-foreground-contrast, #002a00 );
}
.ee-green-background-color.with-accessible-foreground.with-hover:hover,
.ee-green-background-color.with-linear-gradient.with-accessible-foreground.with-hover:hover,
.ee-green-background-color.with-linear-gradient.with-hover:hover,
.ee-green-background-color.with-hover:hover {
    background-image: none;
    background-color: var( --ee-color-green-hover-contrast, #bee7b1 );
}
.ee-green-color {
    color: var(--ee-color-green, #70cc50);
}
.ee-green-color.with-accessible-background {
    background-color: var(--ee-color-green-background-contrast, #002a00 );
}
.ee-green-color.with-accessible-background.with-hover:hover {
    background-color: var(--ee-color-green-hover, #2a2440);
}
.ee-green-background-color.with-linear-gradient {
    background-image: linear-gradient(
        var(--ee-green-background-gradient-x, #80dd5e),
        var(--ee-green-background-gradient-y, #70cc50)
    );
}
.ee-green-background-color.with-linear-gradient.with-accessible-foreground {
    color: var(--ee-color-green-background-contrast, #002a00 );
}
.ee-green-background-color.with-01-transparency {
    background-color: var(--ee-color-green-01-transparency, rgb( 112, 204, 80, 0.01));
}
.ee-green-background-color.with-01-transparency.with-hover:hover {
    background-color: var(--ee-color-green-25-transparency, rgb( 112, 204, 80, 0.25));
}
.ee-green-background-color.with-25-transparency {
    background-color: var(--ee-color-green-25-transparency, rgb( 112, 204, 80, 0.25));
}
.ee-green-background-color.with-25-transparency.with-hover:hover {
    background-color: var(--ee-color-green-01-transparency, rgb( 112, 204, 80, 0.01));
}
.ee-green-background-color.with-50-transparency {
    background-color: var(--ee-color-green-50-transparency, rgb( 112, 204, 80, 0.50));
}
.ee-green-background-color.with-50-transparency.with-hover:hover {
    background-color: var(--ee-color-green-25-transparency, rgb( 112, 204, 80, 0.25));
}
.ee-green-background-color.with-75-transparency {
    background-color: var(--ee-color-green-75-transparency, rgb( 112, 204, 80, 0.75));
}
.ee-green-background-color.with-75-transparency.with-hover:hover {
    background-color: var(--ee-color-green-50-transparency, rgb( 112, 204, 80, 0.50));
}
    /** Dark Green **/
.ee-dark-green-background-color {
    background-color: var(--ee-color-dark-green, #006600);
}
.ee-dark-green-background-color.with-accessible-foreground {
    color: var( --ee-color-dark-green-foreground-contrast, #fcfcfc );
}
.ee-dark-green-background-color.with-accessible-foreground.with-hover:hover,
.ee-dark-green-background-color.with-linear-gradient.with-accessible-foreground.with-hover:hover,
.ee-dark-green-background-color.with-linear-gradient.with-hover:hover,
.ee-dark-green-background-color.with-hover:hover {
    background-image: none;
    background-color: var( --ee-color-dark-green-hover-contrast, #004200 );
}
.ee-dark-green-color {
    color: var(--ee-color-dark-green, #006600);
}
.ee-dark-green-color.with-accessible-background {
    background-color: var(--ee-color-dark-green-background-contrast, #fcfcfc );
}
.ee-dark-green-color.with-accessible-background.with-hover:hover {
    background-color: var(--ee-color-dark-green-hover, #e3e3e3);
}
.ee-dark-green-background-color.with-linear-gradient {
    background-image: linear-gradient(
        var(--ee-dark-green-background-gradient-x, #80dd5e),
        var(--ee-dark-green-background-gradient-y, #006600)
    );
}
.ee-dark-green-background-color.with-linear-gradient.with-accessible-foreground {
    color: var(--ee-color-dark-green-background-contrast, #fcfcfc );
}
.ee-dark-green-background-color.with-01-transparency {
    background-color: var(--ee-color-dark-green-01-transparency, rgb( 0, 102, 0, 0.01));
}
.ee-dark-green-background-color.with-01-transparency.with-hover:hover {
    background-color: var(--ee-color-dark-green-25-transparency, rgb( 0, 102, 0, 0.25));
}
.ee-dark-green-background-color.with-25-transparency {
    background-color: var(--ee-color-dark-green-25-transparency, rgb( 0, 102, 0, 0.25));
}
.ee-dark-green-background-color.with-25-transparency.with-hover:hover {
    background-color: var(--ee-color-dark-green-01-transparency, rgb( 0, 102, 0, 0.01));
}
.ee-dark-green-background-color.with-50-transparency {
    background-color: var(--ee-color-dark-green-50-transparency, rgb( 0, 102, 0, 0.50));
}
.ee-dark-green-background-color.with-50-transparency.with-hover:hover {
    background-color: var(--ee-color-dark-green-25-transparency, rgb( 0, 102, 0, 0.25));
}
.ee-dark-green-background-color.with-75-transparency {
    background-color: var(--ee-color-dark-green-75-transparency, rgb( 0, 102, 0, 0.75));
}
.ee-dark-green-background-color.with-75-transparency.with-hover:hover {
    background-color: var(--ee-color-dark-green-50-transparency, rgb( 0, 102, 0, 0.50));
}
```

Again, I have gone this route many many times before. It's easy to do because it logically makes sense because those classes seem to define the kinds of things that one would want to apply to components they are developing. But here's some of the problems with the above:

 - it's overly complex. That's a lot of classes to have to know about. Mind you, my suggestion will result in a lot of css variables being created, but ultimately we can group those into semantically relevant combinations where possible. So a container with the `.ee-primary` class will automatically possess a cohesive set of WCAG compliant styles. As well, the approach I'm suggesting is easier because an implementer only has to choose between 4 main color groups: 
    - primary - the color used most frequently to highlight elements
    - secondary - the second most used color for highlighting elements - typically complimentary to the above
    - accent - a third and maybe highly contrasting color for drawing attention to special cases
    - default - the colors used for most elements (like dark text on a white background). The above primary, secondary, and accent colors should actually be used more sparingly
so instead of having to know and understand dozens of classes for each color possibility, they only have to work with the above.
 - a lot of those classes explicitly define an element's appearance, which as previously stated, is not semantic nor flexible, and clutters up the markup. Giving a button a class of `.ee-green-background-color` is explicitly controlling its appearance and can not change accordingly when the theme changes. What happens if someone wants to use the "Blue" WP Admin Colour Scheme where buttons are supposed to be blue? There are of course cases where we will want to have color coded buttons, but again, this should be done more semantically through the use of classes like `.create-new-thing` (green button?) or `.remove-thing` (red button?).
 - it appears that accessibility is reliant on the use of specific classes like `.with-accessible-foreground` and `.with-accessible-background` but what happens if an element does not utilize those classes? Again, I think we can achieve WCAG compliance as the default without the use of special classes.
 - a more specific problem with the above that does not pertain to the general approach is that using transparency to change the background color of an element is not a good approach at all because it allows other elements behind it to show through. This can potentially lead to muddied and difficult to read text. What happens when an element using the `.with-25-transparency` is overtop of another element and they both have text? The bottom text will appear to be 25% lighter in color due to the transparent background above it (assuming a white background). That can majorly conflict with the text on top and make things really difficult to read which would surely become an issue with the accessibility for that element.

> Also when you use the phrase "our stylesheets" are you referring to the global stylesheets that are being created or are you referring to stylesheets that may be defined with specific components?

either or, basically any stylesheet that we create whether it is a global stylesheet or one for a specific component. Both should use the css variables we define like `--primary-textColor`.

> I'm wary about getting theming working sooner because we haven't even worked out the styles for the default theme yet. I think we should be focusing on working out those styles and then start using them to see where there may be flaws in the approach and what improvements can be made. If we implement theming then we've got to maintain all those different themes and any changes we make to the process will have to be adjusted in all the various theme configs as well. We've got too much other stuff on our plate to add to it right now. Let's get one solid theme working well first ('default').

Right. I totally agree with you that that is the best approach. To clarify things though, what I meant by getting theming working does not mean that we would necessarily have to develop more than a default theme. I was thinking more about the underlying mechanisms for being able to add multiple themes. In other words, we should at least investigate how that process might work so that we don't trap ourselves in some system that can not be easily be expanded on later. So again, not necessarily create more than one theme, but just be certain that that will work.

> Also, I'm not sure how getting theming working sooner can help with accessibility? Can you expand on what you mean by that? Shouldn't all themes be accessible?

Yes but there are still levels of accessibility right?
For some people, their accessibility needs are really extreme and would require style changes that would not be desirable for anyone else. For example, I had a neighbour once that was legally blind but still had some vision. When they used their computer they would have their face about six inches from this enormous screen and had to magnify things so that the smallest font size was like... 30px maybe? Like the only thing they would have been able to read on this page would be the "New CSS style build system" title at the top of the page. So someone like that needs to have everything massively magnified and have super high contrasting colors. That however would interfere with most other people's ability to view a site. Having the ability to switch to an extremely accessible theme that implements larger font sizes, extreme color contrasts, thicker borders, etc, would be more helpful to people like that, but would be a hindrance to most others. So theming can make things more accessible by providing the ability to cover extreme edge cases.

> Some next steps:
> - I'd like it if you can go through and do the second pass of making any changes to variables/styles that are implemented in this branch to get it the way you think it should be. Nothing beats having something concrete (vs pseudo examples) code to help illustrate what you intend to be used.
> - For the above it'd probably be good to implement what you need for REM that you can implement.
> - I'd prefer if we hold off on getting "theming" working for the reasons I noted in addressing your point about it and focus on getting this branch merged with some basic styles being used right now. We can iterate on this work as we encounter more "global" styles we need to expose in things we build and then once we're confident the default global theme styles work well, we can implement theming (if/when that is something prioritized).

Ya I'll work on solidifying my ideas for variables and such and how those should interact, and then cut a branch from this one to implement them.

nerrad:
Assigned to @br3nt due to this comment he made:

> Ya I'll work on solidifying my ideas for variables and such and how those should interact, and then cut a branch from this one to implement them.

tn3rb:
OK I totally reworked the generated stylesheets as well as the demo.

My primary goals were to:

- generate a large number of consistently named CSS variables for use in stylesheets
- avoid the creation of CSS classes for applying individual styles which leads to presentation details being hardcoded directly in the DOM (ie; classes like `.with-green-background`) as well as a DOM that is cluttered by hundreds of additional CSS classes
- create a small collection of semantically named CSS classes that group several presentation details together to demonstrate how I envision utilizing all of this in components (we can add to this as needed)
- reduce the complexity of design choices to make for both faster development and a more aesthetically pleasing final result
- and finally, and possibly most importantly, create a system that will be accessible out of the box without having to rely on additional CSS classes

So the idea is to replace HTML like this:

```html
<button class="ee-button ee-green-background-color with-accessible-foreground with-hover with-border-radius-small with-small-dark-grey-box-shadow">click me</button>
```

with HTML like this:

```html
<button class="ee-button ee-button-primary small">click me</button>
```

and have the theme take care of defining backgrounds, borders, radii, shadows, and most importantly, WCAG compliant color combinations, so that simply changing the theme can dramatically change the appearance of the system without compromising things like accessibility.

The main thing I did to achieve most of the above was to create a set of color utility functions found in `/bin/css-builder/color-utils.js` that can take a set of predefined hex colors for a theme and dynamically generate WCAG compliant color alternatives. So instead of having to manually find an acceptable text color to use for a particular background color, the system does this automatically. We then generate a pair of CSS variables that utilize the following naming schema:

- `--ee-color-{color name}` - used for setting background color
- `--ee-text-on-{color name}` - a WCAG compliant contrast for the above color

So classes that need to set a specific background color, can do so using CSS like:

```css
.ee-status-background-color-RAP {
    background-color: var(--ee-color-green);
    color: var(--ee-text-on-green);
}
```

and know that regardless of what color hex code is used by the theme to define "green", any text that is used on top of that will be WCAG compliant because it has been dynamically computed to meet that specification. We are also generating a pair of high and low contrast alternatives for each color to allow for a little bit of extra design versatility.

Although I didn't do anything to make theming usable in terms of being able to switch between themes or manage them in any way, I wanted to make sure that all of the dynamic color generation was working properly, so the default theme config has some "pseudo" theme colors defined but commented out. There not full on themes though as they only redefine the four main colors that themes use so the change in appearance won't be as dramatic as what we will be able to do with a full theme change. If you run the demo, you can switch between the pseudo themes by commenting out the default and uncomment one of the others, and feel free to try your own color combinations to see if you can find flaws with the color generation.

Anyways, my branch has been pushed to: FET/introduce-css-component-building-II and it's likely best to view the demo to get a cohesive explanation of things, since that is the purpose of the demo.

Doesn't really matter to me how we proceed from here. We can merge my branch into yours and then work on it from there, or I can create a new pull for my branch, or discuss a totally different approach, or whatever.

tn3rb:
Here's a screenshot of the new demo

CSS Demo

nerrad:
I force pushed your branch to this pull @tn3rb and mostly just did some minor cleanup.  Really like all your changes and I think this gives us a solid base to go forward with.

**Really** like the effort you put into the demo templates.  I think it'll be SUPER useful going forward.

PR #1133, opened 2019-04-24
Use i18n datetime methods to build the date range for the ticket.

Pebblo:
Working on https://github.com/eventespresso/eea-wpuser-integration/issues/16 I found the output that isn't i18n ready was `$ticket->date_range()`, meaning the fix is in core.

So I just switched out methods used by the above for the 'new' i18n datetime methods.

## How has this been tested
Set the site language to something other than English, load `[ESPRESSO_MY_EVENTS]`

With master branch when you click the cog icon for an event and view the 'Your tickets', the date range for the ticket will be in English.

With this branch, it will in whatever language you set.

## Checklist

* [x] My code is tested.


Pebblo:
Sorry Darren, I realised there was a change in output with this code as now includes times.

We often get users stating that events over multiple days show up as one long continuous event and apparently people find it confusing:

7th March 2018 8:00 am -  9th March 2018 9:00 pm

So I've pulled the date format and passed it to the new method, it can still be overridden if passed to `date_range`

PR #1134, opened 2019-04-24
Improve removeRelationForEntity

nerrad:
While working on REM, @tn3rb noticed that in the new editor, detaching tickets from datetimes would cause the ticket entity to be removed as well.  This is because the current behaviour of `removeRelationForEntity` will not only remove the given relation but also the relation entity existing in the state.

This pull remedies that by leaving the relation entity in the state.

**Note:** @tn3rb I remember now why I had it this way originally and it was because there are certain entities that are only exposed for usage via their relation.  For example, in the case of tickets, the only time tickets show up anywhere in the Event Espresso UI is if:

- they are a default ticket (in which case they show up everytime the editor is loaded and in the Default Tickets tab.
- they are attached to a datetime.

When designing the data stores, I had this in mind and was thinking that if a relation was removed then there is increased chance the data could be orphaned (i.e. ticket being detached from a datetime) and thus we don't want to _persist_ that data to the db when a persist action is invoked since it would never appear anywhere in the UI/ux and would populate the db unnecessarily.

I still think the change I made in this pull is necessary, but I think we need to consider the orphaned state behaviour (perhaps in another pull/issue).  Some ideas:

- leave it up to client code to manage that (eg new event editor will not persist tickets that are new and not attached to a datetime - and the logic would be contained in the client implementation).
- add some automatic logic to the stores that don't persist entities that are defined as "relation dependent" somehow.  A few issues with this approach though are:
    * How do we indicate a "relation dependent" model?  Ideally probably would be good to be a property of the php model and expose that property via the REST schema. 
    * There would be exceptions to the rule.  Example is default tickets that might get persisted that don't necessarily have to be attached to a datetime (btw default tickets toggling is something you'll have to work out for the new editor as well).  How do we allow for these exceptions?  Maybe a special action that ignores the usual rules surrounding `relation dependency?`  Or maybe we have a special action that allows for receiving a conditional callback that receives the model entity and returns a boolean on whether it can be persisted or not (i.e. `persistEntityIfAllowed( modelName, entityId, canPersistCallback )`).  I think the latter will give the most flexibility for client code to control what is persisted or not but we will have to be careful about what side-effects run in that callback.

## How has this been tested

* [x] Updated and verify unit-tests pass.
* [x] Brent verified it fixed the issue in his REM branch.

## Checklist

* [x] My code is tested.


tn3rb:
re:
>  I think we need to consider the orphaned state behaviour (perhaps in another pull/issue). Some ideas:
> 
> leave it up to client code to manage that (eg new event editor will not persist tickets that are new and not attached to a datetime - and the logic would be contained in the client implementation).

the new editor has error handling in place to prevent a ticket <=> datetime relation from being removed if:
 - the ticket has no other datetime relations ie: only 1 datetime assigned to ticket (datetime may have other tickets)
 - the datetime has no other tickets ie: only 1 ticket assigned to datetime (ticket may have other datetimes)

> add some automatic logic to the stores that don't persist entities that are defined as "relation dependent" somehow. 

we _could_ embed similar logic directly in the stores to block relation removal if no other relations are found, and the only problem I see is that this is dependent on previous queries having returned those relations. I think in most cases this would never be a problem since all of the current queries return all relations for an entity. The only time this could become an issue is if multiple people are editing a site's data at the same time, ie: admin A removes ticket X from datetime Y and admin B removes ticket X from datetime Z, in separate but concurrent sessions, which would effectively orphan ticket X. Although possible, this seems highly unlikely. I think the better solution in this scenario is to rely on orphan protection server side and not even worry about it with the data stores.

tn3rb:
Here's an image of the kind of error that is displayed if you try to remove the last ticket from a datetime
Ticket Assignments Manager - relation error

PR #1144, opened 2019-04-26
Fix missing capabilities in default roles (fixes #829)

### LINKED ISSUE #829: Fix registrations list table using the `ee_delete_registration` meta cap
Originally reported in Codebase 11270 where @Pebblo reported:

> 
> Reported here: https://eventespresso.com/topic/other-admins-cannot-trash-registrations/
> 
> If you have a second Administrator account on your site, log into it and view the registrations list table. You'll find the 'Trash' link missing for the registrations - http://take.ms/gpXkZ (that screenshot is from the account assigned to the event those registrations are for, you won't see that link)
> 
> The same happens if you use an account with our 'Events Administrator'.
> 
> I checked and the link currently uses `ee_delete_registration` for the capability check, so only the user linked to the event the registration is for can trash them. If you have multiple admins (be that 'Administrators' or 'Event Administrators') should they all be able to trash those registrations in EE? 
> 
> If so it would then need to use `ee_delete_registrations` but does that also mean we need another cap for more control over this, `ee_delete_others_registrations`?
> 
> Side note, the bulk action for 'Trash Registrations' already uses `ee_delete_registrations`

I did a quick check and it looks like there is already a `ee_delete_others_registrations` cap, so this could be a quick fix.

nerrad:

See #829 for context.  This pull adds the missing `ee_delete_others_registrations` capability to the defaults for the `administrator` and the `ee_events_administrator` roles.

On the surface this looks like a regression at some point because I wasn't able to reproduce the original issue because my administrator roles on my development site already had the `ee_delete_others_registrations`.  This may have been why it slipped by testing in other cases as well.

However, I was able to reproduce by removing that capability from the role.  

**Note:** this will not retroactively fix the issue for users experiencing it.  They will have to reset their capabilities using the tool in the maintenance admin page or they will need to manually add the capability to the administrator roles.

## How has this been tested

*  Remove (or ensure) the `ee_delete_others_registrations` capability from the Administrator role.
*  Log in as a secondary administrator on your site and verify you are not seeing the trash link on the registrations list table for registrations belonging to an event the logged in administrator did not author.
* While checked in on this branch, reset the EE capabilities for the site.
* [x] Verify that the trash link is now visible for that user when refreshing the registration list table.

## Checklist

* [x] My code is tested.


joshfeck:
Verified the fix.

PR #1151, opened 2019-04-26
Refactor editor to block editor dependency (fixes #1059)

### LINKED ISSUE #1059: WP 5.2+: Refactor any dependencies on wp.editor that should be changed to wp.block-editor
As a part of the WordPress 5.2 release, there was some refactoring of editor related packages so that the block-editor is independent from the post-editor context.  You can [read more about this here](https://make.wordpress.org/core/2019/04/09/the-block-editor-javascript-module-in-5-2/).

Since back-compat is preserved via proxies, there's no urgency to updating any of our components or dependencies on wp-editor but to support future iterations of the editor being used throughout WordPress, it would be good for us to review where we have dependencies on `wp.editor`and the proxies there that should be switched to `wp.blockEditor` instead.

nerrad:

See #1059 for details.

This pull refactors any imports of `@wordpress/editor` that are deprecated and moved to `@wordpress/block-editor`.  As a result of this refactor:

- EE version requirements have been bumped to WordPress 5.2 in this pull (because of the impact).
- I've removed the wp 4.5 jobs from the travis config because that's no longer applicable.
- I had to tweak jest configuration to mock `simple-html-token` which is a dependency of the `@wordpress/blocks` module block validation.  There's something odd breaking snapshot tests of blocks (see [reported issue here](https://github.com/WordPress/gutenberg/issues/13364)).  Since I'm not caring about this module in the snapshot I just mocked it to fix the issue.
- It's expected (as of the moment this post was created) for a number of jobs to fail because the latest WP version is 5.1.1 which means EE won't activate with the min requirement bump.
- I added new configuration to the wp install script for travis that will accept a `WP_VERSION=next` environment variable value.  This allows us to add a (non-blocking fail) job to our tests against that upcoming release version of WP (in this case 5.2-RC1).  Helpful in this case because we can verify tests pass for this particular branch.

## How has this been tested

* [ ] Currently as of posting this ticket it is expected that php unit tests will fail for the majority of jobs (because of the version bump), so before merging this into master we do need to wait for WP 5.2 to drop so we can verify tests run against those builds of WP.
* [x] However, this can be verified to pass against the new `next` build in travis for php unit tests.
* [x] All js tests should pass for the js unit test builds.
* [x] Verify the EventAttendees block works as expected against a WP 5.2 release.

## Checklist

* [x] My code is tested.


mnelson4:
I just got unassigned so I'm not doing a big review. How many of our users are already on WP 5.1? 
Personally, I'm hesitant to bump the minimum version from 4.5 to 5.2 without adding any new features for users, only tidying up our codebase. This doesn't seem to really benefit users, just us.
Seeing how there's no urgency for this, I'd be inclined to wait a release or two. That's my two bits anyway.

tn3rb:
Code changes are fine so I've approved the code review but I agree with Michael's hesitation regarding bumping the minimum version required to 5.2.
According to the WP post about the the block editor changes, things are backwards compatible and they may not even begin showing deprecation notices until version 5.4, so it seems somewhat premature to force this now.

nerrad:
@mnelson4 and @tn3rb just to be clear, I am aware that WP has put in backward compatibility and mentioned that in the original issue this pull is related to.  I only did this because it was the last remaining thing in my NEXT queue so I assumed that Josh decided he was fine with doing this and bumping the minimum WP version.

nerrad:
> This doesn't seem to really benefit users, just us.

With the current and upcoming WordPress version (WP 5.2) this is true.  But WordPress is working on the new widget interface that IS user facing and WILL require this change.  So eventually, it will be a requirement if we want the `EventAttendees` block to be usable in the new widget interface.

We certainly do not have to merge this into master immediately, but there will be other reasons to bump to 5.2 such as changes in the `wp.data` api which improve error handling control among other things.

nerrad:
@joshfeck next steps for this pull:

- Can test against current 5.2 RC or wait until 5.2 is released - your call :)
- Once 5.2 is released, the travis build should be restarted (likely after rebasing master into this branch) so we can verify all unit tests pass.
- It'll be your call on when this actually gets merged to master so that you can plan the communication to customers about the minimum version bump requirement etc.

joshfeck:
This needs to be put on hold until we get the green light to require WP 5.2.

nerrad:
Once https://github.com/WordPress/gutenberg/pull/15246 is merged I can remove the fix done in 8916a77 (which will also require updating the related `@wordpress/block-editor` and `@wordpress/blocks` packages.

nerrad:
Just noting that the back-compat selectors/actions will start showing deprecation warnings in the console beginning with the next release of the GB plugin and the next major release of WP (5.3): https://github.com/WordPress/gutenberg/pull/15770

So if there's still no desire to bump minimum WP core version requirements for EE yet, we probably have one or two more major version of WP we can roll with before having to make these changes.  I do recommend bumping sooner rather than later though.  

Side note, this branch is now really out of date so will probably be easier to just cherrypick changes into a new branch (if/when we decide to do a release with them).

tn3rb:
not applicable anymore

PR #1153, opened 2019-04-26
Split up javascript builds and shorten build times

nerrad:
This is a build-tools improvement intended to accomplish the following:

- Shorten javascript build times on travis by not setting up a WP install since that's not needed.
- Split up js linting and js build tests from the unit test runs (this allows for easier spotting of what went wrong)
- only do full `npm install` where necessary (not necessary for lint test).


nerrad:
On the surface this doesn't seem to decrease build times that much, but I was monitoring things and the boot up process of the vm on travis is a LOT slower than usual.  The actual scripted build portion went a lot faster once the vm was booted up.  So I think this can be merged as is and we can revisit if we find its not really helping things that much.

nerrad:
It is an improvement though when compared with other pulls not having these changes (where unit test runs for js are taking 6+min)

PR #1164, opened 2019-04-28
should be nvm install not npm install for npm installation!

nerrad:
Pretty self-explanatory little fix.

No review needed, if tests still work on travis then this can be merged.

This should also speed up travis builds because `npm install` won't be running twice!


nerrad:
MUCH faster :)

Issue #1087, opened 2019-04-15
Fix issue where additional attendee questions column headings are omitted in CSV report


joshfeck:
**Describe the bug**

Questions that belong to question groups that are asked of Additional Attendees **only** do not have column headings in the CSV report. The answers are included in the report, so the report ends up with mismatched columns when this happens.

**Screenshots**

![Screen Shot 2019-04-15 at 3 54 08 PM](https://user-images.githubusercontent.com/891156/56161231-07f97480-5f97-11e9-957b-1185e23c5e1a.png)

**To Reproduce**
1. Add a question in Event Espresso > Reg Form > Questions
2. Add a question group in Event Espresso > Reg Form > Question Groups
3. Add an event and set the question group so it's used in the forms for additional attendees **only**
4. Go to the event page, select at least 2 tickets, then complete the reg forms
5. Go to Event Espresso > Registrations for the event and download the CSV report
You'll note that there's no column heading for the additional attendee-only question 

**Expected behavior**

Additional Attendee-only question should also have a column heading in the CSV report

**Environment:**
 EE4 current master

**Additional context**

mnelson4:
Oh cool.

mnelson4:
Hmmm I'm initially having trouble reproducing this issue, see https://drive.google.com/file/d/1i-zrcct-h1f7GtpitNsUcaYJ3BOj4wtK/view Can you see anything I'm doing wrong in order to reproduce @joshfeck?

One way I could see this issue being reproduced is by having a registration marked as INCOMPLETE, and that registrant being the only one asked a certain question.

mnelson4:
I also tried defining `EE_USE_OLD_CSV_REPORT_CLASS` to true to see if that was the difference, but nope. I am not reproducing, and looking at the code, I don't see how this would happen (it fetches all the questions for any registrations that would appear in the report, including those who are 
@joshfeck are you still able to reproduce this issue? If so, could you maybe send me a DB dump, or set it up on a publicly-accessible site so I can play with it?

joshfeck:
Looks like this is related to a migration from EE3 issue, the registrations CSV report were from registrations made with EE3 that were migrated to EE4. I'll see if the latest migration fixes happen to resolve this.

joshfeck:
Updated steps:

1. Using EE3, add a question in Event Espresso > Questions
2. Add a question group in Event Espresso > Question Groups, and assign the question to that group
[...]

PR #1182, opened 2019-05-02
Bug/missing columns in csv report

mnelson4:
Fixed how EE4 Registrations CSV reports could have more data columns than header columns when there were questions for additional attendees only imported from EE3. See https://github.com/eventespresso/event-espresso-core/issues/1087 for more info on the original issue.
Also fixed how system questions imported from EE3 would appear twice in the EE4 registrations CSV reports (once as an attendee column, and again with the custom questions).

The first issue was because custom questions imported from EE3 would have `QST_system=null`, whereas custom questions created in EE4 would have `QST_system=""`. We now always set `QST_system=""`, but the registrations-report-generating code is accommodating to both.
The second issue was because answers to custom questions imported from EE3 would have a row in esp_answer, whereas answers to custom questions in EE4 answers would not. We now don't add rows to esp_answer for answers to custom questions imported from EE3, but the registrations-report-generating code is, again, accommodating to both.

## How has this been tested

* [x] Follow josh's instructions from https://github.com/eventespresso/event-espresso-core/issues/1087#issuecomment-488355381 and verify there is a column header for custom questions asked for additional attendees only
* [x] also after that, verify there aren't two columns for system questions like the attendee first name, last name, and email (system questions should just be mentioned in the columns for attendee fields, NOT custom questions)
* [x] if you have direct DB access, check there are no rows in esp_question where `QST_system` is null
* [x] and no rows in esp_answer for system questions (eg questions with IDs 1-10)

## Checklist

* [x] My code is tested.

### DIFF
--- a/core/data_migration_scripts/4_1_0_stages/EE_DMS_4_1_0_answers.dmsstage.php
+++ b/core/data_migration_scripts/4_1_0_stages/EE_DMS_4_1_0_answers.dmsstage.php
@@ -62,7 +62,7 @@ class EE_DMS_4_1_0_answers extends EE_Data_Migration_Script_Stage_Table
         }
         // as inefficient as this sounds, we create an answer per REGISTRATION, (even if the registrations use the same attendee)
         foreach ($regs as $new_reg_id) {
-            $new_answer_id = $this->_insert_new_answer($old_row, $new_reg_id);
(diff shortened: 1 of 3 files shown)


mnelson4:
GitHub thinks Darren's been in these files most recently, and I thought I'd give Brent a break from reviewing.

PR #1184, opened 2019-05-03
Add withLatestCheckin HOC

nerrad:

This exposes a higher order component that makes it simpler for checkin related components to receive and act on the latestCheckin for a given registration and datetime id.  Currently this is utilized in the new barcode scanner app (see eventespresso/eea-barcode-scanner#7)

## How has this been tested

* [x] New tests for the withLatestCheckin hoc
* [x] Existing tests pass

Note:  This does not impact any current released user-facing features for Event Espresso so if javascript tests pass and this passes review it can be merged into master.

## Checklist

* [x] My code is tested.


assets/src/higher-order-components/with-latest-checkin.js
```diff
@@ -0,0 +1,71 @@
...
+					registration,
+					'registration'
+				) ) {
+					return {};
+				}
+				const { getLatestCheckin } = select( 'eventespresso/core' );
+				const { hasFinishedResolution } = select( 'core/data' );
+				return {
+					checkinEntity: getLatestCheckin( registration.id, datetimeId ),
+					hasResolvedCheckin: hasFinishedResolution(
```
tn3rb: won't this increase the number of renders in wrapped components because it will be returning every result that gets yielded by the generators?

In the admin refactor stuff I have primarily been using `hasFinishedResolution()` to determine whether to return the final results as opposed to the default "loading" results.

For example I would have done something like the following with this:

```jsx
const checkinEntity = getLatestCheckin( registration.id, datetimeId );
const hasResolvedCheckin = hasFinishedResolution(
	'eventespresso/core',
	'getLatestCheckin',
	[ registration.id, datetimeId ]
);
return hasResolvedCheckin ?
	{ checkinEntity, loadingCheckin: false } :
	{ checkinEntity: null, loadingCheckin: true};
```

which means that the wrapped component only sees its props change when `hasFinishedResolution()` returns `true`.

Is this not a good way to do things?

What are the benefits to how you are doing things?

assets/src/higher-order-components/with-latest-checkin.js
```diff
@@ -0,0 +1,71 @@
...
+ */
+const withLatestCheckin = createHigherOrderComponent(
+	compose( [
+		withSelect(
+			( select, { registration, datetimeId } ) => {
+				if ( ! isModelEntityOfModel(
+					registration,
+					'registration'
+				) ) {
+					return {};
```
tn3rb: what if the registration is invalid? 
what about using a warning here so at least there will be some kind of indication what the problem is?

assets/src/higher-order-components/with-latest-checkin.js
```diff
@@ -0,0 +1,71 @@
...
+				return {
+					onClick() {
+						if (
+							isModelEntityOfModel(
+								registration,
+								'registration'
+							)
+						) {
+							toggleCheckin( registration.id, datetimeId );
+						}
```
tn3rb: same deal as earlier, what if the registration is invalid?

maybe a warning here would be good so at least there will be some kind of indication what the problem is, otherwise you would just be clicking on whatever and getting absolutely no feedback.

assets/src/higher-order-components/with-latest-checkin.js
nerrad: I think that's something that should be left up to the wrapped component to do.  The messaging may differ on how a wrapped component handles when a registration isn't valid (wrapped components would still receive the `registration` prop).  This is predominately in place to short circuit the `withLatestCheckin` logic when the registration isn't valid because there's nothing it can do.

assets/src/higher-order-components/with-latest-checkin.js
nerrad: Well I did it my way predominately because I figured that provides some flexibility for the implementing component to decide things but I think your approach is better because it _ensures_ consuming components don't implement things in a way that are less performant.  So ya good point and I'll implement something like what you suggested.

assets/src/higher-order-components/with-latest-checkin.js
nerrad: My explanation for your other comment for the `withSelect` applies here as well.  I think it's still up to the wrapped component to decide how to handle when the incoming `registration` prop is invalid (as messaging may differ depending on the context).

assets/src/higher-order-components/with-latest-checkin.js
nerrad: I went to implement the warning anyways but then remembered again that I intentionally did things this way because there may be legitimate reasons for a parent component to pass through a `null` value for the registration (or when this hoc is composed with other hocs that retrieve a registration).

assets/src/higher-order-components/with-latest-checkin.js
nerrad: So what I ended up implementing in 7cdf751 was simply normalizing an `undefined` result to `null` for the `checkinEntity` property since that's most likely what you see in debugging.  When a selector with a resolver is resolving the initial returned value from the select call is `undefined`.  The return value when there is no checkIn record is `null`, so I normalized things so that `null` is swapped in for `undefined` to keep renders to a minimum.

I didn't change the `hasResolvedCheckin` prop because essentially that will either be true or false anyways, so splitting it out into a separate variable accomplishes nothing different.

PR #1186, opened 2019-05-04
Fix EE_Admin_Hooks not setting the correct value for _current_route (fixes #1185)

### LINKED ISSUE #1185: Ensure cpt-to-people.css is enqueued where needed
@joshfeck commented on [Mon Apr 15 2019](https://github.com/eventespresso/eea-people-addon/issues/16)

Discovered this while testing https://github.com/eventespresso/event-espresso-core/pull/1070

What happens is the "number of people assigned to this event" column is hidden when viewing a filtered list of events. 

Screenshot of the initial view where everything looks good:
Screen Shot 2019-04-12 at 6 38 31 PM

Then after clicking the Filter button (no filter settings changed, just clicked Filter)
Screen Shot 2019-04-12 at 6 38 51 PM

Then after clicking the Reset button the columns are OK again:
Screen Shot 2019-04-12 at 6 39 12 PM

After comparing & checking the source from the above views, the cpt-to-people.css stylesheet was found to not be loading when the list was filtered.

nerrad:

See #1185 for background.  Essentially the `EE_Admin_Hooks::_current_route` property was not getting set correctly.  The fix essentially [does similar to what is done](https://github.com/eventespresso/event-espresso-core/blob/ecc783579f7d81e0630bf3cd86bd592cb5ac6a8b/core/admin/EE_Admin_Page.core.php#L507) in `EE_Admin_Page::_page_setup()` for it's `_req_action` property.

## How has this been tested

This impacts anything using the EE_Admin_Hooks api.  So testing steps cover that impact:

* [x] Verify that this branch fixes the issue outlined (with reproducible steps) in #1185.
* [x] PHP unit tests should pass
* [x] Verify that Ticket and Datetime metaboxes load on the Event editor for both new and existing (edit) events.

## Checklist

* [x] My code is tested.


core/admin/EE_Admin_Hooks.core.php
```diff
@@ -357,12 +358,26 @@ public function enqueue_scripts_styles()
...
         $this->_ajax_func = $this->_init_func = $this->_metaboxes = $this->_scripts = $this->_styles = $this->_wp_action_filters_priority = array();
-        $this->_current_route = isset($_REQUEST['action']) ? $_REQUEST['action'] : 'default';
+        $this->_current_route = $this->getCurrentRoute();
         $this->caller = get_class($this);
         $this->_extend = stripos($this->caller, 'Extend') ? true : false;
     }
 
 
+    /**
+     * A helper for determining the current route.
```
mnelson4: I'd suggest giving a summary of the logic in the method description, so we can know what query parameters to set in order to set the current route without needing to read through the code. Eg

```suggestion
     * A helper for determining the current route. Uses `$_REQUEST['action']` if present and not set to `-1`, otherwise falls back to using `$_REQUEST['route']`.
```

nerrad: I'm not really sure how much the expanded comment in the header helps over reading the code and the comments in the code above the conditions.  I understand it's intended as a summary of the logic, however the summary isn't entirely accurate (because it doesn't include that it also considers whether `$_REQUEST['route']` is set).  Improving the summary essentially results in english form of the condition which I think would be superfluous imo, i.e.

> If an `action` argument is in the request and does not equal -1, then that value is considered the route.  Otherwise, the value for route will be the value of the `route` argument in the request if it exists.  Otherwise, the fallback value will be 'default'.

I don't think that gives much more helpful information than what is already concisely described in the actual conditions :)

mnelson4: > I'm not really sure how much the expanded comment in the header helps over reading the code and the comments in the code above the conditions.

ya no worries, the idea was just to not require people to read the actual code and inline comments if they want to use that method. (IMO you want to learn how to use a method, you shouldn't need to read through its code and inline comments. You only do that if you're to learn *how* it works.)

joshfeck:
I tested this and it's ready for merging.

PR #1187, opened 2019-05-04
Add helper method for exposing filtered EE brand name and also expose to js (fixes #1183)

### LINKED ISSUE #1183: Create a helper method for the `Event Espresso` brand text
Originally reported in https://events.codebasehq.com/projects/event-espresso/tickets/10748

where @nerrad wrote:

We shouldn't be doing this in localized strings:

```
esc_html__('Event Espresso is awesome', 'event_espresso');
```

Instead, we should be doing this:

```
sprintf( esc_html__('%1$s is awesome', 'event_espresso'), 'Event Espresso'));
```

Or even better:

```
sprintf( esc_html__('%1$s is awesome', 'event_espresso'), EventEspresso\core\domain\Constants::brandName());
```

And add the `brandName` method to the Constants class in core:

```
public static function brandName()
{
     return apply_filters( 'FHEE__EventEspresso_core_domain_Constants__brandName', 'Event Espresso' );
}
```

For now, we can:

- [ ] add the helper method 

- [ ] add to the developer documentation to outline how to proceed going forward with handling i18n strings that include "Event Espresso" within

After this is merged, we can start to use this method to update existing strings that appear on the **admin side** while working on other things when we encounter a string that contains "Event Espresso". Any existing front-end strings should be left alone to avoid breaking existing translations.

We'll need this for the Attendee Importer add-on, and can possibly use this for REM too.

nerrad:

See #1183 for background.  In this pull:

* `EventEspresso\core\domain\Domain::brandName()` static method is introduced that returns a filtered `Event Espresso` brand name.
* This new value is exposed to javascript via the `eejsdata` object on `eejs.data.brandName`.  So modules can import it via doing something like this:

```js
import { __, sprintf } from '@wordpress/i18n';

const { brandName } = data;
const someLabel = sprintf( __( 'Provided by %s', 'event_espresso' ), brandName );
```

For php usage:

```php
use EventEspresso\core\domain\Domain;

$some_label = sprintf( __( 'Provided by %s', 'event_espresso' ), Domain::brandName() );
```

### Next Steps

* Once this pull is merged the new static method can be implemented in places referencing the Event Espresso brand in either add-ons or core.
* While I support this change myself (I think the benefits are significant), I do think this is something that needs to run by EE owners (cc @garthkoyle and @sethshoultes) because implementing this everywhere will make it much easier for third parties to white-label Event Espresso.
* I didn't add any documentation for this yet because I'm not sure where it should live.  As far as I can tell we don't have any documentation on proper usage of i18n in EE (that I could remember or find in a brief search) so I wonder if documentation can be done in a separate pull (and really anyone can do it - I'm kind of burned out from doing this kind of documentation).

## How has this been tested

* [x] Load up any ui view that enqueues `eejsdata` in the page output (the plugins page is a good one).  In the browser dev tools console, type `eejs.data.brandName` and you should see `Event Espresso` returned.  This verifies both the php output and the js output (because `eejsdata` gets the string from the new php method).

## Checklist

* [x] My code is tested.

### DIFF
--- a/core/domain/Domain.php
+++ b/core/domain/Domain.php
@@ -66,4 +66,16 @@ class Domain extends DomainBase implements CaffeinatedInterface
         $this->caffeinated = (! defined('EE_DECAF') || EE_DECAF !== true)
             && is_readable($this->pluginPath() . 'caffeinated/brewing_regular.php');
     }
+
+
+    /**
+     * This should be used everywhere the Event Espresso brand name is referenced in public facing interfaces
+     * to allow for filtering the brand.
+     *
+     * @return string
+     */
+    public static function brandName()
+    {
+        return (string) apply_filters('FHEE__EventEspresso_core_domain_Domain__brandName', 'Event Espresso');
+    }
 }
--- a/core/domain/services/assets/CoreAssetManager.php
+++ b/core/domain/services/assets/CoreAssetManager.php
@@ -9,6 +9,7 @@ use EE_Template_Config;
 use EED_Core_Rest_Api;
 use EEH_DTT_Helper;
 use EEH_Qtip_Loader;
+use EventEspresso\core\domain\Domain;
 use EventEspresso\core\domain\DomainInterface;
 use EventEspresso\core\domain\values\assets\JavascriptAsset;
 use EventEspresso\core\exceptions\InvalidDataTypeException;
@@ -312,6 +313,8 @@ class CoreAssetManager extends AssetManager
                 'admin_url' => admin_url('/'),
             )
         );
+        // Event Espresso brand name
+        $this->registry->addData('brandName', Domain::brandName());
         /** site formatting values **/
         $this->registry->addData(
             'site_formats',


joshfeck:
> implementing this everywhere will make it much easier for third parties to white-label Event Espresso.

We're not implementing this everywhere just yet though. For now, we're going to make use of this in the Attendee Importer add-on.

nerrad:
> For now, we're going to make use of this in the Attendee Importer add-on.

That's good, but I still think there should be some review by the code owners to aid with some sort of directive on how extensive we will implement this.  I'm reluctant to start using this in any new code or refactors for php or js until that's direction is clear.  I appreciate that the major impetus for this change is easier switching of the Event Espresso with Event Smart for things we put on EventSmart (starting with the Attendee Importer add-on), but there's still the real side-effect that someone could easily white-label that add-on for their own use-case as well.  Personally, I think that's okay.  But I'm wary of assuming that will be okay to Garth and Seth.

joshfeck:
@sethshoultes or @garthkoyle can you respond to Darren's concern about whether this change will be OK?

sethshoultes:
I'm okay with someone white labeling the plugin or addons for their own use case. I think we could make an addon available for sale that helps with the process.

joshfeck:
This is tested & is ready for merging.

Issue #1193, opened 2019-05-07
Fix issue with Attendee Block's Ticket select - post migration from EE3


joshfeck:
**Describe the bug**
After migrating a site from EE3 to EE4, the attendee block UI hangs when selecting a ticket. The error in the console is:
`Unhandled promise rejection TypeError: "Instance of Currency required"`
```
assertCurrency ee-valueObjects.276cc7a17fa0c373bfe1.dist.js:2894
    js ee-valueObjects.276cc7a17fa0c373bfe1.dist.js:3399
    setCurrency ee-valueObjects.276cc7a17fa0c373bfe1.dist.js:3001
    Money ee-valueObjects.276cc7a17fa0c373bfe1.dist.js:2986
    maybeConvertToValueObject ee-model.fc93494192c32c8ac7cc.dist.js:3014
    derivePreparedValueForField ee-model.fc93494192c32c8ac7cc.dist.js:3089
    setInitialEntityFieldsAndValues ee-model.fc93494192c32c8ac7cc.dist.js:2664
    createEntityGettersAndSetters ee-model.fc93494192c32c8ac7cc.dist.js:2462
    Lodash 4
    createEntityGettersAndSetters ee-model.fc93494192c32c8ac7cc.dist.js:2451
    BaseEntity ee-model.fc93494192c32c8ac7cc.dist.js:1732
    _class ee-model.fc93494192c32c8ac7cc.dist.js:1850
    fromExisting ee-model.fc93494192c32c8ac7cc.dist.js:1941
    fullEntities resolvers.js:64
    buildAndDispatchEntitiesFromResponse$ resolvers.js:63
    tryCatch runtime.js:45
    invoke runtime.js:271
    method runtime.js:97
    yieldValue Redux
    run wp-polyfill.js:4399
    notify wp-polyfill.js:4416
    flush wp-polyfill.js:1501
wp-polyfill.js:4434:11
```

**To Reproduce**

1. Start with an EE3 site, then activate EE4 and migrate the data
2. Go to add a page, add an attendee block
3. Configure attendee block: select an event, then datetime, then on the ticket select

If you go to the Event Espresso > General Settings page and click Save, then no more errors. So it seems a bug is exposed where there's a difference in the config that affects the Attendee block.

**Screenshots**

**Environment:**
 - Version of Event Espresso:  master
 - Version of WordPress: 5.2 RC2 -- Gutenberg plugin not activated
 - Browser: Firefox
[...]

PR #1195, opened 2019-05-07
Fix timezone error for datetime value object with EventAttendees block usage when site is only using an offset for timezone.

nerrad:

While working on debugging #1193 I discovered that if I have my site timezone set to a UTC offset (eg UTC+1 or UTC-1) then using the  EventAttendees will result in this error:

```
assertions.js:108 Uncaught (in promise) InvalidTimezone: Invalid argument provided. The timezone string provided is not valid.
    at Module.assertTimezoneIsValid (http://ee.test/wp-content/plugins/event-espresso-core/assets/dist/ee-valueObjects.276cc7a17fa0c373bfe1.dist.js:483:11)
    at Function.assertTimezoneIsValid (http://ee.test/wp-content/plugins/event-espresso-core/assets/dist/ee-valueObjects.276cc7a17fa0c373bfe1.dist.js:1250:72)
    at new DateTime (http://ee.test/wp-content/plugins/event-espresso-core/assets/dist/ee-valueObjects.276cc7a17fa0c373bfe1.dist.js:655:16)
    at Function.fromISO (http://ee.test/wp-content/plugins/event-espresso-core/assets/dist/ee-valueObjects.276cc7a17fa0c373bfe1.dist.js:1437:14)
    at maybeConvertToValueObject (http://ee.test/wp-content/plugins/event-espresso-core/assets/dist/ee-model.fc93494192c32c8ac7cc.dist.js:3010:88)
    at derivePreparedValueForField (http://ee.test/wp-content/plugins/event-espresso-core/assets/dist/ee-model.fc93494192c32c8ac7cc.dist.js:3089:10)
    at setInitialEntityFieldsAndValues (http://ee.test/wp-content/plugins/event-espresso-core/assets/dist/ee-model.fc93494192c32c8ac7cc.dist.js:2664:133)
    at http://ee.test/wp-content/plugins/event-espresso-core/assets/dist/ee-model.fc93494192c32c8ac7cc.dist.js:2462:7
    at http://ee.test/wp-includes/js/dist/vendor/lodash.js?ver=4.17.11:4911:15
    at baseForOwn (http://ee.test/wp-includes/js/dist/vendor/lodash.js?ver=4.17.11:2996:24)
```

Essentially this is because the javascript was not correctly handling when the server provides an empty string for the `eejsdata.default_timezone.timezone` property.  That string comes from `get_option( 'timezone_string' )` server side which is just an empty string when the site is set to an offset only.

Since this is an indicator of offset, then the solution is to coerce the `timezone_string` value to `UTC` and the `DateTime` vo (which was throwing the error) will be constructed fine.

## How has this been tested

To reproduce the bug against master:

* Set your timezone in the WordPress general options to an offset (UTC+0, UTC+1 etc).  **Note:** the `UTC` (only that no +/-) option in the dropdown is actually a timezone string, so the bug won't reproduce if you select that.
* Visit the post page and attempt to add an EventAttendees block, you'll see the above error in the console and the block won't function.

* [x] Verify after switching to this branch and retesting that there is no error and the block works as expected.
* [x] Verify javascript unit tests pass.

## Checklist

* [x] My code is tested.


joshfeck:
tested and all is good!

PR #1201, opened 2019-05-10
Don't use ExceptionStackTraceDisplay() when catching an exception

Pebblo:
Whilst looking into an issue with the attendee mover, I did a little digging into the ['same ticket exception'](https://github.com/eventespresso/eea-attendee-mover/issues/10) ticket.

When we catch an exception within the SequentialForm for the attendee mover, in the catch we use [ExceptionStackTraceDisplay()](https://github.com/eventespresso/event-espresso-core/blob/master/core/exceptions/ExceptionStackTraceDisplay.php#L26-L33) which either displays a stack trace if WP_DEBUG is enabled or it just throws the exception if it's not. I think that as this is where we are catching the exceptions if that function throws it goes uncaught and so can't be used here (at least as it is now), right?

I've just removed it from the catch so the error is displayed on the form with no exception for now.

See: https://github.com/eventespresso/eea-attendee-mover/issues/10

## How has this been tested
With ee core on master branch, use the attendee mover and try to select the exact same ticket the user is already on.

With WP_DEBUG you'll see a nice exception trace, with it disabled you'll get and uncaught exception error:

```
Fatal error: Uncaught EventEspresso\core\exceptions\InvalidFormSubmissionException: The data for the "Registrations can not be moved if you select the exact same ticket that the registration already has! Please select a different ticket." form, is either missing or was not submitted properly. in /home/pebbloco/public_html/ee4dev/wp-content/plugins/eea-attendee-mover/form/SelectTicket.php:159 
Stack trace: 
#0 /home/pebbloco/public_html/ee4dev/wp-content/plugins/event-espresso-core-reg-git/core/libraries/form_sections/form_handlers/SequentialStepFormManager.php(555): EventEspresso\AttendeeMover\form\SelectTicket->process(Array) 
#1 /home/pebbloco/public_html/ee4dev/wp-content/plugins/event-espresso-core-reg-git/core/libraries/form_sections/form_handlers/SequentialStepFormManager.php(413): EventEspresso\core\libraries\form_sections\form_handlers\SequentialStepFormManager->processCurrentStepForm(Array) 
#2 /home/pebbloco/public_html/ee4dev/wp-content/plugins/eea-attendee-mover/EED_Attendee_Mover.module.php(509): EventEspresso\co in /home/pebbloco/public_html/ee4dev/wp-content/plugins/eea-attendee-mover/form/SelectTicket.php on line 159
```

Use this branch and in either case, you'll just see an error notice at the top of the page like so - https://monosnap.com/file/VOP7IJF3RA04htaflCQRGhbQ23VAWO

## Checklist

* [x] My code is tested.


joshfeck:
Tested and verified the fix

PR #1203, opened 2019-05-10
Improve BaseEntity.forClone method to allow for cloning primary key too.

nerrad:

While working on #961 I encountered a use-case where a clone of a `BaseEntity` instance is needed _including_ the id of the entity and with a state other than `SAVE_STATE.NEW`.   This use-case was:

- A form is loaded, clone the entity so that changes are specific to the clone.
- Form provides option to discard changes and if changes are discarded the clone is discarded and original entity is untouched.
- If changes are accepted, dispatch cloned entity to replace existing entity in the state.

For the above, the cloned entity needed to have the same ID and inherit the same save state as the clone.

So in this pull:

- Entities with `SAVE_STATE.NEW` have a setter for the primary key id (but ONLY if the state is new).
- the `BaseEntity.clone` getter is now a function that accepts a `keepId` boolean argument. When this is true, the entity is cloned completely including the save state and the primary key value.

## How has this been tested

* [ ] Automated tests for behaviour

Currently the clone functionality is not used in any released user-facing features so should be safe to break back-compat, also can be merged after passing tests.

## Checklist

* [ ] My code is tested.

### DIFF
--- a/assets/src/data/model/entity-factory/base-entity.js
+++ b/assets/src/data/model/entity-factory/base-entity.js
@@ -137,7 +137,7 @@ class BaseEntity {
 			const newEntity = factory.createNew( this.forClone );
 			if ( keepId ) {
 				newEntity.id = this.id;
-				setSaveState( this, SAVE_STATE.CLEAN );
+				setSaveState( this, newEntity.saveState );
 			}
 			return newEntity;
 		};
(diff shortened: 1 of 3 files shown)


assets/src/data/model/entity-factory/base-entity.js
```diff
@@ -127,13 +127,20 @@ class BaseEntity {
...
-		return new BaseEntity(
-			this.modelName,
-			this.forClone,
-			{ $schema: {}, properties: this.schema },
-			this.fieldPrefixes,
-			true
-		);
+		return ( keepId = false ) => {
+			// @todo memoize this
+			const factory = createEntityFactory(
```
nerrad: I plan on memoizing this, forgot to add it for the pull.  Memoizing will ensure a factory is only ever created once per model here.

PR #1209, opened 2019-05-14
Adds support for bin-2 mastercard card numbers

mnelson4:

Adds support for bin-2 mastercard card numbers.
Simplifies `EE_Credit_Card_Validation_Strategy` to just provide a regex to `EE_Text_Validation_Strategy`. That regex matches the old one, plus what was mentioned on  https://eventespresso.com/topic/2-series-mastercards-marked-as-invalid/.

## How has this been tested

* [x] Use any onsite payment method that has a credit card input (and not one with any fancy JS, like Stripe; I mean use one like PayPal Pro or AIM or the test onsite payment method). Use the credit card number "2212345678901234". That was considered invalid before, but now is valid.
* [x] Register again, but this time use "6011123456789012" (the next pattern in the regex). It worked before and should continue to work. Actually, any number that worked before should continue to work.

## Checklist

* [ ] My code is tested.

### DIFF
--- a/core/libraries/form_sections/strategies/validation/EE_Credit_Card_Validation_Strategy.strategy.php
+++ b/core/libraries/form_sections/strategies/validation/EE_Credit_Card_Validation_Strategy.strategy.php
@@ -21,47 +21,11 @@ class EE_Credit_Card_Validation_Strategy extends EE_Text_Validation_Strategy
         if (! $validation_error_message) {
             $validation_error_message = __("Please enter a valid credit card number", "event_espresso");
         }
-        parent::__construct($validation_error_message);
-    }
-
-    /**
-     * just checks the field isn't blank
-     * @param $normalized_value
-     * @throws EE_Validation_Error
-     * @return void
-     */
-    public function validate($normalized_value)
-    {
-        if ($normalized_value && ! $this->verify_is_credit_card($normalized_value)) {
-            throw new EE_Validation_Error($this->get_validation_error_message(), 'required');
-        }
-    }
-
-
-
-    /**
-     * gets additional validation rules for use in the jQuery validation JS corresponding to this field when displaying.
-     *
-     * @return array
-     */
-    public function get_jquery_validation_rule_array()
-    {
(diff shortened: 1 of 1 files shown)


core/libraries/form_sections/strategies/validation/EE_Credit_Card_Validation_Strategy.strategy.php
```diff
@@ -21,47 +21,11 @@ public function __construct($validation_error_message = null)
...
-
-
-    /**
-     * gets additional validation rules for use in the jQuery validation JS corresponding to this field when displaying.
-     *
-     * @return array
-     */
-    public function get_jquery_validation_rule_array()
-    {
-        return array('creditcard'=>true, 'messages' => array( 'creditcard' => $this->get_validation_error_message() ) );
```
tn3rb: the rule array in the parent class looks like this:

```php
array( 'regex'=> $this->regex_js(), 'messages' => array( 'regex' => $this->get_validation_error_message() ) );
```

Do we need to change any JS code anywhere to account for this?
I would think that there might be some code out there trying to use `creditcard` as a key for that array but will no longer find it.

mnelson4: Thanks for inquiring. 

Ya so this array was getting dumped into a JS variable and sent to jQuery validate, instructing it to use jQuery's "credit card" validation on the input.
So, with this change, we're now no longer using jQuery's credit card validation, instead just supplying jQuery validate with a REGEX to use on credit card fields.

We're doing this because it seems `creditcard` validation basically had some REGEX built into jQuery validate, which had trouble with the new mastercard numbers. This new regex accounts for the new mastercard numbers.

The new regex is the exact same server-side as it is client-side, so this will make it more consistent too (before it might have been possible for a card to pass validation client-side but not server-side, or vice-versa).

joshfeck:
Tested the AIM gateway and prior supported numbers still work, and the validation allow for numbers like 2212345678901234 now.

PR #1219, opened 2019-05-17
add a check to ensure that the registrations have been updated

joshfeck:

If a transaction is interrupted between the time of payment and return to the site the payment will be shown in the Event Espresso Registration list, but the reg status will remain RPP.

Screen Shot 2019-05-17 at 4 31 11 PM

This change attempts to ensure the registration status gets updated later when the related cron task fires

## How has this been tested

This was tested using PayPal Express. The session expiration was set to 5 minutes i.e. add
```
add_filter( 
'FHEE__EE_Session__construct___lifespan', 
'ee_change_reg_time_limit' 
);
function ee_change_reg_time_limit() {
 return 5 * MINUTE_IN_SECONDS;
}
```
to a code snippet plugin.

Immediately after submitting the payment on PayPal.com, close the tab. 
## Checklist

* [ ] My code is tested.


tn3rb:
there's a mistake in the code I gave you in chat...
i copied over some existing code from SPCO and forgot to reset the last param from false to true...
which just so happens to control whether notifications are being sent out, which is why you were not seeing them in your initial tests.
hat tip to @Pebblo for noticing that

so the code needs to be changed to:

```php
$transaction_processor->update_transaction_and_registrations_after_checkout_or_payment(
    $transaction,
    $transaction->last_payment(),
    [
        ['Registration.STS_ID' => 'RPP'],
        'order_by' => ['REG_count' => 'ASC']
    ]
);
```

which will hopefully also restrict notifications only going out for registrations that were pending payment, but if that does not work, then simply try:


```php
$transaction_processor->update_transaction_and_registrations_after_checkout_or_payment(
    $transaction,
    $transaction->last_payment()
);
```

tn3rb:
going to merge as @joshfeck has already tested this

PR #1220, opened 2019-05-18
Fix normalization of eventespresso/core store relation updates

nerrad:
## Problem this Pull Request solve

During the review of #961 I discovered a bug within the `eventespresso/core` reducer logic for relations:

The normalization logic involved with updating relation state was not correctly accounting for all scenarios.

If a relation had already been added for a relation to a different entity than the one triggering the update then there'd end up being entries for the relation in both `index` and `entityMap` states.  
Normalization is supposed to ensure there's only ever a keyed entry for a model name in one of those state.

Example:

The following:

```
createRelations( 'event', 10, datetime, [ ...datetimes ] );
createRelations( 'datetime', 20, 'tickets', [ ticket ]  );
```

Results in this _incorrect_ state:

```js
const state = {
	entityMap: {
		event: {
			10: {
				datetimes: [ 20, 30 ]
			}
		},
		datetime: {
			20: {
				tickets: [ 1 ]
			}
		}
	},
	index: {
		datetimes: {
			20: {
				event: [ 10 ],
			},
			30: {
				event: [ 10 ]
			}
		},
		tickets: {
			10: {
				datetime: [ 20 ]
			},
		},
	}
};
```

The state **should** be "normalized" so that the datetime model only exists in either `entityMap` or `index` as a top-level key:

```js
const state = {
	entityMap: {
		event: {
			10: {
				datetimes: [ 20, 30 ]
			}
		},
		ticket: {
			1: {
				datetimes: [ 20 ]
			}
		}
	},
	index: {
		datetimes: {
			20: {
				event: [ 10 ],
				ticket: [ 1 ],
			},
			30: {
				event: [ 10 ]
			}
		},
	}
};
```

This is important behaviour to have correct because the purpose of having these two states tracking relations is for quicker lookup.  If everything is in one state then selectors will consistently have to loop through all indexes to find the relation.  With this setup, the selector can just check top level keys.

With that said, I realized after initial fixes in this branch that the current state creates more problems (see [comment](https://github.com/eventespresso/event-espresso-core/pull/1220#issuecomment-493692701)).  So I ended up refactoring the state shape for relations so that every relation has a mirror copy.  Essentially ending up with something like this:

```js
const relationsState: {
       event: {
		10: {
			datetimes: [ 20, 30 ]
		},
	},
	datetime: {
		20: {
			events: [ 10 ]
		},
		30: {
			events: [ 10 ]
		}
	},
}
```

This accomplishes a couple things:

- eliminates need to "normalize" between two maps (which ultimately makes the logic much more straightforward than it was)
- keeps fast lookups as now there will always be a 1 step match to incoming relation queries via the relation selectors (regardless of order)

## How has this been tested

* [x] Ensure existing unit tests pass (with updates as necessary) and added new tests for the new state shape.
* [x] Ensure that this new state shape works well in the work being done in #961 (and it works swimmingly). 

**Note:** the changes in this branch fixed the issue I experienced on #961 due to the bug.  Also, these changes do not impact any user facing released features so it's safe to merge once tests pass.

## Checklist

* [x] My code is tested.


nerrad:
**Note**:

While this fixes the issue experienced in #961, I realized as I was working on this that there's a flaw in the logic for this kind of state.  The flaw is it assumes there will only ever be single relation from a model to another model when it's possible that a single model may be related to multiple models (which in turn are related to other things).  This then means its easily possible for client code to set up an invalid state.

Here's an example (for this example I'm just using ids for the 4th argument in the createRelation call instead of its expected model entity for illustration purposes).

```js
createRelation( 'event', 10, 'datetimes', [ 20 ] );
createRelation( 'datetime', 20, 'tickets',  [ 1 ] );
createRelation( 'registration', 100, 'events' [ 10 ] );
createRelation( 'registration', 100, 'tickets', [ 1 ] );
```
results in the "incorrect" state:

```
const state = {
	entityMap: {
		event: {
			10: {
				datetimes: [ 20 ],
			}
		},
		ticket: {
			1: {
				datetimes: [ 20 ],
			}
		},
		registration: {
			100: {
				events: [ 10 ],
				tickets: [ 1 ],
			}
		}
	},
	index: {
		datetimes: {
			20: {
				event: [ 10 ],
				ticket: [ 1 ],
			}
		},
		events: {
			10: {
				registration: [ 100 ],
			}
		},
		tickets: {
			1: {
				registration: [ 100 ],
			}
		}
	}
};
```

But if you do it this way:

```js
createRelation( 'event', 10, 'datetimes', [ 20 ] );
createRelation( 'datetime', 20, 'tickets',  [ 1 ] );
createRelation( 'event', 10, 'registrations' [ 100 ] );
createRelation( 'registration', 100, 'tickets', [ 1 ] );
```

Then you get the correct state:

```js
const state = {
	entityMap: {
		event: {
			10: {
				datetimes: [ 20 ],
				registrations: [ 100 ],
			}
		},
		ticket: {
			1: {
				datetimes: [ 20 ],
				registrations: [ 1 ],
			}
		},
	},
	index: {
		datetimes: {
			20: {
				event: [ 10 ],
				ticket: [ 1 ],
			}
		},
		registrations: {
			100: {
				event: [ 10 ],
				tickets: [ 1 ],
			}
		}
	}
};
```

So, I think there's still some work needed in this ticket.

nerrad:
As I do more testing, I'm realizing that the state shape needs to change.  Instead of keeping two keys in the `relations` index (`entityMap` and `index`) I need to just always save relations from either direction.  For example, `createRelations( 'event', 10, 'datetime', [ 20, 30 ] )` will result in the following in the `relations` state:

```js
const relationsState: {
       event: {
		10: {
			datetimes: [ 20, 30 ]
		},
	},
	datetime: {
		20: {
			events: [ 10 ]
		},
		30: {
			events: [ 10 ]
		}
	},
}
```

That way there's always a mirror of the relation record recorded for faster lookups and we're the logic doesn't have to do any "normalization".  There will always be a correct relations record for whatever set of queries come in.

nerrad:
> in other words, shouldn't the above just need to return:

Yup you're spot on and I'll make that fix.

nerrad:
K @tn3rb , 9b24cf5 has the changes that should address what you pointed out.

PR #1233, opened 2019-05-22
Fixes #1231 Missing Thank You Page Transaction Details

### LINKED ISSUE #1231: Transaction details table no longer loads on the thank you page
"Dude, wheres my transaction details table?"

From 4.9.76.p onwards, the 'Transactions Details Table' no longer loads on the thank you page.

See - https://monosnap.com/file/4Dx31kMNGzcYcH8YLMrPsitf2hT4X8

Looks like its from this commit - https://github.com/eventespresso/event-espresso-core/commit/340c62d8268348b941a3a2f441aec2a932edbf73

I did some digging and found that here:

https://github.com/eventespresso/event-espresso-core/blob/master/core/domain/services/admin/ajax/ThankYouPageIpnMonitor.php#L89

'still_waiting' wasn't set, so then returns the response from [paymentDetails()](https://github.com/eventespresso/event-espresso-core/blob/master/core/domain/services/admin/ajax/ThankYouPageIpnMonitor.php#L184), which only populates the transaction details section if `$since` is set (which it isn't here).

If I pass paymentDetails the `$response` variable when its called, I get all of the details expected on the thank you page but I'm not so sure that's the correct fix.

tn3rb:
see #1231 
@Pebblo correctly identified that the issue was caused because the `$response` array was not being passed along to `EventEspresso\core\domain\services\admin\ajax\ThankYouPageIpnMonitor::paymentDetails()`.
I noticed that there were also a few places where the response array was being completely overwritten, which could also erase data, so I did some additional refactoring so that any data added to the response array is preserved.

## How has this been tested
browser based monkey testing only 

### testing

I think we should test every possible thank you page scenario just to be sure that nothing else is broken

 - registrations with the following outcomes:
        - deferred payment (ie: invoice)
        - offsite payment with successful payment
        - onsite payment with successful payment
        - offsite payment with a declined/failed payment prior to successful payment (ie: expired CC date)
 - return visit with pending payment registrations
 - return visit with approved registrations
 - ???
- experiment with the thank you page in this branch and attempt to find any other potential issues

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -139,10 +139,8 @@ jobs:
       env: WP_MULTISITE=1
       if: NOT commit_message =~ js_only AND NOT commit_message =~ dependabot
     # wp 4.5 builds
-    - php: 5.6
-      env: WP_VERSION=4.5
-      if: NOT commit_message =~ js_only AND NOT commit_message =~ dependabot
     - php: 5.5
+      env: WP_VERSION=4.5
       if: NOT commit_message =~ js_only AND NOT commit_message =~ dependabot
     - php: nightly
       if: NOT commit_message =~ js_only AND NOT commit_message =~ dependabot
--- a/.travis.yml
+++ b/.travis.yml
@@ -134,8 +134,6 @@ jobs:
       if: NOT commit_message =~ js_only AND NOT commit_message =~ dependabot
     - php: 5.6
       if: NOT commit_message =~ js_only AND NOT commit_message =~ dependabot
-    - php: 5.5
-      if: NOT commit_message =~ js_only AND NOT commit_message =~ dependabot
     # multisite
     - php: 7.2
       env: WP_MULTISITE=1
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -144,6 +142,8 @@ jobs:
     - php: 5.6
       env: WP_VERSION=4.5
       if: NOT commit_message =~ js_only AND NOT commit_message =~ dependabot
+    - php: 5.5
+      if: NOT commit_message =~ js_only AND NOT commit_message =~ dependabot
```
nerrad: you need to include `env: WP_VERSION=4.5` if you only want to test against wp 4.5.

Also, I think we might as well just have a single PHP5.5 and WP 4.5 job.  I think also testing php5.6 on 4.5 is unnecessary.  The biggest changes in php happened going from PHP 5.3->5.4 and 5.6->7.0.  Php 5.6 and 5.5 are _very slow_ builds for our test suite so whatever we can eliminate will help speed up test runs.

tn3rb: makes sense... will change

PR #1234, opened 2019-05-22
Fet/new css framework tweaks

tn3rb:

Changes in this pull for the new CSS framework include:

 - adding units to the new CSS variables so that you don't have to use `calc()` to do so in stylesheets

for example, currently you would have to do something like this in your stylesheet to add the units:

```css
font-size: calc(var(--ee-font-size-default) * 1px);
```

but these changes mean you only have to do this:

```css
```

 - adding missing entity status colors for tickets
 - added a few extra options for border radius sizes
 - changed the class name for the new button styles so that they don't conflict with an existing class we use in the admin
 - and made a few tweaks to the default theme
 - changed the color for sold out tickets and datetimes from yellow to purple, because:
    - yellow is too harsh of a color and is typically used to represent "warning" which is how we use it for abandoned transactions and Incomplete registrations
    - purple is a more positive color (ie: royalty) which is a better representation for sold out dates and tickets (which is good)
    - purple is currently used as the status color for Wait List registrations, which only happens when a datetime or ticket is sold out, so this would be a more consistent pattern (ie: purple registrations go with purple datetimes/tickets)

## How has this been tested

browser based monkey testing and via actual use in REM and EDTR


assets/src/components/ui/styles/root/effects.css
```diff
@@ -0,0 +1,20 @@
+:root {
+
+    --ee-animation-spin: ee-animation-spin 1s linear infinite;
```
nerrad: why the extra line spaces here?

assets/src/components/ui/styles/root/effects.css
```diff
@@ -0,0 +1,20 @@
...
+
+@-webkit-keyframes ee-animation-spin {
+    100% { -webkit-transform: rotate(360deg); }
+}
+
+@keyframes ee-animation-spin {
+    100% {
+        -webkit-transform: rotate(360deg);
+        transform: rotate(360deg);
+    }
```
nerrad: you don't need to add browser specific rules because that is taken care of during the compile step.

tn3rb:
> re:  For reviews of the css system its extra important to keep generated commits separate from actual commits (and preferably the last thing in the commit history) so that the review can make sure there aren't manual edits being made to generated files

sure that sounds like a good rule of thumb to follow 

> re: I think we should use something other than the espresso prefix for classnames.

we've either been using no namespace on css classes or `ee-` for ages now so not sure how `espresso` is more likely to clash than those others. The reason I changed that button class to `espresso-button` is because there is an existing `ee-button` class used in our admin. Again, i don't see `espresso-button` being more likely to have conflicts than `ee-button`

> re: I'm not sure I agree with the hardcoded px or rem values in the generated variables. This gives less flexibility to local theme overrides that might want to use something other than px.

good point. Added unit properties to the theme config so that those can be overridden

>  using pixels for things in the web is generally not a good idea with the modern web moving to more fluent and flexible design

that's actually been the case for a long time (10 years maybe?) but the problem is that `rem` units, which afaik is the best choice for most things, is relative to the `html` element in the DOM and requires a font-size to be set on that element. If that is not the case, then it *may* not work as expected.

HA! Just tested this out as it's been ages since I've experimented with this, and it appears that browsers set the base font size correctly so that 1 rem === 16px which is exactly what we want, so I've changed font sizes over to use rem as well. We may encounter problems with people using old browsers, but those ppl need to learn a lesson!

PR #1259, opened 2019-05-29
move withFormContainerAndPlaceholder into components

tn3rb:
`withFormContainerAndPlaceholder` is exported on `@eventespresso/higher-order-components` but is using components from `@eventespresso/components` which creates a circular dependency because `@eventespresso/components` depends on `@eventespresso/higher-order-components`. 

To resolve, `withFormContainerAndPlaceholder` has been moved into the components module and components utilizing it have been updated.


assets/src/editor/events/ticket-assignments-manager/ticket-assignments-manager-modal.js
```diff
@@ -11,13 +11,13 @@ import {
 	CalendarPageDate,
 	ResponsiveTable,
 	twoColumnAdminFormLayout,
+	withFormContainerAndPlaceholder,
```
nerrad: no need to import this if it's being removed from the compose later?

tn3rb: it's not being removed from the compose,
it's being removed from the import statement from `'@eventespresso/higher-order-components'`

https://github.com/eventespresso/event-espresso-core/pull/1259/files/f4a3276003e0a679660f76fe6fcae9ed683651de#diff-7d99057378317300eaa982ecc3fb4b09L22

nerrad: oops ya I misread things.

PR #1261, opened 2019-05-30
update css to use new framework

tn3rb:
begins to convert existing components to use new css framework (not totally done)


tn3rb:
this was mostly cherry picked commits from my other branch so these changes were made prior to your comments on #1234 where you brought up the issue about namespacing. I can make another pass later to solidify classnames.

PR #1264, opened 2019-05-30
Filter out line items not related to the registration for registration details (fixes #1210)

### LINKED ISSUE #1210: Fix the 'Registration items' showing all tickets/line items related to a transaction.
If you have multiple tickets within a group, all of those tickets show on the 'Registration Items' table when editing a registration.

It's easier to see this so here are the steps to reproduce:

Create an event with multiple ticket types, it doesn't matter how many but its more obvious if you have a few. My setup used 6 (Ticket 1-5) and I added another to test the attendee mover (initially we thought it was an issue from that add-on, but turns out it's not).

Add at least 1 ticket from each of your tickets to a group, for example - https://take.ms/dPQWx

Register and pay for the group.

Now edit any of the 5 registrations created for that group and look at the 'Registration items' table it'll look something like this - https://take.ms/4FHVv

That's editing the registration for Ticket 4, so it shows a price for it, all of the others show 0.00

If you move a registration using the attendee mover to a ticket you hadn't selected before (this is where my additional ticket come into play) you end up with an additional ticket type shown on that list for every registration which can appear to be from no-where.

---

I did some digging into this and it looks like it's from [EE_Admin_Table_Line_Item_Display_Strategy.strategy.php](https://github.com/eventespresso/event-espresso-core/blob/master/core/libraries/line_item_display/EE_Admin_Table_Line_Item_Display_Strategy.strategy.php) looping over all of the children from a sub-total line item, it gets to a line item related to an event and its children are all of the ticket line items related to it.

Just noting in case its an obvious fix that I can't see in the above :)

**Expected behavior**
Display the line items only related to the current registration being edited?

nerrad:

See #1210 for details.  Essentially this pull filters out line items that are not related to the registration. Note:

- This should account for promotions line items (if their parent is either Event or Ticket)
- This should account for any line item objects that are tickets or events.
- Anything outside of those two parameters cannot be dynamically derived because there's no way for this section of code to know how the line item related object relates to the registration.

In all _current_ cases for the issue described, this should fix.

## How has this been tested

* [x] Verify the original reported issue is fixed
* [x] Verify the original reported issue is fixed for tickets purchased where a promotion applied.
* [x] Verify no regression on the issue from https://events.codebasehq.com/projects/event-espresso/tickets/5580, namely:
- Create an event with default reg status set to "Not Approved"
- Createt wo registrations for event with $1 tickets.
- TXN total should be $2
- Cancel one registration their registration and therefore can not / should not have to pay.
- TXN total should now be $1, not $2

## Checklist

* [x] My code is tested.


nerrad:
I wasn't aware of this filter so thanks for making me aware of it.  However from the comment block for that filter:

```
Filters line items to remove :
 *        subtotals with a total of 0
 *        line items with a quantity of 0
```

It doesn't look like I want to use this because the purpose of this ticket is to filter out line items _not related_ to the registration being viewed.  We still want to keep line items that _do_ belong to the registration (for instance, won't `EE_Non_Zero_Line_Item_Filter` filter out _free ticket_ line items for tickets related to that registration?

mnelson4:
> I wasn't aware of this filter

Yeah I don't blame you, I don't think they're documented pretty well anywhere and I think it was mostly Brent and I worked on them.

> It doesn't look like I want to use this because the purpose of this ticket is to filter out line items not related to the registration being viewed.

Right, the previously-applied line item filter, `EE_Single_Registration_Line_Item_Filter` took care of zero-ing out all the line items that applied to other registrations (ie, setting their quantity to zero), the filter `EE_Non_Zero_Line_Item_Filter` actually removes the line items.

> won't EE_Non_Zero_Line_Item_Filter filter out free ticket line items for tickets related to that registration?

Good question, but no, because they don't have a quantity of 0. It's assumed that you would have already used another filter like `EE_Single_Registration_Line_Item_Filter`, which the code does in this case. 

Here's a registration for a free ticket, which still appears fine on this branch: https://drive.google.com/a/eventespresso.com/file/d/1NQ5nlEpvAHpMim47F95krbTp6cHT--T8/view?usp=drivesdk.

Also I tried it with a promotion and a transaction for multiple tickets, and it also worked as hoped:  https://drive.google.com/a/eventespresso.com/file/d/1I5NmHgQRbywZoy_qLpxaFXc6LQL5018K/view?usp=drivesdk. 

Can you think of any other problem? If not, I could create a new PR with this fix if you like.

nerrad:
I'm also okay with you pushing commits directly to this pr Mike :)  Sounds like using the additional filter is _exactly_ what we want 👍

mnelson4:
I actually noticed that before https://github.com/eventespresso/event-espresso-core/commit/dfd3ba157ad we WERE USING `EE_Non_Zero_Line_Item_Filter`, but it was commented-out. I can't tell from the ticket, https://events.codebasehq.com/projects/event-espresso/tickets/5580, why that was done, other than describing an issue (that seems to have been resolved, but these changes don't cause an regression). I've added that ticket's issue as a testing item onto this ticket.

@tn3rb do you know if it will introduce any problems by re-adding `EE_Non_Zero_Line_Item_Filter`?

mnelson4:
Sorry to bother you Brent, but I think you need to give this a thumbs up before it keeps moving ahead because of what I mentioned in my previous comment.

tn3rb:
hmmm... ya I can't remember why `EE_Non_Zero_Line_Item_Filter` was commented out in that old ticket.
My guess is that it's because `EE_Admin_Table_Line_Item_Display_Strategy` doesn't display subtotals with no sub line items, so I thought the extra filter was unnecessary maybe?

I think testing the original issue as well sounds like a wise approach though, so if it fixes the new issue without any regressions, then super good call.

mnelson4:
(And thanks for chiming in Brent)

nerrad:
Looks good @mnelson4 ! thanks for doing the commits.  I agree that testing should cover whatever the previous codebase issue was supposed to fix.

joshfeck:
I tested this and found it works well, with no regressions. Ready for merging.

PR #1274, opened 2019-05-31
fix path to server-date-time

tn3rb:

discovered that the export path for the `ServerDateTime` VO was incorrect in the `/vo/index.js` file

## How has this been tested

* [ ] Verify that the `EventAttendees` block works as expected (and has the correct times for dates in the datetime selector as per the set site timezone).
* [ ] Existing unit tests for javascript should pass.


nerrad:
@tn3rb the primary issue exposed by failing unit tests was that the `DateTime.validateIsDateTime` function (`vo/date-time/datetime.js`) was using our `instanceOf` helper and only checking for the instance having a `constructor.name` matching `DateTime`.  I did do some investigation to see if I could get the parent prototype name (and thus adjust the `instanceOf` helper to assist with that) without much luck.  So for now, the trivial more naive fix is to just also check for `ServerDateTime` on the incoming instance. 

There's a couple of other more minor expectation fixes in the `test/base-entity.js`  test as well.

joshfeck:
@tn3rb Can you resolve the conflicts with master?

joshfeck:
I tested this and all is well.

PR #1276, opened 2019-05-31
Add a url scheme check to file_url() method

knazart:
`EE_PMT_Base::_set_file_url()` was using the `WP_PLUGIN_URL` without checking the URL scheme. On sites that are on SSL where the `siteurl` option was not updated to be https:// or on multisite installs the `EE_PMT_Base::file_url()` was returning non secure URLs. This PR fixes this.

Also fixes https://github.com/eventespresso/eventsmart.com-website/issues/446

## How has this been tested
This was tested on single and multi site WP installs with both: SSL on and off.
* [x] Activate EE on a clean WP install, go to the Payment Methods settings and check the Button URL option on PMs like Invoice and PayPal Express.
* [x] Check the button images on the registration checkout page.

## Checklist

* [x] My code is tested.


lorenzocaum:
I tested this with HTTPS and without and it set up the button URL which worked during registration checkout.

Please merge to master.

PR #1278, opened 2019-05-31
Escape All i18n Strings, Fixes #1275

### LINKED ISSUE #1275: Prepare all translation-ready strings with `wp_strip_all_tags()` before passing to `wp_localize_script()`
Need to review core (and add-ons) to make sure strings do not include tags before they're passed to wp_localize_script().

An example that needs changing:

`EE_Registry::$i18n_js_strings['links_disabled'] = __('All the links on this page have been disabled because this is a generated preview message for the purpose of ensuring layout, style, and content setup.  To test generated links, you must trigger an actual message notification.', 'event_espresso');`

can be changed to:

`EE_Registry::$i18n_js_strings['links_disabled'] = wp_strip_all_tags(__('All the links on this page have been disabled because this is a generated preview message for the purpose of ensuring layout, style, and content setup.  To test generated links, you must trigger an actual message notification.', 'event_espresso'));`

joshfeck:
see #1275 
## How has this been tested

- [ ] view transaction reports on a fresh site
- [ ] click links in message editor preview
- [ ] do a migration
- [ ] use the Event Espresso Maintanence > Reset page to reset something
- [ ] edit an event, but don't save and hit the back button to view the warning

## Checklist

* [x] My code is tested.


mnelson4:
Just noticed this ticket (when stalebot tagged it.) In light of [new documentation around translations](https://github.com/eventespresso/event-espresso-core/blob/master/docs/A--Best-Practices/php-standards.md#translations), why is it better to strip HTML tags from translated strings instead of just escaping it? Does this just apply to strings passed into `wp_localize_script`, or is it all translated strings?

joshfeck:
@mnelson4 This was something that Darren and I discovered when I fixed this [other related issue](https://github.com/eventespresso/event-espresso-core/pull/1277/commits/bcc4deb2965ee70f342395be7263055244737f47). It only really applies to strings passed to `wp_localize_script()`. If a translation that gets passed to `wp_localize_script()`, it has tags, and those tags get escaped, we found that the tags could get converted back to html in the rendering of the page. Which could be a problem for the JavaScript. So in this case it's safer to strip any tags out.

mnelson4:
Thanks for the explanation Josh. Ok so I've documented that https://github.com/eventespresso/event-espresso-core/blob/master/docs/A--Best-Practices/php-standards.md#translations for future reference

tn3rb:
rebased this onto `decaf-fixes/main` since that requires similar changes

knazart (approved):
👍

PR #1282, opened 2019-06-03
Fix broken master due to #1272

nerrad:

https://github.com/eventespresso/event-espresso-core/pull/1272 broke master because the Clean Webpack Build plugin was not only cleaning the build directory _once_.  The update of that plugin broke the fragile implementation using it.  So in this pull I've pulled it out in favor of a more direct approach to cleaning the directory on each build.

## How has this been tested

I rebuilt the dev builds and tested the following:

* [ ] plugin exit modal loads
* [ ] EventAttendeesBlock works 
* [ ] verify no console errors (related to our code)

## Checklist

* [ ] My code is tested.


bin/asset-dependency-maps.js
```diff
@@ -0,0 +1,48 @@
+const requestToExternal = ( request ) => {
```
nerrad: dang it, this is supposed to be in the other pull, I'm going to remove it

PR #1286, opened 2019-06-04
Add automatic dependency management/externals generation to build process 🎉(fixes #1052)

### LINKED ISSUE #1052: Automatic dependency setup for scripts/styles
I'm filing this issue so I don't forget it.  GB is going to be experimenting with having a webpack plugin generate a json file that can be read by php to automatically configure dependencies using the right script handles for various dependencies.  I _really_ like that idea as it reduces the chance for human error with manual dependency configuration.

The current issue to track is here https://github.com/WordPress/gutenberg/issues/14837
Related pull: https://github.com/WordPress/gutenberg/pull/14869

For our needs it will be complicated a bit because we have (I think) a fairly robust php asset management system but I think it's reasonable that we could have a `getDependenciesFor( 'script-handle' )` utility method that parses from this json file.

nerrad:

See #1052 for background.  This pull will make php dependency management for scripts and styles, and js webpack configuration for externals much easier (and less prone to human error).  

Two primary benefits introduced in this pull:

### Dependencies for scripts and styles are automatically registered php side.

With a few exceptions, you'll never have to touch php again for any current registered scripts as it's all handled automatically.  The exceptions are:

- Any script dependent on `eventespresso-data-stores`.  Since this is not imported in implementations, it doesn't get picked up by the dependency extraction.  So there may be a _few_ cases where it may need added as a dependency manually.  However, in most cases (since it, itself is dependent on `wp.data`) things should still get enqueued in the correct order.
- Any style dependent on `eventespresso-core-default-theme` will need to have that dependency added manually because it is not imported in any implementations (I have some ideas for improving this in future iterations, but its tricky so I'll leave out of this pull for now)

### Externals no longer need to be explicitly defined in the webpack config 🎉 

Externals are picked up automatically via the dependency extraction webpack plugin (and the configuration for it).  If any new packages are added they'll need to be updated in the configuration (I'll take care of doing this for the EDTR branch), but externals are automatically generated correctly for built files using the new webpack plugin.

To make all this possible, there were a number of corollary changes made in php files and js configuration including:

- new methods (php side) for getting the dependencies from the json files.
- making the chunk name strings and asset handle labels consistent (reduces the complexity of mapping).
- some work on making the webpack configs more dry.
- I also removed redux and react-redux from builds (making the vendor file much smaller) because we don't currently use them anywhere.

## How has this been tested

* [x] Builds should run fine
* [x] js unit tests should pass

This impacts all current released code using built files so:

* [x] Verify plugin exit modal works as expected
* [x] Verify the `EventAttendees` block works as expected.

## Checklist

* [x] My code is tested.


core/domain/services/assets/CoreAssetManager.php
```diff
@@ -585,17 +591,21 @@ private function registerAdminAssets()
...
         )
         ->setRequiresTranslation();
 
         $this->addStylesheet(
             CoreAssetManager::JS_HANDLE_WP_PLUGINS_PAGE,
-            $this->registry->getCssUrl($this->domain->assetNamespace(), 'wp-plugins-page')
+            $this->registry->getCssUrl($this->domain->assetNamespace(), CoreAssetManager::JS_HANDLE_WP_PLUGINS_PAGE),
+            $this->registry->getCssDependencies(
+                $this->domain->assetNamespace(),
+                CoreAssetManager::JS_HANDLE_WP_PLUGINS_PAGE
```
tn3rb: looks like this was an existing issue, but we are using a JS handle for a stylesheet asset

core/services/assets/Registry.php
```diff
@@ -449,6 +475,125 @@ public function getCssUrl($namespace, $chunk_name)
...
+            $css_chunks += [
+                'wp-components',
+                'wp-block-editor',
+                'wp-block-library',
+                'wp-edit-post',
+                'wp-edit-widgets',
+                'wp-editor',
+                'wp-format-library',
+                'wp-list-reusable-blocks',
+                'wp-nux',
```
tn3rb: are we certain this will always be the same for ALL CSS assets?

Should this be a static property at the top of the class for greater visibility and easier access if it requires editing?

webpack.prod.js
```diff
@@ -6,51 +6,89 @@ const webpack = require( 'webpack' );
...
+		output: path.resolve( __dirname,
+			'assets/dist/build-manifest.json',
+		),
+		assets: assetsData,
+	} ),
+	new miniExtract( {
+		filename: '[name].[contenthash].dist.css',
+	} ),
+];
+const pluginsConfigWithoutExtraction = const pluginsConfigWithExtraction = [
```
tn3rb: is this right?
this is redeclaring `pluginsConfigWithExtraction` ?!?!? (see L15 above)

core/domain/services/assets/CoreAssetManager.php
nerrad: ya I thought the same thing.  It's this way because the handles are the same.  I think we'll leave for now until/when the handles vary.  Might be good for cases like this to just have a more generic `ASSET_HANDLE_WP_PLUGINS_PAGE` const instead for cases where there's no difference between the css and js handle.  We can do in a separate pull.

core/services/assets/Registry.php
nerrad: Sure I can make it a static property at the top of the class for easier visibility.  Essentially this is just a map to _check_ against for dependencies.  It _will_ change if wp core has additional css handles.

This whole css dependency generation isn't ideal, the dependency extraction webpack plugin _only_ considers javascript imports.  However, since there is consistency in WP between javascript and css handle names (and we have consistency with EE handles), I'm able to infer (using this hardcoded map) what css dependencies might exist on a bundled file.

webpack.prod.js
nerrad: nope, not right (and travis caught it) will fix.

core/domain/services/assets/CoreAssetManager.php
```diff
@@ -475,17 +386,23 @@ private function loadCoreCss()
             CoreAssetManager::CSS_HANDLE_CORE_CSS_DEFAULT,
             $this->registry->getCssUrl(
                 $this->domain->assetNamespace(),
-                'core-default-theme'
+                CoreAssetManager::CSS_HANDLE_CORE_CSS_DEFAULT
             ),
             ['dashicons']
         );
         $this->addStylesheet(
             CoreAssetManager::CSS_HANDLE_COMPONENTS,
```
tn3rb: could this not use the new `addCss()` method you added?

core/domain/services/assets/CoreAssetManager.php
```diff
@@ -585,17 +502,21 @@ private function registerAdminAssets()
     {
         $this->addJavascript(
```
tn3rb: these could use `addJs()` no?

core/services/assets/BlockAssetManager.php
```diff
@@ -155,13 +155,19 @@ public function addEditorScript($handle, array $dependencies = array())
         if($this->assets->hasJavascriptAsset($handle)){
             return $this->assets->getJavascriptAsset($handle);
         }
-        return parent::addJavascript(
+        return $this->addJavascript(
```
tn3rb: these look like they could use the new helpers you added as well

core/services/assets/Registry.php
```diff
@@ -70,17 +74,58 @@ class Registry
...
+    /**
+     * @var DomainInterface
+     */
+    private $domain;
+
+    /**
+     * This is a known array of possible wp css handles that correspond to what may be exposed as dependencies in our
+     * build process.  Currently the dependency export process in webpack does not consider css imports, so we derive
+     * them via the js dependencies (WP uses the same handle for both js and css). This is a list of known handles that
+     * are used for both js and css.
```
tn3rb: good description 👍

core/services/assets/Registry.php
```diff
@@ -70,17 +74,58 @@ class Registry
...
-     * @param I18nRegistry    $i18n_registry
+     * @param AssetCollection      $assets
+     * @param I18nRegistry         $i18n_registry
+     * @param DomainInterface|null $domain
+     * @throws InvalidArgumentException
+     * @throws InvalidDataTypeException
+     * @throws InvalidInterfaceException
      */
-    public function __construct(AssetCollection $assets, I18nRegistry $i18n_registry)
+    public function __construct(AssetCollection $assets, I18nRegistry $i18n_registry, DomainInterface $domain = null)
```
tn3rb: hmmmm...

the `Registry` is used by ALL `AssetManagers` so have you investigated/considered how this would affect addons that are registering their assets under their own domain?

I looked through things and don't see any potential issues, but just want to be 150% sure this won't cause problems.

It looks like we are only using the domain for `distributionAssetsPath()` which likely will not change for addons, so I think this is fine for now as long as addons won't require their own bundled dependencies (which doesn't make sense).

thoughts?

nerrad: Great catch.  We discussed this in slack and arrived at the conclusion this injection is not needed because `ManifestFile` has the path exposed. See 15b8d47.

core/domain/services/assets/CoreAssetManager.php
nerrad: Yup, see 3cc025c

core/domain/services/assets/CoreAssetManager.php
nerrad: Yup, see 3cc025c

core/services/assets/BlockAssetManager.php
nerrad: Yup, see 3cc025c

nerrad:
I gave this a priority of high because the longer this lingers the greater the chance of conflicts.  Also, I'd like to merge this into the EDTR refactor work being done (which will already be a tricky merge).

lorenzocaum:
@nerrad 

I tested the Event Attendees block and that worked:

Image 2019-06-04 at 4 09 21 PM

I also tested the exit modal with the skip option and then the yes option:

Image 2019-06-04 at 4 10 20 PM

bin/asset-dependency-maps.js
```diff
@@ -0,0 +1,44 @@
+const requestToExternal = ( request ) => {
```
sirreal: 😎

nerrad:
> Very cool to see this, hope it's working well 😁

Hey! It's working very well! Thanks for creating the plugin.

Issue #1289, opened 2019-06-05
An NVP/SOAP API username from PayPal with a plus symbol will be rejected


lorenzocaum:
**Summary**
A customer has a plus symbol in their API username and this causes the credentials to be rejected in PayPal Express.

This works: `hello-facilitator_api1.example.com`

This causes an issue: `hello+facilitator_api1.example.com`

When I look in PayPal's sandbox, I can see that the email address is used as the API username so if the email address has a "+" then the API username will also get that plus symbol.

![Image 2019-06-05 at 8 42 36 AM](https://user-images.githubusercontent.com/1079138/58957090-e2564380-876d-11e9-9368-73e2349885ab.png)

**Steps to reproduce**
See above

**Reported or seen on**
https://secure.helpscout.net/conversation/866535185/653920/

**Other notes**
I suggested setting up PayPal Smart Buttons since it uses another set of credentials.

joshfeck:
The error shown in the validation message is:

`Your PayPal API credentials appear to be invalid. PayPal said "Security header is not valid (10002)". Please see tips below.`

tn3rb:
likely just needs a regex to filter the API username according to PayPal's requirements, but Michael or Nazar might know a better alternative

mnelson4:
Thanks Brent, but on the related PR, we discovered that even if we fix things on our end (don't double URL-encode the API credentials), PayPal itself has an internal server error when an API username has a plus sign in it. So although they haven't documented that it's illegal, their system blows up with it. So this seems to be a won't-fix type situation.

PR #1293, opened 2019-06-05
Don't double-URL-encode query parameters

mnelson4:

Don't URL-encoding the credentials sent to PayPal twice (doesn't actually fix https://github.com/eventespresso/event-espresso-core/issues/1289 because PayPal seems to have an internal error when we send an API username with a plus symbol in it)

## How has this been tested

* [x] Set up PayPal with a sandbox account that has a + in its API username, and try to pay with PayPal Pro. You should get a 10004 error (Internal error) instead of "Your PayPal API credentials appear to be invalid. PayPal said "Security header is not valid (10002)". While not a ton better, at least the problem is on PayPal's end, not ours.
* [x] Pay using an account that doesn't have a + in its API username
* [ ] Pay using an account that has other interesting but valid characters in the API username (eg maybe accented characters)

## Checklist

* [x] My code is tested.


mnelson4:
I hope it's ok I'm passing to you @lorenzocaum to test seeing how you originally reported you.

lorenzocaum:
Hey Mike, when I use the account that has the plus, then I get an 10004; Internal Error after making a payment.

When I swap to the account without the plus, then it works.

Do you see that 10004 on your end?

mnelson4:
Hi @lorenzocaum yeah actually I was getting that too (some details in the first testing note). I had my fingers crossed that it was just some problem with my setup but it sounds like it isn't. 
So, I think this branch solves the problem on OUR end, but it appears there's still a problem on PayPal's end.
So I think we'll just need to report this to PayPal somehow, I don't see how we can avoid passing the user's credentials... I'm pretty sure their response will primarily be "don't use plus signs in your email"...

lorenzocaum:
Hey Mike, since it is out of our control, then what about leaving it as-is?

I checked with the customer that mentioned this and they created another sandbox account to get around it. I also suggested PayPal Smart Buttons since they had access to that extension as well.

mnelson4:
I'd slightly prefer to include this change. It actually does appear to fix the issue ON OUR END (but it appears PayPal still chokes on plus signs, for PayPal Express and PayPal Pro too). It's possible that URL encoding the credentials TWICE may cause problems with other characters too (but not I haven't confirmed that). The only downside is that it takes a little bit of effort to test for regression, and it is possible there's some problem with this I don't see. But in my best judgment, this looks like an improvement. 

And it might be good to document that having plus signs in account IDs is a problem. 

Or we could do all the following only after this gets reported again, and is shown to not be an isolated issue.

My suggestion is to move ahead with this (just test it doesn't cause any regressions).

Thoughts @lorenzocaum, @joshfeck ?

joshfeck:
Yes I think you've identified a problem that needs to be fixed (instead of the double encoding, it should really be encoded once).

mnelson4:
Thanks Josh. So ya @lorenzocaum I say let's just verify this doesn't cause any regressions with what's already working (see testing notes)

mnelson4:
Hey @lorenzocaum I've updated the testing notes to reflect the new expectation that this (unfortunately) only switches the error from a 10002 error (bad credentials, ie our mistake) to a 10004 error (internal server error, ie a problem in PayPal). So if you get a chance to assert the testing notes are passing, we can merge this into master.

lorenzocaum:
Hey Mike, I tested with a plus symbol and saw the 10004 error.

Then I tested without it and the payment worked.

On the last item, when I try to set up a new account with an accented character, then it doesn't let me save the account:

Image 2019-07-15 at 2 46 27 PM

I tried a few accented characters. I think we can move forward and merge this in.

PR #1302, opened 2019-06-07
allow for EE_State_Inputs to use state abbreviations or names

mnelson4:

Adds support for `EE_State_Input`'s option keys to use the state's name or abbreviation, in addition to ID (the default). Previously, you could provide the options array directly, but when you use the "add state" module, and it uses Javascript to add your new state to the list of options, it would still use the state's ID. This fixes that issue by providing metadata to the Javascript so it knows whether to use the state's ID, abbreviation, or name, as the HTML option tag's "value" attribute.
These changes are needed for the upcoming upgrade to EE4 Stripe for using Payment Intents/Elements.

## How has this been tested

* [x] Add address info question group to an event and register for it. The chosen state should be saved fine.
* [x] Repeat the above, but add a new state. It should work like on master
* [x] Pay using an onsite paymetn method that has a state input. The state should be chosen fine.
* [x] Repeat the above but add a new state in the BILLING form

This was also tested with https://github.com/eventespresso/eea-stripe-gateway/pull/14, which requires this.

## Checklist

* [x] My code is tested.


modules/add_new_state/assets/add_new_state.js
```diff
@@ -206,7 +208,13 @@ jQuery(document).ready(function($) {
 					//SPCO.console_log( 'select_id', select_id, false );
 					var set_selected = $(this).attr('id') === state_select_id;
 					// set target select's value to this input's value
-					EE_ANS.add_option_to_dropdown( select_id, new_state.id, new_state.name, set_selected, new_state.country_name );
+					var new_state_option_value = new_state_option_value = new_state.id;;
```
tn3rb: copy pasta???

the var name and semicolons are duplicated

core/libraries/form_sections/inputs/EE_State_Select_Input.php
```diff
@@ -12,10 +12,20 @@
...
 {
+    /**
+     * @var string the name of the EE_State field to use for option values in the HTML form input.
+     */
+    protected $valueFieldName;
 
     /**
-     * @param array $state_options
-     * @param array $input_settings
+     * @param EE_State[]|array|null $state_options. If an flat array of string is provided,
```
tn3rb: should be:

> If a flat array

modules/add_new_state/assets/add_new_state.js
mnelson4: 👽 that's weird. Ugh thanks!

core/libraries/form_sections/inputs/EE_State_Select_Input.php
mnelson4: 💰

mnelson4:
Thanks for catching those issues @tn3rb, passing off for another review.

mnelson4:
ok passing off for triage/testing.

mnelson4:
noticed something else was needed so passin goff for re-review

mnelson4:
Passing off for triage/testing (I was delayed because there was an issue on eventespresso/eea-stripe-gateway#14 and I wanted to confirm this wasn't the cause; it wasn't.)

joshfeck:
The Authnet Accept gateway is broken with this branch (if a State field is in the reg form)

```
[2019-06-28 16:45:37]  Exception Details
Message: Subsection &#039;state&#039; is not an instanceof EE_Form_Input_Base on form &#039;EventEspresso\Authnet\Accept\forms\AcceptBilling&#039;. It is a &#039;NULL&#039;
Code: EE_Form_Section_Proper - get_input - 513
File: /home/eeservice/www/josh.eventespresso.com/wp-content/plugins/event-espresso-core/core/libraries/form_sections/base/EE_Form_Section_Proper.form.php
Line No: 513
```

mnelson4:
good catch @joshfeck! I managed to reproduce just by changing the AIM payment method settings to EXCLUDE the state input, then got that fatal error as soon as the billing form was being displayed. This latest commit fixed it for me.

joshfeck:
retested and all is OK now

PR #1303, opened 2019-06-10
remove single quotes from line items and descriptions sent to AIM

mnelson4:

If you had a ticket name or description that had a single quote in it, it would cause an error making a payment with AIM. In order to resolve this, single quotes are removed when sending ticket names and descriptions to AIM.
FYI: double quotes, accented characters, and all the other special characters I could think to use did NOT cause problems.

## How has this been tested

* [x] Create an event and ticket and put all the crazy characters you can think of into their name and descriptions (eg  `J'ai été là déjà!@#$%^&*()"`). Activate AIM. Register for that event and pay with AIM. It should work fine.
## Checklist

* [ ] My code is tested.

### DIFF
--- a/caffeinated/payment_methods/Aim/EEG_Aim.gateway.php
+++ b/caffeinated/payment_methods/Aim/EEG_Aim.gateway.php
@@ -215,7 +215,7 @@ class EEG_Aim extends EE_Onsite_Gateway
                     $line_item->unit_price(),
                     'N'
                 );
-                $order_description .= $line_item->desc().', ';
+                $order_description .= str_replace('\'', '', $line_item->desc() . ', ');
             }
             foreach ($total_line_item->tax_descendants() as $tax_line_item) {
                 $this->addLineItem(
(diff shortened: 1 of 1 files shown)


caffeinated/payment_methods/Aim/EEG_Aim.gateway.php
```diff
@@ -446,6 +447,21 @@ private function _log_and_clean_response($response_obj, $payment)
...
+     * @param $text
+     * @return string
+     */
+    private function prepareStringForAuthnet($text)
+    {
+        return str_replace(
+            ['\''],
+            [''],
+            $text
+        );
```
knazart: Not critical, but do we want to sill keep the `$search` and `$replace` as arrays with only one value in each ? Also, a bit nit-picky, but this `str_replace()` could fit in one line.

mnelson4: oh ya, the arrays aren't necessary. Thanks

PR #1308, opened 2019-06-11
Add parameter onto RequestInterface::requestUri to get uri relative to WP root

mnelson4:

Fixes the Basic Auth not authenticating requests to subdirectory multisite installations, and WordPress in subdirectories. 
Basic Auth wasn't getting loaded on REST API requests when the site (or current blog on Multisite) had its home URL in a subdirectory (eg "http://mysite.com/wp/") because we were comparing the REST Prefix (usually "wp-json") to the request's literal URI (continuing the previous example, it would have been "wp/wp-json"). Whereas we should have been removing the site's path before comparison.
Introduces a new parameter `$relativeToWpRoot` onto \EventEspresso\core\services\request\RequestInterface::requestUri()` to indicate whether you want the full path/uri (the default, the old behaviour) or the full/uri relative to WordPress' home URL (or the current site's home URL).

## How has this been tested

* [x] Setup a subdirectory multisite install. Try to use the mobile apps for a subsite . They should work fine. (Or if you don't want to setup a publicly accessible site like that, then a local one will do, and just check that you can use Basic Auth to authenticate a REST API request like "GET http://localsite.test/subsite/wp-json/ee/v4.8.36/registrations")
* [x] Use the mobile apps from a regular site, they should also work fine.
* [ ] Setup a single-site WordPress install in a subdirectory. Verify you can use the mobile apps with it 

## Checklist

* [x] My code is tested.


mnelson4:
Hmm ya having it be a parameter would be good too... Although I'm also realizing I'm not sure if we're accounting for single-site WordPress being in a subdirectory... going to test that...

mnelson4:
Yeah I just tried it on a local site with WordPress in a subdirectory and it had this same problem. Let's see if we can fix both problems while we're at it...

mnelson4:
The latest fix also works with single-site WordPress in a subdirectory.

tn3rb:
> The latest fix also works with single-site WordPress in a subdirectory.

ooo nice catch 👍

mnelson4:
:sunglasses: Passing off for testing

joshfeck:
Tested auth'd requests to /registrations endpoint on a subdirectory Multisite and verified the fix. Also tested the apps on regular WP site and single WP site in a subdirectory.

PR #1312, opened 2019-06-12
Event list shortcodes should allow 'Organization' shortcodes.

Pebblo:
Long story short, if you use most of the Organization shortcodes in the 'Event List' section of a 'Registration' based message type (Reg Approved), most of them don't work.

For example `[CO_EMAIL]` and `[CO_PHONE]` works but `[CO_ADD1], [CO_CITY], [CO_STATE], [CO_ZIP]` just returned the shortcode.

I think (can't follow some of how it's parsing) the reason some of them worked is just that other sections had the `[CO_EMAIL]` and `[CO_PHONE]` already parsed, so the shortcode/values arrays in EEH_Parse_Shortcodes.helper.php had the values already.

This just adds 'organization' shortcodes to the valid shortcodes on the Event List shortcode library.

## How has this been tested
Add [CO_ADD1], [CO_CITY], [CO_STATE], [CO_ZIP] to the registration approved event list section.

With master they don't work, with this branch they do.

## Checklist

* [x] My code is tested.


Pebblo:
Thanks @nerrad,

Originally I thought that was the case, but CO_PHONE (which worked for me) is only parsed in `EE_Organization_Shortcodes`, that shortcode is already parsed elsewhere in main content so I figured it was from that, but not entirely sure.

Pebblo:
Tested this and confirmed it fixed the issue for me.

I'll leave this for Josh to double check, he's already aware of the issue as he's the one that found it :)

joshfeck:
This works, but I did run into a snag when testing. So before this was fixed, if you had a`[CO_STATE]` shortcode set, but no Organization State selected, it'd parse to the shortcode. Now it throws a fatal error:

```
Fatal error: Uncaught Error: Call to a member function name() on null in /srv/www/wordpress-default/public_html/wp-content/plugins/32-core/core/libraries/shortcodes/EE_Organization_Shortcodes.lib.php:88 Stack trace: #0 /srv/www/wordpress-default/public_html/wp-content/plugins/32-core/core/libraries/shortcodes/EE_Shortcodes.lib.php(184): EE_Organization_Shortcodes->_parser('[CO_STATE]') #1 /srv/www/wordpress-default/public_html/wp-content/plugins/32-core/core/helpers/EEH_Parse_Shortcodes.helper.php(234): EE_Shortcodes->parser('[CO_STATE]', Object(EE_Messages_Addressee), Array) #2 /srv/www/wordpress-default/public_html/wp-content/plugins/32-core/core/helpers/EEH_Parse_Shortcodes.helper.php(74): EEH_Parse_Shortcodes->_parse_message_template() #3 /srv/www/wordpress-default/public_html/wp-content/plugins/32-core/core/libraries/messages/EE_Messages_Generator.lib.php(732): EEH_Parse_Shortcodes->parse_message_template(Array, Object(EE_Messages_Addressee), Array, Object(EE_Registration_message_type), Object(EE_Email_messenger), Ob in /srv/www/wordpress-default/public_html/wp-content/plugins/32-core/core/libraries/shortcodes/EE_Organization_Shortcodes.lib.php on line 88```

If there's a selected state in Event Espresso > General Settings, everything is fine. Also, all other Organization shortcodes seem fine if you use them but they have no values, they just parse to an empty string with this branch. 

Since there's a chance someone has a [CO_STATE] shortcode somewhere and doesn't realize it's not parsing, how about we fix the fatal error in this PR too?

**Comment by nerrad** [comment]
Just removed my assignment because this should be something Tony can fix.  Tony if you need my help feel free to pull me back in :)

**Comment by joshfeck** [comment]
@nerrad I was kind of hoping you could do a quick review on the fix I pushed.

File: core/libraries/shortcodes/EE_Organization_Shortcodes.lib.php [hunk 1]
```diff
@@ -85,7 +85,7 @@ protected function _parser($shortcode)
                 $state = EE_Registry::instance()->load_model('State')->get_one_by_ID(
                     EE_Registry::instance()->CFG->organization->STA_ID
                 );
-                return $state->name();
+                return ! empty($state) ? $state->name() : '';
```
nerrad: Would be better here to do:

```
return $state instanceof EE_State ? $state->name() : '';
```

sethshoultes:
@Pebblo @joshfeck  I am seeing some weird issues with this on MENW. I created a new issue here:
https://github.com/eventespresso/eventsmart.com-website/issues/543

Any ideas?

joshfeck:
I checked on a locally hosted site and the same shortcodes parse as expected there. I've followed up with an idea in the [other issue](https://github.com/eventespresso/eventsmart.com-website/issues/543).

PR #1313, opened 2019-06-12
EDTR REM Tweaks n Fixes

tn3rb:
some more tweaks to core EDTR code for REM as well as some fixes for various minor issues after recent EDTR refactors

## How has this been tested


assets/src/components/ui/icon-menu-item/entity-action-menu-item.js
```diff
@@ -0,0 +1,17 @@
...
+/**
+ * EntityActionMenuItem
+ * just a wrapper for an IconButton and additional component
+ * (most likely a modal component) so that it can be added to a menu
+ */
+class EntityActionMenuItem extends Component {
+	render() {
+		return this.props.children;
+	}
+}
```
nerrad: hmm... this would be better as a functional component:
```jsx
const EntityActionMenuItem = ( { children } ) => children;
```

What is the use-case for this though (i.e. how would it be implemented?  It's not clear from the description or the code why it's needed.  It "appears" it's just used for conditional checks?

I'm probably going to be implementing the slot/fill pattern for the action menu as well (which will work a bit better for extensibility than the current filter hook method).

assets/src/editor/events/dates-and-times/editor-date/edit-form/date-entity-form-submit-handler.js
```diff
@@ -41,4 +40,5 @@ export const dateEntityFormSubmitHandler = (
 	dateEntity.order = parseInt( getValue( 'order', 0 ), 10 );
 	dateEntity.parent = parseInt( getValue( 'parent', 0 ), 10 );
 	dateEntity.deleted = !! getValue( 'deleted', false );
+	return dateEntity;
```
nerrad: Why return the `dateEntity` here?  Entity instances are mutable so you get back the same instance?  Returning gives the illusion that there's a different value being returned when that's not the case.

assets/src/editor/events/dates-and-tickets-metabox/with-event-date-and-ticket-entity-filter-state.js
```diff
@@ -91,7 +91,7 @@ const withDateAndTicketEntityFilterState = createHigherOrderComponent(
 			}
 			return <WrappedComponent
 				allTicketEntities={ allTicketEntities }
-				ticketEntitiesLoaded={ ticketEntitiesLoaded }
+				loadingTicketEntities={ ! ticketEntitiesLoaded }
```
nerrad: Just a sanity check - have you verified this change in props is propagated throughout any children?  A simple thumbs up reaction is sufficient.

assets/src/editor/events/dates-and-times/editor-date/edit-form/date-entity-form-inputs.js
```diff
@@ -17,21 +17,20 @@ const { MODEL_NAME: DATETIME } = dateTimeModel;
...
 	exclude = [],
-	currentValues = [],
+	values = {},
 	FormInput,
 ) => {
 	const inputs = [];
 	if ( ! isModelEntityOfModel( dateEntity, DATETIME ) ) {
 		return inputs;
 	}
-	const values = currentValues;
```
nerrad: ya this makes sense :)

assets/src/editor/events/tickets/editor-ticket/editor-ticket-entities-list.js
```diff
@@ -132,29 +153,17 @@ export default compose( [
...
+/**
+ * a trimmed down ticket list: only includes ticket price calculator
+ */
+export const EditorTicketEntitiesOnlyList = compose( [
+	withPriceTypeEntities,
+	withNewTicketEntity,
+	withTicketPriceCalculatorFormModal,
+	withTicketEntityFormModal,
+	createNewTicketEntity,
+] )( EditorTicketEntitiesList );
```
nerrad: Would be helpful in the review to know why this is needed.

assets/src/vo/date-time/datetime.js
```diff
@@ -294,7 +294,7 @@ export default class DateTime {
 	 * Given an indefinite number of DateTimes as arguments, this will return a
 	 * new DateTime that represents the latest point in time.
 	 * @param {...DateTime} datetimes
-	 * @return {DateTime} A new DateTime representing the latest point of time.
+	 * @return {DateTime|ServerDateTime} A new DateTime representing the latest point of time.
```
nerrad: Changes in the VO file should be made in a pull off master otherwise you increase the risk of conflicts for work in other branches off master.

assets/src/components/ui/icon-menu-item/entity-action-menu-item.js
tn3rb: ya I severely want to limit what can be injected into the entity action menus.
basically we only ever want icon buttons, but even amongst the WP core components there are a few that are viable (like the drop down menu), so I unfortunately can't limit it to just `IconButton`s.
This is essentially just a wrapper that can be used, which I realize doesn't actually prevent someone from injecting anything they want.
I've considered using this as a parent class that others could inherit, but that's somewhat frowned upon in React in favour of composition, which doesn't really apply in this use case.

I'm open to other ideas for improving this

nerrad: Yup I grok the desire to limit what can be added to this area.  I think we could explore some further restriction ideas when I get to the slot/fill extension task.

assets/src/editor/events/dates-and-tickets-metabox/with-event-date-and-ticket-entity-filter-state.js
tn3rb: yes in fact the child components were expecting an incoming prop named `loadingTicketEntities` but were not getting it.
I could have changed the children to use `ticketEntitiesLoaded ` instead, but this change makes it work exactly the same way as the corresponding date entity list and filter state HOC.

assets/src/editor/events/tickets/editor-ticket/editor-ticket-entities-list.js
tn3rb: I need to display a list of tickets in REM but don't want the full kit n kaboodle (ya whipper snapper)

assets/src/vo/date-time/datetime.js
tn3rb: oh dang it... right...

assets/src/editor/events/tickets/editor-ticket/editor-ticket-entities-list.js
nerrad: hmm, I'm wary of exporting components like this specifically for add-ons to use (what if another add-on needs something different?) and instead just export the individual parts so that other add-ons can build their own composed component as necessary?

nerrad: So with my previous comment, this is an example of something that could go in its own pull so we could discuss what api we want to expose for add-ons (third parties) to use here (and allow you to get the other changes in this pull merged quicker).

tn3rb: or fix it later so I can continue on REM

nerrad: sure, will you create a card in the project/issue outlining what you did here, why you did it, and what's needed for REM?  I **did** see the inline comment with the code but:

- It doesn't explain _why_ this separate export, only _what_ it is. 
- It's not really accurate, it contains more than the ticket price calculator.
- As far as I can tell the only thing missing from the composition is the paginated filters?  If that's the case, then a possible option here is that the paginated filters are _added_ in somewhere else?  Maybe its an indicator that the structure of the components needs modified (if a common pattern/usage for this component will be without paginated filters?)

Anyways, I'm not going to comment on this further, your call what happens.

PR #1325, opened 2019-06-13
@eventespresso/data: switch to normalize store state to always use singular form of model names

nerrad:

While using the data store in various things in progress with development, I've realized that internally the store shape of the data store has unnecessary complexity with always have "relation" references in the plural.  So this pull normalizes all references to model names to the singular form.

This doesn't affect any of the external api so there's no impact to released code, it's mostly just a simplification of the internal code (which should have adequate coverage on unit tests).

## How has this been tested

Released EE core features are not affected by these changes.

* [ ] Unit test covers the changes in this pull.

## Checklist

* [ ] My code is tested.


assets/src/data/eventespresso/core/resolvers/relations.js
```diff
@@ -67,8 +67,9 @@ export function* getRelatedEntities(
 	if ( entity.isNew ) {
 		return DEFAULT_EMPTY_ARRAY;
 	}
-	const modelName = entity.modelName.toLowerCase();
+	relationModelName = singularModelName( relationModelName );
 	const pluralRelationName = pluralModelName( relationModelName );
```
tn3rb: was this missed? we don't want plurals anymore right? or are they still used in some places?

assets/src/data/eventespresso/core/resolvers/relations.js
```diff
@@ -67,8 +67,9 @@ export function* getRelatedEntities(
...
+	const modelName = entity.modelName.toLowerCase();
 	const relationResourceProperty = pluralRelationName + 'Resource';
```
tn3rb: here's another plural that escaped the purge, although this one looks more legit

assets/src/data/eventespresso/core/resolvers/relations.js
```diff
@@ -250,18 +249,17 @@ export function* getRelatedEntitiesForIds(
 	if ( ! response.length ) {
 		return DEFAULT_EMPTY_ARRAY;
 	}
-	const relationPrimaryKey = getPrimaryKey(
-		singularModelName( relationName )
-	);
-	const modelPrimaryKey = getPrimaryKey( singularModelName( modelName ) );
+	const relationPrimaryKey = getPrimaryKey( relationName );
+	const modelPrimaryKey = getPrimaryKey( modelName );
+	const pluralRelationName = pluralModelName( relationName );
```
tn3rb: ok this looks very intentional

assets/src/data/eventespresso/schema/resolvers.js
```diff
@@ -231,8 +237,8 @@ export function* getRelationType( modelName, relationName ) {
  */
 export function* getRelationSchema( modelName, relationName ) {
 	modelName = singularModelName( modelName );
+	relationName = singularModelName( relationName );
 	const pluralRelationName = pluralModelName( relationName );
```
tn3rb: this ok? just making sure nothing was missed

assets/src/data/eventespresso/core/resolvers/relations.js
nerrad: No not missed.  In this case its needed to extract the relation resource from the endpoint (see two lines after this).

assets/src/data/eventespresso/core/resolvers/relations.js
nerrad: see two lines up.

assets/src/data/eventespresso/core/resolvers/relations.js
nerrad: Yes, a relation record may be in the response in its plural form.  This is an example where the rest api is "inconsistent" (in quotes, because from a certain perspective this might actually be seen as consistent).  Relation records might be singular or plural depending on they type of relation.

assets/src/data/eventespresso/schema/resolvers.js
nerrad: Here it's okay too because a few lines below, it's being used to potentially retrieve the schema for the relation from the schema properties.  Similar to when relations are included in a response, the schema response may have relation records indexed by the singular form or the plural form of the relation.

PR #1327, opened 2019-06-13
Include all tickets when performing a sold out check.

Pebblo:
When checking if an event has spaces available, we need to include sales from all tickets, including any that have been archived.

This pull just changes the query to pull all tickets with no `default_where_conditions`.

## How has this been tested
Create an event with a datetime limit of 8.

Create a ticket with a limit of 8, then register onto 4 of that ticket (you can do so in a single group).

Now edit the event again and change the price of the ticket, the original ticket will be archived and you'll have another with a qty of 4 available. (I created an additional ticket here with a qty of 4 again).

Update the event and register onto the remaining 4 tickets on your one of your new tickets (again this can be done with a group of 4 tickets, no need for singles).

With master, when you view the event it will still have a status of upcoming.
The datetime will show sold out - https://monosnap.com/file/Q0mb2xMG37m8rFgcCWUxxc60oI7eeX
Yet the event status doesn't update, even if you hit update on the editor.

Do the same with this branch and the event status will update to sold out.

## Checklist

* [x] My code is tested.


joshfeck:
@Pebblo I actually cannot replicate this issue on master.

Screen Shot 2019-06-17 at 8 55 47 PM


Is there another step required to reproduce the issue the PR is intended to fix?

Pebblo:
Apologies Josh I missed that you need another ticket on that event for this to show.

You basically need a combination of sold tickets + archived ticket sales to be equal to the datetime limit.

Your getting set to sold-out there because the only ticket it's using in the calculation is sold out, it's **not** using the archived ticket and working out there are 8 paces taken but has calculated 0 spaces available on the only 'active' ticket for the datetime, meaning its sold out.

Add another ticket to the event (duplicate the above if you wish) and update, your event will update to upcoming even though your datetime is still sold out.

Before - https://monosnap.com/file/Lka2rYXs5PFcPhOsFJIdDU1F3S7k2V

After - https://monosnap.com/file/F7kOiFLXYz4owzuwg1CFODV2vPuZdv

Switch to this branch and update again - https://monosnap.com/file/CQiQ92fVXjSSoYs0mxHuMRdv1gNjca

joshfeck:
I got it now!

PR #1340, opened 2019-06-15
More EDTR improvements and fix bug with Ticket Assigments.

nerrad:

> **Note:** I know I've been harping on keeping things in smaller pulls when possible, but this is one of those pulls where the changes are inter-related for the type of refactor happening so we end up with something that actually works :).  There's still a number of things I want to do in later iterations that _will_ be done in separate pulls.

This pull started out as a fix for a Ticket Assignments manager modal bug where if you used a search term for datetimes, the tickets for the datetime results would not update in the tickets filtered view (when chained).  However, once I started getting into the bug fix, I hit a wall again with the way the state was being maintained in the app and the props being passed through components.  So I ended up switching gears in this pull to add additional improvements to state flow and usage in the editor which not only fixed the bug that started this but also will likely end up closing a number of the open cards in the [project](https://github.com/eventespresso/event-espresso-core/projects/4).

## Motivations for the changes in the pull:

- awareness of the number of issues related to tricky state and prop flow in the app.
- planning for eventual/easier hydration from the server.
- clearer encapsulation of state with specific components needing it so prop usage is cleaner.
- elimination of duplicated logic in various places.
- some improvements in the number of re-renders that happen because of prop changes (however, I think with these changes it will be easier to add performance improvements in future iterations). Some examples of things that improved here are: `withTicketEntitiesForAllDateEntities` now only processes once and thus there's no re-rendering of `DateAndTicketMetabox` everytime state changes;  filteredTicketEntities are now only affecting the ticket entity list (and vice versa for the `filteredDateEntities`)... so there's less frequent re-rendering of either of those entity lists when filtering changes happen (or there are changes to the state).

### Introduction of an `initialization` layer.

When the editor first loads, there's an initial state that needs setup in the store in order for it to be usable.  Notably the following needs to happen:

- Event Entity retrieved from server and added to state.
- Related Date entities to that event retrieved from server and added to state.
- Related Ticket entities to all the loaded dates retrieved from the server and added to state.

Once the above entities are setup in the state, the selectors used to retrieve those things should never have to be invoked again and all further work happens with what is in the state (including what might get created and added to state _via in the client_).

So the initialization layer has been setup to take care of the initial initialization and notifying the app when things are initialized.

Everything new related to the initialization layer is found in `src/editor/events/initialization`

#### Benefits

- `withEditorInitialization` can be wrapped on any editor component and then the wrapped components can use the `useEditorInitialized` hook to receive whether the editor has been initialized or not (see `DatesAndTicketMetabox` using this hook).  This will have great utility for add-ons that inject their own ui/ux into the editor and they want to only start up when the editor has been initialized or for the eventual work we do on the entire editor where we may want various ui/ux paused until the editor is initialized.
- By abstracting this into its own layer, we can have more control over what defines "initialized" state.  Right now it's a simple boolean, but this could be iterated on in the future to potentially expose an object that has various initialization states exposed.
- With this being in its own layer, we can now also use it for a future iteration where hydration of the state is provided from an object sent via the server on page load.  While it will likely be _much_ faster than a bunch of rest queries, there still will likely be some delay on initial load to parse the object from the server and ensure the state get's setup correctly.
- A future iteration can introduce the ability to more easily "reset" initialization state.  Notably this _may_ happen after persistence to the server (but we may find we don't need to do this).
- Future iterations introducing `undo/redo` state have a clear point from which to start tracking changes.

### Introduction of `src/editor/events/hocs` folder and moving of various editor related hocs into this folder.

Previously, there were a number of `data` folders in various subfolders found in the `editor/events` path that contained hocs related to the subfolder tree.  However, with this refactoring, hocs have more utility at a subtree level in various paths so it is much easier to have main hocs located in one folder.

As far as naming schema, if I included `editor` in the name of the hoc, it generally means that it's fairly specific to the `editor` context and thus should only be used when the editor has been initialized.  Whereas, if "editor" is not in the name of the hoc, that means that it is something that might still interact with the server to retrieve entities or is not reliant on the store state being initialized. For example, `withEditorDateEntities` is used to retrieve all the date entities currently in the store state and thus it is something that would really only be called in the context of the editor.  Whereas `withPriceTypeEntities` will retrieve all price type entities from the server if the server hasn't been queried yet.

#### Benefits

- Main hocs are easily discoverable and more easily imported.
- Makes it a bit clearer what hocs might be things we can extract to a more top-level package at some point (for example `@eventespresso/editor-hocs` or `@eventespresso/hocs`).
- less folder structure to maintain.

### Tracking filtered entities in the `eventespresso/filter-state` store.

A large part of the reason for the prop pass thru happening on components was because of the delicate dependency on the filters being provided by the filter hocs.  What I did was started storing the filtered results (entity ids) in the `filter-state` whenever filter state changes, and this allows for targeting the various applied filters to different trees independently (relying on filter-state for updates).

#### Benefits

- less need for a tighter tree structure.  Filter state is independent so this allows for various independent components to update from changes happening in the filter state.
- ^^ could have immediate benefit for add-ons that want to react to how things are filtered in the main entity lists outside of those entity lists.

## How has this been tested

Testing has involved mostly just trying all the various editor behaviour and ensuring things work without errors (or any "unknown" errors popup).  On the whole things seem to work really well and this does fix the ticket assignments issue originally prompting this ticket.  The reproduction steps for the ticket assignments bug was:

- Set things up so you have at least two dates, and two tickets which each ticket assigned to a unique date. Ensure the dates have names that are unique.
- In the date filter bar, type a search for one of the dates.
- The only ticket that should show in the ticket entity list (when chained) is the ticket(s) attached to the date matching the search in the date entity list.

## Note about possible breaking change for REM.

In `editor-ticket-entities-list` there had been some work you did @br3nt to export two versions of the EditorTicketsList.  One with the filterbar, and one without.  Since I'm still not sure how this was implemented in REM, it's likely that the "trimmed down ticket list" won't work as expected after these changes.  So we may need to collaborate on what REM needs are exactly so I can suggest an approach to take (if one isn't already apparent to you after reviewing this pull).


assets/src/editor/events/dates-and-tickets-metabox/dates-and-tickets-metabox.js
```diff
@@ -96,13 +80,8 @@ const MetaBox = ( {
 							className="ee-editor-event-tickets ee-form-row">
 							<div>
 								<EditorTicketEntitiesList
-									loading={ ! ticketEntitiesLoaded }
-									entities={ filteredTicketEntities }
-									allDateEntities={ allDateEntities }
-									isChained={ isChained }
-									prefiltered
+									loading={ ! initialized }
```
nerrad: I realize this is a change in that now both entity lists will show the loading indicator until all dates and tickets are initialized from the server.  In a future iteration we can look at making the initialized context value more granular so we can stagger the loading.  However, I want to hold off doing that until I introduce the hydration mechanism to the initialization layer to see whether we need the granular initialization state exposed.

assets/src/editor/events/dates-and-times/editor-date/actions-menu/editor-date-entity-actions-menu.js
```diff
@@ -92,8 +92,8 @@ class EditorDateEntityActionsMenu extends Component {
 	 * @param {boolean} ticketEntitiesLoaded
 	 * @return {IconMenuItem}    View Tickets for Event Date IconMenuItem
 	 */
-	viewTicketsMenuItem = ( dateEntity, ticketEntities, ticketEntitiesLoaded ) => {
-		const tooltip = ticketEntitiesLoaded && isEmpty( ticketEntities ) ?
+	viewTicketsMenuItem = ( dateEntity, ticketEntities ) => {
```
nerrad: `ticketEntitiesLoaded` is no longer needed because this component won't be mounted until the editor is initialized.  So we can just use the `ticketEntities` check here.

assets/src/editor/events/dates-and-times/editor-date/filter-bar/with-date-entities-list-filter-bar.js
```diff
@@ -44,74 +46,85 @@ export const getFilteredDateEntitiesList = ( dateEntities, showDates, datesSorte
...
 				);
-			}
-		};
+			},
+			[
+				dateEntities,
+				searchDateName,
+				showDates,
+				datesSortedBy,
+			]
```
nerrad: This ensures that we're only updating the filteredDateEntities value when any of the above props changes, which reduces re-renders on the entity list.

assets/src/editor/events/dates-and-times/editor-date/filter-bar/with-date-entities-list-filter-bar.js
```diff
@@ -44,74 +46,85 @@ export const getFilteredDateEntitiesList = ( dateEntities, showDates, datesSorte
...
+		);
+		// whenever filtered entities changes let's update the date ids
+		// in the state.
+		useEffect( () => {
+			setFilteredDateEntities(
+				filteredEntities.map( ( dateEntity ) => dateEntity.id )
+			);
+		}, [ filteredEntities ] );
```
nerrad: Whenever the filtered date entities (`filteredEntities`) changes, the state in the store will be updated.

assets/src/editor/events/dates-and-times/editor-date/filter-bar/with-date-entities-list-filter-state.js
```diff
@@ -59,6 +62,14 @@ export default createHigherOrderComponent(
...
 					datesView
 				),
+				filteredDateEntities: getEntitiesByIds(
+					'datetime',
+					store.getFilter(
+						'event-editor-dates-list',
+						'filteredDateIds',
+						DEFAULT_EMPTY_ARRAY
+					)
+				),
```
nerrad: I was able to use the existing structure of the `eventespresso/filter-state` store to track the filtered entities.  Note this value retrieves the actual entities matching the stored ids from the state.

assets/src/editor/events/dates-and-times/editor-date/filter-bar/with-filtered-date-entities.js
```diff
@@ -2,41 +2,27 @@
...
+					'datetime',
+					store.getFilter(
+						'event-editor-dates-list',
+						'filteredDateIds',
+						DEFAULT_EMPTY_ARRAY
+					)
+				),
+			};
+		}
+	),
```
nerrad: This hoc just makes it easier for consumers to get the current (and subscribe to the) filtered date entities in the store.

assets/src/editor/events/hocs/with-editor-date-entities.js
```diff
@@ -0,0 +1,12 @@
...
+import { createHigherOrderComponent } from '@wordpress/compose';
+
+export default createHigherOrderComponent(
+	withSelect( ( select ) => {
+		return {
+			dateEntities: select( 'eventespresso/core' )
+				.getDatetimes(),
+		};
+	} ),
+	'withEditorDateEntities'
```
nerrad: Notice this is simply getting all the datetimes currently in the state and it will always return the complete list when that changes in the current client session.  Think of this replacing the `allDatetimes` prop that was being passed around everywhere.

assets/src/editor/events/hocs/with-editor-event-entity.js
```diff
@@ -0,0 +1,23 @@
...
+			return { eventEntity };
+		}
+		const events = select( 'eventespresso/core' ).getEvents();
+		return events.length > 0 ?
+			{
+				eventEntity: events.slice( -1 ).pop(),
+			} :
+			DEFAULT_OBJECT;
+	} ),
+	'withEditorEventEntity'
```
nerrad: Think of this replacing the `eventEntity` prop being passed down through from the top-level component.  This hoc retrieves the event entity from the store state.  Note, I did introduce some logic here so that if there is already an `eventEntity` being passed through as props on a component that is composed with this hoc, then that `eventEntity` will take priority.  It basically prevents an unnecessary query to the store state (even though that won't be expensive).

assets/src/editor/events/hocs/with-editor-ticket-entities.js
```diff
@@ -0,0 +1,12 @@
...
+import { createHigherOrderComponent } from '@wordpress/compose';
+
+export default createHigherOrderComponent(
+	withSelect( ( select ) => {
+		return {
+			ticketEntities: select( 'eventespresso/core' )
+				.getTickets(),
+		};
+	} ),
+	'withEditorTicketEntities'
```
nerrad: Think of this hoc replacing the need to pass down the `allTicketEntities` prop everywhere.

assets/src/editor/events/hocs/with-price-type-entities.js
```diff
@@ -0,0 +1,30 @@
...
+			const priceTypeEntities = getEntities( 'price_type' );
+			const { hasFinishedResolution } = select( 'core/data' );
+			const priceTypeEntitiesLoaded = hasFinishedResolution(
+				'eventespresso/lists',
+				'getEntities',
+				[ 'price_type' ]
+			);
+			return { priceTypeEntities, priceTypeEntitiesLoaded };
+		}
+	),
```
nerrad: This still loads price type entities from the server on demand.  I think this is something that we _could_ move to the initialization layer but I don't think it creates too much of a problem having it load on demand either.

assets/src/editor/events/hocs/with-ticket-price-entities.js
```diff
@@ -0,0 +1,48 @@
...
+						noBasePrice: isEmpty( priceEntities ),
+						priceEntitiesLoaded,
+						priceEntities,
+					};
+				}
+			}
+			return EMPTY_OBJECT;
+		}
+	),
+	'withTicketPriceEntities'
```
nerrad: Ticket price entities should only be loaded on demand otherwise initialization would take way too long.  Hence keeping this hoc essentially as is for what was already existing in the app.

assets/src/editor/events/initialization/editor-initialization.js
```diff
@@ -0,0 +1,40 @@
...
+	children,
+} ) => {
+	useEffect( () => {
+		if ( eventEntityLoaded && dateEntitiesLoaded && ticketEntitiesLoaded ) {
+			setInitialized( true );
+		}
+	}, [ eventEntityLoaded, dateEntitiesLoaded, ticketEntitiesLoaded ] );
+	return <Provider value={ editorInitialized }>
+		{ children }
+	</Provider>;
```
nerrad: Essentially the `EditorInitialization` component (utilized via the `withEditorInitialization` hoc) wraps children in a Context Provider exposing the initialization state.  You can see here that currently the editor is considered "initialized" when the eventEntity, related dates to that event, and related tickets to those dates are all retrieved from the server and added to the state.  If we want to change what is exposed on the initialization context (or what criteria makes something "initialized" this is where we'd do it).

assets/src/editor/events/initialization/index.js
```diff
@@ -0,0 +1,25 @@
+export {
+	useEditorInitialized,
+	Consumer,
+} from './context';
```
nerrad: These two exports allow for add-ons, or other distinct component trees in editor interfaces to utilize the "initialized" context.  As long as the component using these is within a `Provider`, they can be utilized.  The provider is exposed via `withEditorInitialization` hoc.  So typical usage would go something like this for `useEditorInitialized` (the hook method).

```jsx
const MyComponent = ( props ) => {
	const initialized = useEditorInitialized();

	return initialized ? <Loading { ...props } /> : <NonLoading { ...props }/>;
};

export default withEditorInitialization( MyComponent );
```

Or this way using the Consumer:

```jsx
const MyComponent = ( { initialized, ...props } ) => {
	return initialized ? <Loading { ...props } /> : <NonLoading { ...props }/>;
};

const App = ( props ) => {
	return <Consumer>
		{ ( initialized ) => (
			<MyComponent initialized={ initialized } { ...props } />
		) }
	</Consumer>
};

export default withEditorInitialization( App );
```

assets/src/editor/events/initialization/use-editor-initialization.js
```diff
@@ -0,0 +1,8 @@
+import { useState } from '@wordpress/element';
+
+const useEditorInitialization = () => {
+	const [ editorInitialized, setEditorInitialized ] = useState( false );
+	return [ editorInitialized, setEditorInitialized ];
```
nerrad: This is an internal only hook and not exported.

assets/src/editor/events/initialization/with-event-date-entities.js
```diff
@@ -5,24 +5,24 @@ import { withSelect } from '@wordpress/data';
...
- * withEvent() DATA HOC has completed and resolved true
- *
  * @function
  */
 const withEventDateEntities = createHigherOrderComponent(
 	withSelect(
-		( select, { eventEntity, eventEntityLoaded } ) => {
+		( select, { eventEntity, eventEntityLoaded, editorInitialized } ) => {
 			if (
+				! editorInitialized &&
```
nerrad: This ensures that we're only ever invoking this particular logic until the editor is initialized (same for the `withEventEntity`, and `withTicketEntitiesForAllDateEntities` hocs in the initialization layer.  If this was not added, then they selectors will continue to query the store state everytime the state changes.  We don't need that in this case.

assets/src/editor/events/ticket-assignments-manager/ticket-assignments-manager-modal.js
```diff
@@ -848,74 +860,46 @@ export default compose( [
...
+	ifCondition( ( { editorOpen } ) => editorOpen ),
 	withSelect( ( select, ownProps ) => {
 		const {
-			editorOpen,
 			dateEntity,
-			allDateEntities,
+			dateEntities,
 			ticketEntity,
-			allTicketEntities,
+			ticketEntities,
```
nerrad: Notice that `dateEntities` and `ticketEntities` are now provided via the `withEditorDateEntities` and `withEditorTicketEntities` hocs in the composite.  Thus the `all*` variations don't need to be passed down from components the ticket assignments manager is attached to and it can be self-contained (only receiving things from parent components that affect it's output).

assets/src/editor/events/ticket-assignments-manager/with-ticket-assignments-manager-modal.js
```diff
@@ -19,61 +19,42 @@ import TicketAssignmentsManagerModal from './ticket-assignments-manager-modal';
...
+	( WrappedComponent ) => ( props ) => {
+		const {
+			dateEntity,
+			ticketEntity,
+		} = props;
+		const [ showManager, setShowManager ] = useState( false );
+		const toggleManager = useCallback( ( event ) => {
+			if ( event ) {
+				event.preventDefault();
+				event.stopPropagation();
```
nerrad: I converted this to a functional component because I think the logic is a bit clearer and we can take advantage of React hooks here.

assets/src/editor/events/tickets/editor-ticket/editor-ticket-entities-list.js
```diff
@@ -115,7 +116,7 @@ const createNewTicketEntity = withDispatch(
 			const newTicket = await createEntity( 'ticket', {} );
 			const newBasePrice = await createEntity(
 				'price',
-				{ prtId: basePriceTypeEntity.id }
+				{ PRT_ID: basePriceTypeEntity.id }
```
nerrad: When creating new entities, the _actual_ field names must be used (not the aliases).

assets/src/editor/events/tickets/editor-ticket/filter-bar/with-ticket-entities-for-filtered-date-entities.js
```diff
@@ -11,10 +10,18 @@ const withTicketEntitiesForFilteredDateEntities = createHigherOrderComponent(
 		{
 			filteredDateEntities = DEFAULT_EMPTY_ARRAY,
 			ticketEntities = DEFAULT_EMPTY_ARRAY,
+			isChained = true,
 		}
 	) => {
-		const dateIds = getDateEntityIds( filteredDateEntities );
+		if ( ! isChained ) {
+			return { ticketEntities };
+		}
```
nerrad: So the nice thing here, is if the ticket entity list is not chained to the date entity list, then any date state updates are not going to have any effect on the ticket entity list.

assets/src/editor/events/tickets/editor-ticket/filter-bar/with-ticket-entities-list-filter-bar.js
```diff
@@ -42,80 +47,98 @@ export const getFilteredTicketEntitiesList = ( ticketEntities, showTickets, tick
...
+				ticketsSortedBy,
+			]
+		);
+		// whenever filtered entities changes let's update the ticket ids
+		// in the state.
+		useEffect( () => {
+			setFilteredTicketEntities(
+				filteredEntities.map( ( ticketEntity ) => ticketEntity.id )
+			);
+		}, [ isChained, filteredDateEntities ] );
```
nerrad: What this logic indicates is that there is only need to update the filtered ticket entities (ids)  in the state if either `isChained` value changes or if `filteredDateEntities` changes.  However, this may need revisited because this means the state will not get updated if the ticket entity is not chained and filter selections are updated for tickets.

I have this the way it is currently because `ticketEntities` without this, the effect fires really frequently (add a console.log inside the effect).  That may have been just something in an earlier iteration of this pull and won't happen now but we probably should have this also include any of the filter state toggles as dependencies too.

nerrad:
## Possible todos

- `with-ticket-entities-list-filter-bar` has some fragile logic in the `useEffect`.  See [this comment](https://github.com/eventespresso/event-espresso-core/pull/1340/files#r294063228)

assets/src/editor/events/tickets/editor-ticket/filter-bar/with-ticket-entities-for-filtered-date-entities.js
tn3rb: what about date ticket relations though?

for example, if the ticket list is unchained and I add or remove ticket assignments for a particular datetime. Would the "item count" (the little number bubble that appears over the calendar icon in the ticket action menu) for the affected tickets change?

assets/src/editor/events/tickets/editor-ticket/filter-bar/with-ticket-entities-list-filter-bar.js
tn3rb: i checked out the branch and played around with the filters and everything seems to work ok (except for a bug in the sort chronologically filter but that has nothing to do with this branch)

tn3rb:
oops... i was a little overzealous with my response and approved the changes even though I had comments that will likely result in changes...
so will have to pass back for another review after pushing changes. sorry!

assets/src/editor/events/dates-and-tickets-metabox/dates-and-tickets-metabox.js
nerrad: Right, I expected those improvements because there's in editor-dates-tickets-refactor the three hocs now in the initialization layer were repeatedly getting triggered and thus causing re-renders in the component tree because of the change in props.  Thanks for posting this though, it's nice to get some hard date to confirm it helped things 👍

assets/src/editor/events/initialization/editor-initialization.js
```diff
@@ -0,0 +1,40 @@
...
+};
+
+export default compose( [
+	( WrappedComponent ) => ( props ) => {
+		const [ editorInitialized, setInitialized ] = useEditorInitialization();
```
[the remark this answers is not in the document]
nerrad: If it was going to be exported standalone yeah maybe, however because this is exported as part of a composite, for the purpose of debugging the `EditorInitialization` name will be present.  So unnecessary in this case.

assets/src/editor/events/initialization/index.js
nerrad: I don't think there's any harm in having both here.  The maintenance overhead is virtually nil as `Consumer` is something exposed via the `React.Context` api.  For our _own_ use, we'll likely use the hook more because it's an easier api to work with (imo), but there's no harm in exposing `Consumer` for third-party devs that want to use it because they are familiar with it.

assets/src/editor/events/ticket-assignments-manager/ticket-assignments-manager-modal.js
nerrad: > this file is still a mess (not your fault) as there's lots of logic that should be extracted into separate functional components and files, but these changes clean up a few things nicely

Agreed, but at least now we're better setup for future refactors that can touch a smaller part of the codebase 👍

assets/src/editor/events/tickets/editor-ticket/filter-bar/with-ticket-entities-for-filtered-date-entities.js
nerrad: > for example, if the ticket list is unchained and I add or remove ticket assignments for a particular datetime. Would the "item count" (the little number bubble that appears over the calendar icon in the ticket action menu) for the affected tickets change?

Yes it will because that bubble is reacting now to _state_ change in the store, not passed through props.

nerrad: Oh wait, I think I might have misunderstood.  Is the expectation that if `isChained` that the bubble only on the ticket only shows the count of dates matching the dates in the _filtered date list_ (even if the ticket may be assigned to more dates)?  If so, I'm not sure it currently behaves that way.  However, if that's the way yo _do_ want it, then it should be fairly trivial to simply ensure we use `withFilteredDatetimes` in the composite for the component displaying that count as a part of the Ticket Assignments Manager Modal.  I'd like to implement/fix that in a separate pull if needed?

assets/src/editor/events/tickets/editor-ticket/filter-bar/with-ticket-entities-list-filter-bar.js
nerrad: > except for a bug in the sort chronologically filter but that has nothing to do with this branch

Is that bug documented anywhere?

assets/src/editor/events/tickets/editor-ticket/filter-bar/with-ticket-entities-for-filtered-date-entities.js
tn3rb: > Is the expectation that if isChained that the bubble only on the ticket only shows the count of dates matching the dates in the filtered date list (even if the ticket may be assigned to more dates)?

oh no, that's not what I meant at all.

I think if the ticket is assigned to 10 dates, then the item count bubble should read 10 regardless of whether any of those related dates are displayed in the dates list or not.

My question was whether that item count bubble on a ticket would decrement to 9 if you removed that ticket assignment via the date when the ticket list was unchained.

nerrad: > My question was whether that item count bubble on a ticket would decrement to 9 if you removed that ticket assignment via the date when the ticket list was unchained.

Ahh ok, then in that case yes it will update correctly because the bubble count is not relying on passed through props but getting the number of dates related to the ticket via the store state.

assets/src/editor/events/tickets/editor-ticket/filter-bar/with-ticket-entities-list-filter-bar.js
tn3rb: no but I will be making a pull after this gets merged

assets/src/editor/events/dates-and-tickets-metabox/dates-and-tickets-metabox.js
tn3rb: oh and just for clarity, those numbers I posted only apply to the one particular event I have been using for development purposes, so those numbers are a reflection of the number of dates, tickets, and relations for that event, and would not be the same for any other event.

That said, though this branch does appear to be a good 40% faster which is a very significant improvement, so ya, good work!

assets/src/editor/events/tickets/editor-ticket/filter-bar/with-ticket-entities-for-filtered-date-entities.js
tn3rb: perfect

PR #1346, opened 2019-06-17
display ticket price in entity list

tn3rb:
currently the Ticket Entity List view does not display the ticket price and some of the css needs updating from when it was first converted over to using the new css variables.

This pull adds a Price column to the ticket entity list view table, renames the "Start Date" column to "Goes On Sale" and removes the "End Date" column since screen real estate is already completely cramped and the ticket price is a far more important detail than when the ticket is no longer on sale.

### DIFF
--- a/assets/src/editor/events/tickets/editor-ticket/list-view/editor-ticket-entity-list-item.js
+++ b/assets/src/editor/events/tickets/editor-ticket/list-view/editor-ticket-entity-list-item.js
@@ -86,18 +86,22 @@ class EditorTicketEntityListItem extends Component {
 					</div>
 					<div className="ee-ticket-list-item">
 						<span className="ee-ticket-list-item-label">
-							{ __( 'Start Date:', 'event_espresso' ) }
+							{ __( 'Price:', 'event_espresso' ) }
 						</span>
 						<span className="ee-ticket-list-item-value">
-							{ ticketEntity.startDate.toFormat( 'ddd MMM YY h:mm a' ) }
+							{
+								ticketEntity.price.formatter.formatNumber(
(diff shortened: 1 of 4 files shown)


assets/src/editor/events/tickets/editor-ticket/list-view/editor-ticket-entity-list-item.js
```diff
@@ -86,18 +86,22 @@ class EditorTicketEntityListItem extends Component {
...
 						<span className="ee-ticket-list-item-label">
-							{ __( 'Start Date:', 'event_espresso' ) }
+							{ __( 'Price:', 'event_espresso' ) }
 						</span>
 						<span className="ee-ticket-list-item-value">
-							{ ticketEntity.startDate.toFormat( 'ddd MMM YY h:mm a' ) }
+							{
+								ticketEntity.price.formatter.formatNumber(
+									ticketEntity.price.amount.toNumber()
+								)
```
nerrad: You should be able to just use `ticketEntity.price.toString()` to output the formatted value?

tn3rb: does that apply all of the currency symbol formatting, etc? 
if so, then ya, I'll do that, thnx

nerrad: It should:
```js
toString() {
		return this.formatter.format(
			this.amount.toNumber(),
			this.formatter.settings
		);
	}
```

Also worth noting that this is a "magic" method.  If you concatenate the object with a string (or cast it to a string) it will call `toString()` and output the formatted money value.

PR #1351, opened 2019-06-19
attempt to add  unit test to demonstrate issue

tn3rb:

If a recurrence pattern is edited in REM, the  resulting list of generated dates and tickets may differ from the existing dates and tickets associated with that recurrence pattern. So we basically need to loop through everything, compare the existing entities against what is required by the new recurrence pattern, and then generate a list of actions that fall into the following three categories:

- existing entities that need to be updated because of changes to some of their properties
- new entities that need to be created because no existing entities were found that matched
- existing entities that need to be deleted because they no longer belong as part of the recurrence pattern

during this process I need to retrieve all related entities for every entity I'm dealing with, such as tickets for each datetime, etc.
To ensure that queries are performant as possible I first collect the IDs for a set of entities and use `getRelatedEntitiesForIds()` to retrieve all of their relations so that calls to `getRelatedEntities()` on individual entities does not trigger multiple requests.

The logic for the above appears to work correctly, but of course triggers a re-render for the entity lists. When this happens an error is triggered in `getRelatedEntitiesForIds()`... long **after** the call to `getRelatedEntitiesForIds()`. I have console logged out the results of all of my function calls while stepping through all of the logic that REM performs when you submit the pattern editor, and I can see the results from the call to `getRelatedEntitiesForIds()` as well as the results from everything that happens after it. The error only appears when the REM editor modal closes and the entity list goes to re-render.

I've tried to write a unit test to demonstrate the issue but am getting a different error than what I'm seeing in the browser.

The process I'm attempting to use in the unit test is as follows:

- using the mock state plus event and datetime entities from `core/test/fixtures`
- add relation for datetime b to event a (and vice versa) so that event a is now related to both datetime a and datetime b
- leave datetime b's relation to event b in place
- call `getRelatedEntitiesForIds()` using the IDs for datetimes a & b to verify that they are related to events a & b
- trash event a
- call `getRelatedEntities()` on datetime b to retrieve its related events, which should just be event b now

this last call to `getRelatedEntities()` fails with an error which is different than what I am seeing in REM, but I'm not recreating the exact scenario that is happening in REM because that is far far **far** too complicated to recreate in a unit test. I'm also not sure if the above error is a result of my not performing all of the necessary steps to successfully trash event a and remove it from state. That said, the error could be a symptom of the same thing I'm seeing in REM.

Sorry I can't recreate this better.


nerrad:
The problem you were encountering in the tests was from incorrect test setup.  A tip for creating tests like this is to look at what the default state passed into the reducer is.  The final reducer registered on the store is composed from various separate substate reducers in the files.  So when you test an individual sub reducer you need to feed it only the _portion_ of the state it expects.  For example, anything in `core/reducers/dirty-entities.js` is only dealing with the `state.dirty.trash` (for trashes) and `state.dirty.delete` (for deletes) trees.

With the fixes I pushed, the test pass so on the surface it _looks_ like the store is working as expected here.

tn3rb:
thnx for making those fixes, didn't realize that only partial state was being worked on. Makes sense though.

I tried changing the test around to see if I could tease out the error I was seeing in REM, but have not been successful with that.

I pushed my changes because I think this test case is more interesting and useful than what I had before (still similar to some other tests though).

Here's what it does now:

- using the mock state plus event and datetime entities from core/test/fixtures
- adds relation for datetime b to event a (and vice versa) so that event a is now related to both datetime a and datetime b
- leave datetime b's relation to event b in place
- call `getRelatedEntitiesForIds()` using the IDs for **events** a & b to verify that they are related to datetimes a & b
- trash datetime a
- call `getRelatedEntitiesForIds()` on events a & b **again** (ie: exact same function call as above) to retrieve their related datetimes, which should just be datetime b now since datetime a was trashed

PR #1352, opened 2019-06-19
EDTR: Improve ticket assignments orphan control

nerrad:

One of the bugs with the EDTR is that this flow is possible:

- create a new ticket
- close the ticket modal which will open the ticket assignments manager modal
- close the ticket assignments manager modal which will leave the ticket unassigned to any dates.

Similar behaviour occurs for the creating a new date too except that currently closing the date creation modal will not bring up the ticket assignments modal.  Also, the flow for dates isn't as bad as the flow for tickets because the date is still assigned to the event.  In the case of tickets, if they are not assigned to a date then they are effectively orphaned and even if they got persisted when that work is done, the ticket would not show up on refresh of the event.

In order to fix this issue I needed to do some _initial_ refactoring of the ticket manager modal (TAM) so that I could re-order the composition of the component.  This is needed so that the TAM could inject behaviour into the modal toggle callback that aborts closure with an alert if there are no assignments.  This refactor involved:

- converting the class component to a functional component so we could use hooks.
- implementing hooks to replace the state in the class component
- implemented a new hook `useCountsManager` for deriving various state/props needed for informing ui about error state for orphaned tickets/dates.
- various other improvements via inline hocs to assist with improving the ticket manager behaviour.

The work results in improving the flow so that the user is properly prevented from closing the ticket assignments modal if there are no assignments (no matter what the context is) and allowing closure once that condition is satisfied:

gif showing improvements

This also should show the correct error message given the context the TAM is opened up in and the conditions producing the error.

### What is NOT done in this pull

- It's still possible to _create_ dates without any ticket assignments.  I'm intentionally leaving this to another pull because there's refactoring that will happen in the date editor modal to make it submitless.  At that point I'll also address the flow on date creation so ticket assignments is in that flow.
- This is not a _complete_ refactor of TAM.  There's still a lot of logic happening in there that probably can be simplified a bit more.  Also, the composition will need reviewed to see if there's any logic that can be either made internal hocs (for easier testing) or exposed for usage separately as well.  So for now, the new hocs I made are _intentionally_ inline until we can revisit that in individual targeted pulls.

## How has this been tested

Two things were verified through manual testing:

* [x] Ticket assignments work from all three contexts (single date, single ticket, main button in date editor section) and the warning about lack of ticket assignments shows up as expected in the various contexts.
* [x] The new warning and prevention of closing the TAM modal works as expected.
* [x] new unit tests for `useCountsManager` pass.


assets/src/editor/events/ticket-assignments-manager/hooks/use-counts-manager.js
```diff
@@ -0,0 +1,118 @@
...
+ */
+import { useRef, useMemo, useCallback } from '@wordpress/element';
+import { reduce } from 'lodash';
+import { __ } from '@eventespresso/i18n';
+
+// @todo, when the wp.data.useSelect hook is available, we can utilize that in
+// this custom hook and then we'll be able to remove the withSelect hoc usage
+// in the composed `TicketAssignmentsManagerModal` component.
+
+const useCountsManager = (
```
nerrad: I'm not sure on the name of this, so I'm open to suggestions for improving the name.

assets/src/editor/events/ticket-assignments-manager/hooks/use-counts-manager.js
```diff
@@ -0,0 +1,118 @@
+/**
+ * External imports
+ */
+import { useRef, useMemo, useCallback } from '@wordpress/element';
+import { reduce } from 'lodash';
+import { __ } from '@eventespresso/i18n';
+
+// @todo, when the wp.data.useSelect hook is available, we can utilize that in
+// this custom hook and then we'll be able to remove the withSelect hoc usage
+// in the composed `TicketAssignmentsManagerModal` component.
```
nerrad: It would greatly simplify things if we were able to say that the new ticket/date editor will _require_ WordPress 5.3 (which assuming everything goes to plan, will be the version containing the new `useDispatch` and `useSelect` hooks).  So I'm wondering if we should bite the bullet and start developing the editor with the Gutenberg plugin active so we can use those hooks and simplify the code now.

Pros:

- Avoid the current HOC hell we have now (only using them in very necessary places)
- Simplifies the logic flow for a lot of the big important things such as TAM.

Cons:

- we create a hard dependency on a future api that is stable in GB but hasn't landed in WP core yet.  So there's uncertainty there.
- We _will_ need to bump minimum WP requirements to a more recent version (likely WP 5.3+) when this lands.
- We need to remember to develop with the GB plugin active while working on this.
- Some more mental decision angst around whether we build something in a HOC or using new hooks for things we write.

I just bring this up, not because I think the decision needs to be made in this  pull, but as something to keep in mind and think on.  I haven't convinced myself one way or the other yet so we can sit on it.  But I'm sure this will come up again as I continue to go through and improve the logic of the new editor.

tn3rb: I'm almost always for going with the more future forward way of doing things and here's some of the reasons that come to mind for me:

- there's already a hard dependency on a fairly recent version of WP core anyways, so I don't think bumping that up to 5.3 is that big of a deal.
- yes there's a risk that using the new `useDispatch` and `useSelect` hooks may force us to refactor things **IF** those APIs change prior to release, but not doing so and using the existing API means more convoluted code that we are likely going to want to refactor one day anyways. So _potential_ refactors in exchange for easier to deal with code versus guaranteed technical debt and more obtuse code is an easy choice for me.
- activating a plugin for dev purposes may actually prove to be too hard... LOLZ j/k
- more "mental decision angst around whether we build something in a HOC or using new hooks" is nothing compared to the mental angst involved with what to name things, so this isn't a big deal for me either. I think I will likely favour hooks and functional composition over HOCs even though I haven't actually started using them yet.

assets/src/editor/events/ticket-assignments-manager/hooks/test/use-counts-manager.js
```diff
@@ -0,0 +1,627 @@
...
+					},
+					tickets: {
+						...testedAssignedCounts.tickets,
+						30: 1,
+					},
+				},
+			);
+			// original assigned counts should not have changed
+			expect( testedAssignedCounts.dates[ 40 ] ).toBe( 1 );
+			expect( testedAssignedCounts.tickets[ 30 ] ).toBe( 0 );
```
nerrad: One of the issues I discovered in the latest round of changes is that the original assigned counts coming into the custom hook were getting mutated.  This is primarily because I was treating the spread operator as doing a deep clone (🤦‍♂ ).  So I added this expectation (in all tests) which both exposed it, and also verifies that the original assignment counts don't get mutated.

assets/src/editor/events/ticket-assignments-manager/hooks/use-counts-manager.js
```diff
@@ -0,0 +1,141 @@
...
+
+const useCountsManager = (
+	dateEntities,
+	ticketEntities,
+	assignedCounts,
+	assignedState,
+) => {
+	const getMessage = useCallback( ( whichAreEmpty ) => {
```
nerrad: Instead of `whichAreEmpty` being a ref (which was fragile), it is now passed in as an argument to this memoized callback.  Thus it's much less fragile.

assets/src/editor/events/ticket-assignments-manager/hooks/use-counts-manager.js
```diff
@@ -0,0 +1,141 @@
...
+		}
+		return message;
+	}, [ dateEntities, ticketEntities ] );
+
+	const isEmptyAssignedState = useCallback(
+		() => isEmpty( assignedState.assigned ) && isEmpty( assignedState.removed ),
+		[ assignedState ]
+	);
+
+	const [ hasNoAssignments, noAssignmentsMessage, updatedAssignmentsCounts ] = useMemo( () => {
```
nerrad: Instead of `updatedAssignmentsCounts` being a ref, it is now contained in the `useMemo` implementation and returned as a value.  Much less fragile.

assets/src/editor/events/ticket-assignments-manager/hooks/use-counts-manager.js
```diff
@@ -0,0 +1,141 @@
...
+		let howManyEmpty = 0;
+		const whichAreEmpty = { dates: 0, tickets: 0 };
+		const updatedCounts = cloneDeep( assignedCounts );
```
nerrad: These are both reset everytime the `useMemo` is recalculated (which happens when `assignedState` or `assignedCounts` change.  This replaces the `refs` that were being used for these values which is much less fragile.

assets/src/editor/events/ticket-assignments-manager/hooks/use-counts-manager.js
```diff
@@ -0,0 +1,141 @@
...
+		// first let's see if any of the current counts are 0 (but only not
+		// already toggled as empty)!
+		// instead of using whichAreEmpty as a ref, use assignedState (whether its empty or not) to decide whether we
+		// do this here or not. If assigned state is empty then we can do it.
+		if ( isEmptyAssignedState() ) {
```
nerrad: Previously I was using `whichAreEmpty` as a ref for deciding whether to do this initial short-circuit calc.  That was very fragile because relying on the ref to keep counts through state changes was super fragile.  All that was really needed here is to run this short circuit path if `assignedState` is empty (meaning the original assignedCounts could be used instead)

assets/src/editor/events/ticket-assignments-manager/hooks/use-counts-manager.js
```diff
@@ -0,0 +1,141 @@
...
+				}
+			}
+			return emptyCount;
+		};
+		howManyEmpty = reduce( assignedState.assigned, countCallback( true ), howManyEmpty );
+		howManyEmpty = reduce( assignedState.removed, countCallback( false ), howManyEmpty );
+
+		// ensure whichAreEmpty don't go lower than zero
+		whichAreEmpty.dates = Math.max( 0, whichAreEmpty.dates );
+		whichAreEmpty.tickets = Math.max( 0, whichAreEmpty.tickets );
```
nerrad: It is possible for `whichAreEmpty` values to end up negative depending on how many tickets/dates are being processed and what are assigned or not assigned.  Because `whichAreEmpty` is used in determining what message is printed, it's easier to just make sure that we convert negative to zero here.

assets/src/editor/events/ticket-assignments-manager/ticket-assignments-handler.js
```diff
@@ -1,76 +1,56 @@
...
  * @return {number|boolean} index for date id in assigned array or
  *                        boolean if returnIndex is false
  */
-export const isAssigned = ( assigned, date, ticket, returnIndex = false ) => {
+export const isAssigned = ( assigned, dateId, ticketId, returnIndex = false ) => {
 	let index = noIndex;
-	if ( isArray( assigned[ date.id ] ) ) {
-		index = findIndex( assigned[ date.id ], { id: ticket.id } );
+	if ( isArray( assigned[ dateId ] ) ) {
+		index = assigned[ dateId ].indexOf( ticketId );
```
nerrad: The main change in the handler file is to only deal with entity ids instead of full entities.  The primary reason for this is that full entities are _not_ needed and it makes it unnecessarily more complicated when searching for values.  The trigger for this is that `prevState` is being mutated in `assignTicketEntity` and `removeTicketEntity`.  State should never be mutated directly (note that `TicketAssignmentsManagerModal` is passing in a deep cloned `prevState` to the handler).

assets/src/editor/events/ticket-assignments-manager/ticket-assignments-manager-modal.js
```diff
@@ -41,463 +41,180 @@ const {
...
+		( dateId, ticketId ) => {
+			setAssignedState( ( prevState ) => {
+				return handler.removeTicketEntity(
+					cloneDeep( prevState ),
+					dateId,
+					ticketId
+				);
+			} );
+		},
+		[]
```
nerrad: This does not need to have any dependencies because the state setter does not change.

assets/src/editor/events/ticket-assignments-manager/ticket-assignments-manager-modal.js
```diff
@@ -41,463 +41,180 @@ const {
...
+								}
+							>
+							</td>
+						),
+					}
+				);
+			} );
+			return rowData;
+		},
+		[]
```
nerrad: `yearRow` does not need any dependencies because it is not using anything from outside its scope (but only what is passed in on the callback).

assets/src/editor/events/ticket-assignments-manager/ticket-assignments-manager-modal.js
```diff
@@ -578,34 +262,14 @@ class TicketAssignmentsManagerModal extends Component {
...
-			canRemoveAssignment,
-		};
-	};
+	const getActionModifiers = useCallback(
+		( hasTicket, isAssigned, isRemoved ) => {
+			const currentlyAssigned = isAssigned ||
+				( hasTicket && ! isRemoved );
+			return { currentlyAssigned };
+		},
+		[]
```
nerrad: No dependency required because all values used are passed in as arguments.

assets/src/editor/events/ticket-assignments-manager/ticket-assignments-manager-modal.js
```diff
@@ -659,201 +297,359 @@ class TicketAssignmentsManagerModal extends Component {
...
+			} else if ( isAssigned ) {
+				icon = 'tickets-alt';
+				bgColor += ' ee-tam-add-ticket-relation';
+			} else {
+				icon = 'minus';
+				bgColor += ' ee-tam-no-ticket-relation';
+			}
+			return { icon, bgColor };
+		},
+		[]
```
nerrad: No dependencies needed because all variables are passed in as arguments.

assets/src/editor/events/ticket-assignments-manager/ticket-assignments-manager-modal.js
```diff
@@ -659,201 +297,359 @@ class TicketAssignmentsManagerModal extends Component {
...
+						<CalendarPageDate
+							startDate={ dateEntity.start }
+							statusClass={ getDateBgColorClass( dateEntity ) }
+							size={ 'small' }
+						/>
+					</div>
+				),
+			};
+		},
+		[]
```
nerrad: No dependencies needed because:

- `dateEntity` is passed in as an argument
- `getDateBgColorClass` and `CalendarPageDate` are both imports and thus will not change.

nerrad:
More fixes, new screencast: https://cl.ly/71266eea3158

tn3rb:
pushed a branch with my suggested fixes: `EDTR/improve-ticket-assignments-orphan-control-BC`

https://github.com/eventespresso/event-espresso-core/tree/EDTR/improve-ticket-assignments-orphan-control-BC

assets/src/editor/events/ticket-assignments-manager/hooks/use-assignments-calculator.js
```diff
@@ -0,0 +1,148 @@
...
+
+		const countAssignmentChangesForTickets = ( emptyCount, count ) => {
+			if ( count === 0 ) {
+				missingAssignmentCounts.tickets++;
+				emptyCount++;
+			}
+			return emptyCount;
+		};
+		totalMissingAssignments = reduce( updatedCounts.dates, countAssignmentChangesForDates, totalMissingAssignments );
+		totalMissingAssignments = reduce( updatedCounts.tickets, countAssignmentChangesForTickets, totalMissingAssignments );
```
nerrad: Two reducers are needed here because it's possible there may be tickets not assigned to any dates (or in any assigned state change), so we need to consider those for possible missing assignment counts.

assets/src/editor/events/ticket-assignments-manager/ticket-assignments-manager-modal.js
```diff
@@ -865,67 +661,95 @@ export default compose( [
...
+			if ( typeof assignmentCounts.current.tickets[ ticket.id ] === 'undefined' ) {
+				assignmentCounts.current.tickets[ ticket.id ] = getRelatedEntities(
+					ticket,
+					'datetime'
+				).length || 0;
+				// no need to set dtmProps.ticketEntitiesByDateIds here as
+				// those will already have been setup for all dates.
+			}
+		} );
+		return dtmProps;
```
nerrad: So part of the reason the reducer were looping over assignedState _and_ assignedCounts in the original iteration of the custom hook is because it was possible for the passed in `assignedCounts` to not have ticket or date ids in it’s collection that existed in `assignedState` due to how `assignedCounts` was built in the composition of `TicketAssignmentsManagerModal`.  Hence, the refactoring in here.  Now, there is always a count set for the tickets and dates being used in the session.

nerrad:
Okay @tn3rb, I basically copied in the naming changes on the custom hook from the branch you created, so that branch is no longer needed.  I think this incorporates the needed logic fixes and extra tests should pass.  I did some user testing of multiple dates and tickets and it seems to be working okay.  So ready for another review.

Issue #887, opened 2019-01-11
Fix `espresso_event_reg_button()` template tag


joshfeck:
This was reported via Help Scout:

File `template_tags.php` , function `espresso_event_reg_button`, link href is `espresso_event_link_url()` while shoul be `espresso_event_link_url($EVT_ID)`.

Also reported this:

1) `espresso_event_reg_button` function in template tags call function `EEH_Event_View::event_active_status( $EVT_ID );`
2) `event_active_status` call `$event->pretty_active_status()`
3) `pretty_active_status` be default has `$echo = TRUE`
4) if `$echo = TRUE` we get
```
echo $status;
return '';
```
5) Result is that `espresso_event_reg_button` returns event status = ''`
6) and with status < 1 we have
```
$btn_text = $btn_text_if_inactive;
$class = 'ee-grey';
```

PR #1356, opened 2019-06-21
Fix `espresso_event_reg_button()` template tag

joshfeck:

See #887. This template tag isn't used anywhere in EE4 core or as far as I can tell any add-ons. However, if you try to use this template tag you'll find its logic is broken and it doesn't return the correct URL to the event page.

## How has this been tested

Added the following code to a custom template and plugged in a variety of valid event IDs:
e.g. `echo espresso_event_reg_button(null, false, 40298);`
also test an invalid event ID which should display nothing.

## Checklist

* [x] My code is tested.


public/template_tags.php
```diff
@@ -542,7 +550,7 @@ function espresso_event_date_range( $date_format = '', $time_format = '', $singl
 			$time_format = ! empty( $time_format ) ? $time_format : get_option( 'time_format' );
 			$time_format = apply_filters( 'FHEE__espresso_event_date_range__time_format', $time_format );
 			$html = sprintf(
-				__( '%1$s - %2$s', 'event_espresso' ),
+				esc_html__( '%1$s - %2$s', 'event_espresso' ),
```
tn3rb: maybe add the translation context so that translators understand what this is for, either:

- via code comment like: `/* translators: 1: earliest event date, 2: latest event date */`
- or via `esc_html_x()`

joshfeck:
If we change to `esc_html_x()` it will actually break existing translations. I added a note.

sethshoultes:
Tested with a variety of events and the button shows the expected event status:
Event Blah – Events by Espresso 2019-06-26 15-48-58
Event Blah – Events by Espresso 2019-06-26 15-48-20
Event Blah – Events by Espresso 2019-06-26 15-48-09

PR #1357, opened 2019-06-21
additional component keys

tn3rb:
was experiencing some rendering issues with the entity lists while working on REM and thought that the lack of component keys might be contributing. Decided to add a few extras. Was not the cause of the issue, but these can't hurt, so pushed branch anyways.


tn3rb:
just going to kill this

PR #1358, opened 2019-06-21
don't display archived/trashed entities

tn3rb:
Archived/deleted entities still appear in the entity lists when using normal filters but this should not be the case.
This simply prevents trashed entities from being considered when checking the status for an entity.

plz note:

I removed several calls to `assertDateTimeEntity()` and `assertTicketEntity()` because that check is performed by `isTrashed()` and `isArchived()` so the check would be redundant. 

## How has this been tested
- monkey based browser testing 
- added unit tests


assets/src/data/model/datetime/test/status-helper.js
```diff
@@ -0,0 +1,411 @@
...
+		expect( () => isActive( validDateEntity ) ).not.toThrow( TypeError );
+		expect( () => isExpired( validDateEntity ) ).not.toThrow( TypeError );
+		expect( () => isRecentlyExpired( validDateEntity ) ).not.toThrow( TypeError );
+		expect( () => isSoldOut( validDateEntity ) ).not.toThrow( TypeError );
+		expect( () => isTrashed( validDateEntity ) ).not.toThrow( TypeError );
+		expect( () => isUpcoming( validDateEntity ) ).not.toThrow( TypeError );
+	} );
+	describe( 'isActive()', () => {
+		it( 'returns false if Date Entity start and end dates are in future',
+			() => expect( isActive( upcomingDateEntity() ) ).toBe( false )
```
nerrad: This returns the value of the `expect` call.  Test callbacks should not return values (see https://github.com/jest-community/eslint-plugin-jest/pull/136).

assets/src/data/model/ticket/status-helper.js
```diff
@@ -25,31 +25,43 @@ const assertTicketEntity = ( TicketEntity ) => {
...
+ * @param {boolean} includeArchived if true will not filter out archived entities
  * @return {boolean} true if tickets sold meets or exceeds available quantity
  */
-export const isSoldOut = ( TicketEntity ) => {
-	if ( isArchived( TicketEntity ) ) {
+export const isSoldOut = ( TicketEntity, includeArchived = false ) => {
+	if (
+		( includeArchived && ! assertTicketEntity( TicketEntity ) ) ||
+		( ! includeArchived && isArchived( TicketEntity ) )
+	) {
```
nerrad: I have the same feedback here as the status-helper for date-times regarding extracting out this common condition into it's own function that you can call here and elsewhere (to keep code DRY).

assets/src/data/model/datetime/status-helper.js
```diff
@@ -25,42 +25,58 @@ const assertDateTimeEntity = ( DateTimeEntity ) => {
...
  * @return {boolean} true if event date is occurring NOW
  */
-export const isActive = ( DateTimeEntity ) => {
-	return ! isTrashed( DateTimeEntity ) &&
-		DateTimeEntity.start.diffNow().asSeconds() < 0 &&
-		DateTimeEntity.end.diffNow().asSeconds() > 0;
+export const isActive = ( DateTimeEntity, includeTrashed = false ) => {
+	return (
+		( includeTrashed && assertDateTimeEntity( DateTimeEntity ) ) ||
+		( ! includeTrashed && ! isTrashed( DateTimeEntity ) )
```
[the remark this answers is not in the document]
tn3rb: oooo good point... dunno how I didn't notice that!  will do

PR #1360, opened 2019-06-21
fix edit date venue link

tn3rb:
plz see: https://github.com/eventespresso/event-espresso-core/projects/4#card-22142453

The date entity grid items in the new EDTR were using a temporary hardcoded placeholder for the venue name / edit link.
This adds a new HOC for retrieving the venue for an event and using it to build a real link.

## How has this been tested

#### testing

- [ ] confirm venue name and valid working edit venue link appears for events with an associated venue
- [ ] confirm no venue name appears for events with no associated venue


assets/src/editor/events/hocs/if-valid-date-entity.js
```diff
@@ -0,0 +1,12 @@
+import { createHigherOrderComponent, ifCondition } from '@wordpress/compose';
+import { isModelEntityOfModel } from '@eventespresso/validators';
+
+export default createHigherOrderComponent(
```
nerrad: This hoc can go inside the `editor/hocs` folder and thus be exported on `@eventespresso/editor-hocs` aliased package.

Also, there's good opportunity to add unit tests for this (which should be straightforward enough).

assets/src/editor/events/hocs/with-editor-event-venue-entity.js
```diff
@@ -0,0 +1,25 @@
...
+			);
+			return Array.isArray( venueEntity ) && venueEntity[ 0 ] &&
+				isModelEntityOfModel( venueEntity[ 0 ], 'venue' ) ?
+				{ venueEntity: venueEntity[ 0 ] } :
+				DEFAULT_OBJECT;
+		}
+		return DEFAULT_OBJECT;
+	} ),
+	'withEditorEventVenueEntity'
+);
```
nerrad: This creates a tight coupling of the hoc to getting a venue related to the given event.  Since currently, there's only ever one venue for an event, I think you should move this to be defined as a part of the initialization layer so the venue is retrieved at the start of the editor initialization and then introduce a hoc here that retrieves the current venue in the state.

I _do_ realize that eventually we will make venues related to date-times and that will change things, but without there being any timeline on that eventuality I think we should code things for how they are _now_.  Right now, this means that any component displaying the venue (in this case date entity components) will need to also include `withEditorEventEntity`  in their hoc chain.

assets/src/editor/events/hocs/if-valid-date-entity.js
```diff
@@ -0,0 +1,12 @@
...
+	ifCondition(
+		( { dateEntity } ) => isModelEntityOfModel(
+			dateEntity,
+			'datetime'
+		)
+	),
+	'ifValidDateEntity'
+);
```
nerrad: This hoc can go inside the `editor/hocs` folder and thus be exported on `@eventespresso/editor-hocs` aliased package.

Also, there's good opportunity to add unit tests for this (which should be straightforward enough).

assets/src/eejs/routes.js
```diff
@@ -99,6 +99,7 @@ export const ADMIN_ROUTE_ACTIONS = {
 	VENUES: {
 		OVERVIEW: ADMIN_ROUTE_ACTION_DEFAULT,
 		CATEGORIES: 'category_list',
+		EDIT: 'edit',
```
nerrad: I know it's a pain, but I think this should be done separately in a branch off master.

tn3rb: see: #1364

Issue #1326, opened 2019-06-13
Changes sometimes not applied when calling the save method


Pebblo:
Easiest to show with a a quick example:

```
add_action('AHEE__Extend_Events_Admin_Page___duplicate_event__after', 'updateacfanddatetime', 10, 2);
function updateacfanddatetime($new_event, $orig_event) {
  $new_datetimes = $new_event->get_many_related('Datetime');
  foreach ($new_datetimes as $new_dtt) {
      $newdate = new DateTime(date("Y-m-d",strtotime($new_dtt->start_date()))) ;
      $newdateplusone = date_format($newdate->add(new DateInterval('P1D')), 'F d, Y');
      $new_dtt->set_start_date($newdateplusone);
      $new_dtt->set_end_date($newdateplusone);
      $new_dtt->save();
 }            
}
```

The above adds a single date onto your events datetimes when you duplicate an event, however, the change is not saved as `set_end_date` does not set a `_has_changes` flag in `EE_Base_Class.class.php::save()`.

So for the above to work as is, you'd need to add something like:

`$new_dtt->set('DTT_EVT_start', $new_dtt->start_date_and_time());
$new_dtt->set('DTT_EVT_end', $new_dtt->end_date_and_time());`

just before the `save()` call, as `set()` sets the `_has_changes` flag it all then works fine.

Spoke about this a little with Darren, take a look here
See: https://eventespresso.slack.com/archives/C080L8G68/p1560423687230700
(Darren replies a little below) 

Posting convo here in case we hit slack limit:

> tony [12:01 PM]
Found something that (I think) is a little strange with Datetimes.
A user is updating datetimes by 1 day when you duplicate the event and is doing it like this:
`function updateacfanddatetime($new_event, $orig_event) {
  $new_datetimes = $new_event->get_many_related('Datetime');
  foreach ($new_datetimes as $new_dtt) {
      $newdate = new DateTime(date("Y-m-d",strtotime($new_dtt->start_date()))) ;
      $newdateplusone = date_format($newdate->add(new DateInterval('P1D')), 'F d, Y');
      $new_dtt->set_start_date($newdateplusone);
      $new_dtt->set_end_date($newdateplusone);
 }            
}`
(edited)
Which apart from missing `$new_dtt->save();` I would have through worked, right?
Echoing out `$new_dtt->start_date_and_time()` and `$new_dtt->start_end_and_time()` both before and after the change shows the date is updated, but save fails.
[...]

PR #1363, opened 2019-06-24
mark model objects as dirty when datetimes are set too

mnelson4:

If you used one of `EE_Datetime`'s helpers for setting dates (eg `set_start_time()`, `set_end_date()`, etc), the model object wouldn't recognized a change happened (ie, the property `EE_Base_Class::is_changed` wouldn't get set to `true`), so if that was the only change made, and you then called `EE_Base_Class::save()`, the change would NOT get saved.

The fix was to have `EE_Base_Class::_set_date_time()`, the method ultimately used by all those helper methhods, to also set `EE_Base_Class::is_changed`, just like `EE_Base_Class::set()` does.

## How has this been tested

* [x] Use Tony's code snippet from https://github.com/eventespresso/event-espresso-core/issues/1326#event-2434387590, and duplicate an event. The duplicate event's datetime should be one day later than the original
* [ ] Do a normal registration, just to double-check this didn't have an unexpected side effect

## Checklist

* [x] My code is tested.


joshfeck:
fixes #1326

Pebblo:
All good so I've merged this in.

PR #1368, opened 2019-06-26
Add Admin Options and Conditional Loading for EDTR

tn3rb:
Currently the display of the new EDTR is hardcoded into the Event Admin and the existing "Event Tickets and Datetimes" metabox is still displayed. 
This branch adds a new section to the **Events Admin Default Settings** tab that allows the EDTR to be turned on and off, as well as set a couple of options for setting default values for how the Entity Lists function. Filters have also been added so that the REM addon will be able to force usage of the EDTR.

## testing

* log into the admin and visit the **Event Espresso - Events** > **Default Settings** admin page tab
* [ ] confirm that new setting options appear at the top of the form
* change the first option to "Advanced Editor" and click save
* click on the **Overview** tab so that it opens in another browser tab (or window) and click on an event to open its editor
* [ ] confirm that the new EDTR loads and that the old "Event Tickets and Datetimes" metabox does NOT appear
* back in the first tab, set the "Activate Advanced Editor?" input to "Legacy Editor" and click save
* refresh the event editor in the second tab
* [ ] confirm that the new EDTR does NOT load and that the old "Event Tickets and Datetimes" metabox appears again
* try changing some settings for the dates or tickets using the legacy editor and click save
* [ ] confirm that the saving the options was successful and that no errors occurred
* back in the first tab, play around with the "View" and "Per Page" options for the EDTR and click save
* refresh the event editor in the second tab
* [ ] confirm that your selections have been reflected in the appearance of the EDTR


assets/src/editor/events/event-editor.js
```diff
@@ -10,10 +10,20 @@ const editor = document.createElement( 'div' );
...
 
 if ( eventId ) {
 	render(
-		<EventDatesAndTicketsMetabox eventId={ eventId } />,
+		<EventDatesAndTicketsMetabox
+			eventId={ eventId }
+			entityListView={ entityListView }
+			entityListPerPage={ entityListPerPage }
+		/>,
 		editor
```
nerrad: After doing the full review it looks like these are incoming options set server side.  Initially however, from the name I thought `entityListView` was an actual view component, but instead I see that it is indicating whether it is "grid" or "list".

Some thoughts I have:

- If you decide to keep this in, I think a change in variable name might be helpful.
- I'm wary of prematurely setting options this way as I think it as much as possible we should be careful about what props need to be passed in on initialization until we have a better idea of how we want to initialize. I'd be more inclined to have options loaded as a part of the _internal_  initialization process (as opposed to externally via the  so they could be either coming from browser localStorage (so there can be per user/browser defaults) OR defaults from the server (via api or global variable from the dom set by the server).
- Until we're ready to do that, I'd be inclined to simplify the admin option you are adding in this pull to just activating/deactivating the advanced editor and we can spend more time figuring out how we will initialize "defaults" in it's own pull.

admin_pages/events/Events_Admin_Page.core.php
```diff
@@ -2350,26 +2341,32 @@ protected function _default_event_settings_form()
...
                 }
                 if (isset($valid_data['default_max_tickets'])) {
                     $registration_config->default_maximum_number_of_tickets = $valid_data['default_max_tickets'];
                 }
+                do_action(
+                    'AHEE__Events_Admin_Page___update_default_event_settings',
+                    $valid_data,
+                    EE_Registry::instance()->CFG,
+                    $this
+                );
```
nerrad: I know this is a pain, but this seems like something that would be better to add in a branch off master to reduce chances of conflicts etc.  I'll leave it up to you what you do but.

caffeinated/admin/extend/events/Extend_Events_Admin_Page.core.php
```diff
@@ -1269,4 +1301,115 @@ protected function _delete_the_ticket($TKT_ID)
...
+                        esc_html__('Advanced Editor', 'event_espresso'),
+                    ]
+                ),
+                [
+                    'default'         => $this->admin_config->useAdvancedEditor(),
+                    'html_label_text' => esc_html__('Activate Advanced Editor?', 'event_espresso'),
+                    'html_help_text'  => sprintf(
+                        esc_html__(
+                            'Controls whether the Event Espresso Event Editor continues to use the existing legacy editor that functions like the typical older WordPress admin you are used to,%1$sor uses the new Advanced Editor with a more powerful and easier to use interface. This may be automatically turned on in order to utilize advanced features from new addons.',
+                            'event_espresso'
```
nerrad: The "this may be automatically turned on in order to use advanced features" is good text here for when the REM add-on is active.  I wonder if it'd be good to add an hook somewhere here so REM can _disable_ this input and add it's own text explaining why the option is disabled (so that the user knows why they can't change it and why it is forced on)?

tn3rb:
I liked your idea for using localStorage so went with that

assets/src/editor/events/dates-and-times/editor-date/filter-bar/with-date-entities-list-filter-state.js
```diff
@@ -23,8 +23,10 @@ export default createHigherOrderComponent(
 				datesSortedBy = 'chronologically',
 				displayDates = 'start',
 				searchDateName = '',
-				datesPerPage = 6,
-				datesView = 'grid',
+				datesPerPage,
+				datesView,
+				defaultDatesListView,
+				defaultDatesListPerPage,
```
nerrad: Why have two sets of props here for per page and view type?  Establishing this pattern becomes hard to maintain if in the future other props having defaults are also implemented (meaning everytime such prop is added, another default prop is included).  I think it might be better to lift up the state of default management and only pass through what the set prop is OR internally manage the default in this component.

assets/src/higher-order-components/filter-bar/entity-list-filter-bar.js
```diff
@@ -93,50 +102,88 @@ class EntityListFilterBar extends Component {
...
+			</label>
+			<IconButton
+				id={ `ee-grid-view-btn-${ this.props.instanceId }` }
+				className={ view === 'grid' ? 'active-list' : '' }
+				icon="screenoptions"
+				tooltip={ __( 'grid view', 'event_espresso' ) }
+				onClick={ setGridView }
+			/>
+		</Fragment>
+	);
```
nerrad: I'm wary of having a separate area for setting defaults.  I think we should _consider_:

- a unified options/settings toggle that would be used for the entire editor eventually for controlling various defaults  in the editor (default per page for things) etc.  What would that look like, how would options be registered with it.
- Does per-page and view type even NEED controllable defaults.  Could it not just remember what the user has selected in the existing toggles between sessions and across events? For example, user selects 5 as the per-page value for date list and 8 as the per-page value for ticket list.  They go and edit another event and that value is automatically used for that new event.  OR  could remembered specific to an event/user.

assets/src/higher-order-components/local-storage/use-local-storage.js
```diff
@@ -0,0 +1,16 @@
+/**
+ * External imports
+ */
+import { useState, useEffect } from '@wordpress/element';
+
+const useLocalStorage = ( key ) => {
```
nerrad: Rather than rolling our own implementation, I was thinking we could use something like `https://github.com/WordPress/gutenberg/tree/master/packages/data/src/plugins/persistence` to simply persist the filter-state store between sessions.

assets/src/higher-order-components/local-storage/with-default-dates-list-settings.js
```diff
@@ -0,0 +1,35 @@
...
+		] = useLocalStorage( 'defaultDatesListPerPage' );
+
+		return (
+			<EntityListWithFilterBar
+				{ ...otherProps }
+				defaultDatesListView={ defaultDatesListView || 'grid' }
+				setDefaultDatesListView={ setDefaultDatesListView }
+				defaultDatesListPerPage={ defaultDatesListPerPage || 6 }
+				setDefaultDatesListPerPage={ setDefaultDatesListPerPage }
+			/>
```
nerrad: Rather than adding additional props, I think this should override the props passed through to whatever implements this.  So for example, you'd end up with something like:

```jsx
( EntityListWithFilterBar ) => ( {
	datesListView,
	datesListPerPage,
	...otherProps
} ) => {
	const [
		defaultDatesListView,
		setDefaultDatesListView,
	] = useLocalStorage( 'defaultDatesListView' );
	const [
		defaultDatesListPerPage,
		setDefaultDatesListPerPage,
	] = useLocalStorage( 'defaultDatesListPerPage' );

	return (
		<EntityListWithFilterBar
			{ ...otherProps }
			datesListView={ datesListView || defaultDatesListView }
			setDefaultDatesListView={ setDefaultDatesListView }
			datesListPerPage={ datesListPerPage || defaultDatesListPerPage }
			setDefaultDatesListPerPage={ setDefaultDatesListPerPage }
		/>
	);
};
```

But also, if you implement the persistence plugin for the filter state as I referenced in a previous comment I think this hoc would become unnecessary.

_note: comments here apply to the similar work done for tickets list_

caffeinated/admin/extend/events/Extend_Events_Admin_Page.core.php
nerrad: Did you see this comment I added ^^?

admin_pages/events/Events_Admin_Page.core.php
tn3rb: no we do not want to add options to the config before they are going to be used.
as it turns out, based on other feedback you have given I have removed a couple of these options. Had I added these to master and merged them in prior to deciding to go a different direction with how and where options are saved, then it would require another PR to fix master.

caffeinated/admin/extend/events/Extend_Events_Admin_Page.core.php
tn3rb: oops mistook which block of text this comment was made on, and thought it applied to one of the options that has been removed. Thought it would be obvious why I didn't respond to a comment regarding something that was removed, but that was not the case. My apologies. I've added a filter for the input settings as well.

admin_pages/events/Events_Admin_Page.core.php
nerrad: I think you are misunderstanding.  My comment applies specifically to the changes the comment is attached to.  You've repositioned the initialization of the `$registration_config` variable here (good!) and added an action (good).  I still believe those could have been added in master without issue.  I do agree you don't add new unused options in a branch off master for options not used, but that's not what I was suggesting here.

tn3rb:
just ripped everything out.
this is now just the single admin option for turning the EDTR on and off

assets/src/editor/events/dates-and-times/editor-date/filter-bar/index.js
```diff
@@ -1,7 +1,10 @@
 /**
  * External imports
  */
-import { withEntityPagination } from '@eventespresso/higher-order-components';
+import {
+	withDefaultDatesListSettings,
```
nerrad: Is this still supposed to be here?

assets/src/editor/events/dates-and-times/editor-date/filter-bar/index.js
```diff
@@ -17,6 +20,7 @@ const withPaginatedDateEntitiesListAndFilterBar = (
 	paginationConfig = {}
 ) => createHigherOrderComponent(
 	compose( [
+		withDefaultDatesListSettings,
```
nerrad: Is this still supposed to be here?

assets/src/editor/events/dates-and-times/editor-date/filter-bar/with-date-entities-list-filter-bar.js
```diff
@@ -61,6 +61,10 @@ export default createHigherOrderComponent(
 		setDatesListView,
 		setDatesGridView,
 		setFilteredDateEntities,
+		defaultDatesListView,
+		defaultDatesListPerPage,
+		setDefaultDatesListView,
+		setDefaultDatesListPerPage,
```
nerrad: are all these props still supposed to be here?

assets/src/editor/events/dates-and-times/editor-date/filter-bar/with-date-entities-list-filter-state.js
nerrad: Are all these props still supposed to be here?

assets/src/editor/events/tickets/editor-ticket/filter-bar/index.js
```diff
@@ -1,7 +1,10 @@
 /**
  * External imports
  */
-import { withEntityPagination } from '@eventespresso/higher-order-components';
+import {
+	withDefaultTicketsListSettings,
```
nerrad: Is this still supposed to be here?

assets/src/editor/events/tickets/editor-ticket/filter-bar/with-ticket-entities-list-filter-bar.js
```diff
@@ -65,6 +65,10 @@ const withTicketEntitiesListFilterBar = createHigherOrderComponent(
 		setTicketsListView,
 		setTicketsGridView,
 		setFilteredTicketEntities,
+		defaultTicketsListView,
+		defaultTicketsListPerPage,
+		setDefaultTicketsListView,
+		setDefaultTicketsListPerPage,
```
nerrad: Are these props still supposed to be here?

assets/src/editor/events/tickets/editor-ticket/filter-bar/with-ticket-entities-list-filter-state.js
```diff
@@ -23,8 +23,10 @@ export default createHigherOrderComponent(
 				displayTicketDate = 'start',
 				isChained = true,
 				searchTicketName = '',
-				ticketsPerPage = 6,
-				ticketsView = 'grid',
+				ticketsPerPage,
+				ticketsView,
+				defaultTicketsListView,
+				defaultTicketsListPerPage,
```
nerrad: Are these props still supposed to be here?  Should the original be restored?

assets/src/higher-order-components/filter-bar/entity-list-filter-bar.js
```diff
@@ -93,50 +102,88 @@ class EntityListFilterBar extends Component {
...
-			onClick={ setGridView }
+			className={ 'ee-save-filter-settings' }
+			icon={ 'admin-generic' }
+			tooltip={ __(
+				'click to save current "per page" and "view" settings',
+				'event_espresso'
+			) }
+			onClick={ () => {
+				setDefaultView( view );
+				setDefaultPerPage( perPage );
```
nerrad: Is the handling for setDefaultView and setDefaultPerPage still supposed to be here?

assets/src/higher-order-components/filter-bar/with-entity-list-filter-bar.js
```diff
@@ -41,6 +41,10 @@ export default createHigherOrderComponent(
 					setPerPage,
 					setListView,
 					setGridView,
+					defaultView,
+					setDefaultView,
+					defaultPerPage,
+					setDefaultPerPage,
```
nerrad: Are these props still supposed to be here?

assets/src/higher-order-components/filter-bar/with-entity-list-filter-state.js
```diff
@@ -19,8 +19,8 @@ export default createHigherOrderComponent(
 		withSelect( ( select, ownProps ) => {
 			const {
 				searchText = '',
-				perPage = 6,
-				view = 'grid',
+				perPage,
+				view,
```
nerrad: should this be restored?

assets/src/higher-order-components/index.js
```diff
@@ -4,6 +4,7 @@ export {
 	PaginatedEntityListWithFilterBarAndState,
 	EntityListFilterBar,
 } from './filter-bar';
+export * from './local-storage';
```
nerrad: is this supposed to still be here (along with all the related files)?

admin_pages/events/Events_Admin_Page.core.php
tn3rb: ahh I see what you mean now... we could but this code hasn't been touched in years and the chances of it getting updated in another pull are probably nill so I'm not going to go to the extra trouble to prevent something that is just not going to happen.

tn3rb:
> @tn3rb Did you miss pushing changes.

eeeeeyaa.... not sure what happened... all of my changes that I had meant to push were sitting in an auto-stash ?!?!?
It was the end of the day on Friday when it happened, so I must have not been paying attention and did something stupid.

changes have been committed and branch has been rebased onto latest master

assets/src/editor/events/dates-and-tickets-metabox/dates-and-tickets-metabox.js
```diff
@@ -91,13 +92,11 @@ const MetaBox = ( { eventId } ) => {
 	);
 };
 
-const DatesAndTicketsMetabox = withEditorInitialization( MetaBox );
-
-MetaBox.propTypes = {
+DatesAndTicketsMetabox.propTypes = {
```
nerrad: Hmm, I think we should have the propTypes defined for the composed component (which I know wasn't the case originally either).

assets/src/editor/events/dates-and-times/editor-date/filter-bar/with-date-entities-list-filter-bar.js
```diff
@@ -45,7 +45,7 @@ export const getFilteredDateEntitiesList = ( dateEntities, showDates, datesSorte
  * @param {Object} EntityList
  * @return {Object} EntityList with added DateListFilterBar
  */
-export default createHigherOrderComponent(
+const withDateEntitiesListFilterBar = createHigherOrderComponent(
```
nerrad: curious why this change?  It was unnecessary?

assets/src/editor/events/dates-and-tickets-metabox/dates-and-tickets-metabox.js
tn3rb: what are the benefits and/or what is the logic behind doing that?
Those props are part of of the component, not the HOC, so logically (to me) it doesn't make sense to associate them with the HOC.

As well, from a troubleshooting point of view, if a notice appeared in the editor saying there was a bad prop type for `withEditorInitialization` one might open up `/assets/src/editor/events/initialization/index.js` instead of this file, which would potentially just cause confusion and waste time by forcing a search for the real source of the problem. Leaving it as is will indicate an issue with the props for **this** component and **this** file.

assets/src/editor/events/dates-and-times/editor-date/filter-bar/with-date-entities-list-filter-bar.js
tn3rb: you're right... but that's kinda been the convention in every other file so I just changed it to match everything else.

assets/src/editor/events/dates-and-tickets-metabox/dates-and-tickets-metabox.js
nerrad: I'm not saying you add the prop types on to `withEditorInitialization`.  I'm saying you add the PropTypes to the **composed** component (`withEditorInitialization( DatesAndTicketsMetabox )` ).  The composed component is what gets exported for the public api, so imo that's what should have the propTypes. Does that clarify?

tn3rb: I realize that, although my comment about a console warning appearing for `withEditorInitialization` does look like I thought you meant to define the prop types there. I was assuming that a call stack for any warning would indicate that the error occurred when attempting to render `withEditorInitialization` even though it wasn't happening in the file that contains that HOC. I could be completely wrong about that and any notices would be 100% clear that the issue was in `dates-and-tickets-metabox.js` but I assumed that was not the case and that the potential for confusion exists.

So that said, I still don't see the logic or any benefit to defining the prop types for the composed component.

> re: The composed component is what gets exported for the public api,

sure that's currently the case, but what if that changes? What if another HOC gets added and we switch to using compose?
Similar to my comment below, what I'm doing here is in line  with the way we have been doing things in every other file, so why the change here? If there is a good reason to make this change then are we going to do so in every other composed component?

nerrad: - The public api exposes `withEditorInitialization( DatesAndTicketsMetabox )`.  So usage will be `<ComposedComponent eventId={ 10 } />`.  I think any errors should be on what is publicly exposed.

> but what if that changes? What if another HOC gets added and we switch to using compose?

Then any additional prop type would still be defined on the composed component.

> Similar to my comment below, what I'm doing here is in line with the way we have been doing things in every other file, so why the change here? If there is a good reason to make this change then are we going to do so in every other composed component?

While consideration _should_ be given for what has already been done, I don't think that precludes noticing things in reviews and improving what's being done.  I noticed this, commented on it and made a suggestion for improving it.  We definitely don't _have_ to go through and make the changes everywhere else (which I don't think are as big a deal as you might think... proptypes themselves have been used inconsistently so far), but I do think its worthwhile to make any appropriate changes as they are spotted.

With that said, I'll reiterate what I said in my approval comment:

>  I'll leave it up to you whether you action on the comments.

PR #1370, opened 2019-06-27
Fix Ticket Assignments Manager Display Issues

tn3rb:
The Ticket Assignments Manager was experiencing some display issues with regards to CUIDs being displayed. They were previously being filtered out by parsing IDs as integers which removed string values but that was causing other problems. This issue was fixed by changing how and where the IDs were being displayed.

As well, the Ticket Assignments Manager could easily become visually overwhelming when large numbers of dates and tickets were being loaded. To remedy this issue I have added some simple filters for removing expired and archived dates and tickets from the grid.

## How has this been tested


tn3rb:
did you try an event that didn't have expired dates and tickets?

nerrad:
> did you try an event that didn't have expired dates and tickets?

No I did not.  I currently start all my tests with expired dates and tickets because it's a case that's easy to miss if we're only testing with active/future events/tickets.

Unfortunately the latest branch still doesn't work.  Here's a screencast: https://cl.ly/a04ba81a0623

In the one place I see something that looks like an interface where I might be able to assign something (when triggering TAM from a ticket), it's not a clickable area.

I should note the following about my test environment:

- I'm running with latest version of Google Chrome.
- I'm testing with the Gutenberg plugin active.
- To ensure this is something specific to this branch, I've tested against a build on the current main EDTR branch (`editor-dates-and-tickets-refactor`) and everything is working fine there.

nerrad:
So I realized that the reasons why nothing was showing for me was because I needed to toggle the filters.  However there's still other issues.

- chosen toggles do not persist between opening/closing TAM for the session.
- I think the TAM view should inherit whatever filtered options are set for the main view.
- There's inconsistency in what appears in my screencast when toggling TAM from a single ticket.  i.e. with the default filters, _some_ data still shows when it probably should be expected that _no_ data shows.

nerrad:
Still seem to be a few issues:

**When toggling TAM from the context of an expired date item**

expired date item view

**During the creation of a new date (that is expired)**

expired date creation flow

**When toggling TAM from the context of a not on sale ticket**

not on sale ticket flow

**During the creation of a new ticket**

creating new ticket

**Triggerd by All Ticket Assignments Button**

from all ticket assignments button flow

I also wonder if the "Update Ticket Assignments" or "Cancel" buttons should even show (or maybe be disabled?) when there is nothing displayed?

tn3rb:
> re: margins 

good catch, I had checked really small screens and larger laptop screens but nothing in between. This is fixed now.

> re: multiple notices

ya I think if someone is doing multiple things incorrectly, then multiple notices is okay, and combining them into one big message runs the risk of the user reading it once and then not noticing if it changes.

> re: ticket info appearing but no assignment button

I was okay with that, but have added some logic so that no UI appears unless the user is capable of acting on it

> re: I also wonder if the "Update Ticket Assignments" or "Cancel" buttons should even show (or maybe be disabled?) when there is nothing displayed?

I've left the "Cancel" button as is, because it can still be used to close the modal window, but have disabled the "Submit" button if there are no dates or tickets available to act on. Plz also see: https://github.com/eventespresso/event-espresso-core/pull/1374

assets/src/editor/events/ticket-assignments-manager/ticket-assignments-manager-modal.js
```diff
@@ -637,6 +769,66 @@ export default compose( [
...
+		dateEntities,
+		ticketEntities,
+		...otherProps
+	} ) => {
+		const unfilteredDates = dateEntities;
+		const unfilteredTickets = ticketEntities;
+		const [ showArchivedDates, setShowArchivedDates ] = useState( false );
+		const [ showExpiredDates, setShowExpiredDates ] = useState( false );
+		const [ showArchivedTickets, setShowArchivedTickets ] = useState( false );
+		const [ showExpiredTickets, setShowExpiredTickets ] = useState( false );
```
nerrad: When we get this amount of state setters/values, I think it might work a bit better to implement [`useReducer`](https://reactjs.org/docs/hooks-reference.html#usereducer) especially for actions that belong in the same "type" of ui/ux behaviour (which in this case would be controlling filters for displayed data.

Example:

Outside the HOC you can define the reducer:

```js
const DEFAULT_STATE = {
	showArchivedDates: false,
	showExpiredDates: false,
	showArchivedTickets: false,
	showExpiredTickets: false,
}

const filterReducer = ( state, action ) => {
	switch ( action.type ) {
		case 'SHOW_ARCHIVED_DATES':
			return { ...state, showArchivedDates: ! state.showArchivedDates };
		case 'SHOW_EXPIRED_DATES':
			return { ...state, showExpiredDates: ! state.showExpiredDates };
		case 'SHOW_ARCHIVED_TICKETS':
			return { ...state, showArchivedTickets: ! state.showArchivedTickets };
		case 'SHOW_EXPIRED_TICKETS':
			return { ...state, showExpiredTickets: ! state.showExpiredTickets };
	}
	throw new Error( sprintf( 'The action %s has no handling implemented', action.type ) );
};
```
Then inside this HOC you would define the following instead of the `useState` calls:

```js
const [ filterState, dispatch ] = useReducer( filterReducer, DEFAULT_STATE );
```

Then you'd just pass through `dispatch` to the wrapped component (or you could name it something like `dispatchFilterState`).  Then in components implementing this you'd just do something like `dispatch( { type: 'SHOW_ARCHIVED_DATES' } );` (or if you rename, `dispatchFilterState`) on whatever toggles the action.

Advantages:

- keep state managed in a single object
- less props passed through (just a single dispatch action)
- you could move the logic for how the `show*` value is determined via passed in arguments to the dispatch (for example you could pass arguments on the action object that help calculate the state change.
- allows for easier updating of additional filter state down the road without having to add additional props everywhere.

assets/src/editor/events/tickets/editor-ticket/price-calculator/ticket-price-calculator-form-data-map.js
```diff
@@ -5,6 +5,7 @@ import { isArray, isEmpty, uniq } from 'lodash';
 import { isModelEntityOfModel } from '@eventespresso/validators';
 
 import { TICKET_PRICE_CALCULATOR_FORM_INPUT_PREFIX } from './constants';
+import { shortenCuid } from '../../../../utils';
```
nerrad: `shortenCuid` seems like something that would be good to move into the  `@eventespresso/helpers` module.  I'd prefer it done in a separate pull so that any place using it can be updated in that pull and unit tests could also be added for it.

assets/src/components/ui/responsive-table/table-row.js
```diff
@@ -23,15 +23,14 @@ const TableRow = ( {
...
-	rowType += 'RowClass';
 	htmlId = htmlId ?
 		`${ htmlId }-row-${ rowNumber }` :
 		`ee-rTable-row-${ rowNumber }`;
 	htmlClass = classNames(
 		htmlClass,
 		`ee-rTable-${ rowType }-row`,
 		`row-${ rowNumber }`,
-		classes[ rowType ],
+		classes[ `${ rowType }RowClass` ],
```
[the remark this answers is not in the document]
tn3rb: I can tackle fixing those classes in another pull. 
I only had to make that change in order to fix some of the display issues for TAM

assets/src/editor/events/ticket-assignments-manager/ticket-assignments-manager-modal.js
tn3rb: me likey

assets/src/editor/events/tickets/editor-ticket/price-calculator/ticket-price-calculator-form-data-map.js
tn3rb: sure, will add to EDTR project board

assets/src/editor/events/ticket-assignments-manager/filters/with-ticket-assignments-filters.js
```diff
@@ -0,0 +1,156 @@
...
+						showArchivedTickets={ showArchivedTickets }
+						showExpiredTickets={ showExpiredTickets }
+						toggleArchivedTickets={
+							() => setFilter( showArchivedTicketsAction )
+						}
+						toggleExpiredTickets={
+							() => setFilter( showExpiredTicketsAction )
+						}
+						showDateFilters={ showDateFilters() }
+						showTicketFilters={ showTicketFilters() }
```
nerrad: hmm better, but I was still thinking we could reduce the amount of props passes through to the Filter components.  Think of cases where there may be additional filter state passed through (right now the Filter components are hardcoded to only support toggling archived or expired state)?  For now, in the interest of keeping momentum and getting this pull merged we can just use it as is as I don't have the energy at the moment to propose alternatives 😄

assets/src/editor/events/ticket-assignments-manager/filters/with-ticket-assignments-filters.js
```diff
@@ -0,0 +1,156 @@
...
+				showExpiredTickets
+			),
+			[
+				ticketEntity,
+				filteredTicketEntities,
+				unfilteredTicketCount,
+				showArchivedTickets,
+				showExpiredTickets,
+			]
+		);
```
nerrad: This (and the other usages of `useCallback` here)  looks like something that could be calculated on render as there is no runtime argument passed in on the callback. so likely could simply be implementing `useMemo` instead.

assets/src/editor/events/ticket-assignments-manager/filters/with-ticket-assignments-filters.js
tn3rb: There's some interplay between the different filters so I've had no choice but to "lift state" to the highest component in the tree (or lowest? ie: closer to trunk vs top of tree!?!) so at some point we have to pass the props down. 

The only way to reduce things more would be to bundle the props into an object and pass that, but imho that's not good because it hides the implementation details / component signature and potentially interferes with memoization.

I think the biggest win with this branch is the fact that the filters are at least all self contained and the only things being passed to `TicketAssignmentsManagerModal` is the new `withTicketAssignmentsFilters` HOC and a single prop for displaying the filters. If we need to add more filters or completely change their configuration or whatever, the effects on TAM will be minimal if anything at all.

assets/src/editor/events/ticket-assignments-manager/filters/with-ticket-assignments-filters.js
tn3rb: switched to `useMemo` in some places but kept `useCallback` in others to keep function construction from triggering re-renders

assets/src/editor/events/ticket-assignments-manager/filters/date-filters.js
```diff
@@ -14,22 +22,25 @@ const { FormColumn } = twoColumnAdminFormLayout;
...
-	showTicketFilters,
+	dateFiltersOffset,
+	setFilter,
 } ) => {
-	const dateFiltersOffset = showDateFilters && showTicketFilters ? 2 : 7;
-	return showDateFilters ? (
+	const toggleArchivedDates = useCallback(
+		() => setFilter( showArchivedDatesAction ),
+		[]
+	);
```
nerrad: This should have `showArchivedDatesAction` and `setFilter` as dependencies

assets/src/editor/events/ticket-assignments-manager/filters/date-filters.js
```diff
@@ -14,22 +22,25 @@ const { FormColumn } = twoColumnAdminFormLayout;
...
+	const toggleExpiredDates = useCallback(
+		() => setFilter( showExpiredDatesAction ),
+		[]
+	);
```
nerrad: This should have `setFilter` and `showExpiredDatesAction` as dependencies.

assets/src/editor/events/ticket-assignments-manager/filters/ticket-filters.js
```diff
@@ -14,22 +22,25 @@ const { FormColumn } = twoColumnAdminFormLayout;
...
-	showTicketFilters,
+	ticketFiltersOffset,
+	setFilter,
 } ) => {
-	const ticketFiltersOffset = showDateFilters && showTicketFilters ? 0 : 7;
-	return showTicketFilters ? (
+	const toggleArchivedTickets = useCallback(
+		() => setFilter( showArchivedTicketsAction ),
+		[]
+	);
```
nerrad: This should have `setFilter` and `showArchivedTicketsActions` as dependencies.

assets/src/editor/events/ticket-assignments-manager/filters/ticket-filters.js
```diff
@@ -14,22 +22,25 @@ const { FormColumn } = twoColumnAdminFormLayout;
...
+	const toggleExpiredTickets = useCallback(
+		() => setFilter( showExpiredTicketsAction ),
+		[]
+	);
```
nerrad: This should have `setFilter` and `showExpiredTicketsAction` as dependencies.

assets/src/editor/events/ticket-assignments-manager/filters/with-ticket-assignments-filters.js
nerrad: > kept useCallback in others to keep function construction from triggering re-renders

I'm curious about this statement, can you expand on what you mean by that?  `useCallback` should only be used for when you want to construct a memoized function to pass through as a prop (to avoid unnecessary renders) and the dependency arguments provided to it control when it the function is recreated... otherwise reference equality on the prop it's added to will remain the same.  `useMemo` is used when you want the value to be calculated _immediately_ on render and that same value is used across re-renders unless any of the provided dependencies change at which point it's recalculated.

assets/src/editor/events/ticket-assignments-manager/filters/date-filters.js
tn3rb: even if those two will not change?
are the setters returned from `useReducer` not referentially equal from render to render?

assets/src/editor/events/ticket-assignments-manager/filters/date-filters.js
tn3rb: see response above

assets/src/editor/events/ticket-assignments-manager/filters/ticket-filters.js
tn3rb: see response above

assets/src/editor/events/ticket-assignments-manager/filters/ticket-filters.js
tn3rb: see response above

tn3rb:
> ```kept useCallback in others to keep function construction from triggering re-renders```
> 
> I'm curious about this statement, can you expand on what you mean by that? useCallback should only be used for when you want to construct a memoized function to pass through as a prop (to avoid unnecessary renders)...

I mean the exact same thing that you stated...

`to keep function construction from triggering re-renders` === `to construct a memoized function to pass through as a prop (to avoid unnecessary renders)`

I just left out "pass through as a prop" because that is not always the case, ex:

```
const aCallback = useCallback( () => doSomething(), [] );
const buttonA = useMemo( () => <button onClick={ aCallback } />, [ aCallback ] );
const buttonB = useMemo( () => <button onClick={ aCallback } />, [ aCallback ] );
```

assets/src/editor/events/ticket-assignments-manager/filters/date-filters.js
nerrad: The thing is, they are outside this component and this component has no awareness of where those values are coming from.  So if you don't add the dependencies and this component just happened to get used somewhere else (or a future maintainer uses this as a guideline for copy/pasting) that could add to unexpected behaviour.  So, yes I would add the dependencies, it doesn't hurt and it prevents the _potential_ problems I outlined.

tn3rb: kk fair enuf... will do...

PR #1371, opened 2019-06-27
EDTR: Implement submit-less edit date forms.

nerrad:

> **NOTE:** The work in this pull requires the Gutenberg plugin to be active.  I tested against the latest released version of the plugin.  When it is merged, we will need to develop with the plugin active going forward.

The work in this pull implements submit-less edit date forms when creating new dates or editing existing dates.  This means that changes to the date is reflected immediately in the dom (but won't persist until persist action is triggered).

Also in this pull, the Ticket Assignments Modal will automatically pop up after the initial edit during the create date flow (but not popup when editing an existing date entity).

## How has this been tested

* [x] Try creating a date and the flow after creating the date.  Ensure that editing inputs etc works as expected and the resulting entity in the grid/list view appears as expected.
* [x]  Try editing a date and ensure any changes reflect in the list after closing it.


assets/src/editor/events/dates-and-times/editor-date/actions-menu/editor-date-entity-actions-menu.js
```diff
@@ -3,196 +3,184 @@
...
+			sidebarMenuItems.push(
+				viewTicketsMenuItem()
+			);
+			return applyFilters(
+				'FHEE__EditorDateEntityActionsMenu__SidebarMenuItems',
+				sidebarMenuItems,
+				dateEntity
+			);
+		},
+		[ mainDropDownMenu, editDateMenuItem, viewTicketsMenuItem ]
```
tn3rb: this is currently using `applyFilters()` and I know you have suggested changing this to utilize slot fill instead,
but regardless of which method is employed, how does it affect the dependencies for `useCallback()` ?
What if the listed dependencies don't change but the conditions in the logic filtering `sidebarMenuItems` do, and it needs to change what is returned?
For example, what if REM changes the state for what it is injecting here while the listed dependencies remain the same?

I'm not expecting you to have an answer and am content with having to just wait and see what happens.

assets/src/editor/events/dates-and-times/editor-date/editor-date-entities-list.js
```diff
@@ -1,15 +1,19 @@
...
-import { compose } from '@wordpress/compose';
-import { withDispatch } from '@wordpress/data';
-import { Component } from '@wordpress/element';
+import { compose, createHigherOrderComponent } from '@wordpress/compose';
+import { useDispatch } from '@wordpress/data';
+import {
+	useEffect,
+	useMemo,
+	useState,
+	useCallback, useReducer,
```
tn3rb: needs formatting fix (surprised linting didn't catch this)

assets/src/editor/events/dates-and-times/editor-date/grid-view/editor-date-entity-details.js
```diff
@@ -13,57 +13,58 @@ import {
...
+				'ee-editor-date-name-heading',
+				{
+					'ee-long-title': dateEntity.name && dateEntity.name.length > 40,
+				}
+			);
+			return (
+				<h1 className={ htmlClass }>
+					<InlineEditInput
+						htmlId={ `event-date-name-${ dateEntity.id }` }
+						type="text"
```
tn3rb: noticed earlier that you were changing prop formatting like this to `type={ 'text' }` 
is that something we want to follow as a style guide?
I originally was just using `"` for string values, but have switched to braces and `'` for better consistency

assets/src/editor/events/dates-and-times/editor-date/actions-menu/editor-date-entity-actions-menu.js
nerrad: Your observations/questions are good.  I had the same thoughts when I was implementing this and forgot to leave inline comments about it.  It's a good example why implementing filters in reactive components is difficult because there is no way to reliably handle scenarios where things hooking into a filter may  introduce other dependencies in react hooks etc.  I decided to leave as is for now because with the refactor to slot/fill it will become a moot issue.

assets/src/editor/events/dates-and-times/editor-date/grid-view/editor-date-entity-details.js
nerrad: No real reason other than I get in a habit of using braces and phpStorm (for me at least) automatically adds them when it detects I'm doing a jsx element.

PR #1377, opened 2019-07-02
Add venue model/entity fixtures

nerrad:
This pull adds various fixtures for venue data to use in javascript tests.   This has zero impact on released code.


nerrad:
travis is having some build problems, since the files here would only be covered under the linting and js tests (which passed in the build) I'm going to go ahead and merge this.

PR #1380, opened 2019-07-02
EDTR: Refactor higher order components to hooks (event, event date relation, event venue)

nerrad:

I initially started out with implementing a hook for using instead of `withUpdateEventDateRelation` but then realized:

- to properly implement, I'd need a hook to replace `withEditorEventEntity`.
- to properly implement, I'd need a hook to replace `withEventVenueEntity`.
- Part of the implementation of the new hooks required refactoring a class component to a functional component and simplifying the composition of some components meant replacing the above hocs with hooks.

So, in this pull the following three new hooks were created:

- `useEditorUpdateEventDateRelation` - replaces `withEditorUpdateEventDateRelation`
- `useEditorEventEntity` - replaces `withEditorEventEntity`
- `useEventVenueEntity` - replaces `withEventVenueEntity`.

The new hooks were added to a `hooks` subfolder and also have unit tests covering them.

## How has this been tested

* [x] Unit tests covering hooks
* [x] Tested built files on the editor manually to ensure there was no regressions to behaviour.


assets/src/editor/events/hooks/use-event-venue-entity.js
```diff
@@ -0,0 +1,55 @@
...
+
+/**
+ * A custom hook for retrieving the venue related to the given event
+ *
+ * @param {BaseEntity} eventEntity  An instance of an event entity.
+ *
+ * @return {VenueResponse} An object containing details about the retrieved
+ *                         related venue to the event.
+ */
+const useEventVenueEntity = ( eventEntity ) => {
```
nerrad: I think one of the things I'd like to do in a separate pull(s) when the new editor gets close to "done" (there's a number of refactors to do yet) is review what can go into other modules.  We may want to create a new package module named `@eventespresso/component-hooks` (as opposed to just "hooks" to avoid confusion with WordPress js hooks) for holding hooks like this which have a broader application.

assets/src/editor/events/dates-and-times/editor-date/list-view/editor-date-entity-list-item.js
```diff
@@ -29,26 +28,33 @@ const { getBackgroundColorClass } = dateTimeModel;
...
+					return 'ee-datetime-active';
+				case 'DTE' :
+					return 'ee-datetime-expired';
+				case 'DTS' :
+					return 'ee-datetime-sold-out';
+				case 'DTU' :
+					return 'ee-datetime-upcoming';
+			}
+		},
+		[ dateEntity ]
```
tn3rb: Would it not be better to simply use `dateEntity.status` for the dependency?
ie: it wouldn't be necessary to recalculate this if the date's name or other property changed

assets/src/editor/events/hooks/use-editor-event-entity.js
```diff
@@ -0,0 +1,32 @@
...
+ */
+const useEditorEventEntity = ( existingEventEntity = null ) => {
+	const eventEntity = useSelect( ( select ) => {
+		// if already have an event entity, then return it.
+		if ( isModelEntityOfModel( existingEventEntity, 'event' ) ) {
+			return existingEventEntity;
+		}
+		const events = select( 'eventespresso/core' ).getEvents();
+		return events.length > 0 ?
+			events.slice( -1 ).pop() :
```
tn3rb: this is copying the last element of the `events` array into a new array then popping the last element off...

why not just `events.pop()` ? ie: is there a need to preserve the `events` array here?

assets/src/editor/events/hooks/use-editor-update-event-date-relation.js
```diff
@@ -0,0 +1,55 @@
...
+ * @return {function}  A function for updating the event date relation.
+ */
+const useEditorUpdateEventDateRelation = ( eventEntity = null ) => {
+	eventEntity = useEditorEventEntity( eventEntity );
+	const { createRelation } = useDispatch( 'eventespresso/core' );
+	const updateEventDateRelation = useCallback( ( dateEntity ) => {
+		if ( ! isModelEntityOfModel( eventEntity, EVENT ) ) {
+			throw new Error(
+				__(
+					'Unable to process the Event Date Entity form because an invalid Event Entity was supplied. ',
```
tn3rb: should trim space at end of string

assets/src/editor/events/hooks/use-editor-update-event-date-relation.js
```diff
@@ -0,0 +1,55 @@
...
+					'event_espresso'
+				)
+			);
+		}
+		if ( ! isModelEntityOfModel( dateEntity, DATETIME ) ) {
+			throw new Error(
+				__(
+					'Unable to process the Event Date Entity form because an invalid Date Entity was supplied. ',
```
tn3rb: should trim space at end of string

assets/src/editor/events/hooks/use-event-venue-entity.js
tn3rb: totally agree

assets/src/editor/events/hooks/use-event-venue-entity.js
```diff
@@ -0,0 +1,55 @@
...
+	const { venueEntity, venueEntityLoaded } = useSelect( ( select ) => {
+		const { getRelatedEntities, hasFinishedResolution } = select(
+			'eventespresso/core'
+		);
+		if ( isModelEntityOfModel( eventEntity, 'event' ) ) {
```
tn3rb: what about displaying a warning if the incoming event is invalid to help with troubleshooting?
as is this would just return the default object with no indication of why the venue isn't loading

assets/src/editor/events/dates-and-times/editor-date/list-view/editor-date-entity-list-item.js
nerrad: Ya good point.  It also wouldn't be necessary to recalculate if the dateEntity changed, but the status was the same.

assets/src/editor/events/hooks/use-editor-event-entity.js
nerrad: I think I had that originally but the tests exposed issues with mutating the returned array (as `events.pop()` would actually _change_ the length of the original array).  Hence switching to doing this.

assets/src/editor/events/hooks/use-editor-update-event-date-relation.js
nerrad: oops, good catch!

assets/src/editor/events/hooks/use-event-venue-entity.js
nerrad: Ya good point, will do.

assets/src/editor/events/dates-and-times/editor-date/list-view/editor-date-entity-list-item.js
tn3rb: oh HA! ya didn't even think of that!
ya sold out is sold out regardless of what date you're referring to.

assets/src/editor/events/hooks/use-editor-event-entity.js
tn3rb: why are the tests concerned about the length of the array or whether it is mutated?

like I said above, is there a need to preserve the events array here?

Looking at the full code for this function, it appears that this would only run once for any individual event, meaning the events array is basically just an unimportant stepping stone and can be disposed of once a `venueEntity` has been retrieved.

nerrad: > why are the tests concerned about the length of the array or whether it is mutated?

I'm not just talking about automated tests.  I'm talking about the user land testing I did as well.  The selector for getEvents is memoized which means it returns the same array for events if the state has not changed.  Thus if  you mutate that array, future calls to the selector will return the mutated value.  I didn't realize that until I tested and discovered that and thus made this change.

>  it appears that this would only run once for any individual event, meaning the events array is basically just an unimportant stepping stone and can be disposed of once a venueEntity has been retrieved.

I'm not sure where that reasoning comes from.  This is a hook that could be implemented anywhere, so the assumption that it would only run once is incorrect - it depends on how and where its implemented.

PR #1391, opened 2019-07-08
Add static function for creating a Money VO from a number or string.

nerrad:

There will be cases in apps where the value for a money value object will come from a form input and will thus be in a string format (and might have the currency sign on it).  The helper added in this pull  assists with converting that value to the value object.

Usage:

```js
import { SiteCurrency as Currency, Money } from '@eventespresso/value-objects';

const value = '$10.10';
const moneyValue = Money.fromMoneyValue( value, Currency );

// outputs 10.1
console.log( moneyValue.toNumber() );
```
## How has this been tested
* [x] Changes are covered by unit tests and this has no impact on existing released user interfaces.

## Checklist

* [x] My code is tested.

### DIFF
--- a/assets/src/vo/money.js
+++ b/assets/src/vo/money.js
@@ -7,6 +7,7 @@ import isShallowEqual from '@wordpress/is-shallow-equal';
 import { Exception } from '@eventespresso/eejs';
 import { isEmpty } from 'lodash';
 import { instanceOf } from '@eventespresso/validators';
+import { sprintf } from '@eventespresso/i18n';
 
 /**
  * Asserts if incoming value is an instance of Money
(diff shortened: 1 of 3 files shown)


assets/src/vo/money.js
```diff
@@ -515,4 +516,45 @@ export default class Money {
...
+	 * @param {string|number} moneyValue
+	 * @param {Currency} currency
+	 *
+	 * @return {Money} An instance of a money value object
+	 */
+	static fromMoneyValue = ( moneyValue, currency ) => {
+		assertCurrency( currency );
+		// detect if incoming value has a currency sign not matching provided
+		// currency.  This doesn't provide full protection from improper
+		// values sent in but is an initial safeguard.
```
nerrad: This check:

- throws an error if the incoming money value has more than one non numeric string (not counting commas or periods).
- throws an error if there is only one non numeric (not counting commas or periods) string that does not match the provided currency value object currency sign value.

Thus it's still possible to end up with an unexpected value if something like `10,10` is provided for a currency that uses `.` as the decimal place separator.  I think that would be an extreme edge case though so not worrying about it here.

assets/src/vo/money.js
```diff
@@ -515,4 +516,45 @@ export default class Money {
...
+		if ( typeof moneyValue === 'string' ) {
+			const match = moneyValue.match( /[^\d\.\,\s]+/ );
+			if ( match && match[ 0 ] !== currency.sign ) {
```
tn3rb: is this assuming the currency sign is the first character in the string?
what if the currency sign is at the end of the string?

assets/src/vo/money.js
```diff
@@ -515,4 +516,45 @@ export default class Money {
...
+				// returned which is likely the currency symbol.  Otherwise,
+				// give a more generic message.
+				const message = match[ 0 ].length === 1 ?
+					sprintf(
+						'The provided money value has a %1$s sign in it, but the provided currency value object defines %2$s as the currency sign.',
+						match[ 0 ],
+						currency.sign
+					) :
+					sprintf(
+						'The provided money value has non numeric strings in it (%1$s), please double-check the value.',
```
tn3rb: might be better as :

> The provided money value has non numeric characters in it

as only programmers will understand "strings"

assets/src/vo/money.js
nerrad: Nope, `match[ 0 ]` will be whatever gets returned by the regex expression (which will match the currency sign no matter where it is in the string).

assets/src/vo/money.js
nerrad: > as only programmers will understand "strings"

Only programmers should be seeing this (which is why its not going through `wp.i18n` too).

nerrad: This is a low level api, so I think any higher level apis (in ui/ux etc) should catch any errors and pass on any user friendly message _if needed/desired_.

nerrad:
The tests passed on travis but the status is  not updating in the pull, merging anyways.

PR #1401, opened 2019-07-09
Refactor Entity Actions Menus and Remove Reliance on Filter Hooks

tn3rb:

Currently the `EditorDateEntityActionsMenu` and `EditorTicketActionsMenu` components were utilizing `applyFilters()` for injecting new menu items from external sources which is fragile and not a great way to do things. As well, these components included a number of functions for rendering menu items that would work better as separate functional components that could then be imported and memoized. This branch makes these changes.

## How has this been tested

- monkey based browser testing
- REM was changed to use `registerEntityActionsMenuItem()` instead of the `FHEE__EditorDates__EditorDateSidebar__SidebarMenuItems` filter


assets/src/editor/events/hocs/with-editor-event-date-entities.js
```diff
@@ -0,0 +1,34 @@
...
+				);
+				return { dateEntities };
+			}
+			return DEFAULT_OBJECT;
+		}
+	),
+	'withEditorEventDateEntities'
+);
+
+export default withEditorEventDateEntities;
```
nerrad: This component should not be needed.  `withEditorDateEntities` and `useEditorDateEntities` already exist and already return all date entities existing in the state for the editor instance (which will be for a specific event - because only the only dates in the store will be for that event).

tn3rb: hmmmm... ya ok... I didn't see any logic pertaining to the current event being edited, but that would likely be the case regardless... can't think of any cases where that would be different... will change

nerrad: This is related to the work I did in creating the Editor Initialization layer (and why these hocs have the word `Editor` in them).  Essentially in the initialization layer, all the datetimes for the event are fetched and added to the state.  From that point forward any datetime in the state is related to the event being edited, so no need to get the datetimes related to the event (you can instead just get all the datetimes in the store state).

tn3rb:
I looked into the slot/fill pattern and thought the following:

- it's still subject to timing issues just like filters and what I'm doing in this branch. This is why your Provider has to be declared in the root of your application because that needs to happen first or else things don't work. If an addon or third party app loaded prior to our Provider declaration and tried to add a slot it wouldn't work. A similar thing happens if you use `createSlotFill()`. The addon or third party app has to do the following in order to add a `Fill`:
    - Import the registerPlugin method from wp.plugins.
    - Import the slotFill we want from wp.editPost.

    this means that your slot/fill needs to be created **before** that can happen... which is a timing issue!!!

- slot/fill appears unnecessarily complex for what is required. ie: more code to achieve the same thing ("is there some other benefit I'm not seeing?")

re:
> I would implement it for ALL menu items.

registering a menu item via `registerEntityActionsMenuItem()` triggers a re-render because it changes what is returned from `entityActionsMenu()` so I can't do that from within `EditorDateEntityActionsMenu` because it causes recursion. I guess I could try registering the menu items within their own component files and see what happens.

re:
>  It would also remove the need for the indexOffset argument.

that was only done because the `IconMenuItem` component currently requires a `number` index. I could just change that to allow strings and then use the date ID appended to the menu item name or something similar. Was simply trying to avoid changing more files but ya... maybe I'll just do that instead anyways.

tn3rb:
looked into slot/fill some more to try and get it working but it just does not function the way we need it to work in the EDTR. 

The biggest reason is this:

**there is no way to pass data from the position of the Slot to the components that become the Fills**

you can pass props from your components that render the Fill content to the component that renders the Fills themself, but not the other way around.

Here's a good example that illustrates this
see: https://github.com/WordPress/gutenberg/pull/13361/files

The first step is to declare both the Slot and Fill components:

`// packages/edit-post/src/components/sidebar/plugin-document-setting-panel/index.js`
```jsx
/**
 * WordPress dependencies
 */
import { createSlotFill, PanelBody } from '@wordpress/components';
import { compose } from '@wordpress/compose';
import { withPluginContext } from '@wordpress/plugins';
import { withDispatch, withSelect } from '@wordpress/data';

export const { Fill, Slot } = createSlotFill( 'PluginDocumentSettingPanel' );

const PluginDocumentSettingFill = ( {
    isEnabled,
    opened,
    onToggle,
    className,
    title,
    icon,
    children
} ) => {
	if ( ! isEnabled ) {
		return null;
	}
	return (
		<Fill>
			<PanelBody
				className={ className }
				title={ title }
				icon={ icon }
				opened={ opened }
				onToggle={ onToggle }
			>
				{ children }
			</PanelBody>
		</Fill>
	);
};

const PluginDocumentSettingPanel = compose(
	withPluginContext( ( context, ownProps ) => {
		return {
			icon: ownProps.icon || context.icon,
			panelName: `${ context.name }/${ ownProps.name }`,
		};
	} ),
	withSelect( ( select, { panelName } ) => {
		return (
			{
				opened: select( 'core/edit-post' ).isEditorPanelOpened( panelName ),
				isEnabled: select( 'core/edit-post' ).isEditorPanelEnabled( panelName ),
			}
		);
	} ),
	withDispatch( ( dispatch, { panelName } ) => (
		{
			onToggle() {
				return dispatch( 'core/edit-post' )
					.toggleEditorPanelOpened( panelName );
			},
		}
	) ),
)( PluginDocumentSettingFill );

PluginDocumentSettingPanel.Slot = Slot;
export default PluginDocumentSettingPanel;
```

`PluginDocumentSettingFill` is what will wrap the content that gets added for each Fill (represented by the `children` prop) and `PluginDocumentSettingPanel.Slot` is where the Fills will actually appear when the component that uses it gets rendered. Here's where that happens in the `SettingsSidebar` component:

`// packages/edit-post/src/components/sidebar/settings-sidebar/index.js`
```jsx
const SettingsSidebar = ( { sidebarName } ) => (
	<Sidebar
		name={ sidebarName }
		label={ __( 'Editor settings' ) }
	>
		<SettingsHeader sidebarName={ sidebarName } />
		<Panel>
			{ sidebarName === 'edit-post/document' && (
				<>
					<PostStatus />
					<PluginDocumentSettingPanel.Slot />
					<LastRevision />
					<PostLink />
					<PostTaxonomies />
					<FeaturedImage />
					<PostExcerpt />
					<DiscussionPanel />
					<PageAttributes />
					<MetaBoxes location="side" />
				</>
			) }
			{ sidebarName === 'edit-post/block' && (
				<PanelBody className="edit-post-settings-sidebar__panel-block">
					<BlockInspector />
				</PanelBody>
			) }
		</Panel>
	</Sidebar>
);
```

as you can see, `<PluginDocumentSettingPanel.Slot />` does **NOT** receive any props or data from where it is actually used.

But all of the above is only one half of the coin, to actually inject something into that Slot, you do something like this:

`// third party plugin/theme/addon/etc`
```jsx
 const { registerPlugin } = wp.plugins;
 const { PluginDocumentSettingPanel } = wp.editPost;

const MyDocumentSettingTest = () => (
	<PluginDocumentSettingPanel className="my-document-setting-plugin">
 		<p>My Document Setting Panel</p>
 	</PluginDocumentSetting>
 );
 
registerPlugin( 'document-setting-test', { render: MyDocumentSettingTest } );
```

the `className` prop passed to `PluginDocumentSettingPanel` corresponds to the `classname` prop here:

```jsx
const PluginDocumentSettingFill = ( {
    isEnabled,
    opened,
    onToggle,
    className,
    title,
    icon,
    children
} ) => {
```

one could pass ALL of those props from the where the Fill content is created, but in this example they are using HOCs to generate values for the rest of the props.


Anyways, the point is that the one way data flow goes like this:

`Fill Content` => `Fill Component` => `Rendered Slot`

or using the component names in this example:

`MyDocumentSettingTest` => `PluginDocumentSettingFill` => `PluginDocumentSettingPanel.Slot`

**HOWEVER**

in the EDTR (specifically for REM), we need to be able to inject menu items into the action menus for each date entity (and ticket) AND the code that is doing the injecting needs to receive the corresponding `dateEntity` in order to do so. ie: REM needs to know what date entity it is dealing with so that it can retrieve the corresponding recurrence data in order to populate the REM editor. This simply will not work utilizing the Slot/Fill pattern, unless of course there is some hidden functionality allowing this that has not been documented anywhere.

PR #1405, opened 2019-07-10
Add filter hook to allow filtering output of [ANSWER] shortcode

joshfeck:

See forum topic:
https://eventespresso.com/topic/checkboxes-in-email-display-inline-instead-of-block/

## How has this been tested

- [ ] Set up a checkbox or multi select question type
- [ ] Register for the event and select multiple answers
- [ ] With the default Registration Approved message template active, check the email 
The answer output should be the same (comma separated & inline)

- [ ] Now add this filter function:
```
add_filter(
    'FHEE__EE_Question_Shortcodes___parser__answer',
    'the_example_function_for_testing',
    10, 
    3
);
function the_example_function_for_testing($answer, $question, $data) {
    if($question->get('QST_type') == 'CHECKBOX' || $question->get('QST_type') == 'MULTI_SELECT' ) {
            $answer = implode('<br>', $data->get('ANS_value'));
    }
    return $answer;
}
```
And resend the message, the answer output should have each selected answer on its own line.

## Checklist

* [x] My code is tested.

### DIFF
--- a/caffeinated/core/libraries/shortcodes/EE_Question_Shortcodes.lib.php
+++ b/caffeinated/core/libraries/shortcodes/EE_Question_Shortcodes.lib.php
@@ -85,7 +85,12 @@ class EE_Question_Shortcodes extends EE_Shortcodes
                         break;
 
                     default:
-                        $answer = $this->_data->get_pretty('ANS_value', 'no_wpautop');
+                        $answer = apply_filters(
+                            'FHEE__EE_Question_Shortcodes___parser__answer',
+                            $this->_data->get_pretty('ANS_value', 'no_wpautop'),
+                            $question,
+                            $this->_data
+                        );
                         break;
                 }
 
--- a/caffeinated/core/libraries/shortcodes/EE_Question_Shortcodes.lib.php
+++ b/caffeinated/core/libraries/shortcodes/EE_Question_Shortcodes.lib.php
@@ -27,11 +27,11 @@ class EE_Question_Shortcodes extends EE_Shortcodes
      */
     protected function _init_props()
     {
-        $this->label = __('Attendee Shortcodes', 'event_espresso');
-        $this->description = __('All shortcodes specific to attendee related data', 'event_espresso');
+        $this->label = esc_html__('Attendee Shortcodes', 'event_espresso');
+        $this->description = esc_html__('All shortcodes specific to attendee related data', 'event_espresso');
         $this->_shortcodes = array(
-            '[QUESTION]' => __('Will parse to a question.', 'event_espresso'),
-            '[ANSWER]'   => __('Will parse to the answer for a question', 'event_espresso'),
+            '[QUESTION]' => esc_html__('Will parse to a question.', 'event_espresso'),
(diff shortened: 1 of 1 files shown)


caffeinated/core/libraries/shortcodes/EE_Question_Shortcodes.lib.php
```diff
@@ -27,11 +27,11 @@ class EE_Question_Shortcodes extends EE_Shortcodes
...
     {
-        $this->label = __('Attendee Shortcodes', 'event_espresso');
-        $this->description = __('All shortcodes specific to attendee related data', 'event_espresso');
+        $this->label = esc_html__('Attendee Shortcodes', 'event_espresso');
+        $this->description = esc_html__('All shortcodes specific to attendee related data', 'event_espresso');
         $this->_shortcodes = array(
-            '[QUESTION]' => __('Will parse to a question.', 'event_espresso'),
-            '[ANSWER]'   => __('Will parse to the answer for a question', 'event_espresso'),
+            '[QUESTION]' => esc_html__('Will parse to a question.', 'event_espresso'),
+            '[ANSWER]'   => esc_html__('Will parse to the answer for a question', 'event_espresso'),
```
mnelson4: I had thought switching from `__()` to `esc_html__()` invalidated the translated strings, right? If so, I'd just like to confirm you're ok with that.

joshfeck: It does not invalidate strings, nope.

joshfeck: Where it would invalidate strings is if I were to change `__()` to something like `_x()` or `esc_html_x()`.

mnelson4: ooooh ok great. Thanks. FYI @tn3rb and @knazart  ^^ (maybe you were aware, and I was leading you astray with regards to that)

Pebblo:
Can we move this filter to always filter `$answer`?

Meaning the filter applies here:

https://github.com/eventespresso/event-espresso-core/blob/9f9265eb265614cf3092f8986f8b23295f50921f/caffeinated/core/libraries/shortcodes/EE_Question_Shortcodes.lib.php#L97

Right now we only 2 other options in the switch, state and country but in the future we may add more and then possibly need an additional filter(s) so filtering `$answer` its returned covers all bases.

joshfeck:
how's that look @Pebblo ?

Pebblo:
Looks (and works) great! 

Thanks, Josh. I wasn't sure if you wanted to specifically target the case used at first so I didn't change it.

I did have to change your snippet though as `$answer` at the point would always be a string as `get_pretty()` will have already done its thing and also included multi_select inputs

Do you not fancy adding this into core rather than through a filter? (I think the filter should stay but just do the above by default)

joshfeck:
If more users request line breaks instead of commas we can revisit this and possibly change the defaults.

PR #1407, opened 2019-07-11
EDTR: Implement persistence of entities relations in state (fixed #1180)

### LINKED ISSUE #1180: Persist entity changes to the database when the user hits submit
When the user hits submit all the entities edited via the datetime/ticket editor need to run through the models and redux store to persist to the database.

nerrad:

This pull implements a custom react hook for initializing the new dates and tickets editor with a listener hooked into the post editor form submit action so when that is triggered dirty state in the `eventespresso/core` store is persisted to the server.  Currently this is working with the following caveat, it has only been tested against _existing_ events.  

###  New Things:

- new `getRelationModelsQueuedForAddition` and `getRelationModelsQueuedForDeletion` selectors (along with tests)
- new `persistAllAddRelations` and `persistAllDeleteRelations` generator actions (along with tests)
- new `persistAllEntities` generator action (along with tests) 
- new `useEditorPersistence` custom react hook.

### Known Issues:

I may get this fixed before submitting this for review, but in case I don't, I'm documenting the known issues here:

* [x] The publish/update action spinner doesn't show on initial submission click while the persisting of the store is happening.  This should show up immediately so there is a visual indicator to the user that something is happening.  Likely just need to trigger whatever event shows the spinner early (see wp core's `post.js` for clues).
* [x] If you create a ticket and assign it to a new created datetime (thus having a `cuid`), the assignment to the new datetime does not persist, but any assignments to _existing_ datetimes does.  I think this _might_ be a race condition thing (and thus the dispatches happening in the new action generators need to be using `resolveDispatch` not `dispatch` to force asynchronous to synchronous action).

## How has this been tested

* [x] unit tests for new selectors and actions.
* [x] browser based testing of persisting behaviour (watching what happens in the store state and how persisted items show up in the db after save).


assets/src/data/eventespresso/core/actions/persist-entity-generators.js
```diff
@@ -280,11 +280,40 @@ function* persistAllDeletes() {
...
+	} );
+	while ( entitiesByModel.length > 0 ) {
+		const [ modelName, entities ] = entitiesByModel.pop();
+		while ( entities.length > 0 ) {
+			const entity = entities.pop();
+			yield dispatch(
+				CORE_REDUCER_KEY,
+				'persistEntityRecord',
+				modelName,
+				entity
```
tn3rb: maybe this could be 

```js
		while ( entities.length > 0 ) {
			yield dispatch(
				CORE_REDUCER_KEY,
				'persistEntityRecord',
				modelName,
				entities.pop()
			);
```

like you did below in `persistAllAddRelations()` ???

saves the creation of a variable...

assets/src/editor/events/initialization/use-editor-persistence.js
```diff
@@ -0,0 +1,42 @@
...
+		persistAllDeleteRelations,
+		persistAllEntities,
+	} = useDispatch( 'eventespresso/core' );
+	const persistAll = useCallback( async ( event ) => {
+		if ( eventProcessed.current === false ) {
+			event.preventDefault();
+			await persistAllDeletes();
+			await persistAllAddRelations();
+			await persistAllDeleteRelations();
+			await persistAllEntities();
```
tn3rb: wouldn't you need to call `persistAllEntities()` before `persistAllAddRelations()` and `persistAllDeleteRelations()` so that any newly created entities have the correct ID?

assets/src/data/eventespresso/core/actions/persist-entity-generators.js
nerrad: yup good spot 👍

assets/src/editor/events/initialization/use-editor-persistence.js
nerrad: The relation persisting automatically takes care of persisting any dirty entities first before persisting the relations, so the `persistAllEntities` here is just to catch any entities that have no relations.  However, reflecting on your comment, in the context of the current state of the editor, it will likely be more performant if we persist all entities before doing the relation persisting so I'll do that (I do think I need to still fix the bug reported in "known issues" around relation connection before doing this though so it doesn't surface in other contexts).

Should note also, if you find out this logic works okay, you can probably create an aggregate action generator in the store for handling this kind of `persistAllState` behaviour.

assets/src/data/eventespresso/core/actions/persist-entity-generators.js
nerrad: Actually, this is different because I'm using destructure assignment for the entities from the `entitiesByModel` array (which is an array of arrays).  So this needs to stay as is.

tn3rb: just curious...

how does the creation of the `entity` variable affect what's happening in the `entitiesByModel` array?

regardless of whether you assign the return value from `entities.pop()` to the `entity` variable or just call `entities.pop()` directly within the dispatch, you are removing the last entity from the `entities` array (which has already been popped off the end of the `entitiesByModel` array).

Yes? No? Am I missing something?

I just want to understand how this is different.

nerrad: doh, sorry, I was basing my comment on the `entities` assignment from the `entitiesByModel.pop()` destructure.  For some reason (on this second look at things) I was focusing on the `entities` value instead of the `entity` value.  So ya I can push a change as you illustrated in your example.

tn3rb: phew... thought I was going crazy... I could not for the life of me think of how things would be affected differently by my suggestion.
totally shook my understanding of JS (as limited as it already is).

nerrad: Ha! Sorry about that.  That was entirely my bad there.

PR #1408, opened 2019-07-12
EDTR:  Fix bug with Ticket Assignment updates not sticking.

nerrad:

#1401 introduced a bug in the Ticket Assignments Manager.  From the context of either a date entity or ticket entity, triggering the ticket assignments manager and assigning (or unassigning) relations between tickets and date times would not reflect in the ui after closing the modal.  This happened for me when I created a new date entity and then assigned a ticket to it, although the store was updated, the ui/ux did not reflect the change on closing the TAM modal.

This pull fixes that issue.

The root cause was a couple of factors:

- Although I know it was implemented at my recommendation in the original pull's review, `useMemo` was implemented incorrect as the value returned from it would always be `undefined`.  Thus the dependency array for the `getActionsMenuForEntity` call in another `useMemo` would never get re-invoked when the callbacks for actions changed (this ultimately is what led to the buggy behaviour).
- In examining the component lifecycle and behaviour for this component, none of the `registerEntityActionsMenuItem` calls need to memoized.  Ultimately, the only thing that matters is that `getActionsMenuForEntity` is memoized (since it's result that becomes the children on the menu). 

## How has this been tested

- Tested creating a new date and following through the creation process along with assigning tickets and ensuring after the editor modals have closed that the ui reflects the changes.
- Did the same as the above from the context of creating a ticket.
- Tested editing an existing date and ensuring any changes are reflected in the ui after closing the modal (along with other action menu items for dates).
- Did similar testing on ticket entity action items.


assets/src/editor/events/tickets/editor-ticket/actions-menu/editor-ticket-actions-menu.js
```diff
@@ -118,9 +117,9 @@ export default compose( [
 				_x(
 					'Date Assignments for Ticket:  %1$s',
 					'Date Assignments for Ticket:  Ticket name',
-					'event_espresso'
+					'event_espresso',
```
tn3rb: ???

assets/src/editor/events/tickets/editor-ticket/actions-menu/editor-ticket-actions-menu.js
```diff
@@ -118,9 +117,9 @@ export default compose( [
...
 				),
-				ticketEntity.name
+				ticketEntity.name,
```
tn3rb: why commas after function call parameters?

PR #1422, opened 2019-07-17
BUG: Fix block styles not loading on frontend (fixed #1375)

### LINKED ISSUE #1375: No attendee block styles on the front end?
Looks like the Attendee block front end styles aren't loading, for example, if you look here: 

https://ee4m.eventespresso.com/attendee-block-testing/

The UL has a list-style-type of disc but the `ee-block-event-attendees` CSS sets it to none.

If you view the styles that apply on the page above its only styles from the theme.

nerrad:

See #1375 for details.

The fix implemented here may or may not be the chosen route for long term fixes.  

In #1286 I introduced the automatic dependency generation script and logic which in turn introduced the bug in #1375 because the frontend script for blocks (`/src/blocks/index-frontend.js`) doesn't list any dependencies on components, hence the  `@eventespresso/components`  built module is not loaded on the frontend.  That module is where the styles for the component the event attendees block uses (`components/event-attendees/style.css`) are found.  There are two temporary solutions:

1. Manually add `@eventespresso/components` as a dependency for the frontend css php side.
2. Manually import any styles we want included into `/src/blocks/index-frontend.js` (which would then get built to the `/src/blocks/index-frontend.css` file).

I chose the second option for the following reasons:

- Option 1 would result in a bigger package loaded over the wire on the frontend because ALL the styles for eventespresso components are loaded.  Compounded with the yet unresolved problem of loading block assets on every route, this is not really desired.
- Option 2 allows for targeting _only_ what is needed for frontend block loads.  The downside of course is that in the backend, styles will be duplicated in both the "frontend" css and the main blocks css (loaded for the editor).  I think for the backend that's okay.

Going forward as you plan for more blocks (not for this pull - but for future pulls), I would suggest evaluating:

- building block assets to their own files (I think this is what Jetpack does).
- look into only loading block assets on the routes they are needed.  I would register block assets on every route, but then hook into the block parser code server side to determine whether to `wp_enqueue` the block assets by what blocks are detected.  There is a `pre_render_block` filter in wp core that you might be able to use to detect when blocks are being rendered for post content and use that to enqueue the script/style (print style inline or enqueue script for footer). The other option is to use the `parse_blocks` function in core on an early hook (as soon as you know post_content) to detect if the post content has blocks (which is what you might want to use on a single post view). 

## How has this been tested

* [x] Verify that the Event Attendee block assets load and display as expected on the frontend and the backend (editor).

## Checklist

* [x] My code is tested.


package-lock.json
```diff
@@ -12056,8 +12056,8 @@
 			}
 		},
 		"del": {
-			"version": "git+https://github.com/sindresorhus/del.git#51662acf8c0b0632ed577fc2bb64a65a677ead9b",
-			"from": "git+https://github.com/sindresorhus/del.git#51662acf8c0b0632ed577fc2bb64a65a677ead9b",
+			"version": "git+https://github.com/sindresorhus/del.git#ca05c65ca582b47a20e1b64539dd5ef45a0e8419",
+			"from": "git+https://github.com/sindresorhus/del.git",
```
nerrad: I expected this to happen for the package-lock in this case because of all the dependency updates merged into master without any rebuilds (dependabot updates)

Pebblo:
Tested and confirmed the attendee block styles are applied in the admin and on the front end.

PR #1425, opened 2019-07-18
BUG: Build files with wp element not react (fixes #953)

### LINKED ISSUE #953: Test if React.createElement is being used instead of wp.element.createElement in our built files.
**Describe the bug**

I don't know for sure if this is a bug or not, but this is something I want to look into to verify. I just realized today that I have the webpack config setup to use the `ProvidePlugin` for exposing the react import automatically in built files.  Effectively this removes the need to import react wherever we're using jsx.  However, I'm concerned that this means the built files for things used in the wp5.0+ environment are using `React.createElement` instead of `wp.element.createElement` (which is currently a thin wrapper around `React.createElement`). Some _potential_ side effects of this would be:

- unnecessary increase in bundle size.
- increased dependence on react when WP intentionally exposed this wrapper so react could be swapped out for something else down the road.

Currently, a downside to fixing this (if I found it was doing what I thought it might) is that the we'd only be able to use our built files in WP 5.0+ (because of the `wp.element` dependency).  I don't think it'd be worth the extensive work required to make things back-compat with earlier WP versions (meaning we'd have to manually import the `@wordpress/element` package and build two versions of built files for use in different environments).

With the WP version requirement in mind, I'm just filing this issue and putting it in the 
icebox until we bump our minimum WP version requirement.

**To Verify**

* Verification involves looking at built files for any of our components that use jsx and make sure the components are using `wp.element.createElement` rather than `react.createElement`.

nerrad:

See #953 for background.  Basically this pull removes a configuration from the babel configuration file so that built files for the EE assets use `wp.element.createElement` instead of `React.createElement`.  Note, any dependencies that import `react` will still use `react.createElement` but that still utilizes the WP external via the dependency extractor plugin (which effectively means that the version of react used will always be whatever is bundled with WP).

## How has this been tested

* [x] Impacts Event Attendees block so ensure it still works as expected (breakage would result in it not loading/working at all, so easy to verify).
* [x] Impacts exit modal on plugins page for EE deactivation because the exit modal package depends on React. Easy to verify, if it shows up then it's working.
* [x] Make sure there are no console errors related to the above when testing (there are some in the block editor but are unrelated to these changes and pre-existing).

## Checklist

* [x] My code is tested.


babel.config.js
```diff
@@ -7,7 +7,6 @@ module.exports = function( api ) {
 		],
 		plugins: [
 			/** this is needed because plugin exit modal uses react not wp.element */
-			'@babel/transform-react-jsx',
```
tn3rb: what about the comment right above this line?

> this is needed because plugin exit modal uses react not wp.element

nerrad: oh ya, the comment could have been removed too!  Dunno how I missed that.

nerrad: Fixed in ffb080c

PR #1444, opened 2019-07-29
Fet/credit card string changes

mnelson4:

Changes a few strings for label texts for billing inputs as [Lorenzo suggested](https://github.com/eventespresso/eea-stripe-gateway/pull/14#issuecomment-515713345). 
Specifically, 
* onsite billing forms' "Email" becomes "Email Address" (to match the registration form; btw it shouldn't break translations because that exact string is already in EE)
* `EE_Credit_Card_Input`s label text will now be "Card Number", so client code doesn't need to set it (or accidentally set it to something different from this standard; btw this also shouldn't break translations as, again, that exact string is already in-use)
* "CVV" becomes "Card Security Code" (this DOES break translations; so I'm not 100% certain it's worth breaking the translation for this change, but [Lorenzo thought so](https://github.com/eventespresso/eea-stripe-gateway/pull/14#issuecomment-516124369))

Consistency lends itself to credibility, and so I think its best to default to always using the same input names where possible.

## How has this been tested

* [ ] Just use a variety of billing forms from different gateways. Card numbers, email addresses, and card security codes' labels should use the correct label.

## Checklist

* [x] My code is tested.

### DIFF
--- a/caffeinated/payment_methods/Aim/EE_PMT_Aim.pm.php
+++ b/caffeinated/payment_methods/Aim/EE_PMT_Aim.pm.php
@@ -53,8 +53,7 @@ class EE_PMT_Aim extends EE_PMT_Base
             'name'=>'AIM_Form',
             'subsections'=>array(
                 'credit_card'=>new EE_Credit_Card_Input(array(
-                    'required'=>true,
-                    'html_label_text' => __('Card Number', 'event_espresso')
+                    'required'=>true
                 )),
                 'exp_month'=>new EE_Credit_Card_Month_Input(true, array(
                     'required'=>true,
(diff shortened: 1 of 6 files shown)


mnelson4:
hey @joshfeck, before this goes for code review and testing, can you get behind these string changes? (Like I mentioned above, one of them will break translations; the others shouldn't though.)

joshfeck:
No, I'm afraid I cannot get behind these string changes.

Any of these changes may break existing translations. It might seem like they will not, but they could. For example, what happens on a site where they do have a translation for "Email", but they do not have a translation for "Email Address"? We cannot assume that every translation-ready string has a translation.

mnelson4:
Ok thanks for the quick and clear reply. I've made those changes to just the new strings in Stripe. I like things being consistent, but also prioritize not breaking folks' translated sites more.

PR #1001, opened 2019-03-06
Wrap the phone numbers shown in the reg and txn page in an a tag using 'tel:'

Pebblo:
This is mainly for mobile, it just wraps the phone number in an a tag so that you can click on it to automatically load the number on your phone ready to call.

Most mobile browsers will try to auto-detect a number and do this for you anyway, but its not always 100%. I believe iOS adds styles to those numbers but Android (or at least Chrome for android) does not and can be a little fiddly imo.

## How has this been tested
Loaded the Edit Registration and TXN page, then clicked on the number in Android. Loads the number into the Phone application.

## Checklist

* [x] I have read the documentation relating to systems affected by this pull request, see https://github.com/eventespresso/event-espresso-core/tree/master/docs
* [x] User input is adequately validated and sanitized
* [x] all publicly displayed strings are internationalized (usually using `esc_html__()`, see https://codex.wordpress.org/I18n_for_WordPress_Developers)
* [x] My code is tested.
* [ ] My code follows the Event Espresso code style.
* [ ] My code has proper inline documentation.
* [ ] My code accounts for when the site is in Maintenance Mode (MM2 especially disallows usage of models)

PR #1477, opened 2019-08-08
Use correct variable for Primary Reg Phone

Pebblo:
This fixes a silly mistake on my part from #1001 

I tried to put this in an add-on but it felt a bit hacky!

## How has this been tested
Go to Event Espresso ->Transactions.

View a transaction and in the Primary Contact metabox is a 'Phone #' field.

Hover over the phone number and it should be `tel: {phone_num}` but in master it'll be `tel:` and/or a notice showing undefined variable.

## Checklist

* [ ] My code is tested.


tn3rb:
> I tried to put this in an add-on but it felt a bit hacky!

oh come on ! no excuse !

Issue #1490, opened 2019-08-08
Add `TKT_reverse_calculate` Field to Ticket Model


tn3rb:
## Issue Overview
Currently a ticket's total price is calculated after applying all price modifiers to the ticket's base price, but the new EDTR introduces the ability to reverse calculate the ticket base price from the ticket total. However, the direction of calculation is not saved anywhere in the ticket data making it impossible to know which method of calculation to use. The new Ticket Price Calculator in the EDTR has a button for controlling the direction of calculation, but this is reset as soon as you exit the calculator.

This branch adds a new field to our ticket model for saving the direction of calculation.

PR #1491, opened 2019-08-08
Add `TKT_reverse_calculate` Field to Ticket Model 

tn3rb:
plz see: #1490

## How has this been tested
monkey based browser testing in EDTR branch only

# !!! IMPORTANT !!!

this db change in this branch was added to the 4.10.0 DMS but will need to be moved as this db change will force a version bump, but the exact version can not yet be known as other version bumps may occur before this.

So once this looks like we are ready to merge the EDTR branch into master, a new DMS will need to be made for this.


joshfeck:
Is this something we could instead merge to master so it's ready to go, and in a soon to be released version with its version bump?

joshfeck:
Further to the above, instead of branching off FET/editor-dates-tickets-refactor, could this be something we branch off master, review, then merge into Master?

tn3rb:
discussed in chat and the decision was to keep as is

Issue #891, opened 2019-01-15
Junk Transaction delete query needs preparing


tn3rb:
**Describe the bug**
The following comment can be found in `EEM_Transaction::delete_junk_transactions()`

```php
// let's get deletin'...
// Why no wpdb->prepare?  Because the data is trusted.
// We got the ids from the original query to get them FROM
// the db (which is sanitized) so no need to prepare them again.
```

but here is where the IDs are actually coming from:

```php
$txn_ids = apply_filters(
    'FHEE__EEM_Transaction__delete_junk_transactions__transaction_ids_to_delete',
    EEM_Transaction::instance()->get_col($ids_query, 'TXN_ID'),
    $time_to_leave_alone
);
```

aaaaaaaannnnndddd... filtered content can not be trusted, so the following query:

```php
$query   = '
    DELETE
    FROM ' . $this->table() . '
    WHERE
        TXN_ID IN ( ' . implode(",", $txn_ids) . ')';
$deleted = $wpdb->query($query);
```

simply needs to be run through `wpdb->prepare`

joshfeck:
resolved in #1501

PR #1501, opened 2019-08-12
Add $wpdb->prepare to junk delete query

raphaelkross:

Solves: https://github.com/eventespresso/event-espresso-core/issues/891

Add $wpdb-prepare to junk transactions delete query, since IDs come from a filtered variable and can't be trusted.

## How to Test

- Start registration process at a couple events but DO NOT finish the registration process - this will generated abandoned transactions (junk) at Dashboard -> Event Espresso -> Transactions.

- After generating a few transactions, we'll test the sanitization of the query (feature implemented by this PR), edit the following snippet to include a couple Transaction IDs (replace 1 by some real IDs - but ensure to keep the strings, since we need to prove the query do not blow up with weird values:

```php
add_filter( 'FHEE__EEM_Transaction__delete_junk_transactions__transaction_ids_to_delete', function($ids) {
    return [ 1, 'junk', 'hwoowow', 'injection' ];
});

add_action( 'admin_init', function() {
    \EE_Cron_Tasks::clean_out_junk_transactions();
});
```

Paste the code at your theme `functions.php` and refresh Transactions page, the transactions you've added in the filter should be gone and no error should be displayed.

## Checklist


core/db_models/EEM_Transaction.model.php
```diff
@@ -380,18 +380,18 @@ public function delete_junk_transactions()
...
             // the db (which is sanitized) so no need to prepare them again.
-            $query   = $wpdb->prepare('
-				DELETE
-				FROM %s
-				WHERE
-                    TXN_ID IN ( %s )',
-                $this->table(),
-                implode(",", $txn_ids)
-            );
+            $query   = $wpdb->prepare('DELETE FROM %s WHERE TXN_ID IN ( %s )', $this->table(), implode(',', $safe_txn_ids));
```
tn3rb: this still won't work with `$wpdb->prepare` because it is still declaring a placeholder for a string:

```php
WHERE TXN_ID IN ( %s )
```

which will result in the actual query looking something like:

```php
WHERE TXN_ID IN ( '1,2,3,4' )
```

whereas we need it to look something like:

```php
WHERE TXN_ID IN ( 1,2,3,4 )
```

notice the latter version does not have quotes around the TXN IDs, which means the database will now interpret those as numbers instead of a single string like `'1,2,3,4'`.

so you need to change how the placeholders are generated so that it produces something like:

```php
WHERE TXN_ID IN ( %d,%d,%d,%d )
```

Now that said, running all of the TXN IDs through `intval()` and then imploding the `$safe_txn_ids` array in the query is appropriate sanitization but just won't work with  `$wpdb->prepare`. So you could revert back to the original code that did not prepare the query and use your `$safe_txn_ids` idea instead if you like, or properly prepare the placeholders like I have previously described.

Your choice.

tn3rb:
So after passing code review, the next step is to pass the PR (pull request) on to support for quality assurance testing.

This will require you to write some testing notes that the support staff can use to verify that the fix works and does not introduce any regressions or new issues. This should be consist of fairly detailed instructions that include any areas that the code changes may affect. Here's simple example from a PR that is currently awaiting testing: https://github.com/eventespresso/event-espresso-core/pull/1278#issue-284183623 (see the *How has this been tested* section). It is preferred that you write testing notes in the initial pull request so that the code reviewer can better understand what areas of the codebase are affected, which may help them think of potential gotchas that may not have been considered.

For this issue, the work affects the query that deletes junk transactions from the db. Junk transactions are basically the result of an abandoned cart where someone clicks to start the registration process but then immediately abandons it. So we want to ensure two basic things:

- that junk transactions are still deleted correctly
- that the `FHEE__EEM_Transaction__delete_junk_transactions__transaction_ids_to_delete` filter can not pollute the query with bad data

In order to test that first point, we will need support staff to create some abandoned transactions by starting then immediately stopping the registration process and verifying that some junk data appears in their db (checking the EE admin UI is sufficient) and then manually trigger the garbage collection process, which is normally done via a Cron task. This can be done by adding a function somewhere that will run during a request (such as your theme's `functions.php` file) that calls `\EE_Cron_Tasks::clean_out_junk_transactions()` at an appropriate time (maybe the `admin_init` hook).

In order to test the new filter sanitization works, we need to inject some garbage into the query using the `FHEE__EEM_Transaction__delete_junk_transactions__transaction_ids_to_delete` filter, then confirm that the query does not blow up. This could be done by rerunning the process described above, but running that filter just prior to calling `\EE_Cron_Tasks::clean_out_junk_transactions()`.

So first let's get you to walk through the process of creating some junk transactions, by starting and stopping the registration process on the frontend, and then using the admin UI to view the data collected, and then see if you can write up a function that can be used to delete that junk data, and also test your new sanitization by injecting garbage via that filter.

tn3rb:
this branch also needs the latest changes from master to be merged into it. This can easily be done by clicking on the *Update branch* button below, or can be done locally if you prefer (will require a new push)

raphaelkross:
@joshfeck assigning this one for you for testing triage :)

joshfeck:
All right this works well! I also tested the regular cron-fired junk transactions clean up and all is well.

PR #1514, opened 2019-08-15
make sure validation defaults are initialized when loading billing forms

joshfeck:

Discovered this while testing the new Stripe add-on. What was happening was all required fields on the billing form were highlighted with green, which shouldn't happen when the field is empty.
Screen Shot 2019-08-15 at 3 58 03 PM

After a quick chat about this in Slack, @tn3rb came up with the solution in this PR.

## How has this been tested

Test SPCO reg form, there should be no change to the validation behavior.
then test a few onsite payment method forms, switching to and from other PMs. Blank required fields shouldn't initially be highlighted with green. They should be highlighted with green only when complete & pass validation.

Screen Shot 2019-08-15 at 5 49 40 PM

## Checklist

* [x] My code is tested.

### DIFF
--- a/modules/single_page_checkout/js/single_page_checkout.js
+++ b/modules/single_page_checkout/js/single_page_checkout.js
@@ -225,8 +225,6 @@ jQuery(document).ready( function($) {
 					if ( ! $(element ).hasClass('spco-next-step-btn') ) {
 						if ( $( element ).attr( 'value' ) ) {
 							$( element ).removeClass( 'ee-needs-value' ).addClass( 'ee-has-value' );
-						} else if ( $( element ).hasClass( 'ee-required' ) ) {
-							$( element ).removeClass( 'ee-needs-value' ).addClass( 'ee-needs-value' );
 						}
 					}
 				},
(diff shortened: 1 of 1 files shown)


tn3rb:
> were both removing then adding the class ee-needs-value just to double-check its set

that's correct, as far as I know there is no built-in logic for preventing duplicates. 

from the jQuery docs at https://api.jquery.com/addClass/:

```
It's important to note that this method does not replace a class. It simply adds the class, appending it to any which may already be assigned to the elements.
```

sethshoultes:
Here's how it looks using Authorize.net:
Registration Checkout – Events by Espresso 2019-08-15 21-33-24

@Pebblo can you please check a couple of the other gateways?

sethshoultes:
Tested with Stripe Elements and the payment forms look correct now:
Registration Checkout – Events by Espresso 2019-08-16 05-47-01

Pebblo:
Looks good to me!

Phone fields don't seem to validate when auto-filled - https://monosnap.com/file/fMJe7Lg8V8HkFQGR8561zi2eOtb4o7

It doesn't prevent the payment and validates on click but just thought I'd mention it.

Other than that I've not had any issues with the payment methods I tested

Infusionsoft, Vanco, AIM, Quickbooks all work fine.

Issue #848, opened 2018-12-20
Remove option to log to a file


joshfeck:
In Event Espresso > General Settings > Admin Options there's a setting to Enable Full Logging.

This particular option hasn't been useful, and could become a disk space problem if turned on and left on. 

So it'd be best to remove the option and its connected functionality.
![Screen Shot 2019-08-09 at 10 42 37 AM](https://user-images.githubusercontent.com/891156/62787531-e7e44880-ba92-11e9-8b6e-bbd6cd4427ae.png)

joshfeck:
When we remove this option we should also remove all mention of requestb.in and postbin.org. These are mentioned by the Remote Logging options. Both of these services are no longer available.

PR #1531, opened 2019-08-20
FET/deprecate log functionality

raphaelkross:

It removes local logging functionality to avoid issues with file size and others. Solves: https://github.com/eventespresso/event-espresso-core/issues/848

## How to Test
The option and related functionality was deleted, there's nothing left to test related to this. Only Remote Logging is enabled.

Public facing functions were deprecated with proper alerts.

To ensure this is working, when going to In Event Espresso > General Settings > Admin Options there's **NOT** a setting called "Enable Full Logging" anymore.

* [x] There was some potential for these changes to have introduced warnings during the 4.1 DMS. So it would be good to do a migration from EE3 to EE 4.1 (no need to go further) and verify no warnings or errors are produced

## Checklist

* [x] My code is tested.

### DIFF
--- a/admin_pages/general_settings/help_tours/Admin_Options_Help_Tour.class.php
+++ b/admin_pages/general_settings/help_tours/Admin_Options_Help_Tour.class.php
@@ -27,15 +27,6 @@ class Admin_Options_Help_Tour extends EE_Help_Tour
             10 => array(
                 'content' => $this->_start(),
             ),
-            20 => array(
-                'id'      => 'use_full_logging',
-                'content' => $this->_use_full_logging_stop(),
-                'options' => array(
-                    'tipLocation'    => 'right',
-                    'tipAdjustmentY' => -50,
-                    'tipAdjustmentX' => 15,
-                ),
-            ),
             30 => array(
                 'id'      => 'use_remote_logging',
                 'content' => $this->_use_remote_logging_stop(),
@@ -77,15 +68,6 @@ class Admin_Options_Help_Tour extends EE_Help_Tour
                         'event_espresso'
                     ) . '</p>';
         return $content;
-    }
-
-    protected function _use_full_logging_stop()
-    {
-        return '<p>'
-               . __(
-                   'When enabled, a log file is created which can be useful for debugging. Please disable this option when you are finished debugging.',
-                   'event_espresso'
-               ) . '</p>';
     }
 
     protected function _use_remote_logging_stop()
--- a/core/EE_Log.core.php
+++ b/core/EE_Log.core.php
@@ -79,7 +79,7 @@ class EE_Log
     public function verify_filesystem()
     {
         $msg = esc_html__(
-            'The Local File Logging functionality was removed permanentely. Remote Logging is recommended instead.',
+            'The Local File Logging functionality was removed permanently. Remote Logging is recommended instead.',
             'event_espresso'
         );
         EE_Error::doing_it_wrong(
@@ -137,7 +137,7 @@ class EE_Log
     public function write_log()
     {
         $msg = esc_html__(
-            'The Local File Logging functionality was removed permanentely. Remote Logging is recommended instead.',
+            'The Local File Logging functionality was removed permanently. Remote Logging is recommended instead.',
             'event_espresso'
         );
         EE_Error::doing_it_wrong(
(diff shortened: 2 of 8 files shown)


admin_pages/general_settings/help_tours/Admin_Options_Help_Tour.class.php
```diff
@@ -27,15 +27,6 @@ protected function _set_tour_stops()
...
             ),
-            20 => array(
-                'id'      => 'use_full_logging',
-                'content' => $this->_use_full_logging_stop(),
-                'options' => array(
-                    'tipLocation'    => 'right',
-                    'tipAdjustmentY' => -50,
-                    'tipAdjustmentX' => 15,
-                ),
-            ),
```
tn3rb: good job snagging this stuff too 👍

core/EE_Log.core.php
```diff
@@ -111,22 +78,15 @@ private function __construct()
...
-                    ' &nbsp; &nbsp; ' . $e->getMessage()
-                ),
-                __FILE__,
-                __FUNCTION__,
-                __LINE__
-            );
-            return;
-        }
+        $msg = esc_html__(
+            'The Local File Logging functionality was removed permanentely. Remote Logging is recommended instead.',
```
tn3rb: typo, should be:

> was removed permanently.

same deal with the other deprecation messages.

core/EE_System.core.php
```diff
@@ -854,11 +854,6 @@ public function load_core_configuration()
         // load and setup EE_Config and EE_Network_Config
         $config = $this->loader->getShared('EE_Config');
         $this->loader->getShared('EE_Network_Config');
-        // setup autoloaders
-        // enable logging?
-        if ($config->admin->use_full_logging) {
```
tn3rb: hmmmm... we should likely change this to:

```php
if ($config->admin->use_remote_logging) {
    $this->loader->getShared('EE_Log');
}
```

instead of removing it because otherwise that won't work at all

core/data_migration_scripts/4_1_0_stages/EE_DMS_4_1_0_org_options.dmsstage.php
```diff
@@ -256,9 +256,6 @@ private function _handle_org_option($option_name, $value)
             case 'use_event_timezones':
                 $c->admin->use_event_timezones = ($value == 'Y');
                 break;
-            case 'full_logging':
-                $c->admin->use_full_logging = ($value == 'Y');
-                break;
```
tn3rb: hmmmmm... not sure if it's kosher to remove stuff from older DMS's 

@mnelson4 
do you know if removing this `3.1 => EE4` config conversion from `EE_DMS_4_1_0_org_options` will mess anything up?

I have the mildest of concerns that it might have the potential to cause errors during migrations if anything happens to reference that option and it's not there, although I'm also skeptical of that actually happening. I also think it won't be an issue at all to simply leave this in place.

raphaelkross: @mnelson4 my point here is that $use_full_logging was removed - so perhaps would cause problem if tried to be accessed even during a migration.

All the $use_full_logging uses were removed as well - together with the class' property.

core/EE_System.core.php
raphaelkross: Nice catch!

raphaelkross:
Just sent back some adjustments. Just waiting for @mnelson4 feedback 👍

core/data_migration_scripts/4_1_0_stages/EE_DMS_4_1_0_org_options.dmsstage.php
mnelson4: Brent's right that we usually leave the older data migration scripts alone, seeing how they're not supposed to be bringing older database tables' data into conformity with the newest database schema, but the database tables that existed at the time.
BUT the config object is an exception because it's as soon as EE is installed, they're using the current class (not the class that existed during EE 4.1).
So in this case, seeing how the property has been removed from the `EE_Admin_Config` class, if we didn't remove this line, we'd get a warning during the migration (because that property wouldn't be defined).
So it turns out, where the data migration scripts are using config classes, like in this case, they actually DO need to be updated when the config classes get changed.
So this is good.

mnelson4:
oh also I added another test case that I thought would be good

mnelson4:
Thanks for the thought you put into this @raphaelkross :+1:

joshfeck:
@mnelson4 no one is using this feature. If they did, we'd get at least one question about how to interpret the logs it generates (in which case, we'd tell them there's nothing useful being logged).

sethshoultes:
This all checks out. I didn't see any issue while performing migrations either.

Issue #892, opened 2019-01-15
Fatal error in `ProcessTicketSelector::addTicketsToCart()`


tn3rb:
**Describe the bug**
While troubleshooting another issue, the following error was spotted in the EventSmart error logs:

```
PHP Fatal error: Uncaught TypeError: 
Argument 1 passed to EventEspresso\modules\ticket_selector\ProcessTicketSelector::addTicketsToCart() must be of the type array, boolean given,
called in /dom16036/wp-content/plugins/event-espresso-core/modules/ticket_selector/ProcessTicketSelector.php on line 213
and defined in /dom16036/wp-content/plugins/event-espresso-core/modules/ticket_selector/ProcessTicketSelector.php:403
```
The error occurs because if an invalid Ticket Selector Event ID is passed to `EventEspresso\modules\ticket_selector\ProcessTicketSelector::validatePostData()`, then that method returns false, whereas it otherwise returns an array. The boolean `false` value is therefore not being handled properly by `addTicketsToCart()`.

To correct the issue, the following steps should be taken:

- the return data type for `validatePostData()` should be consistent and return an empty array instead of `false` 
- `addTicketsToCart()` should then detect an empty array being returned and not proceed with processing ticket selections

joshfeck:
I wonder if someone manipulated the form inputs and removed the `tkt-slctr-event-id` input from the form? In any case this sounds like it'd be a quick, zero-impact fix. This could probably be fit between larger high priority work.

joshfeck:
This is actually an error that happens if the event is expired. In other words, the ticket sale end date hasn't passed yet, but the event's datetime end date is past. 

Also reported here:
https://eventespresso.com/topic/manual-registration-from-admin-error-message/
and here:
https://eventespresso.com/topic/add-new-registrations-to-the-past-event/

joshfeck:
Maybe we could change this to allow adding registrations in the admin if the tickets are still on sale, but the event is expired?

joshfeck:
A minor clarification: It doesn't appear to matter whether the ticket sale end dates are past, present or future. The "Add Tickets and Continue to Registrant Details" button is always shown regardless of whether the event is past, present, between datetimes, or future.

joshfeck:
Mentioned earlier:

> To correct the issue, the following steps should be taken:

- [ ] the return data type for validatePostData() should be consistent and return an empty array instead of false
[...]

PR #1536, opened 2019-08-26
Change return from bool to always array

raphaelkross:

Solved https://github.com/eventespresso/event-espresso-core/issues/892 - it will always return an empty array at validatePostData(), this way the bahavior is not broken at addTicketsToCart() - which also checks now if the array is not empty before processing the tickets.

## How has this been tested

I've tested mocking an invalid event ID and the Fatal Error is not triggered after the changes.

To validate that at QA you might try to add tickets to cart in an expired event, for example and the Fatal Error should not be triggered anymore.

## Checklist

* [x] My code is tested.


raphaelkross:
@joshfeck notices should be fixed 👍

joshfeck:
@sethshoultes can you retest and verify no notices now?

sethshoultes:
Notices are gone now. Thanks!

Issue #731, opened 2018-10-15
Ability to add registrations in the admin after event is expired


sethshoultes:
See Codebase tickets:
https://events.codebasehq.com/projects/event-espresso/tickets/7227
https://events.codebasehq.com/projects/event-espresso/tickets/8509

Grinder Discussion:
http://p2.eventespresso.com/2014/12/18/adding-registrations-in-the-admin-after-event-is-expired/

If a ticket expires, it also expires in both admin and front end. Many customers are requesting thatr the ticket should remain open in the admin (but expired front end) so any manually added attendees can be applied to that (expired) ticket.

tn3rb:
setting estimate at easy for the time being, but won't know until this is investigated.
it's possible this could be more convoluted if overriding expired entities needs to "trickle down" through a ton of code

joshfeck:
See also #892

joshfeck:
Currently in order to add a registration to an event after it's expired you need to:

1) Go in and edit the event and either change the datetime or add a new datetime so the event is no longer expired
2) edit the ticket sale end date to today or later
3) then add the registrations
4) Go back edit the event and remove/change the end date back

We'd like to change this so all they'd need to do is add the registrations, they won't need to edit ticket sale end dates or edit datetimes.

joshfeck:
Here's how I see things working:
- [ ] The issue reported in #892 should get fixed first and it its own PR
- [ ] No changes to the front end, the only changes will happen in the admin
- [ ] If a ticket sale date is past, the event admin will continue to be able to select that ticket 
- [ ] If the last datetime's end date is past, the event admin will continue to be able to select tickets for that event using the ticket selector in the admin
- [ ] No changes to the the admin ticket selector's behavior if tickets are sold out, one day we might add support for the admin side within the wait list add-on

joshfeck:
tracking additional requests for this feature:
https://eventespresso.com/topic/book-user-in-expire-event/

PR #1537, opened 2019-08-26
Fet/Enable add registration at admin

raphaelkross:
Solves https://github.com/eventespresso/event-espresso-core/issues/731 - allowing to add Registrations to an expired event when using Dashboard.

## How has this been tested

This task modified the TicketQtySelector behavior, so it should not bring any regression to BE or FE.

Main new functionality scenario:

- Given you have an expired event (with available tickets) - you should be able at **admin** add more registrations going to: Event - Registrations - Add New Registration and being able to add more registrations according to the quantity EVEN if the event is expired.

- That behavior is only available at admin level, at FE you should not be able to register when the event is expired.

- If the event is sold out, you should not be able to select any ticket.

- If the event has available tickets, you should be able to select and proceed to Registration and Payment.

Side effects:

Since this affects the ticket selector it's worth to test at FE the following situations:

- Available tickets works as usual, allowing you to pick the qty you want from the dropdown.
- Goes on Sale should work as expected, not allowing you to buy until the proper date.
- Min and Max tickets should work as expected.
- "Dude Where's My Ticket Selector?" events (1 ticket with max attendees set to 1)

## Checklist

* [x] My code is tested.


raphaelkross:
@tn3rb @mnelson4 I'm investigating this one.

So far, I've tried the easiest path as possible that is using is_admin() just to go around the expiration logic at some points.

Right now I'm facing a infinite loop at ` Fatal error: Allowed memory size of 268435456 bytes exhausted (tried to allocate 140509224 bytes) in /var/www/html/wp-content/plugins/event-espresso-core/modules/ticket_selector/TicketSelectorRowStandard.php on line 351`

Because `$this->max;` is set to INFINITE when it should be equal than the max available (or perhaps we should use a custom input field here? - any suggestions?).

raphaelkross:
@tn3rb made changes to the logic according to your suggestions - but still experiencing the infinite loop stack overflow caused by $max being equal to positive infinite.

raphaelkross:
@tn3rb I need your opinion about the new approach following your feedback BUT more than that, I'm still having the infinite loop issue because TKT_MAX is set to EE_INF even when limit is set to 20, for example.

Can you reproduce this error at your end using this branch?

Line: `for ($i = $this->min; $i <= $this->max; $i++) {` at `TicketSelectorRowStandard.php@ticketQuantitySelector`

Any help with that would be highly appreciated, since if this works, the ticket can be closed. The whole flow is working as expected, but this loop thing.

I've tried to debug that last week but without much success unfortunately.

tn3rb:
sry for not responding earlier, I hadn't checked this PR because it was not assigned to me. In the future, if you require feedback from someone, you should add that person as an **Assignee** to the issue/PR (pull request) and also add the `status:feedback-needed` label to really catch their attention.

So the issue with the infinite loop is because of the following conditional at the top of `\EventEspresso\modules\ticket_selector\TicketSelectorRowStandard::getHtml()`:

```php
if ($this->ticket->is_on_sale() && $this->ticket->is_remaining()) {
    $this->setTicketMinAndMax($remaining);
} else {
```

what's happening is `EE_Ticket::is_on_sale()` is returning `false` because the ticket is expired, and then `\EventEspresso\modules\ticket_selector\TicketSelectorRow::setTicketMinAndMax()` is not getting called.

We either need an `is_admin()` check added to that or... I'm tempted to have you simply remove the `if else` aspect of that and just call the contained logic directly, ie: change that code to:

```php
$this->setTicketMinAndMax($remaining);
// set flag if ticket is required (flag is set to start date so that future tickets are not blocked)
$this->required_ticket_sold_out = $this->ticket->required() && ! $remaining
    ? $this->ticket->start_date()
    : $this->required_ticket_sold_out;
```

with no conditional so that it just always runs. Yeah let's try that.

modules/ticket_selector/DisplayTicketSelector.php
```diff
@@ -218,7 +218,7 @@ public function display($event = null, $view_details = false)
             return $this->ticketSalesClosedMessage();
         }
         // is the event expired ?
-        $template_args['event_is_expired'] = $this->event->is_expired();
+        $template_args['event_is_expired'] = is_admin() ? false : $this->event->is_expired();
```
tn3rb: this needs to be reversed so that the "falsey" result is the last part of the ternary, as is this is saying: "If true, then false"

modules/ticket_selector/DisplayTicketSelector.php
```diff
@@ -363,6 +363,11 @@ protected function ticketSalesClosedMessage()
      */
     protected function getTickets()
     {
+        $show_expired_tickets = is_admin() || (
+            EE_Registry::instance()->CFG->template_settings->EED_Ticket_Selector instanceof EE_Ticket_Selector_Config
+            && EE_Registry::instance()->CFG->template_settings->EED_Ticket_Selector->show_expired_tickets
+        ) ? true : false;
```
tn3rb: don't need to explicitly return `true` and false`

sethshoultes:
@joshfeck sounds good. Also, I just noticed that I pasted the wrong image for the "dude where's my ticket selector" in the testing notes above, so I updated that.

Issue #1020, opened 2019-03-14
Add option to toggle display of copy attendee info template


joshfeck:
We get a number of support topics started because the site admin wants to remove this feature and there's currently no option to remove it.

Current workarounds include hiding it with CSS, copying over a blank copy_attendee_info.template.php into the wp theme, or using a function like this:

```
add_filter('FHEE__EEH_Template__locate_template__full_template_paths', 'my_copy_attendee_info_template_do_not_load', 10, 2);
function my_copy_attendee_info_template_do_not_load($full_template_paths, $file_name) {
  if( $file_name == 'copy_attendee_info.template.php') {
    return;
  }
  return $full_template_paths;
}
```

There is a todo in a comment in the generate_reg_form() method, followed by some logic already in place:

```
// print_copy_info ?
if ($primary_registrant && ! $this->checkout->admin_request && count($registrations) > 1) {
    // TODO: add admin option for toggling copy attendee info,
    // then use that value to change $this->_print_copy_info
    $copy_options['spco_copy_attendee_chk'] = $this->_print_copy_info
        ? $this->_copy_attendee_info_form()
        : $this->_auto_copy_attendee_info();
```

### This change affects Eventsmart documentation

After this is rolled out to Eventsmart, the documentation there will need to change to show how to use the option (instead of the CSS hack)

joshfeck:
tracking requests:

https://eventespresso.com/topic/split-change-registrant-for-event-multiple-registrants-create-contact-button/

PR #1555, opened 2019-08-30
Fet/add toggle display copy info option

raphaelkross:
Solves https://github.com/eventespresso/event-espresso-core/issues/1020 adding option to toggle the copy attendee info at Registration Form.

## How has this been tested

To test that, toggle the option at Dashboard - EventEspresso - Registrations - Registration Form :: "Allow copy #1 attendee info to extra attendees?"

And check if the copy attendee option is working according to the Yes/No set at option, when trying to register in an event w/ more than 1 attendee (purchasing more than one ticket).

## Checklist

* [x] My code is tested.


raphaelkross:
@tn3rb ok - first try adding and using a new option at the Dashboard - waiting your feedback :)

core/EE_Config.core.php
```diff
@@ -2332,6 +2332,13 @@ class EE_Registration_Config extends EE_Config_Base
      */
     public $gateway_log_lifespan = '1 week';
 
+    /**
+     * Enable copy attendee info at form
+     *
+     * @var boolean $enable_copy_attendee
+     */
+    public $copy_attendee_info;
```
tn3rb: These config classes were originally setup with `public` properties but that was a bad choice as it exposes the internals of the class.

For any new properties that we add, we want to not have them `public`, and instead use `private` or `protected` (you likely need to use `protected` if you want to mock the class for testing purposes) and then use public setters and getters for controlling access. My apologies for not explaining this to you earlier. My bad.

A good example that you can follow is the `EE_Registration_Config->track_invalid_checkout_access` property. 

As well, the config classes do not use the model system like our entities do (like events, etc) and so their properties need to be sanitized (another super important reason to use a setter). Since this new property you are adding is a boolean, you can use the PHP `filter_var()` function to sanitize it. Again, see the setter for `track_invalid_checkout_access` as an example.

This also means that you will need to change the places that retrieve this value to use the new getter instead of directly accessing the property.

As well, for any new code we want to follow the [PSR guidelines](https://www.php-fig.org/psr/psr-2/) for formatting and styles, so any new methods you add should use camelCase for the naming pattern.

caffeinated/admin/extend/registration_form/Extend_Registration_Form_Admin_Page.core.php
```diff
@@ -1144,6 +1152,118 @@ protected function _update_reg_form_settings()
...
+                // capture form data
+                $copy_attendee_info_settings_form->receive_form_submission();
+                // validate form data
+                if ($copy_attendee_info_settings_form->is_valid()) {
+                    // grab validated data from form
+                    $valid_data = $copy_attendee_info_settings_form->valid_data();
+                    if (isset($valid_data['copy_attendee_info'])) {
+                        $copy_attendee_info = $valid_data['copy_attendee_info'];
+
+                        $EE_Registration_Config->copy_attendee_info = $copy_attendee_info;
```
tn3rb: see below for more info, but we need to change this to use a setter, and also don't need to capture the form value in a variable, ie:

```php
$EE_Registration_Config->setCopyAttendeeInfo( $valid_data['copy_attendee_info'] );
```

modules/single_page_checkout/reg_steps/attendee_information/EE_SPCO_Reg_Step_Attendee_Information.class.php
```diff
@@ -121,7 +121,12 @@ public function initialize_reg_step()
      */
     public function generate_reg_form()
     {
-        $this->_print_copy_info = false;
+        /**
+         * @var $reg_config EE_Registration_Config
+         */
+        $reg_config = LoaderFactory::getLoader()->getShared('EE_Registration_Config');
+    
+        $this->_print_copy_info = $reg_config->copy_attendee_info;
```
tn3rb: this will need to change to be a getter

modules/single_page_checkout/reg_steps/attendee_information/EE_SPCO_Reg_Step_Attendee_Information.class.php
```diff
@@ -322,8 +320,14 @@ private function _registrations_reg_form(EE_Registration $registration)
...
                 );
+
+                /**
+                 * @var $reg_config EE_Registration_Config
+                 */
+                $reg_config = LoaderFactory::getLoader()->getShared('EE_Registration_Config');
+
                 // if we have question groups for additional attendees, then display the copy options
-                $this->_print_copy_info = $attendee_nmbr > 1 ? true : $this->_print_copy_info;
+                $this->_print_copy_info = $attendee_nmbr > 1 ? $reg_config->copy_attendee_info : false;
```
tn3rb: AWESOME!!!

Good job catching that change in the logic structure 👍

tn3rb:
looks good.

one FYI: in the future, when you pass a pull request to someone for re-review, you have to click the little circular arrow icon to the right of their name that shows the "Re-request review" label when you hover over it.

So this is good to go, but you need to merge the latest changes from master into this branch and wait for tests to pass before you can forward this to support for QA testing.

This can be done in either of the following two ways:

- using GitHub's UI and clicking the "**Update branch**" button below
- using your local Git client to pull the latest changes from master, merge them into this branch, and then push the updated branch

the most ideal way to do this is to use your local git client and rebase your branch's commits onto master (after pulling the latest changes)  and then force push the changes. This essentially replays your commits onto master and results in a much cleaner git history. Don't do this if you are unfamiliar with rebasing and NEVER do this with a branch that has already been shared (such as master branch) as it rewrites the history for the branch, which is okay to do with a new PR (pull request) like this one. We always use Squash and Merge when bringing PRs into master so ultimately this doesn't really matter.

So here's your next steps with this PR:

- update the branch as explained above and wait for automated tests to pass
- verify that you have adequate testing notes written and that they do not require any changes 
- change the pipeline for the PR to "Q/A Testing"
- remove the `status:review` label and add the `status:needs-testing` label
- assign to Josh for testing triage

joshfeck:
Works great! This will be much better than telling people to load a blank template or hide the box with CSS.

PR #1556, opened 2019-08-30
avoid incorrectly triggering DMS scripts when on a version 4.x.0.decaf

mnelson4:

If you install 4.10.0.decaf brand new (with absolutely no traces of EE anywhere on the site before), and then activate any add-on, you'll be asked to do the 4.10.0 DMS, even though you clearly have no previous data to migrate (it's a brand new install of EE). The changes in this branch fix that in the DMS scripts themselves.
The reason for the problem was the 4.10 DMS thought it should run if the previous version of EE core installed was lower 4.10.0.rc.000 (ie, it's looking for any version of 4.9.x). Unfortunately, I forgot 4.10.0.decaf IS lower than 4.10.0.rc.000! So if you're installing EE 4.10.decaf, it thinks your data needs to be migrated. (Note: this is an issue with the version string, not any other code in decaf. So adding `define('EE_DECAF', true);` to your `wp-config.php` file doesn't reproduce the issue. You need a version of EE whose version string is actually 4.10.0.decaf.) The fix was for the DMS' version comparison to use 4.10.0.decaf instead of 4.10.0.rc.000.

### Why Didn't we notice this before?
This has actually been an easy for each first release of decaf on a new minor version (ie 4.9.0.decaf, 4.8.0.decaf, etc), but the issue goes away with the second micro version (ie 4.9.1.decaf, 4.8.1.decaf). So yes, just bumping the version to 4.10.1.decaf will also fix the issue... but it will come up again in 4.11.0, assuming we copy-and-paste the same buggy code. So this should prevent it in the future too.

## How has this been tested

* [x] Add `add_filter('FHEE__espresso__espresso_version',function($true_version){return '4.10.0.decaf';});` to your theme's functions.php.
*  activate a brand new install of EE (there must be no options leftover from previous installs of EE... using the maintenance page's "permanently delete all EE data" actually doesn't do it because we leave a trace of previous versions of EE that were installed. It's easiest to just make a brand new WP site, and install EE on it.)
* Activate a payment method add-on like Sage Pay. On master the site would be put into maintenance mode and want to run the 4.10 DMS. On this branch it won't.
* [x] Remove that filter. Activate EE 4.9.x, then switch to this branch. The 4.10 DMS should trigger correctly and the DMS should work as expected.

## Checklist

* [x] My code is tested.


mnelson4:
Correct, there was no 4.4 migration. (And we might not have yet had the policy of only doing minor version updates on DMSs yet, I don't remember.)

sethshoultes:
I was able to trigger the migration and running the migration worked fine.

joshfeck:
Seth says this tested OK, but I wonder if the built files for cardinal commerce should have been included in this branch?

mnelson4:
> Seth says this tested OK, but I wonder if the built files for cardinal commerce should have been included in this branch?

Oh! Totally not. Sheesh my bad. I removed those. 

Those build files wouldn't have affected anything, so I don't think testing was invalidated by that. But I went through the testing checklist again anyway myself.

mnelson4:
So now the question is if someone else should retest. I'm satisifed not retesting because it's theoretically unnecessary and I restested anyway. But no worries if you want to assign someone else to retest.

Pebblo:
I retested this and all is good!

Merging in.

PR #1566, opened 2019-08-30
Show the more specific error when trouble setting redirect URL in gateway

mnelson4:

If there is an error inside `EE_Offsite_Gateway::set_redirection_info()`, SPCO currently doesn't show that error message (like it does if there's an error with an onsite gateway), it just shows a generic error message saying something was misconfigured and they should contact the admin.
This adds the gateway's response, if available, to the displayed error message.
This seemed important to add for the new Sage Pay offsite gateway, which can have an error while setting the redirection info (although I just realized PayPal Express can likewise have an error at this step.)

## How has this been tested

* [x] Activate PayPal Express but don't save it (ie it shouldn't have any credentials). Then try to pay with it. When you expect to be redirected, master will show a generic error message. On this branch, it includes the exact error code and the response from PayPal.
* [x] Now add your creds to PayPal Express payment method, and make a successful payment

## Checklist

* [x] My code is tested.

### DIFF
--- a/modules/single_page_checkout/reg_steps/payment_options/EE_SPCO_Reg_Step_Payment_Options.class.php
+++ b/modules/single_page_checkout/reg_steps/payment_options/EE_SPCO_Reg_Step_Payment_Options.class.php
@@ -2410,17 +2410,22 @@ class EE_SPCO_Reg_Step_Payment_Options extends EE_SPCO_Reg_Step
                 $payment->set_status(EEM_Payment::status_id_pending);
                 $payment->save();
             } else {
-                // not a payment
-                $this->checkout->continue_reg = false;
-                EE_Error::add_error(
-                    sprintf(
+                // we couldn't redirect the user. Let's tell them why.
+                if($payment instanceof EE_Payment && $payment->gateway_response()){
+                    $error_message = $payment->gateway_response();
+                } else {
+                    $error_message = sprintf(
                         esc_html__(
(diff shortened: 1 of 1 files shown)


modules/single_page_checkout/reg_steps/payment_options/EE_SPCO_Reg_Step_Payment_Options.class.php
```diff
@@ -2410,17 +2410,21 @@ private function _post_payment_processing($payment = null)
                 $payment->set_status(EEM_Payment::status_id_pending);
                 $payment->save();
             } else {
-                // not a payment
+                // we couldn't redirect the user. Let's tell them why.
+                $error_message = sprintf(
+                    esc_html__(
+                        'It appears the Off Site Payment Method was not configured properly.%sPlease try again or contact %s for assistance.',
```
tn3rb: We should always use numbered placeholders (`%1$s`, `%2$s`, etc) when there are more than one just in case some other language needs to rearrange the order of things.
I know it seems unlikely in this case, given the sentence, but you never know, and it's a good habit we should follow.

mnelson4: Thanks for the reminder. Although I rearranged the code, that exact string existed previously, and I'm pretty sure changing it to numbered placeholders would break the existing translations. Assuming that's right, I think it's best to leave that string as it was before this PR.
But ya, I'm on board that new strings should always use the numbered placeholders as a habit.

Issue #981, opened 2019-02-25
Add a venue filter to the event list admin overview


joshfeck:
Originally reported in [9226](https://events.codebasehq.com/projects/event-espresso/tickets/9226)

This will be a useful feature to help find events in the admin when there are many events happening at different venues. 
<img width="1102" alt="screen shot 2019-02-25 at 6 53 12 pm" src="https://user-images.githubusercontent.com/891156/53377158-fa2c5900-392e-11e9-81d5-8410c958850a.png">

We might also consider adding a column to the venue list table that shows the number of events it belongs to linked to the filtered event view.

PR #1586, opened 2019-09-10
Add venue filter to events

raphaelkross:

Solves https://github.com/eventespresso/event-espresso-core/issues/981 - adding a new Venue filter to Events archive.

## How has this been tested

Go to Events page at admin and now you might filter the events using the new Venues dropdown.

## Checklist

* [ ] My code is tested.

### DIFF
--- a/caffeinated/admin/extend/events/Extend_Events_Admin_Page.core.php
+++ b/caffeinated/admin/extend/events/Extend_Events_Admin_Page.core.php
@@ -967,6 +967,9 @@ class Extend_Events_Admin_Page extends Events_Admin_Page
             $filters[] = $this->active_status_dropdown(
                 isset($this->_req_data['active_status']) ? $this->_req_data['active_status'] : ''
             );
+            $filters[] = $this->venuesDropdown(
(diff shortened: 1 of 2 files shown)


caffeinated/admin/extend/events/Extend_Events_Admin_Page.core.php
```diff
@@ -1017,6 +1020,31 @@ public function active_status_dropdown($current_value = '')
...
+     * @return string
+     */
+    protected function venuesDropdown($current_value = '')
+    {
+        $select_name = 'venue';
+        $values = array(
+            '' => esc_html__('All Venues', 'event_espresso'),
+        );
+        // populate the list of venues.
+        $vnumdl = EE_Registry::instance()->load_model('Venue');
```
tn3rb: Please try to avoid using abbreviations for variable names, as there is zero benefit nowadays to having super short variable names like this. 
Readability is far more important, so this should be changed to something like `$venue_model`.

Abbreviations that are instantly recognizable because they are used extensively, such as `qty`, `min`, `max`, etc, are acceptable

PR #1588, opened 2019-09-11
Doc/eejs dependency implied gutenberg dependency

mnelson4:

Updates documentation to warn against depending on some stellar EE JS dependencies if the code needs to work with WP 4.9. I'm not 100% certain what I'm saying is correct so passing for review

## How has this been tested

no testing needed, it's just a change to docs.

## Checklist

* [ ] My code is tested.

### DIFF
--- a/docs/AA--Javascript-in-EE/eejs-core-dependency.md
+++ b/docs/AA--Javascript-in-EE/eejs-core-dependency.md
@@ -1,5 +1,14 @@
 ## The `eejs-core` dependency.
 
+> Update from Mike, September 11 2019:
+>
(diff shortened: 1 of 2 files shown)


mnelson4:
Ya it sounds to me like he was saying all EE JS needs to declare `eejs-core` or `ee-manifest` as a dependency. But I don't think either of those are correct seeing how Stripe Elements JS and PayPal Smart Buttons JS don't declare either one as a dependency and work fine. Oh well, maybe we'll revisit in a couple months when we're more familiar with Darren's old code.

mnelson4:
FYI @joshfeck this didn't affect any released code (just docs) so I didn't put it in a milestone

Issue #1527, opened 2019-08-16
Add support for ticket/date ordering (drag and drop)


joshfeck:
In the "legacy" ticket/date editor, there is the ability to set the order of dates and tickets which is typically used for presentation purposes. Currently in the new ui/ux there is no way to re-order things.

Brent remarked:
> in the `EDTR/drag-n-drop-tables` branch `react-beautiful-dnd` is integrated with our responsive table component, but the drag n drop doesn't fire the reorder callback when it should. I have created an issue on their repo: https://github.com/atlassian/react-beautiful-dnd/issues/1408
but have not rebuilt everything in their sandbox for troubleshooting purposes as requested.
As well, the current version of React implemented by WP is just below the min version required by `react-beautiful-dnd`. So it's totally possible that things may not work until WP updates their version of React. I created a pull with GB that does this:
https://github.com/WordPress/gutenberg/pull/16622
but there has been zero response

8/9/2019 Update: https://github.com/WordPress/gutenberg/pull/16982

Issue #1525, opened 2019-08-16
REFACTOR: Implement `useDispatch` and `useSelect` hooks in the TAM composition


joshfeck:
to eliminate multiple hocs.

Pre-requisites:

- #1396
- #1397

See also #1524

tn3rb:
refactor as part of #1589

Issue #1524, opened 2019-08-16
Review prop usage and ensure we're not passing through props unnecessarily through intermediary components


joshfeck:
(taking advantage of withSelect and withDispatch to grab state where necessary. For instance, it may be beneficial to introduce state for things like: current_filtered_date_ids (for the chained ticket results), selected_date/ticket (for actions such as copy and trash date and future ordering behaviour). Reducing the amount of props being passed through intermediaries helps cut down on the number of unnecessary renders due to prop changes.

tn3rb:
completed as part of #1589

Issue #1523, opened 2019-08-16
Refactor/Replace EntityActionMenuItem


joshfeck:
After replacing the `FHEE__EditorDates__EditorDateSidebar__SidebarMenuItems` filter with the slot/fill pattern we should investigate ways to restrict what components can be injected into entity action menus.

see: https://github.com/eventespresso/event-espresso-core/projects/4#card-22142568

tn3rb:
completed as part of #1589

PR #1589, opened 2019-09-12
Convert Classes and HOCs to Functional Components and Hooks

tn3rb:
The bulk of the EDTR work was done prior to React releasing Hooks. Therefore most if not all of the non-presentational logic was either in HOCs or embedded in the components themselves, and all localized state was implemented within class components. This resulted in some unnecessary code duplication but also in too may properties having to "trickle down" through the component tree causing an inordinate amount of re-renders.

This pull request refactors many of the EDTR class components into functional components and extracts much of the non-presentational logic into Hooks allowing for greater re-usability and more direct use which prevents prop trickle down.

Since many of the outstanding issues in the EDTR were likely the result of inefficient and/or improper prop passing and misuse of localized state, I have attempted to fix as many of those problems during this refactoring.

These issues include, but are not limited:

-  #1489
-  #1200
-  #1506
-  #1509
-  #1512
-  #1515
-  #1517
-  #1520
-  #1523
-  #1524
-  #1525
-  #1527

## How has this been tested


tn3rb:
will fix unit tests in EDTR/master after merge so that Rafael can start on other EDTR issues

Issue #754, opened 2018-10-24
Add filters to allow overriding help tours


lorenzocaum:
**Summary**
Here is the feedback that Nazar shared here (https://github.com/eventespresso/eventsmart.com-website/pull/275#issuecomment-432587177):

> It is not working because a wrong filter is used for this. Most templates are generated with `EEH_Template::display_template()` and that is where the `FHEE__EEH_Template__display_template__template_path` filter resides (which is used in this PR).
> 
> Help tours, on the other hand, are generated with a different method - `EEH_Template::help_tour_stops_generator()` and I don't see an obvious and proper hook.
> 
> The path to each help tour class looks to be hard-coded in the `_add_help_tour()` method in the `EE_Admin_Page.core.php`. And there are also no filters in the help tour classes to make those required changes.
> 
> I think that we will need more EE core changes to be able to filter the help tours.
>

sethshoultes:
@knazart is this something you are comfortable with adding?

joshfeck:
We have some documentation on naming conventions to use when adding new hooks:

https://github.com/eventespresso/event-espresso-core/blob/master/docs/A--Best-Practices/ee-actions-and-filters.md

tn3rb:
looks like the filter needs to be applied to `$this->_help_tour` right at the very end of `EE_Admin_Page::_add_help_tour()`

PR #1597, opened 2019-09-16
Add filter that allows to override tours

raphaelkross:

Solves https://github.com/eventespresso/event-espresso-core/issues/754 - adding a new filter to the tours variable, allowing it to be extended.

## How has this been tested

It adds only a new filter, the behavior of tours is not modified, therefore there's no need for extensive test - the tours should just keep working as expected.

```php
$tours = apply_filters('FHEE__EE_Admin_Page___add_help_tour___help_tour', $tours);
```

The new filter may be added to documentation as a way to extend the help tours.

## Checklist

* [x] My code is tested.


core/admin/EE_Admin_Page.core.php
```diff
@@ -1402,6 +1402,14 @@ protected function _add_help_tour()
                 $this->_help_tour[ $route ][] = EEH_Template::help_tour_stops_generator($end_stop_tour);
             }
         }
+
+        /**
+         * Allow extending the help tours variable.
+         *
+         * @param Array $tours The array containing all tour items to be displayed.
+         */
+        $tours = apply_filters('FHEE__EE_Admin_Page___add_help_tour__tours', $tours);
```
tn3rb: i think this filter will need to be applied to the ` $this->_help_tour` array after the conditional below.
You will notice in the code above that some things like:

```php
$this->_help_tour[ $route ][] = EEH_Template::help_tour_stops_generator($end_stop_tour);
```

are getting added directly to ` $this->_help_tour` and therefore would not be editable via the filter you have added.

PR #1606, opened 2019-09-18
Fix RECIPIENT_DATETIME_LIST shortcodes

Pebblo:
Currently, if you try to use RECIPIENT_DATETIME_LIST in messages, it will fail and output nothing.

Its pretty obvious what the problem is in this commit but to sum it up if `$registrations_on_attendee` is empty, we try to pull a registration from `$data` using `$data->reg_obj` but if it wasn't empty (meaning we had registrations for the attendee) we set it to an empty array... oops.

## How has this been tested
I used the Ticket Notice template (but you can use any that allow RECEIPIENT_DATETIME_LIST shortcodes in main content)

Add `[RECIPIENT_DATETIME_LIST]` to main content, then in DATETIME_LIST add something like `[ICAL_LINK_*]` (you can add some text before each of those shortcodes just to confirm the sections are being called if preferred).

In master, nothing is output.

In this branch, it should work.

Test a few of the DATETIME_LIST shortcodes if prefered but if one works, they all should.

## Checklist

* [x] My code is tested.


core/libraries/shortcodes/EE_Recipient_List_Shortcodes.lib.php
```diff
@@ -170,9 +170,9 @@ private function _get_recipient_datetime_list_parsed(EE_Messages_Addressee $data
     {
         // first get registrations just for this attendee.
         $att = $data->att_obj;
-        $registrations_on_attendee = $att instanceof EE_Attendee ? $data->attendees[ $att->ID() ]['reg_objs'] : null;
+        $registrations_on_attendee = $att instanceof EE_Attendee ? $data->attendees[ $att->ID() ]['reg_objs'] : array();
         $registrations_on_attendee = empty($registrations_on_attendee) && $data->reg_obj instanceof EE_Registration
-            ? array($data->reg_obj) : array();
+            ? array($data->reg_obj) : $registrations_on_attendee;
```
tn3rb: the fix is perfect but the formatting is a bit off...

should be:

```php
$registrations_on_attendee = empty($registrations_on_attendee) && $data->reg_obj instanceof EE_Registration
	? array($data->reg_obj)
	$registrations_on_attendee;
```

It's only acceptable to have the ternary options on one line if the entire statement fits on one line, like:

```php
$registrations_on_attendee = $att instanceof EE_Attendee ? $data->attendees[ $att->ID() ]['reg_objs'] : array();
```

is POIFECT!!!

Pebblo: Ah, thanks Brent.

Fixed.

sethshoultes:
Here's what I see output in the message:
‹ Events by Espresso — WordPress 2019-09-26 19-08-27(1)

sethshoultes:
I think this is good to go.

PR #1607, opened 2019-09-18
Fix Stale Date - Ticket Relations

tn3rb:
while demonstrating the new EDTR to Manzoor, I noticed that the relations between tickets and dates were not updating when re-opening the Ticket Assignments Manager after having been previously changed.

This PR simply fixes that issue so that the current relations are always utilized when opening the Ticket Assignments Manager

## How has this been tested

### DIFF
--- a/assets/src/editor/events/ticket-assignments-manager/submit-ticket-assignments-button.js
+++ b/assets/src/editor/events/ticket-assignments-manager/submit-ticket-assignments-button.js
@@ -39,9 +39,7 @@ const SubmitTicketAssignmentsButton = ( {
 	} );
 	return useMemo( () => (
 		<FormSubmitButton
-			onClick={ ( click ) => {
-				processChanges( click );
-			} }
+			onClick={ () => processChanges() }
 			buttonText={ __(
 				'Update Ticket Assignments',
 				'event_espresso'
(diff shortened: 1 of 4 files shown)


assets/src/editor/events/ticket-assignments-manager/submit-ticket-assignments-button.js
```diff
@@ -39,9 +39,7 @@ const SubmitTicketAssignmentsButton = ( {
 	} );
 	return useMemo( () => (
 		<FormSubmitButton
-			onClick={ ( click ) => {
-				processChanges( click );
-			} }
+			onClick={ () => processChanges() }
```
ManzoorWaniEE: This could even be `onClick={ processChanges }`

tn3rb: This was done on purpose this way so as not to pass along the click event to `processChanges()`
The reason being is because `processChanges()` performs some async logic and if passed the click event, will forward that to the `closeEditor()` callback which will attempt to call `preventDefault()` on it. The problem is that the click event is a React Synthetic Event and will have its values nulled out by the time the async functions finish, so calling `preventDefault()` results in a console warning.

ManzoorWaniEE: Sounds great! Thank you for the clarification.

PR #1609, opened 2019-09-19
Fix the date time input component

### LINKED ISSUE #1590: EDTR date/time editor modal closes when a date is selected
Here's a video that shows the issue:

Sep-12-2019 13-53-32

Ideally the modal should stay open until "Submit" is clicked.

Could we also consider allowing to click anywhere outside the modal to save and close? That's what the original datetime & editor does.

ManzoorWaniEE:

- Fixes #1590 
- Removes the submit and cancel buttons from date picker.
- Fixes the date not being updated in the form data.

## How has this been tested
unit tests only

### DIFF
--- a/assets/src/components/form/inputs/date-time-input.js
+++ b/assets/src/components/form/inputs/date-time-input.js
@@ -3,10 +3,9 @@
  */
 import classNames from 'classnames';
 import PropTypes from 'prop-types';
-import { Component } from '@wordpress/element';
+import { useState, useEffect } from '@wordpress/element';
 import { Field } from 'react-final-form';
 import {
-	Button,
 	DateTimePicker,
 	Dropdown,
 	IconButton,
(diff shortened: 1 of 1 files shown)


assets/src/components/form/inputs/date-time-input.js
```diff
@@ -84,162 +46,112 @@ class DateTimeDropdown extends Component {
...
+	const {
+		input,
+		htmlId,
+		htmlClass,
+		helpTextID,
+		dataSet,
+		inputWidth = '',
+		...attributes
+	} = props;
+	delete attributes.initialValue;
```
tn3rb: can we plz move the destructuring to the function declaration?

ie:

```jsx
const DateTimeDropdown = ( {
	input,
	htmlId,
	htmlClass,
	helpTextID,
	initialValue,
	dataSet,
	inputWidth = '',
	...attributes
} ) => {
```

plz note that I also included `initialValue` as one of the props so that we don't have to delete it from the attributes afterwards (no idea why I didn't do that initially?!?!)

I think this will work better as it allows IDEs to detect the props for the component

this of course will also require you to change any place where you access props via: `props.{prop name}`

assets/src/components/form/inputs/date-time-input.js
```diff
@@ -3,10 +3,9 @@
  */
 import classNames from 'classnames';
 import PropTypes from 'prop-types';
-import { Component } from '@wordpress/element';
+import { useState, useEffect } from '@wordpress/element';
```
tn3rb: this line is fine but there is some existing code in the "External imports" that should be moved down under an "Internal imports" comment heading.
ie: this existing code should be moved:

```jsx
import {
	EspressoIcon,
	ESPRESSO_ICON_CALENDAR
} from '../../ui/image/espresso-icon';
```

can you also do that as part of this PR plz?

assets/src/components/form/inputs/date-time-input.js
ManzoorWaniEE: Moved to function declaration. Now it looks good :)

assets/src/components/form/inputs/date-time-input.js
ManzoorWaniEE: Done :)

tn3rb: I see you added an "Internal dependencies" heading, but those imports above from from `'../../ui/image/espresso-icon'` are still listed under the "External imports".

Can you move those as well plz?

PR #1611, opened 2019-09-20
Fix Ticket Registrations Link

tn3rb:
Currently the Ticket Details Panel is displaying the ticket sold count for the registrants field because the Registrations Admin List Table is not capable of filtering for specific tickets. This branch adds a new  `TicketRegistrationsLink` functional component that looks and works nearly the same as the `EventDateRegistrationsLink` (they could be abstracted/combined one day).

Ticket filtering for the Registrations Admin List Table will be pushed in a separate branch off of master immediately after this branch is pushed.

## How has this been tested


tn3rb:
> I guess `target={ '_blank' }` is to keep the event editor open, right?

that is correct

Issue #1449, opened 2019-08-01
Deprecate `html_other_attributes` and use `other_html_attributes` instead


joshfeck:
@mnelson4 originally reported:

Somehow we have a duplicate property on form sections: 
* `EE_Form_Section_Base::_other_html_attributes` with accessors `set_other_html_attributes()` and `other_html_attributes()`
* `EE_Form_Input_Base::_html_other_attributes` with accessors `html_other_attributes()` and `set_html_other_attributeso()`

both are used, but `EE_Form_Section_Base::_other_html_attributes` is used more.

I think we should deprecate `EE_Form_Input_Base::_html_other_attributes` and its accessors, because it's redundant, confusing and leads to bugs (eg I just set `EE_Form_Input_Base::_html_other_attributes` and was wondering why I wasn't seeing those attributes getting output on the form input... it's because the value that's output is  `EE_Form_Section_Base::_other_html_attributes`).

**We'll just need to make sure that we check all add-ons as well**

It's used on vanco, quickbooks, and EE core

PR #1617, opened 2019-09-23
Deprecate html_other_attributes

raphaelkross:
Solves https://github.com/eventespresso/event-espresso-core/issues/1449 deprecating the `html_other_attributes` and using `other_html_attributes` instead

## How has this been tested

Nothing should change at custom HTML attributes since we're only migrating the functions being used.

To test nothing changed, there's one case that used the old method and now is using the new one:

- Go to Checkout page of any event
- At info section, expand the Add New State/Provincy
- The field `New State/Province Abbreviation *` should have a `size="24"` HTML attribute.

## Checklist

* [x] My code is tested.


sethshoultes:
Here's what I see when viewing the source of the page in Google Chrome:
image

mnelson4:
@sethshoultes what do you see on master? (FYI I don't see that red text on Chrome, and I don't see why it's being highlighted, I don't see what's wrong with it.)

sethshoultes:
@mnelson4 here's master:
image

Maybe we need to add a space between `style=""` and `"size="24"`?

joshfeck:
> Maybe we need to add a space between style="" and size="24" in this branch?

Can you check into this @raphaelkross ?

raphaelkross:
@tn3rb @joshfeck added an extra spacing when the other_attributes is not empty, this should be enough to add the required spacing.

PR #1619, opened 2019-09-24
Refactor Radio and Checkbox inputs 

### LINKED ISSUE #1618: EDTR Refactor inputs
`CheckboxInput` and `RadioInput` are not functional components.

ManzoorWaniEE:
Fixes #1618

## Checklist

* [ ] My code is tested.
js_only

### DIFF
--- a/assets/src/components/form/inputs/checkbox-input.js
+++ b/assets/src/components/form/inputs/checkbox-input.js
@@ -1,7 +1,7 @@
 /**
  * External imports
  */
-import { Component } from '@wordpress/element';
+import { useState } from '@wordpress/element';
 import PropTypes from 'prop-types';
 
 /**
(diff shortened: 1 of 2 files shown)


assets/src/components/form/inputs/checkbox-input.js
```diff
@@ -12,110 +12,93 @@ import { OptionInputs } from './base/option-inputs';
...
+	name,
+	htmlId,
+	options,
+	required,
+	helpTextID,
+	btnGroup = true,
+	inputWidth = '',
+	...attributes
+} ) => {
+	const [ checked, setChecked ] = useState( resetState() );
```
tn3rb: so `resetState` is being hoisted above this usage... 
but isn't that **just** the declaration?
meaning the memory space is reserved for something named `resetState` but the actual function doesn't get assigned to that name until we reach that in code?
or do function declarations get hoisted immediately?

assets/src/components/form/inputs/checkbox-input.js
```diff
@@ -12,110 +12,93 @@ import { OptionInputs } from './base/option-inputs';
...
 	 */
-	resetState = () => {
+	const resetState = () => {
 		return OptionCheckedState(
-			this.props.htmlId,
-			this.props.options,
-			this.props.initialValue,
+			htmlId,
+			options,
+			attributes.initialValue,
```
tn3rb: should this be introduced as a prop var as well?
or do we still need it passed through as part of the attributes prop?

assets/src/components/form/inputs/radio-input.js
```diff
@@ -14,100 +14,87 @@ import { OptionLabelSize } from './base/option-label-size';
...
-			PropTypes.string,
-		] ),
-		options: PropTypes.array,
-		onChange: PropTypes.func,
-		btnGroup: PropTypes.bool,
-		inputWidth: PropTypes.string,
-		helpTextID: PropTypes.string,
-		dataSet: PropTypes.object,
-	};
+export const RadioInput = ( {
```
tn3rb: same request as above regarding exporting as default

assets/src/components/form/inputs/radio-input.js
```diff
@@ -14,100 +14,87 @@ import { OptionLabelSize } from './base/option-label-size';
...
+	/**
+	 * @function
+	 * @param {mixed} reset
+	 * @return {Object} state
+	 */
+	const setCheckedState = ( reset ) => {
+		return OptionCheckedState(
+			htmlId,
+			options,
+			attributes.initialValue,
```
tn3rb: same comment as above regarding whether this should be converted to a prop...
I'm guessing this was done the way it was originally to avoid having to manage that prop within the class in other ways (ie: dumping into state, etc), but now that this is converted to a functional component, it's trivial to bring it in as its own prop and access it from anywhere.

assets/src/components/form/inputs/checkbox-input.js
```diff
@@ -12,110 +12,93 @@ import { OptionInputs } from './base/option-inputs';
...
-			PropTypes.string,
-		] ),
-		options: PropTypes.array,
-		onChange: PropTypes.func,
-		btnGroup: PropTypes.bool,
-		inputWidth: PropTypes.string,
-		helpTextID: PropTypes.string,
-		dataSet: PropTypes.object,
-	};
+export const CheckboxInput = ( {
```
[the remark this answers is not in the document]
ManzoorWaniEE: I retained the existing behavior. Will change it to default, which is better for single components

assets/src/components/form/inputs/checkbox-input.js
ManzoorWaniEE: Unlike variables, a function declaration doesn't just hoist the function's name. It also hoists the actual function definition.

assets/src/components/form/inputs/checkbox-input.js
ManzoorWaniEE: I will need to check whether it's required to be passed

assets/src/components/form/inputs/checkbox-input.js
```diff
@@ -12,110 +12,93 @@ import { OptionInputs } from './base/option-inputs';
...
+	htmlId,
+	options,
+	required,
+	helpTextID,
+	btnGroup = true,
+	inputWidth = '',
+	...attributes
+} ) => {
+	const [ checked, setChecked ] = useState( resetState() );
+	let { htmlClass } = attributes;
```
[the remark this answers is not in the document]
ManzoorWaniEE: because `htmlClass` is expected to be passed in `attributes` to `OptionInputs`

assets/src/components/form/inputs/checkbox-input.js
ManzoorWaniEE: Yes, you are right. `initialValue` is not needed.

assets/src/components/form/inputs/checkbox-input.js
tn3rb: ahh kk cool... that makes sense then

tn3rb:
this looks good now

PR #1639, opened 2019-09-26
Fix and Standardize Imports and Exports as well as PropTypes

tn3rb:
This branch fixes some errors introduced by recent changes made to prop types and exports and also standardizes the use of explicit default exports. Other notable changes:

- renames `FormSaveCancelButtons` component since it was really not being used for those kinds of buttons
- fixed some confusing logic in the `OptionsState` component which is used for managing state for checkboxes and radio buttons

## How has this been tested


assets/src/components/form/inputs/form-input.js
```diff
@@ -10,16 +10,16 @@ import { OnChange } from 'react-final-form-listeners';
...
-import { DateTimeInput } from './date-time-input';
-import { HiddenInput } from './hidden-input';
-import { IconButtonInput } from './icon-button-input';
-import { RadioInput } from './radio-input';
-import { SelectInput } from './select-input';
-import { Textarea } from './textarea';
-import { TextInput } from './text-input';
-import { ToggleInput } from './toggle-input';
-import { DefaultInputLayout } from './base/default-input-layout';
+import CheckboxInput from './checkbox-input';
```
ManzoorWaniEE: This one needed a fix because it was already exported as `default`.

assets/src/components/form/inputs/form-input.js
```diff
@@ -48,7 +48,7 @@ HTML5_INPUT_TYPES.push( 'iconButton' );
  * @param {Function} changeListener
  * @return {string} rendered date name form row
  */
-export class FormInput extends Component {
+class FormInput extends Component {
```
ManzoorWaniEE: What about refactoring this to functional component?

tn3rb: ya that component is a bit of a hot mess, but didn't feel like tackling it atm

PR #1640, opened 2019-09-27
Better Handling for Entity Dates and Potentially Infinite Values

### LINKED ISSUE #1516: BaseEntity objects are not correctly accounting for infinity type values on construction
The REST response returns infinity values as null but internally we should either convert them to INFINITY js representation or consistently use null or -1 (and that may depend on how writes receive the value too).

Requires #1515

### LINKED ISSUE #1518: Set wpUser field for Ticket
The wpUser field in ticket-entity-form-inputs.js is excluded, but there is no-where (that the reviewer could see) where the correct value is set for this field for created/cloned tickets. In this particular case the current authed user id should probably be used (or rather the set event author).

tn3rb:

- Fixes #1516 
- Fixes #1518 

This branch fixes some subtle issues with entity date fields such as event date start and end dates or ticket sale dates not having default values set correctly. This also includes fixes for values like event date capacities (reg limit) or ticket quantities that can potentially have infinite values.

## How has this been tested


tn3rb:
also handles #1516

PR #1649, opened 2019-09-27
Fix PayPal Pro gateway to send correct value for the State/Province field, fixes #453

### LINKED ISSUE #453: Fix PayPal Pro gateway so it sends the correct value for the State/Province field
Currently the PayPal Pro gateway sends the full state/province name instead of the expected state/province code. Examples:

https://developer.paypal.com/docs/classic/api/state_codes/#usa
https://developer.paypal.com/docs/classic/api/state_codes/#canada

joshfeck:
See #453 

## How has this been tested

- [x] Test payments using PayPal Pro. Do one payment where the Registration form includes address fields (and complete those fields). Do another payment where the registration form does not include address fields.

- [x] After testing payments, check the Transaction records and verify billing info's State field looks correct (it should show the full state/province name).

- [x] Then check the PayPal transactions admin and verify the shipping address now includes something for the State field.

## Checklist

* [x] My code is tested.

### DIFF
--- a/caffeinated/payment_methods/Paypal_Pro/EEG_Paypal_Pro.gateway.php
+++ b/caffeinated/payment_methods/Paypal_Pro/EEG_Paypal_Pro.gateway.php
@@ -260,7 +260,7 @@ class EEG_Paypal_Pro extends EE_Onsite_Gateway
                 ? $attendee->city()
                 : $billing_info['city'], 0, 40),
             'shiptostate'          => substr($use_registration_address_info
-                ? $attendee->state_abbrev()
+                ? $attendee->state_name()
                 : $billing_info['state'], 0, 40),
             'shiptocountry'  => $use_registration_address_info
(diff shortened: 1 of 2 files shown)


joshfeck:
The related issue on GitHub has the links:

https://developer.paypal.com/docs/classic/api/state_codes/#canada
https://developer.paypal.com/docs/classic/api/state_codes/#usa

mnelson4:
d'oh! I was just updating my comment mentioning the same thing.
Most importantly though @joshfeck, I was just suggesting
> We should probably test what happens if an EE state's abbreviation doesn't match the PayPal abbreviation. I'm guessing it will be the same as what's currently happening (so this change can only be an improvement)

joshfeck:
retested Mike's change, with all my now-apparently-unnecessary changes removed.

joshfeck:
Screen Shot 2019-10-01 at 12 58 37 PM

PR #1650, opened 2019-09-28
Display Entity CUIDs in Grid View and show Loading Notice when Creating Entities  (plus some other fixes)

tn3rb:
Currently the only way to view the ID for an entity is to open the edit form which is a bit of a hassle when you just need to know an ID for something. This branch adds a tiny new component called `EntityCuid` which displays the ID for an entity above each card in grid view.

This branch also adds a `LoadingNotice` to the `Add New Date` and `Add New Ticket` button components because loading could sometimes take a few seconds, during which there was no visual feedback that anything was happening. Entity creation was also moved into a `useEffect` hook instead of running immediately during the button's click events which makes the UI run a little smoother since the loading notice can now be displayed immediately.

As well, the "Add New Date" and "Add New Ticket" buttons would only open their corresponding editor modals **once** because the editor ID would not update correctly if the previously created entity was still in the local state. Clearing the cached entity when the editor closes resolves this issue.

This branch fixes the following additional minor issues:

- inline edit input for the ticket price not updating correctly because the value was not being returned
- invalid values for some of the ticket properties when creating a new ticket
- removes an old warning from the ticket edit form regarding the price calculator closing the current modal
- updates the getter used for retrieving a Price entity's base price type ID which changed after some getters were removed from the models
- uses the label for an inline edit input as a placeholder when the input has no value, otherwise nothing appears and there is nothing to click on if you want to enter a new value for that property. You will now see something like `click to enter Date Name` if there is no value.

## How has this been tested


assets/src/components/ui/entity-paper-frame/entity-paper-frame.js
```diff
@@ -0,0 +1,34 @@
...
+/**
+ * EntityPaperFrame
+ * adds a styled frame that gives the appearance
+ * of a piece of paper on a surface
+ *
+ * @return {Object} Entity with added EntityPaperFrame
+ */
+const EntityPaperFrame = ( { instanceId, children, entityID } ) => {
+	return (
+		<div id={ `ee-entity-paper-frame-${ instanceId }` }
```
ManzoorWaniEE: `id` could have been in next line.

assets/src/editor/events/dates-and-times/editor-date/add-new-date-entity-button.js
```diff
@@ -103,13 +114,16 @@ const AddNewDateEntityButton = () => {
 
 	return (
 		<>
+			<LoadingNotice
+				loading={ loading }
```
ManzoorWaniEE: good one :)

assets/src/components/ui/entity-paper-frame/entity-paper-frame.js
tn3rb: oops... ya think I'll make that change before merging. thnx 👍

assets/src/components/ui/entity-list/entity-list.js
```diff
@@ -0,0 +1,86 @@
...
+	entities,
+	EntityGridView,
+	EntityListView,
+	htmlClass,
+	view,
+	noResultsText,
+	...otherProps
+} ) => {
+	// verify array and remove undefined
+	entities = Array.isArray( entities ) ?
```
nicolad: React is using unidirectional data flow also called one-way binding. Mutating prop is a form of two-way data binding which might create some unwanted results in context of react.

assets/src/components/ui/entity-list/entity-list.js
```diff
@@ -0,0 +1,86 @@
...
+		noResultsText = noResultsText !== '' ?
+			noResultsText :
+			__( 'no results found', 'event_espresso' );
+		return (
+			<div className="ee-entity-list-no-results">
+				{ noResultsText }
+			</div>
+		);
+	}
+	htmlClass = classNames( 'ee-editor-entity-list', htmlClass );
```
nicolad: Same as above.

assets/src/components/ui/entity-list/entity-list.js
ManzoorWaniEE: @nicolad thank you for pointing that out. As React docs also mention
>**All React components must act like pure functions with respect to their props.**

assets/src/components/ui/entity-paper-frame/entity-paper-frame.js
nicolad: We have used Prettier, ESLint and Husky in the last project and it was a huge time-saver in the context of formatting.

assets/src/components/ui/entity-list/entity-list.js
nicolad: Yeah, especially with the introduction of hooks, they went all in with FP concepts.

tn3rb: :facepalm: doh

ya thanks @nicolad, I'm well aware that props should never be mutated, but obviously forget from time to time.

nicolad: @tn3rb I also do forget some subtle details and because of that I would rather use a ESLint rule like this: https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/prefer-read-only-props.md
rather than worrying about all the rules and just focus on the problem at hand.

Issue #915, opened 2019-01-24
Add ticket info to admin registration step 2


joshfeck:
Originally from https://events.codebasehq.com/projects/event-espresso/tickets/8075 where @garthkoyle requested:

> We need to show the ticket details like the front-end so admins know which ticket they are adding a registration to: 

![pasted_image_at_2015_05_01_03_22_pm](https://user-images.githubusercontent.com/891156/51704524-46f6da00-1fe7-11e9-8d94-df3d723b6e3c.png)

PR #1661, opened 2019-10-03
Show ticket details at admin

raphaelkross:

Solves https://github.com/eventespresso/event-espresso-core/issues/915 - allowing to see ticket details at Admin Registration step 2 like in FE.

## How has this been tested

Testing at Admin:

- At admin, add a new registration to an event and ensure to include 2 attendees, at least.
- At step two, where you add attendee info - above the form, you should be able to see Tickets details (Name, Qty, Price) that you're filling info to.

At Frontend Checkout (since the templates are shared, we should not see any regression here):

- Buy tickets for 2 attendees for an event at front-end.
- At step two of Checkout, where you fill attendee info, you still should see the tickets details (nothing can be broken at that part) - (Name, Qty, Price).

## Checklist

* [x] My code is tested.

### DIFF
--- a/modules/single_page_checkout/reg_steps/attendee_information/attendee_info_main.template.php
+++ b/modules/single_page_checkout/reg_steps/attendee_information/attendee_info_main.template.php
@@ -3,19 +3,21 @@
 /** @var array $ticket_line_item */
 /** @var string $default_hidden_inputs */
 ?>
-<p id="spco-attendee_information-pg" class="spco-steps-pg small-text drk-grey-text">
-    <?php echo apply_filters(
-        'FHEE__registration_page_attendee_information__attendee_information_pg',
-        sprintf(
-            __(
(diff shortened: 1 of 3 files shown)


modules/single_page_checkout/reg_steps/attendee_information/attendee_info_main.template.php
```diff
@@ -3,19 +3,21 @@
...
-        sprintf(
-            __(
-                'In order to process your registration, we ask you to provide the following information.%1$sPlease note that all fields marked with an asterisk (%2$s) are required.',
-                'event_espresso'
-            ),
-            '<br />',
-            '<span class="asterisk">*</span>'
-        )
-    ); ?>
+<?php if (!is_admin()) : ?>
```
tn3rb: nice.
good thinking to add this

Issue #188, opened 2016-12-21
Add copy attendee info section for admin (backend) registrations


thebeline:
Title says it all: There is no option to "Use First Registrant Answers" when making a Registration via the Administrative interface.  That is going to make it tedious for the people it's most tedious for...  :-/

thebeline:
Ok...

joshfeck:
@nerrad, Brent already has a ticket for this in Codebase. #8076

nerrad:
Thanks josh.

tn3rb:
https://events.codebasehq.com/projects/event-espresso/tickets/8076

joshfeck:
Ideally the copy attendee info option would use the setting we've just added to toggle its display. Then, we can also add a filter hook to override the setting. This way, if the site admin wants to turn off the copy attendee info option for the front-end, but keep it active on the back-end, then they can add simple filter function to activate the feature for the back-end only.

joshfeck:
added in #1669

PR #1669, opened 2019-10-04
Enable copy attendee info at admin

raphaelkross:

Solves https://github.com/eventespresso/event-espresso-core/issues/188 adding the copy attendee at admin registration page.

## How has this been tested

At Dashboard/admin, add new Registration to an event (that asks for more info for extra attendees) and add more than 1 ticket.

At Registration page on Admin, the Copy Info for other Attendee should be displayed and work same way than Frontend Checkout.

- When fields are not filled, validation errors should be displayed and you can't check the toggle.
- After valid info is filled, the info will be copied to other attendee fields.

A new filter is added in case customer wants to override the option from admin to, for example, enable the copy attendee only at admin (using a is_admin inside the filter):

```
FHEE__EE_SPCO_Reg_Step_Attendee_Information___registrations_reg_form___printCopyInfo
```

## Checklist

* [x] My code is tested.


joshfeck:
I tested the feature and it works great. 

There may be an issue with the filter hook though. If I add a filter function to return false, the copy attendee information section still displays. 

It looks like we might need the filter hook near the end of the _registrations_reg_form method?

raphaelkross:
@joshfeck hmmm that's true - I've moved the filter and now to make this work you would use some code like:

```php
add_filter( 'FHEE__EE_SPCO_Reg_Step_Attendee_Information___registrations_reg_form___printCopyInfo', function( $flag ) {
    if ( is_admin() ) {
        return true;
    }

    return $flag;
});
```

In case the customer wants to display it only at admin and at FE just follow the Registration Form settings.

joshfeck:
The new location works and I think we'll want to pass that $attendee_nmbr variable as an additional argument, which will allow for only showing the copy form in the admin **and** only when there are additional attendee forms.

e.g.
`add_filter(
    'FHEE__EE_SPCO_Reg_Step_Attendee_Information___registrations_reg_form___printCopyInfo',
    'toggle_print_copy_info',
    10,
    2
);

function toggle_print_copy_info($flag, $attendee_nmbr) {
    if (is_admin() && $attendee_nmbr > 1) {
        return true;
    }
    return $flag;
}`

If the above is OK with you, then this is ready to merge. (I've tested this again).

raphaelkross:
@joshfeck 👍 makes sense for me! Nice idea!

PR #1682, opened 2019-10-08
Extract Date Input Change Listener and Validation Logic into Hooks

tn3rb:
Consolidates existing logic for updating and validating entity date properties into custom hooks that can be shared by multiple components

## How has this been tested

### DIFF
--- a/assets/src/editor/events/dates-and-times/editor-date/edit-form/use-date-entity-input-config.js
+++ b/assets/src/editor/events/dates-and-times/editor-date/edit-form/use-date-entity-input-config.js
@@ -5,22 +5,63 @@ import warning from 'warning';
 import { useMemo } from '@wordpress/element';
 import { __, sprintf } from '@eventespresso/i18n';
 import { validations } from '@eventespresso/components';
+import {
+	useEndDateAfterStartDateValidator,
+	useEntityDateChangeListeners,
+	useEntityDateChangeValidators,
+} from '@eventespresso/hooks';
 import { parseInfinity } from '@eventespresso/utils';
 import { isModelEntityOfModel } from '@eventespresso/validators';
-import { ServerDateTime as DateTime, Duration } from '@eventespresso/value-objects';
+import {
+	ServerDateTime as DateTime,
+	Duration,
+} from '@eventespresso/value-objects';
 
+/**
+ * input configuration for event date entity edit form
+ *
+ * @function
+ * @param {Object} props
+ * @member {Object} dateEntity
+ * @member {string} prefix partial identifier for React Final Form data schema
+ * @member {Function} updateField callback for editing a field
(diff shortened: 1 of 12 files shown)


assets/src/hooks/use-end-date-change-listener.js
```diff
@@ -0,0 +1,54 @@
...
+			touchField( startDateFormKey );
+			touchField( endDateFormKey );
+		}
+	}, [
+		entity[ dateProps.start ],
+		entity[ dateProps.end ],
+		startDateFormKey,
+		endDateFormKey,
+		updateField,
+		touchField,
```
ManzoorWaniEE: Why should it depend upon `updateField` and `touchField`?

assets/src/hooks/use-start-date-change-listener.js
```diff
@@ -0,0 +1,73 @@
...
+			touchField( startDateFormKey );
+			touchField( endDateFormKey );
+		}
+	}, [
+		entity[ dateProps.start ],
+		entity[ dateProps.end ],
+		startDateFormKey,
+		endDateFormKey,
+		updateField,
+		touchField,
```
ManzoorWaniEE: Same here...

assets/src/editor/events/dates-and-times/editor-date/edit-form/use-date-entity-input-config.js
```diff
@@ -64,45 +105,8 @@ const useDateEntityInputConfig = ( {
...
-					if ( startDate > dateEntity.end ) {
-						return __(
-							'End Date & Time must be set later than the Start Date & Time',
-							'event_espresso'
-						);
-					}
-				}
-			},
+			changeListener: startDateChangeListener,
+			validate: startDateChangeValidator,
```
ManzoorWaniEE: Nice one! It makes the config file clean and simple.

assets/src/hooks/use-end-date-change-listener.js
tn3rb: from the React documentation regarding dependencies:

> The array of dependencies is not passed as arguments to the function. Conceptually, though, that’s what they represent: every value referenced inside the function should also appear in the dependencies array.

My understanding of that is that unless you are 1000% certain that some external code used by a hook is completely static and will not change, then it should be included as a dependency (just to be clear I was not using the term "static" here to mean "a class function not belonging to an instance", but just to mean something that stays the same).

I don't know if `updateField` and `touchField` are completely unchanging and therefore I have included them as dependencies. It's probably fair to assume that those callbacks won't change as they are likely just simple setters... but I'm just not 1000% certain of that...

ManzoorWaniEE: 👍🏻

PR #1694, opened 2019-10-09
Add Utility for Parsing Placeholders in Strings closes #1498

### LINKED ISSUE #1498: Evaluate how to handle situations where i18n is used in components that have html/elements mixed in with the translation strings
(in particular the calendar display components)

Darren did some work on this here: https://github.com/WordPress/gutenberg/pull/16374

tn3rb:
plz see #1498
Adds a new parseHtmlPlaceholders() function to the `@eventespresso/utils` module for substituting placeholders in string with HTML tags because the existing translation functions in WordPress do not work with `sprintf()`.
Not an extensively thorough solution but enough to get us by for the time being.

## How has this been tested
has accompanying unit tests to verify functionality

### DIFF
--- a/assets/src/components/form/base/form-info-base.js
+++ b/assets/src/components/form/base/form-info-base.js
@@ -1,10 +1,10 @@
 /**
  * External imports
  */
-import { isEmpty } from 'lodash';
 import PropTypes from 'prop-types';
 import { __ } from '@eventespresso/i18n';
 import { Dashicon, IconButton } from '@wordpress/components';
+import { parseHtmlPlaceholders } from '@eventespresso/utils';
 
 /**
  * Internal dependencies
@@ -58,26 +58,7 @@ const FormInfoBase = ( {
 			onClick={ onDismiss }
 		/>
 	) : null;
-	// parseHtmlPlaceholders parse-html-placeholders
-	if ( ! isEmpty( formInfoVars ) ) {
-		const formInfoText = [];
-		const chunks = formInfo.split( '%%var%%' );
-		if ( chunks.length !== formInfoVars.length ) {
-			throw new RangeError(
-				'The number of notice placeholders does not match' +
-				' the number of form info variables supplied.'
-			);
-		}
-		for ( let x = 0; x < chunks.length; x++ ) {
-			if ( chunks[ x ] ) {
-				formInfoText.push( chunks[ x ] );
-			}
-			if ( formInfoVars[ x ] ) {
-				formInfoText.push( formInfoVars[ x ] );
-			}
-		}
-		formInfo = formInfoText;
-	}
+	formInfo = parseHtmlPlaceholders( formInfo, formInfoVars );
 	return formInfo ? (
 		<div
 			aria-label={ __( 'important information', 'event_espresso' ) }
--- a/assets/src/utils/parse-html-placeholders.js
+++ b/assets/src/utils/parse-html-placeholders.js
@@ -14,16 +14,14 @@ const PLACEHOLDER = '%%var%%';
  * @return {string} text with placeholders replaced by variables
  */
 const parseHtmlPlaceholders = ( placeholderText, replacements ) => {
-	if ( isEmpty( replacements ) ) {
+	if (
+		isEmpty( replacements ) ||
+		! placeholderText.includes( PLACEHOLDER )
+	) {
 		return placeholderText;
 	}
 	if ( ! Array.isArray( replacements ) ) {
 		replacements = [ replacements ];
-	}
-	if ( ! placeholderText.includes( PLACEHOLDER ) ) {
-		throw new RangeError(
-			'The provided string does not have any placeholders.'
-		);
 	}
 	const finalText = [];
 	const chunks = placeholderText.split( PLACEHOLDER );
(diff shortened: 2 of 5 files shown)


assets/src/components/form/base/form-info-base.js
```diff
@@ -56,19 +56,7 @@ const FormInfoBase = ( {
 			onClick={ onDismiss }
 		/>
 	) : null;
-	if ( ! isEmpty( formInfoVars ) ) {
-		const formInfoText = [];
-		const chunks = formInfo.split( '%%var%%' );
-		for ( let x = 0; x < chunks.length; x++ ) {
```
ManzoorWaniEE: Why not `chunks.forEach()`?

assets/src/utils/parse-html-placeholders.js
```diff
@@ -0,0 +1,47 @@
...
+ */
+const parseHtmlPlaceholders = ( placeholderText, replacements ) => {
+	if ( isEmpty( replacements ) ) {
+		return placeholderText;
+	}
+	if ( ! Array.isArray( replacements ) ) {
+		replacements = [ replacements ];
+	}
+	if ( ! placeholderText.includes( PLACEHOLDER ) ) {
+		throw new RangeError(
```
ManzoorWaniEE: Why should there be an error if there are no placeholders? The default behavior of `sprintf` is that it returns the string if no placeholders are found. I think we should keep the behavior same as that.

That way, we can move this condition inside the first one:
```JS
if ( isEmpty( replacements ) || ! placeholderText.includes( PLACEHOLDER ) ) {
	return placeholderText;
}
```

assets/src/utils/parse-html-placeholders.js
```diff
@@ -0,0 +1,47 @@
...
+			'The provided string does not have any placeholders.'
+		);
+	}
+	const finalText = [];
+	const chunks = placeholderText.split( PLACEHOLDER );
+	if ( chunks.length - replacements.length !== 1 ) {
```
ManzoorWaniEE: Similarly here, if the number of replacements is more than the number of placeholders, it shouldn't be a problem I think. So, it could be
```JS
if ( chunks.length > replacements.length ) {
	...
}
```

assets/src/utils/parse-html-placeholders.js
```diff
@@ -0,0 +1,47 @@
...
+		throw new RangeError(
+			'The number of text placeholders does not match' +
+			' the number of replacement strings supplied.'
+		);
+	}
+	for ( let x = 0; x < chunks.length; x++ ) {
```
ManzoorWaniEE: `chunks.forEach()`? 🤔

assets/src/utils/parse-html-placeholders.js
tn3rb: hmmm... ya good point... will change

assets/src/utils/parse-html-placeholders.js
tn3rb: my thinking was that the number of chunks would be greater than the number of replacements by **one**, ie:

```js
const replacements = [ 'placeholder' ];
const placeholderText = `this text has a %%var%% in it`;
const chunks = placeholderText.split( PLACEHOLDER );
console.log( 'chunks: ', chunks.length );
// chunks: 2
console.log( 'replacements : ', replacements .length );
// replacements : 1
```

...but that may not always be case.

will need to look into this more to ensure that there is always a consistent ration of chunks to replacements, else find a better way to handle this

assets/src/utils/parse-html-placeholders.js
tn3rb: because I forget that `forEach` provides an index !!!

will update

assets/src/components/form/base/form-info-base.js
tn3rb: you want me to edit code that's being removed ;)

j/k

ManzoorWaniEE: Oh my bad 🤦🏻‍♂️

assets/src/utils/parse-html-placeholders.js
ManzoorWaniEE: A strict check isn't a problem though.

tn3rb: kk just confirmed this:

```js
const placeholderText = `sample text%%var%%`;
const chunks = placeholderText.split( '%%var%%' );
console.log( 'chunks: ', chunks.length );
// chunks:  2 (2) ["sample text", ""]

const placeholderText2 = `%%var%%sample text%%var%%`;
const chunks2 = placeholderText2.split( '%%var%%' );
console.log( 'chunks: ', chunks2.length, chunks2 );
// chunks:  3 (3) ["", "sample text", ""]
```

so the number of chunks should always be **one** more than the number of replacements, so I think my original condition:

```js
if ( chunks.length - replacements.length !== 1 ) {
```

should remain as is imho and simply ensuring that there are more chunks than replacements, like:

```js
if ( chunks.length > replacements.length ) {
```

does not protect against something like:

```js
const replacements = [ 'only one placeholder' ];
const placeholderText = `%%var%%sample text%%var%%`;
```

which would error when we attempt to access the replacement for the second placeholder.

That said, I'm still open to other suggestions for improving this, but please note that i didn't intend on having the logic smart enough to determine whether it should reuse the same replacement for multiple placeholders.

ManzoorWaniEE: Great! Thanks

PR #1712, opened 2019-10-16
WIP EDTR DOM Data Dump Fixes #1708

### LINKED ISSUE #1708: EDTR : Dump all event and related entity data to the DOM
Currently the EDTR is performing REST requests to load all of the initial data which requires dozens upon dozens of requests (potentially hundreds in the case of REM) to load all of the initial data required for the EDTR.

To speed up initial loading we are instead going to just dump all of the data into the DOM and then use that to hydrate the Redux stores.

tn3rb:
plz see #1708 

added some comments to the file changes to help explain why things were done

## How has this been tested


admin_pages/events/Events_Admin_Page.core.php
```diff
@@ -701,7 +708,8 @@ public function verify_event_edit($event = null, $req_type = '')
                 );
             }
             return;
-        } elseif ($orig_status === EEM_Event::sold_out) {
```
tn3rb: don't use `else if`after `return`

admin_pages/events/Events_Admin_Page.core.php
```diff
@@ -777,7 +788,7 @@ protected function _create_new_cpt_item()
                 __LINE__
             );
         }
-        return parent::_create_new_cpt_item();
+        parent::_create_new_cpt_item();
```
tn3rb: `parent::_create_new_cpt_item()` does not return

admin_pages/events/Events_Admin_Page.core.php
```diff
@@ -1020,7 +1050,7 @@ protected function _insert_update_cpt_item($post_id, $post)
         $att_success = true;
         foreach ($event_update_callbacks as $e_callback) {
             $_success = is_callable($e_callback)
-                ? call_user_func($e_callback, $event, $this->_req_data)
+                ? $e_callback($event, $this->_req_data)
```
tn3rb: don't need to use `call_user_func()` cuz PHP 5.3 is no more

admin_pages/events/Events_Admin_Page.core.php
```diff
@@ -1110,23 +1150,28 @@ protected function _default_venue_update(\EE_Event $evtobj, $data)
             $rows_affected = $venue_model->update($venue_array, array($update_where));
             // we've gotta make sure that the venue is always attached to a revision.. add_relation_to should take care of making sure that the relation is already present.
             $evtobj->_add_relation_to($venue_id, 'Venue');
-            return $rows_affected > 0 ? true : false;
-        } else {
```
tn3rb: no `else` after `return`

admin_pages/events/Events_Admin_Page.core.php
```diff
@@ -1511,8 +1572,7 @@ protected function _register_event_editor_meta_boxes()
             esc_html__('Event Registration Options', 'event_espresso'),
             array($this, 'registration_options_meta_box'),
             $this->page_slug,
-            'side',
-            'default'
+            'side'
```
tn3rb: removed values that match default parameter values

admin_pages/events/Events_Admin_Page.core.php
```diff
@@ -1810,25 +1889,22 @@ public function get_events($per_page = 10, $current_page = 1, $count = false)
...
-                        'status*restrict_private' => array('!=', 'private'),
-                        'AND'                     => array(
-                            'status*inclusive' => array('=', 'private'),
-                            'EVT_wp_user'      => get_current_user_id(),
-                        ),
-                    );
-                }
-            }
+        } else if (! isset($where['status'])
+            && ! EE_Registry::instance()->CAP->current_user_can('ee_read_private_events', 'get_events')
```
tn3rb: combined `if`s

admin_pages/registrations/Registrations_Admin_Page.core.php
```diff
@@ -88,7 +88,7 @@ public function __construct($routing = true)
     protected function getRegistrationModel()
     {
         if (! $this->registration_model instanceof EEM_Registration) {
-            $this->registration_model = $this->getLoader()->getShared('EEM_Registration');
+            $this->registration_model = $this->loader->getShared('EEM_Registration');
```
tn3rb: parent class already had `loader` property, so no need for `getLoader()`

assets/src/editor/events/event-editor.js
```diff
@@ -9,7 +9,12 @@ const container = document.getElementById( 'normal-sortables' );
 const editor = document.createElement( 'div' );
 editor.className = 'ee-editor-div';
 container.prepend( editor );
-const eventId = window.eeEditorEventId;
+const eventData = window.eeEditorEventData;
+console.log( '%c > eventData: ', 'color: #BCBDAC;', eventData );
+const eventId = eventData.event && eventData.event.EVT_ID ?
+	eventData.event.EVT_ID :
+	0;
+console.log( '%c > eventId: ', 'color: #BCBDAC;', eventId );
```
tn3rb: don't worry about these console logs, they are so Manzoor can see how data is structured

caffeinated/admin/extend/events/Extend_Events_Admin_Page.core.php
```diff
@@ -207,17 +237,7 @@ protected function _extend_page_config()
...
-        add_filter(
-            'FHEE__Events_Admin_Page___default_event_settings_form__form_subsections',
-            [$this, 'advancedEditorAdminFormSection']
-        );
-        add_action(
-            'AHEE__Events_Admin_Page___update_default_event_settings',
-            [$this, 'updateAdvancedEditorAdminFormSettings'],
-            10,
-            2
-        );
```
tn3rb: extracted into classes that now have the callbacks

caffeinated/admin/extend/events/Extend_Events_Admin_Page.core.php
```diff
@@ -482,7 +496,7 @@ public function additional_legend_items($items)
         }
         if (EE_Registry::instance()->CAP->current_user_can('ee_read_global_messages', 'view_filtered_messages')) {
             $related_for_icon = EEH_MSG_Template::get_message_action_icon('see_notifications_for');
-            if (isset($related_for_icon['css_class']) && isset($related_for_icon['label'])) {
+            if (is_array($related_for_icon) && isset($related_for_icon['css_class'], $related_for_icon['label'])) {
```
tn3rb: `$related_for_icon` can sometimes be a `string` so `'css_class'` would be an illegal offset (can only use numeric offsets when treating strings as arrays)

caffeinated/admin/extend/events/Extend_Events_Admin_Page.core.php
```diff
@@ -1236,15 +1301,13 @@ protected function _trash_or_restore_ticket($trash = false)
             // if array has more than one element then success message should be plural
             $success = count($this->_req_data['checkbox']) > 1 ? 2 : 1;
             // cycle thru the boxes
-            while (list($TKT_ID, $value) = each($this->_req_data['checkbox'])) {
+            foreach ($this->_req_data['checkbox'] as $TKT_ID) {
```
tn3rb: got rid of this nonsense which is deprecated in PHP 7.3

core/admin/EE_Admin_Page.core.php
```diff
@@ -632,13 +647,21 @@ private function _do_other_page_hooks()
                                );
                 throw new EE_Error(implode('||', $error_msg));
             }
-            $a = new ReflectionClass($classname);
-            // notice we are passing the instance of this class to the hook object.
-            $hookobj[] = $a->newInstance($this);
+            // // notice we are passing the instance of this class to the hook object.
+            $this->loader->getShared($classname, [$this]);
```
tn3rb: using loader instead of Reflection so that dependencies can be auto loaded
also removed `$hookobj[]` because nothing was done with it

core/admin/EE_Admin_Page.core.php
```diff
@@ -1109,15 +1128,14 @@ protected function _reset_routing_properties($new_route)
      */
     public static function add_query_args_and_nonce(
         $args = array(),
-        $url = false,
+        $url = '',
```
tn3rb: never utilized as boolean or expected to be one

core/admin/EE_Admin_Page.core.php
```diff
@@ -1206,20 +1224,23 @@ protected function _add_help_tabs()
...
-                $_ht['id'] = $this->page_slug;
-                $_ht['title'] = esc_html__('Help Tours', 'event_espresso');
-                $_ht['content'] = '<p>'
-                                  . esc_html__(
-                                      'The buttons to the right allow you to start/restart any help tours available for this page',
-                                      'event_espresso'
-                                  ) . '</p>';
-                $this->_current_screen->add_help_tab($_ht);
+            // if there ARE tour buttons...
+            if (! empty($tour_buttons)) {
```
tn3rb: don't duplicate conditional

core/admin/EE_Admin_Page.core.php
```diff
@@ -1746,7 +1769,7 @@ private function _get_help_content()
             return array();
         }
         // k we're good to go let's retrieve the help array
-        $help_array = call_user_func(array($this, $method_name));
+        $help_array = $this->{$method_name}();
```
tn3rb: call dynamic methods directly

core/admin/EE_Admin_Page.core.php
```diff
@@ -1788,7 +1811,7 @@ protected function _set_help_trigger($trigger_id, $display = true, $dimensions =
                     'event_espresso'
                 ),
             );
-            $help_content = $this->_set_help_popup_content($help_array, false);
+            $help_content = $this->_set_help_popup_content($help_array);
```
tn3rb: removed value that matches default

core/admin/EE_Admin_Page.core.php
```diff
@@ -2409,10 +2436,7 @@ protected function _espresso_ratings_request()
      */
     public function espresso_ratings_request()
     {
-        EEH_Template::display_template(
-            EE_ADMIN_TEMPLATE . 'espresso_ratings_request_content.template.php',
-            array()
-        );
+        EEH_Template::display_template(EE_ADMIN_TEMPLATE . 'espresso_ratings_request_content.template.php');
```
tn3rb: removed value that matches default

core/admin/EE_Admin_Page.core.php
```diff
@@ -2627,9 +2651,7 @@ protected function _set_publish_post_box_vars(
                 $delete,
                 $delete,
                 $delete_link_args,
-                'submitdelete deletion',
-                '',
-                false
+                'submitdelete deletion'
```
tn3rb: removed value that matches default

core/admin/EE_Admin_Page.core.php
```diff
@@ -2746,7 +2768,7 @@ public function _add_admin_page_meta_box(
         }
         // if $create_func is true (default) then we automatically create the function for displaying the actual meta box.  If false then we take the $callback reference passed through and use it instead (so callers can define their own callback function/method if they wish)
         $call_back_func = $create_func
-            ? function ($post, $metabox) {
+            ? static function ($post, $metabox) {
```
tn3rb: only need to build callback once

core/admin/EE_Admin_Page_CPT.core.php
```diff
@@ -86,10 +84,6 @@ abstract class EE_Admin_Page_CPT extends EE_Admin_Page
      */
     protected $_cpt_model_obj = false;
 
-    /**
-     * @var LoaderInterface $loader ;
-     */
-    protected $loader;
```
tn3rb: parent class already has this property

core/admin/EE_Admin_Page_CPT.core.php
```diff
@@ -157,20 +151,6 @@ abstract class EE_Admin_Page_CPT extends EE_Admin_Page
...
-     * @throws InvalidDataTypeException
-     * @throws InvalidInterfaceException
-     */
-    protected function getLoader()
-    {
-        if (! $this->loader instanceof LoaderInterface) {
-            $this->loader = LoaderFactory::getLoader();
-        }
-        return $this->loader;
-    }
```
tn3rb: only added recently by me, so won't be in use by others

core/admin/EE_Admin_Page_Init.core.php
```diff
@@ -386,8 +417,8 @@ protected function _initialize_admin_page()
                 'AHEE__EE_Admin_Page___initialize_admin_page__before_initialization_' . $this->_menu_map->menu_slug
             );
             require_once($path_to_file);
-            $a = new ReflectionClass($admin_page);
-            $this->_loaded_page_object = $a->newInstance($this->_routing);
+            $this->_loaded_page_object = $this->loader->getShared($admin_page, [$this->_routing]);
+            $this->_loaded_page_object->initializePage();
```
tn3rb: split this up because child classes would call parent constructor
but the constructor was running all kinds of logic prior to child classes being able to finish instantiation

core/admin/EE_Admin_Page_Loader.core.php
```diff
@@ -335,21 +342,23 @@ private function _get_installed_pages()
...
                             return $caf_admin_page;
                         }
                     );
                     $extend = true;
                 }
                 // let's do the registered hooks
                 $extended_hooks = $this->_installed_pages[ $page ]->register_hooks($extend);
-                $hooks_ref = array_merge($hooks_ref, $extended_hooks);
+                foreach ($extended_hooks as $key => $value) {
+                    $hooks_ref[] = $value;
```
tn3rb: should not run `array_merge()` inside loops

core/admin/EE_Admin_Page_Loader.core.php
```diff
@@ -464,14 +476,17 @@ private function _load_admin_page($page = '', $path = '')
...
+     * @throws InvalidDataTypeException
+     * @throws ReflectionException
      */
     public function set_menus()
     {
         // prep the menu pages (sort, group.)
         $this->_prep_pages();
         foreach ($this->_prepped_menu_maps as $menu_map) {
             if (EE_Registry::instance()->CAP->current_user_can($menu_map->capability, $menu_map->menu_slug)) {
-                $menu_map->add_menu_page(false);
```
tn3rb: removed value that matches default parameter values

tn3rb:
bah gotta fix some weirdness with the tests

core/admin/EE_Admin_Page.core.php
```diff
@@ -197,6 +197,19 @@ public function __construct($routing = true)
         $this->_req_data = array_merge($_GET, $_POST);
         // routing enabled?
         $this->_routing = $routing;
+    }
+
+
+    /**
```
mnelson4: Seeing how we're introducing `initializePage`, it might be good to differentiate between it and `__construct()` in the comments (eg what goes in which; when should `initializePage` be called, etc). This will help make it clearer where new code should go if we need to add to them someday.
Also I'm a little nervous that this change might break something. I think we should go ahead with it, just try our best to test it, especially with add-ons.

core/admin/EE_Admin_Page.core.php
```diff
@@ -944,7 +970,7 @@ protected function _verify_nonce($nonce, $nonce_ref)
 
     /**
      * _route_admin_request()
-     * Meat and potatoes of the class.  Basically, this dude checks out what's being requested and sees if theres are
+     * Meat and potatoes of the class.  Basically, this dude checks out what's being requested and sees if there are
      * some doodads to work the magic and handle the flingjangy. Translation:  Checks if the requested action is listed
```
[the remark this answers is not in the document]
tn3rb: oh i wonder if "theres" was supposed to be part of the joke?

core/admin/EE_Admin_Page_Loader.core.php
tn3rb: it's not about references it's because `array_merge()` makes copies of all of the arrays which can just eat up system resources when you are doing so inside a loop, so it is not recommended. Outside of a loop it's ok but it does take significantly longer than just adding elements via a loop like I have done. How you do that depends on the the data and what keys there are. These arrays were numerically indexed so what I did is fine. Basically, any time i see simple optimizations like this I go ahead and do them because every little bit adds up over time when we are running a platform like ES.

admin_pages/events/Events_Admin_Page.core.php
```diff
@@ -1247,15 +1292,15 @@ protected function _default_tickets_update(EE_Event $evtobj, $data)
                                 ),
                             ),
                         )
-                    ) > 0 ? true : false;
+                    ) > 0;
                     // let's just check the total price for the existing ticket and determine if it matches the new total price.  if they are different then we create a new ticket (if tkts sold) if they aren't different then we go ahead and modify existing ticket.
-                    $create_new_TKT = $ticket_sold && $ticket_price != $TKT->get('TKT_price')
+                    $create_new_TKT = $ticket_sold && $ticket_price !== $TKT->get('TKT_price')
```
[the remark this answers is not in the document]
tn3rb: good catch... 
i took it one step further and used `EEH_Money::compare_floats()`

admin_pages/events/Events_Admin_Page.core.php
```diff
@@ -1810,25 +1889,22 @@ public function get_events($per_page = 10, $current_page = 1, $count = false)
...
+                'AND'                     => array(
+                    'status*inclusive' => array('=', 'private'),
+                    'EVT_wp_user'      => get_current_user_id(),
+                ),
+            );
         }
-        if (isset($this->_req_data['EVT_wp_user'])) {
-            if ($this->_req_data['EVT_wp_user'] != get_current_user_id()
+
+        if (isset($this->_req_data['EVT_wp_user']) && $this->_req_data['EVT_wp_user'] !== get_current_user_id()
```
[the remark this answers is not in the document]
tn3rb: good call, will do

core/admin/EE_Admin_Page_Loader.core.php
```diff
@@ -565,12 +585,15 @@ private function _prep_pages()
             // reset $pages_array with prepped data
             $pages_array[ $group ] = $menu_maps;
         }
+
         // now let's setup the _prepped_menu_maps property
         foreach ($menu_groups as $group => $group_objs) {
             if (isset($pages_array[ $group ])) {
-                $this->_prepped_menu_maps = array_merge($this->_prepped_menu_maps, $pages_array[ $group ]);
```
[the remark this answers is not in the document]
tn3rb: here's some links describing why this is not recommended:
- https://github.com/dseguy/clearPHP/blob/master/rules/no-array_merge-in-loop.md
- https://dev.to/klnjmm/never-use-arraymerge-in-a-for-loop-in-php-5go1
i think it's more so the case with larger arrays, but I'm just trying to consciously trim things down as much as possible for the benefit of our EventSmart resource usage

core/domain/services/admin/events/default_settings/AdvancedEditorAdminFormSection.php
```diff
@@ -0,0 +1,124 @@
+<?php
+/**
+ *     Event Espresso
```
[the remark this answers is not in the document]
tn3rb: i bumped EE's license to GPLv3 and the documentation for that recommended placing a copyright notice in every file, but i forgot that we had decided a long time ago not to do that. will remove... it's kinda a pain having to scroll past all of the time anyways

core/domain/services/admin/events/default_settings/AdvancedEditorAdminFormSection.php
```diff
@@ -0,0 +1,124 @@
...
+                           'FHEE__Events_Admin_Page___default_event_settings_form__advanced_editor_input_answer_options',
+                           [
+                               esc_html__('Legacy Editor', 'event_espresso'),
+                               esc_html__('Advanced Editor', 'event_espresso'),
+                           ]
+                       ),
+                       apply_filters(
+                           'FHEE__Events_Admin_Page___default_event_settings_form__advanced_editor_input_settings',
+                           [
+                               'default'         => $this->admin_config->useAdvancedEditor(),
```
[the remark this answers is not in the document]
tn3rb: oh I didn't realize we weren't doing it :\
my IDE just does it for me.
it's not part of the PSR spec so it doesn't really matter in terms of passing linting tests, so technically we could each do what we want with respect to indents.

core/domain/services/admin/events/editor/AdvancedEditorEntityData.php
```diff
@@ -0,0 +1,223 @@
...
+            $eventId = $eventId === 0 && $post instanceof WP_Post ? $post->ID : $eventId;
+            if ($eventId) {
+                $data = $this->getAllEventData($eventId);
+                $data = wp_json_encode($data);
+                add_action(
+                    'admin_footer',
+                    static function () use ($data) {
+                        wp_add_inline_script(
+                            EspressoEditorAssetManager::JS_HANDLE_EDITOR,
+                            "var eeEditorEventData={$data}",
```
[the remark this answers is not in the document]
tn3rb: good catch on the `;` 
even after you pointed it out, it took me a minute to realize where it was missing from!

> re: don't we have a helper with a method...

Are you referring to `EventEspresso\core\services\assets\Registry::addData()` ?

that works if you have access to the script you are assigning the data to such as in an AssetManager class (see: `EventEspresso\core\domain\services\assets\CoreAssetManager::loadCoreJs()` for an example). This needs to only run within the event editor but I could likely figure out a way to move this logic into `EspressoEditorAssetManager` but that doesn't seem to jive with what the AssetManager classes were intended to be. This is admittedly a bit hacky so I'll try to think of a better way to handle this.

core/domain/services/admin/events/editor/AdvancedEditorEntityData.php
```diff
@@ -0,0 +1,223 @@
...
+     * @throws ReflectionException
+     * @since $VID:$
+     */
+    protected function getAllEventData($eventId)
+    {
+        $event = $this->spoofer->getApiResults(
+            $this->event_model,
+            [['EVT_ID' => $eventId]]
+        );
+        $event = is_array($event) && $event[0] && $event[0]['EVT_ID'] ? $event[0] : null;
```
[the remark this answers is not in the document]
tn3rb: ya great idea, will do!

core/domain/services/admin/events/editor/AdvancedEditorEntityData.php
```diff
@@ -0,0 +1,223 @@
...
+            $this->datetime_model,
+            [['EVT_ID' => $eventId]]
+        );
+        $eventDateTickets = [];
+        foreach ($eventDates as $eventDate) {
+            if (isset($eventDate['DTT_ID']) && $eventDate['DTT_ID']) {
+                $DTT_ID = $eventDate['DTT_ID'];
+                $eventDateTickets[ $DTT_ID ] = $this->spoofer->getApiResults(
+                    $this->ticket_model,
+                    [['Datetime.DTT_ID' => $DTT_ID]]
```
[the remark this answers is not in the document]
tn3rb: yes I do, same with dates, thnx!

core/domain/services/admin/events/editor/AdvancedEditorEntityData.php
```diff
@@ -0,0 +1,223 @@
...
+                        );
+                    }
+                }
+            }
+        }
+        $price_types = $this->spoofer->getApiResults(
+            $this->price_type_model,
+            [['PRT_deleted' => false]]
+        );
+        return [
```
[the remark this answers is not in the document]
tn3rb: ya hopefully this all gets replaced with something moar betterer

core/domain/services/converters/RestApiSpoofer.php
```diff
@@ -0,0 +1,101 @@
...
+     * @throws RestException
+     * @throws RestPasswordIncorrectException
+     * @throws RestPasswordRequiredException
+     * @throws UnexpectedEntityException
+     * @since $VID:$
+     */
+    public function getApiResults(EEM_Base $model, array $query_params, $include = '')
+    {
+        /** @type array $results */
+        $results = $model->get_all_wpdb_results($query_params);
```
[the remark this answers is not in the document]
tn3rb: awesome!!!

thnx :thumbsup:

core/admin/EE_Admin_Page.core.php
tn3rb: done

tn3rb:
thnx for the great review Michael.
I rebased this branch onto the latest changes from its parent, so if you are plan on viewing it locally, make sure to delete your local branch before checking out these changes.

Some of the changes here are React based so just ignore those as I will be having Manzoor reviewing them. 

I may end up cutting a new branch from master and cherry picking some of these changes over to it so that they get into master sooner rather than later.

PR #1722, opened 2019-10-16
add a bunch of policies regarding translations

mnelson4:

Just documents some policies regarding tranlations in PHP.
I [started a p2 discussion on it a while ago](http://p2.eventespresso.com/2019/09/26/ee-coding-style-use-translators-comment-or-_x-family-of-functions/), although I think there was a bit of confusion (I think we confused "only new strings CAN use _x()" with "new strings SHOULD ALWAYS use _x()"). Hopefully this documentation makes it clearer.

Please let me know what doesn't make sense, what policies I forgot, and what's dead wrong.

## How has this been tested

This only affects documentation so no testing needed.


docs/A--Best-Practices/php-standards.md
```diff
@@ -41,6 +40,39 @@ All new classes and files will follow modern day best practices, including names
...
+ <p>After the event is over, we're all going to Tim Horton's for some donuts...</p>
+ ```
+Notice it's the same translated string with different meanings depending on the context. In French, the first should be translated like "Publier l'événement" and the second as "Après l'événement". In order to allow translators to translate them differently, you must use `esc_html_x()`, like so:
+```php
+ <button><?php echo esc_html_x('Post Event', 'Button to publish the event', 'event_espresso');?></button>
+ ...
+ <h1><?php echo esc_html_x('Post Event', 'Title for information happening after the event', 'event_espresso');?></h1>
+ <p>After the event is over, we're all going to Tim Horton's for some donuts...</p>
+ ```
+ * Be kind to translators and avoid "breaking translations" (invaliding existing translations by changing the translated text). The following will break translations:
```
tn3rb: > invaliding existing translations

???

I'm guessing that's supposed to be "invalidating" ?

mnelson4: oh yeah thanks

Issue #333, opened 2018-04-24
Move all ticket information so it's displayed before the registration form if there's only one attendee information form


sethshoultes:
Codebase: https://events.codebasehq.com/projects/event-espresso/tickets/11099

Requested:

https://eventespresso.com/topic/personal-info-section-embedded-within-ticket-details-on-attendee-info-page/
https://eventespresso.com/topic/multiple-event-registration-checkout-customization/
https://secure.helpscout.net/conversation/458079216/450050?folderId=100570

> Also, the Registration Checkout page displays information in this
> order:
> 
> 1st ticket (Camp 1) personal information 2nd ticket (Camp 2) 3rd
> ticket (English Course)
> 
> I would prefer to to have all tickets together in one list and then
> either before or after collect the payer / attendee information. The
> current layout does not make a lot of sense.

In cases like these I've been sharing [this snippet][3] that embeds some JavaScript to move all the ticket information divs into one place. Which results in something that looks like this:

![attendee info step][4]

> On the Attendee Info page after selecting the “Register Now” button, the screen displays the first ticket details followed by the request for Attendee 1 Personal Information (name, address, email, etc), then the second and third ticket details. This split is confusing. After selecting “Proceed to Payment Options” it summarizes and totals the ticket offerings as expected.

https://eventespresso.com/topic/ee4-multi-ticket-total-vat/

> One questions – is there an option to show all selected tickets on top of the actual registration form. By default we always see the first selected ticket, then the form and then the other ticket(s). Can it be modified?

https://eventespresso.com/topic/edit-registration-checkout-page/

> Edit registration-checkout page to show all the tickets purchased be together. If more than one type of ticket is purchased the details is shown, then the personal Information section is displayed, then the details of any subsequent tickets is shown. Can the detail information not be split?

https://eventespresso.com/topic/event-checkout-page-doesnt-show-all-selected-tickets/#post-228630

> I would like to specify only one address with an order of 2 tickets and show the tickets in a table. The first ticket is displayed at the top, then the query for the personal data comes and then the second ticket appears.
This makes no sense for me if I do not ask a second participant.

Screenshot showing the proposed move:

![move stuff][1]

Possible end result could look like this:

![enter image description here][2]

  [1]: https://www.evernote.com/l/AAR-uPh1i01AxrABeNH6hKhdvhnH_Z-97_0B/image.png
[...]

PR #1723, opened 2019-10-16
Fet/put ticket details together if one attendee

raphaelkross:

Solves https://github.com/eventespresso/event-espresso-core/issues/333 - moving the tickets details together when only one form is available to be rendered.

## How has this been tested

- To test the new feature, add to cart two different tickets from an event that do NOT ask additional questions for extra attendees.

The tickets should be displayed first and the form at bottom, after the form.

- To test the new feature, add to cart two different tickets from an event that DO ask additional questions for extra attendees.

The tickets should be displayed individually and the form after each ticket.

## Checklist

* [x] My code is tested.


joshfeck:
So far so good.

I did find one issue in testing where extra rows are displayed when multiple ticket quantities are selected. For example if an event has two ticket types, and I select 2 of one and 4 of the other:
Screen Shot 2019-10-29 at 4 30 06 PM
Then submit the ticket selector, on the reg page there are extra rows displayed:
Screen Shot 2019-10-29 at 4 30 17 PM

Here's the URL for the event I tested with:
https://josh.eventespresso.com/events/sample-event/

Ideally we'd want to show only one row for each ticket type when no additional forms are displayed.

raphaelkross:
@joshfeck really well noticed! Should be patched after the code I've pushed.

joshfeck:
This works and looks great!

PR #1738, opened 2019-10-21
Refactor class components to functional components js_only

ManzoorWaniEE:
## How has this been tested
Unit tests

## Checklist

* [x] My code is tested.

### DIFF
--- a/assets/src/components/form/base/with-form-container-and-placeholder.js
+++ b/assets/src/components/form/base/with-form-container-and-placeholder.js
@@ -2,7 +2,7 @@
  * External imports
  */
 import { compose, createHigherOrderComponent, withInstanceId } from '@wordpress/compose';
-import { Component, Fragment } from '@wordpress/element';
+import { Fragment } from '@wordpress/element';
 import { FormContainer, FormPlaceholder } from '../';
 
 /**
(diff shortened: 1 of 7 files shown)


assets/src/components/form/base/with-form-container-and-placeholder.js
```diff
@@ -18,27 +18,24 @@ export default compose( [
 	withInstanceId,
 	createHigherOrderComponent(
```
tn3rb: this likely doesn't even need to be a HOC as it doesn't modify the child form in any way.
could likely just use the `children` prop and `React.Children.only(children)` to enforce a single form.
that said, we don't need to do that as part of this PR, just commenting for the sake of commenting

assets/src/components/form/inputs/form-input.js
```diff
@@ -48,215 +48,206 @@ HTML5_INPUT_TYPES.push( 'iconButton' );
...
-			changeListener = null,
-			afterInput = null,
-			...rest
-		} = this.props;
-		let {
-			helpTextID,
-			helpText = '',
-			htmlClass = '',
-			dataSet = {},
-			validations = [],
```
tn3rb: these defaults are not being set anymore which could cause issues if code is expecting these values
this code be done using `propTypes` `defaultProps`

assets/src/components/form/inputs/form-input.js
```diff
@@ -48,215 +48,206 @@ HTML5_INPUT_TYPES.push( 'iconButton' );
...
-			<OnChange name={ name }>
-				{ changeListener }
-			</OnChange>
-		) : null;
-		return InputLayout ? (
-			<Fragment>
-				<InputLayout
-					inputName={ name }
-					formInput={ formInput }
+const FormInput = ( {
```
tn3rb: for some reason i had thought that this one was going to be tricky to convert, 
but it looks like it was really straight forward, good stuff 👍

assets/src/components/ui/dates-and-times/calendar-page-date/index.js
```diff
@@ -116,27 +109,31 @@ class CalendarPageDate extends Component {
...
-			size = 'small',
-			statusClass = '',
-			...otherProps
-		} = this.props;
-		const htmlClass = classNames( {
-			'ee-calendar-page-date-wrapper': true,
-			[ `ee-calendar-page-date-${ size }` ]: true,
+	const htmlClass = classNames( {
+		'ee-calendar-page-date-wrapper': true,
+		[ `ee-calendar-page-date-${ size }` ]: true,
```
tn3rb: don't know why I did this as an object 🤷‍♂ 
would have made more sense to just list the classes...
if if you end up making other changes to this branch, then maybe fix this too

Issue #1707, opened 2019-10-15
Fix copy ticket


joshfeck:
The copy ticket feature needs fixing.

![Oct-15-2019 09-35-13](https://user-images.githubusercontent.com/891156/66836405-3e588200-ef2f-11e9-8e4d-eec157407c3f.gif)

tn3rb:
never had that problem... what's your pagination setting (per page) at?
is it showing up on the next page? and/or are other filter settings preventing it from showing?

something similar can happen when you create a new ticket, because sometimes they won't appear in the list immediately because of the current filter settings, so I'm considering tracking newly created tickets and then prepending them onto the list of tickets after other filters have been applied. Might have to do something similar with dates.

joshfeck:
It's set to 6 per page, and there are currently only 2 tickets. So it's not showing on the next page, and if it's not showing because of filter settings then maybe something isn't getting copied over as it should? I would think datetime relationships and sale start/end dates should get copied onto the new ticket, right?

This is one event where I tried to copy a ticket:

https://brent.eventespresso.com/wp-admin/admin.php?page=espresso_events&action=edit&post=34

PR #1746, opened 2019-10-28
Ticket: fix copy action button.

nicolad:
Fixed this issue: https://github.com/eventespresso/event-espresso-core/issues/1707

## How has this been tested
It has been tested locally:
Peek 2019-10-28 23-35

## Checklist

* [ ] My code is tested.

### DIFF
--- a/assets/src/hooks/use-copy-ticket.js
+++ b/assets/src/hooks/use-copy-ticket.js
@@ -14,27 +14,24 @@ import useCreateRelationsForTicketToEventDates
 import useCreateRelationsForTicketToPrices
 	from './use-create-relations-for-ticket-to-prices';
 
-const PRICE = 'price';
-const PRICES = 'prices';
-const TICKET = 'ticket';
 const falseFunc = () => false;
 
 const useCopyTicket = ( ticketEntity, dateEntities ) => {
 	const { createEntity } = useDispatch( 'eventespresso/core' );
 	const relatedPrices = useSelect( ( select ) => {
 		const { getRelatedEntities } = select( 'eventespresso/core' );
-		return getRelatedEntities( ticketEntity, PRICES );
+		return getRelatedEntities( ticketEntity, 'prices' );
 	}, [ ticketEntity ] );
-	const newPrices = useCloneEntities( relatedPrices, PRICE );
+	const newPrices = useCloneEntities( relatedPrices, 'price' );
 	const updateTicketDateRelations = useCreateRelationsForTicketToEventDates();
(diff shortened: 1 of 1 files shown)


assets/src/hooks/use-copy-ticket.js
```diff
@@ -14,31 +14,34 @@ import useCreateRelationsForTicketToEventDates
 import useCreateRelationsForTicketToPrices
 	from './use-create-relations-for-ticket-to-prices';
 
+const PRICE = 'price';
```
nicolad: @tn3rb let me know if I can pull these somewhere in a shared location.
Or maybe it's not so good idea.

assets/src/hooks/use-copy-ticket.js
```diff
@@ -14,31 +14,34 @@ import useCreateRelationsForTicketToEventDates
...
+const PRICES = 'prices';
+const TICKET = 'ticket';
```
nicolad: @tn3rb let me know if I can pull these somewhere in a shared location.
Or maybe it's not so good idea.

assets/src/hooks/use-copy-ticket.js
```diff
@@ -14,31 +14,34 @@ import useCreateRelationsForTicketToEventDates
...
 	const relatedPrices = useSelect( ( select ) => {
 		const { getRelatedEntities } = select( 'eventespresso/core' );
-		return getRelatedEntities( ticketEntity, 'prices' );
+		return getRelatedEntities( ticketEntity, PRICES );
 	}, [ ticketEntity ] );
-	const newPrices = useCloneEntities( relatedPrices, 'price' );
+	const newPrices = useCloneEntities( relatedPrices, PRICE );
 	const updateTicketDateRelations = useCreateRelationsForTicketToEventDates();
 	const updateTicketPriceRelations = useCreateRelationsForTicketToPrices();
 	return useCallback( async () => {
```
nicolad: here we are already returning a function, no need to return the other one from below

nicolad:
There is a bit of delay after copy ticket button is pressed, that might be solved by adding a spinner or by optimistic UI at some point: https://www.apollographql.com/docs/react/performance/optimistic-ui/

assets/src/hooks/use-copy-ticket.js
nicolad: Yep, I totally agree with you, this change is out of the scope. I had reverted back those changes.
Thanks for the in-depth explanations, it clarified the situation a little bit for me :)

Issue #1743, opened 2019-10-25
Ticket Price Calculator modal improvements


joshfeck:
The first time I used the ticket price calculator modal I wasn't sure why I couldn't select a price type for the base ticket price. Which makes sense to me now because I've realized it's not a price modifier, it's a base price! Maybe we could label it as "Base Price" instead of showing a disabled select input?

It might also be useful to show the name of the ticket so the user knows which ticket they're editing prices for. 

![Screen Shot 2019-10-25 at 6 17 59 PM](https://user-images.githubusercontent.com/891156/67608053-9e7fcd00-f754-11e9-9ee5-51e251f721cc.png)

garthkoyle:
+1 I had a similar thought.

nicolad:
@joshfeck @garthkoyle I didn't make it bold since there is a bit of confusion between that and the column label from above:

![image](https://user-images.githubusercontent.com/1477453/67849528-60e1c200-fb0f-11e9-811a-19bc54e20f25.png)

versus:

![image](https://user-images.githubusercontent.com/1477453/67849560-722ace80-fb0f-11e9-995e-8d6ca8ddd980.png)

joshfeck:
Good call on not making the label bold.

PR #1758, opened 2019-10-30
Ticket Price Calculator modal:  replaced disabled base price with a label.

nicolad:
Fixes this issue: https://github.com/eventespresso/event-espresso-core/issues/1743

## How has this been tested
Locally:
image

## Checklist

* [ ] My code is tested.

### DIFF
--- a/assets/src/components/form/layouts/two-column-admin/two-column-admin.css
+++ b/assets/src/components/form/layouts/two-column-admin/two-column-admin.css
@@ -483,10 +483,6 @@
     margin-top: 3px;
 }
 
-.ee-two-column-admin .ee-visible-label {
-    padding-left: .4rem;
-}
-
 .ee-two-column-admin .ee-hidden-label {
     display: none !important;
 }
--- a/assets/src/components/form/layouts/two-column-admin/two-column-admin.css
+++ b/assets/src/components/form/layouts/two-column-admin/two-column-admin.css
@@ -483,6 +483,10 @@
     margin-top: 3px;
 }
 
+.ee-two-column-admin .ee-visible-label {
+    padding-left: .4rem;
+}
+
 .ee-two-column-admin .ee-hidden-label {
     display: none !important;
 }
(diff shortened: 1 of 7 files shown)


assets/src/components/form/layouts/two-column-admin/two-column-admin.css
```diff
@@ -483,6 +483,10 @@
     margin-top: 3px;
 }
 
+.ee-two-column-admin .ee-visible-label {
+    padding-left: .4rem;
```
nicolad: @tn3rb I am not :100: % happy with this approach.
If there is a better mechanism of handling spacing in EE, please do suggest.
On the previous project, we had something similar to https://getbootstrap.com/docs/4.3/utilities/spacing/

tn3rb:
@nicolad 
so originally the price calculator showed the "base price" label for the first price option, but that broke at some point. 
I just discovered that the actual problem is that the price objects no longer have a `price.prtId` property because of some changes that happened in the models a little while ago, and that property is now named `price.PRT_ID`.

So can I get you to revert your changes you've made so far (sorry) and instead run a find and replace on the `/assets/src/` folder and change usages of `prtId` to `PRT_ID`? Another option if you prefer, is to simply invalidate this PR, delete the branch, and then create a new one.

Your choice, and my apologies for making you redo this.

tn3rb:
oh and here's a screenie of what the price calculator looks like when the base price is detected properly:

price calc base price

nicolad:
> @nicolad
> so originally the price calculator showed the "base price" label for the first price option, but that broke at some point.
> I just discovered that the actual problem is that the price objects no longer have a `price.prtId` property because of some changes that happened in the models a little while ago, and that property is now named `price.PRT_ID`.
> 
> So can I get you to revert your changes you've made so far (sorry) and instead run a find and replace on the `/assets/src/` folder and change usages of `prtId` to `PRT_ID`? Another option if you prefer, is to simply invalidate this PR, delete the branch, and then create a new one.
> 
> Your choice, and my apologies for making you redo this.

I reverted commit and pushed changes here to have more context related to the fix.

Issue #1756, opened 2019-10-29
EDTR : Use Constants for JS model names


tn3rb:
Plz see https://github.com/eventespresso/event-espresso-core/pull/1746#discussion_r339855565 for background

I think my preference is for the second of the two options outlined in that comment but I am open for alternate suggestions

PR #1759, opened 2019-10-30
EDTR : Use Constants for JS model names.

nicolad:
##  Problem this Pull Request solves
Fixes this [issue](https://github.com/eventespresso/event-espresso-core/issues/1756)
This PR implements the second option from this [comment](https://github.com/eventespresso/event-espresso-core/pull/1746#discussion_r339855565)

## How has this been tested
By running `test-unit` command

## Checklist

* [ ] My code is tested.


assets/src/data/model/event/constants.js
```diff
@@ -3,7 +3,8 @@
  */
 import { values } from 'lodash';
 
-export const MODEL_NAME = 'event';
+export const EVENT = 'event';
+export const EVENTS = 'events';
 
 export const EVENT_STATUS_ID = {
```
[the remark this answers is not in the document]
nicolad: Done.

nicolad:
@tn3rb have updated as required, let me know if there is anything else I can do here.

assets/src/hooks/use-copy-ticket.js
```diff
@@ -13,25 +13,27 @@ import useCreateRelationsForTicketToEventDates
...
+const { PRICE, PRICES, TICKET } = MODELS;
 
 const falseFunc = () => false;
 
 const useCopyTicket = ( ticketEntity, dateEntities ) => {
 	const { createEntity } = useDispatch( 'eventespresso/core' );
 	const relatedPrices = useSelect( ( select ) => {
 		const { getRelatedEntities } = select( 'eventespresso/core' );
-		return getRelatedEntities( ticketEntity, 'prices' );
+		return getRelatedEntities( ticketEntity, PRICES );
```
tn3rb: shouldn't this be:

```js
return getRelatedEntities( ticketEntity, PRICE.RELATION );
```

assets/src/hooks/use-copy-ticket.js
```diff
@@ -13,25 +13,27 @@ import useCreateRelationsForTicketToEventDates
...
 	}, [ ticketEntity ] );
-	const newPrices = useCloneEntities( relatedPrices, 'price' );
+	const newPrices = useCloneEntities( relatedPrices, PRICE );
```
tn3rb: `PRICE.MODEL` ?

assets/src/hooks/use-copy-ticket.js
```diff
@@ -13,25 +13,27 @@ import useCreateRelationsForTicketToEventDates
...
 	return useCallback( async () => {
-		if ( ! isModelEntityOfModel( ticketEntity, 'ticket' ) ) {
+		if ( ! isModelEntityOfModel( ticketEntity, TICKET ) ) {
 			return falseFunc;
 		}
 
 		const newTicket = await createEntity(
-			'ticket',
+			TICKET,
 			ticketEntity.forClone
```
tn3rb: `TICKET.MODEL` ?

nicolad:
> changes to the constant files look good, but I think some of the uses are still based on what you had before. As well, I'm thinking that this PR should maybe just change the constants and fix any issues that arise as a result of those changes, but converting existing strings over to use the constants could maybe be part of another PR? That way it can be done in a more incremental manner.
> 
> what do you think?

My bad. Yes, I do agree, this PR is aimed towards limited scope. Converting strings to constants pollute the readability of the PR. I have reverted back those changes.

manzoorwanijk:
@tn3rb we have constants for store reducer keys as well like `'eventespresso/core'` etc. but there hasn't been any consistency in using them everywhere. In most of the hooks they are used as string when passed to `select` from GB `data`. Should we make a convention about such constants?

tn3rb:
@manzoorwanijk 

> we have constants for store reducer keys as well like 'eventespresso/core' etc. but there hasn't been any consistency in using them everywhere. In most of the hooks they are used as string when passed to select from GB data. Should we make a convention about such constants?

ya I didn't even know those existed until I started diving into the model code to search for bugs, because none of those constants were used in the documentation on how to use the stores.

So to answer your question, YES we should move towards greater consistency and replace strings with constants, just not part of this PR.

manzoorwanijk:
> So to answer your question, YES we should move towards greater consistency and replace strings with constants, just not part of this PR.

@tn3rb  Yes, that would be a separate PR.

Issue #1778, opened 2019-10-31
Update usage of getters/setters where needed


joshfeck:
In #1641 the unnecessary getters and setters for every field/property of an entity in the model system were removed. It appears there are a few places where some of the old, removed aliases are still being used. We'll need to find those and update them.

This is likely related to the issue with editing prices as mentioned in this project card:

https://github.com/eventespresso/event-espresso-core/projects/4#card-28578673

manzoorwanijk:
It's actually a bug related to data hydration where the price type list is not updated in `'eventespresso/lists'` store.

manzoorwanijk:
Created #1780 from that card

PR #1781, opened 2019-11-01
Fix price_type hydration

### LINKED ISSUE #1780: JS error: TypeError: priceTypeEntity is undefined when updating a price via the Card UI
Store rehydration does not update the `price_type` list in `'eventespresso/lists'`, which means no price types are available.

manzoorwanijk:
Fixes #1780 #1778 

## How has this been tested
monkey and unit tests

## Checklist

* [x] My code is tested.

### DIFF
--- a/assets/src/editor/events/initialization/data-hydrator.js
+++ b/assets/src/editor/events/initialization/data-hydrator.js
@@ -13,6 +13,7 @@ const {
 } = dispatch( 'eventespresso/core' );
 
 const {
+	receiveEntityResponse,
 	resolveGetEntities,
 } = dispatch( 'eventespresso/lists' );
 const { receiveSchemaForModelAndResolve } = dispatch( 'eventespresso/schema' );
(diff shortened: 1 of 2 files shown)


assets/src/editor/events/initialization/data-hydrator.js
```diff
@@ -136,7 +137,14 @@ const hydrateEntityData = async ( rawData, schemas ) => {
...
 				}
 			);
-			resolveGetEntities( model );
+			if ( model === 'price_type' ) {
+				receiveEntityResponse(
+					model,
+					undefined,
+					Object.values( entities[ model ] ),
+				);
+				resolveGetEntities( model );
```
tn3rb: this was previously being done for all entities but now it is only happening for prices?!?!?!
was this supposed to be after the closing brace that follows it? 
ie: we just need to receive price types prior to calling `resolveGetEntities( model )` ?

manzoorwanijk: We just need that for `price_type`, doing for others doesn't really matter, that's why I moved it inside that block.

PR #1788, opened 2019-11-04
Auto update ticket quantity when the datetime capacity is changed

### LINKED ISSUE #1747: Automatically decrement ticket quantity fields when related datetimes reg limit gets set
This was a feature in the original ticket editor where the value for the registration limit for a datetime would automatically "sync" the related ticket's quantity field. Animated gif to show the ticket fields updating after saving the event's changes:

ticket-datetime-limit

In the new editor, if you change the datetime reg limit to a smaller number, the related ticket quantity field does not automatically update.
Screen Shot 2019-10-28 at 5 31 21 PM

manzoorwanijk:
Fixes #1747 

## How has this been tested
monkey testing

## Checklist

* [ ] My code is tested.

### DIFF
--- a/assets/src/editor/events/dates-and-times/editor-date/edit-form/use-date-entity-input-config.js
+++ b/assets/src/editor/events/dates-and-times/editor-date/edit-form/use-date-entity-input-config.js
@@ -19,6 +19,11 @@ import {
 } from '@eventespresso/value-objects';
 
 /**
+ * Internal dependencies
+ */
+import { updateTicketQtyAfterCapacityChange } from '../../../mutations/ticket-mutations';
+
+/**
  * input configuration for event date entity edit form
  *
  * @function
@@ -133,13 +138,14 @@ const useDateEntityInputConfig = ( {
 			label: __( 'Capacity', 'event_espresso' ),
 			default: -1,
 			changeListener: ( value ) => {
-				const regLimit = parseInfinity(
+				const capacity = parseInfinity(
 					value || -1,
 					true,
 					true
 				);
-				dateEntity.regLimit = regLimit;
-				updateRelatedTickets( { capacity: regLimit } );
+				dateEntity.regLimit = capacity;
+				const updateTicketQty = updateTicketQtyAfterCapacityChange( capacity );
+				updateRelatedTickets( [ updateTicketQty ] );
 			},
 			min: -1,
 			inputWidth: 3,
--- a/assets/src/editor/events/dates-and-times/editor-date/edit-form/use-date-entity-input-config.js
+++ b/assets/src/editor/events/dates-and-times/editor-date/edit-form/use-date-entity-input-config.js
@@ -9,6 +9,7 @@ import {
 	useEndDateAfterStartDateValidator,
 	useEntityDateChangeListeners,
 	useEntityDateChangeValidators,
+	useEventDateUpdateRelatedTickets,
 } from '@eventespresso/hooks';
 import { parseInfinity } from '@eventespresso/utils';
 import { isModelEntityOfModel } from '@eventespresso/validators';
@@ -60,6 +61,7 @@ const useDateEntityInputConfig = ( {
 	const endDateIsAfterStartDate = useEndDateAfterStartDateValidator(
 		dateInputFormProps
 	);
+	const updateRelatedTickets = useEventDateUpdateRelatedTickets( dateEntity );
 	// useEntityDateChangeValidators use-entity-date-change-validators
 	return useMemo( () => [
 		{
@@ -131,11 +133,13 @@ const useDateEntityInputConfig = ( {
 			label: __( 'Capacity', 'event_espresso' ),
 			default: -1,
 			changeListener: ( value ) => {
-				dateEntity.regLimit = parseInfinity(
+				const regLimit = parseInfinity(
 					value || -1,
 					true,
 					true
 				);
+				dateEntity.regLimit = regLimit;
+				updateRelatedTickets( { capacity: regLimit } );
 			},
 			min: -1,
 			inputWidth: 3,
--- a/assets/src/editor/events/dates-and-times/editor-date/grid-view/event-date-details-panel.js
+++ b/assets/src/editor/events/dates-and-times/editor-date/grid-view/event-date-details-panel.js
@@ -13,6 +13,7 @@ import { useEventDateUpdateRelatedTickets } from '@eventespresso/hooks';
  * Internal dependencies
  */
 import EventDateRegistrationsLink from '../event-date-registrations-link';
+import { updateTicketQtyAfterCapacityChange } from '../../../mutations/ticket-mutations';
 
 const EventDateDetailsPanel = ( { eventDate } ) => {
 	const updateRelatedTickets = useEventDateUpdateRelatedTickets( eventDate );
@@ -39,7 +40,8 @@ const EventDateDetailsPanel = ( { eventDate } ) => {
 						onChange: ( cap ) => {
 							const capacity = parseInfinity( cap, true, true );
 							eventDate.regLimit = capacity;
-							updateRelatedTickets( { capacity } );
+							const updateTicketQty = updateTicketQtyAfterCapacityChange( capacity );
+							updateRelatedTickets( [ updateTicketQty ] );
 							return <InfinitySymbol value={ capacity } asInt />;
 						},
 					},
--- a/assets/src/editor/events/dates-and-times/editor-date/grid-view/event-date-details-panel.js
+++ b/assets/src/editor/events/dates-and-times/editor-date/grid-view/event-date-details-panel.js
@@ -37,10 +37,10 @@ const EventDateDetailsPanel = ( { eventDate } ) => {
 						type: 'text',
 						valueType: 'infinite',
 						onChange: ( cap ) => {
-							cap = parseInfinity( cap, true, true );
-							eventDate.regLimit = cap;
-							updateRelatedTickets( { capacity: cap } );
-							return <InfinitySymbol value={ cap } asInt />;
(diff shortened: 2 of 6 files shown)


assets/src/editor/events/dates-and-times/editor-date/edit-form/use-date-entity-input-config.js
```diff
@@ -131,11 +133,13 @@ const useDateEntityInputConfig = ( {
 			label: __( 'Capacity', 'event_espresso' ),
 			default: -1,
 			changeListener: ( value ) => {
-				dateEntity.regLimit = parseInfinity(
+				const regLimit = parseInfinity(
 					value || -1,
 					true,
 					true
 				);
+				dateEntity.regLimit = regLimit;
```
nicolad: Although this is not directly related to this PR.
We need to reconsider how state management is done.
One way of improvement is by using immutability.

manzoorwanijk: It's already immutable :)

manzoorwanijk:
There is a memoization issue though when you update the datetime capacity inline, the change doesn't reflect unless there is a change in the UI filters. I think that can be fixed in another PR.

assets/src/editor/events/dates-and-times/editor-date/edit-form/use-date-entity-input-config.js
tn3rb: > It's already immutable

???

The model entities are not immutable, some fields that use value objects like date or money fields are immutable, but the objects in whole are not.

or are you referring to the state object itself? In which case, yes the state is immutable.

> We need to reconsider how state management is done.
> One way of improvement might be: immutability.

How would that work using the code above as an example? 
Are you suggesting constructing a whole new event date object when the capacity (regLimit) property changes value?

assets/src/hooks/use-event-date-update-related-tickets.js
```diff
@@ -0,0 +1,35 @@
...
+ */
+const useEventDateUpdateRelatedTickets = ( eventDate ) => {
+	const { tickets: relatedTickets, ticketsLoaded } = useEventDateTickets( eventDate );
+	return useCallback( ( { capacity = null } ) => {
+		if ( ticketsLoaded ) {
+			// Make sure that the ticket qty value is compared with
+			// a non negative datetime capacity value in Math.min()
+			const nonNegativeDTTCapacity = parseInfinity( capacity, false, false );
+			relatedTickets.forEach( ( ticket ) => {
+				if ( capacity !== null ) {
```
tn3rb: since we don't need to do anything when capacity is null, then why not merge this with the conditional above and then bypass everything?

ie:

```js
if ( ticketsLoaded && capacity !== null ) {
    // loop through tickets and possibly adjust quantity
}
```

manzoorwanijk: `parseInfinity` already converts `null` to `Infinity` or `-1` by making use of `isNil( value );`

tn3rb: My comment has nothing to do with `parseInfinity` or how it converts values.

Look at the entire function again:


```js
	( { capacity = null } ) => {
		if ( ticketsLoaded ) {
			// Make sure that the ticket qty value is compared with
			// a non negative datetime capacity value in Math.min()
			const nonNegativeDTTCapacity = parseInfinity( capacity, false, false );
			relatedTickets.forEach( ( ticket ) => {
				if ( capacity !== null ) { <<<--- WE ARE CHECKING THIS REPEATEDLY
					const nonNegativeTKTQty = parseInfinity( ticket.qty, false, false );
					const qty = Math.min( nonNegativeDTTCapacity, nonNegativeTKTQty );
					ticket.qty = parseInfinity( qty, true, true );
				}
			} );
		}
	}
```

`capacity` is passed into the callback but is not modified in any way before being compared against null while looping through tickets to decide whether to adjust the ticket qty. But since the value of `capacity` is never changed anywhere in the function, and we don't end up doing anything when it's incoming value is `null`, we are running that same comparison over and over. We could move that comparison into the very first conditional and then bypass **everything** when `capacity` is `null` since we are not going to end up doing anything anyways.

```js
if ( ticketsLoaded && capacity !== null ) { <<<--- ONLY NEED TO CHECK CAP ONCE
	// Make sure that the ticket qty value is compared with
	// a non negative datetime capacity value in Math.min()
	const nonNegativeDTTCapacity = parseInfinity( capacity, false, false );
	relatedTickets.forEach( ( ticket ) => {
		const nonNegativeTKTQty = parseInfinity( ticket.qty, false, false );
		const qty = Math.min( nonNegativeDTTCapacity, nonNegativeTKTQty );
		ticket.qty = parseInfinity( qty, true, true );
	} );
}
```

assets/src/editor/events/dates-and-times/editor-date/edit-form/use-date-entity-input-config.js
nicolad: `Are you suggesting constructing a whole new event date object when the capacity (regLimit) property changes value?`
Yes, that would be faster since old object will be garbage collected and memory got freed.
From what I saw we already use that pattern since we are using `@wordpress/data`

assets/src/hooks/use-event-date-update-related-tickets.js
manzoorwanijk: I was pointing towards how the value actually reaches there. If you check the `onChange` handler for `capacity`,
```JS
onChange: ( cap ) => {
	cap = parseInfinity( cap, true, true );
	eventDate.regLimit = cap;
	updateRelatedTickets( { capacity: cap } );
	return <InfinitySymbol value={ cap } asInt />;
}
```
You can see that the value is not null.
May be we can pass the actual value of `cap` from the `onChange` handler to the hook for updating the ticket quantity.

manzoorwanijk: Why I have used `if ( capacity !== null )`?
As I mentioned in the commit message, it is a generic hook which can/may be used to update other fields in the related tickets. This is also the reason that the arguments are passed as object `( { capacity = null } )` instead of `( capacity = null )`.

assets/src/editor/events/dates-and-times/editor-date/grid-view/event-date-details-panel.js
```diff
@@ -7,61 +7,66 @@ import { parseInfinity } from '@eventespresso/utils';
...
-						return <InfinitySymbol value={ cap } asInt />;
+const EventDateDetailsPanel = ( { eventDate } ) => {
+	const updateRelatedTickets = useEventDateUpdateRelatedTickets( eventDate );
+	return useMemo(
+		() => {
+			const details = [
+				{
+					id: 'ee-event-date-sold',
+					label: __( 'sold', 'event_espresso' ),
+					value: eventDate.sold || 0,
```
nicolad: might `eventDate` object be null ?
if so, then maybe smth like: `value: (eventDate && eventDate.sold) || 0`

assets/src/editor/events/dates-and-times/editor-date/grid-view/event-date-details-panel.js
```diff
@@ -7,61 +7,66 @@ import { parseInfinity } from '@eventespresso/utils';
...
+					label: __( 'capacity', 'event_espresso' ),
+					value: <InfinitySymbol value={ eventDate.regLimit } asInt />,
+					editable: {
+						type: 'text',
+						valueType: 'infinite',
+						onChange: ( cap ) => {
+							cap = parseInfinity( cap, true, true );
+							eventDate.regLimit = cap;
+							updateRelatedTickets( { capacity: cap } );
+							return <InfinitySymbol value={ cap } asInt />;
```
nicolad: What about:
```
const capacity = parseInfinity( cap, true, true );
eventDate.regLimit = cap;
updateRelatedTickets( { capacity } );
return <InfinitySymbol value={ cap } asInt />;
```
this way we don't mutate cap

assets/src/editor/events/dates-and-times/editor-date/grid-view/event-date-details-panel.js
manzoorwanijk: `evenDate` is expected to be non null.

assets/src/editor/events/dates-and-times/editor-date/grid-view/event-date-details-panel.js
manzoorwanijk: Yes it makes the code better. Thanks

manzoorwanijk: Done :)

nicolad: Thanks :bowing_man:

assets/src/hooks/use-event-date-update-related-tickets.js
manzoorwanijk: > 
> 
> Why I have used `if ( capacity !== null )`?
> As I mentioned in the commit message, it is a generic hook which can/may be used to update other fields in the related tickets. This is also the reason that the arguments are passed as object `( { capacity = null } )` instead of `( capacity = null )`.

@tn3rb I understand your point, please read this quoted comment again. What if in future we want to update something other than quantity for the related tickets? We could use the same hook and that time capacity may not need to be changed. I hope I am not trying create an ultra hypothetical use case. 😄

tn3rb: **GO FOR IT!!!**  😆

manzoorwanijk: Done! Please have a look 😊

PR #1790, opened 2019-11-04
Fix the issue caused by passing InfinitySymbol as initial value.

### LINKED ISSUE #1789: Some inline editable fields result in an error if no change is made to the value.
The fields like `capacity` for datetime and `quantity` for ticket when edited inline, result in an error if you decide not to make any change after switching to edit mode. It happens only if the field has a non-infinity value.

**To Reproduce**

1. Click on the capacity of a datetime.
2. Click anywhere else without making any change.
3. See the error in console.

**Expected behavior**
The focus should be taken away from the field and it should go back to display mode.

manzoorwanijk:
Fixes #1789 

## How has this been tested
monkey testing

## Checklist

* [ ] My code is tested.

### DIFF
--- a/assets/src/utils/parse-infinity.js
+++ b/assets/src/utils/parse-infinity.js
@@ -15,20 +15,26 @@ import { isNil } from 'lodash';
 const parseInfinity = ( number, asInt = false, forDb = false ) => {
 	// returns true for any possible value that could represent infinity
 	const representsInfinity = ( value ) =>
-		value === -1 ||
+		value < 0 ||
 		value === '' ||
 		value === 'INF' ||
 		value === Infinity ||
 		isNil( value );
+
 	if ( number && number.type && number.type.name === 'InfinitySymbol' ) {
 		number = number.props.value;
 	}
+
 	number = representsInfinity( number ) ? Infinity : number;
 	number = number !== Infinity && asInt ? parseInt( number, 10 ) : number;
-	// not infinity OR is infinity but not for db
-	number = number !== Infinity || ( number === Infinity && ! forDb ) ?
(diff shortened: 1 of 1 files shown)


assets/src/utils/parse-infinity.js
```diff
@@ -20,13 +20,10 @@ const parseInfinity = ( number, asInt = false, forDb = false ) => {
...
 		value === Infinity ||
 		isNil( value );
-	number = representsInfinity( number ) || (
-		number.type &&
-		number.type.name === 'InfinitySymbol' &&
-		representsInfinity( number.props.value )
-	) ?
-		Infinity :
-		number;
+	if ( number && number.type && number.type.name === 'InfinitySymbol' ) {
```
nicolad: It would be nice to have this plugin: https://babeljs.io/docs/en/next/babel-plugin-proposal-optional-chaining.html
Have used it on previous project, didn't had any issue.
With that plugin, we'd have:
`	if ( numbe?r.type?.name && number.type.name === 'InfinitySymbol' ) { ...`
***
note: this is optional, thus I do approve the PR with the current solution

manzoorwanijk: Yes, optional chaining can be enabled. I have some other suggestions for that part of the code that I want to discuss with you and @tn3rb

tn3rb:
changes look good

we should add a unit test for this utility and verify that it handles all possible conditions correctly

manzoorwanijk:
> 
> 
> changes look good
> 
> we should add a unit test for this utility and verify that it handles all possible conditions correctly

Yes, It needs that.

PR #1792, opened 2019-11-04
Add tests for parseInfinity()

manzoorwanijk:
This PR adds tests for `parseInfinity()` as discussed in #1790 

## How has this been tested
Unit tests

## Checklist

* [ ] My code is tested.

### DIFF
--- a/assets/src/utils/test/parse-infinity.js
+++ b/assets/src/utils/test/parse-infinity.js
@@ -1,0 +1,163 @@
+
+/**
+ * Internal dependencies
+ */
+import parseInfinity from '../parse-infinity';
+
+const testData = [
+	{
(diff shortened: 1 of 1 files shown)


assets/src/utils/test/parse-infinity.js
```diff
@@ -0,0 +1,163 @@
...
+	testData.forEach( ( test ) => {
+		const {
+			doesThis,
+			input,
+			asInt,
+			forDb,
+			expectedResult,
+		} = test;
+		it( doesThis, () => {
+			const result = () => parseInfinity( input, asInt, forDb );
```
nicolad: Is there a specific reason it is a function and not a plain variable?

manzoorwanijk: Ah yes, it could be a plain variable as well.

Issue #1794, opened 2019-11-04
trashed ticket and/or datetime doesn't persist after updating event


joshfeck:
https://slack-files.com/T02SY781D-FPX4FE88H-f94ea2ed32

tn3rb:
This appears to be an issue in the JS models somewhere.

If you perform the following steps:

- edit the name of a date
- open your browser's dev tools and clear the network panel
- click Save/Update in the top right corner of the Event Editor

you will see an API request with the data for the edited date go out prior to the POST submission

Now when the page refreshes do this:

- find the date that you edited the name of
- click on the hamberder menu icon button in the top right corner of the date card and click "trash date"
- open your browser's dev tools and clear the network panel
- click Save/Update in the top right corner of the Event Editor

this time you will NOT see any requests going out prior to the POST submission

So something is going wrong inside the JS models and the date deletion is not getting queued for persistence, or something like that.

PR #1795, opened 2019-11-05
Fix Unit Tests

tn3rb:
for dramatic heart wrenching back story see: https://travis-ci.com/eventespresso/event-espresso-core/pull_requests


nicolad:
Would be nice to have also some e2e tests, at least for some actions like these:
https://github.com/eventespresso/event-espresso-core/issues/1794

Issue #1773, opened 2019-10-30
Need one default ticket and one default datetime initially set up when first creating a new event


joshfeck:
The default ticket should use the default price set up in Event Espresso > Pricing -or- use the default ticket set previously if another event's ticket's "Default Ticket?" option is set to true.

PR #1796, opened 2019-11-05
EDTR : Add Default Entities for New Events

tn3rb:
Currently the EDTR doesn't really work with new events because the draft Post ID is not getting detected, nor do any dates or tickets appear.

This branch adds draft Post ID detection and default Dates, Tickets, and Prices for New Events.


joshfeck:
PR for #1773

assets/src/editor/events/tickets/editor-ticket/use-tickets-for-event-editor-ticket-list.js
```diff
@@ -37,6 +37,9 @@ const useTicketsForEventEditorTicketList = ( isChained ) => {
 	}
 
 	return useSelect( ( select ) => {
+		if ( filteredDateIds === null ) {
+			return { tickets: EMPTY_ARRAY, ticketsLoaded: false };
+		}
```
manzoorwanijk: Good one

assets/src/editor/events/tickets/editor-ticket/use-tickets-for-event-editor-ticket-list.js
```diff
@@ -26,7 +26,10 @@ const useTicketsForEventEditorTicketList = ( isChained ) => {
 
 	const [ finishedResolution, setFinishedResolution ] = useState( false );
 	const { resolveGetRelatedEntitiesForIds } = dispatch( 'eventespresso/core' );
-	if ( filteredDateIds.length && ! finishedResolution ) {
+	if ( ! finishedResolution &&
+		Array.isArray( filteredDateIds ) &&
+		filteredDateIds.length
+	) {
```
manzoorwanijk: I didn't realize that `filteredDateIds` could be a non array, thanks 👍

assets/src/editor/events/tickets/editor-ticket/edit-form/use-ticket-form-loading-state.js
```diff
@@ -0,0 +1,19 @@
...
+ */
+const useTicketFormLoadingState = () => {
+	const [ loading, setLoading ] = useState( false );
+	return {
+		loadingTicketForm: loading,
+		setLoadingTicketForm: setLoading,
+	};
+};
+
+export default useTicketFormLoadingState;
```
manzoorwanijk: Why is this specific to Ticket form? It can have a generic use. Also, does it it do anything more than `useState()`?

assets/src/editor/events/tickets/editor-ticket/editor-ticket-entities-list.js
```diff
@@ -91,6 +92,10 @@ const EditorTicketEntitiesList = ( { ...otherProps } ) => {
 		ticketsSortedBy,
 		Array.isArray( tickets ) ? tickets.length : 0,
 	] );
+	const {
+		loadingTicketForm,
+		setLoadingTicketForm,
+	} = useTicketFormLoadingState();
```
manzoorwanijk: Why not `useState()`?

assets/src/editor/events/tickets/editor-ticket/use-tickets-for-event-editor-ticket-list.js
tn3rb: i didn't either but then i got an `undefined` error at one point :shrug:

assets/src/editor/events/tickets/editor-ticket/editor-ticket-entities-list.js
tn3rb: good question!! :laughing:

I think maybe I initially thought that more was going to need to happen in that hook and then I just didn't realize how idiotic the final code was :(

tn3rb:
i was trying to fix the Travis tests yesterday from my laptop which isn't set up as well as my office cpu so there was a lot of hit n miss and guesswork happening.

assets/src/editor/events/tickets/editor-ticket/edit-form/new-ticket-form-modal.js
```diff
@@ -0,0 +1,71 @@
...
+		}
+	} );
+	// because we have to wait for a valid ticket entity to be created,
+	// we can't simply open the editor via the Add New Ticket click event,
+	// so instead we toggle the following flag to indicate this
+	if ( toggleTicketEditor ) {
+		openTicketEditor();
+	}
+	// once the ticket editor has been opened, we can flip that toggle to off
+	const onEditorOpen = useCallback( () => {
```
nicolad: `useCallback` [returns a memoized callback](https://reactjs.org/docs/hooks-reference.html#usecallback).
This isn't an expensive function, we can simply have:
`const onEditorOpen = () => setToggleTicketEditor( false );`

assets/src/editor/events/tickets/editor-ticket/edit-form/new-ticket-form-modal.js
```diff
@@ -0,0 +1,71 @@
...
+		setToggleTicketEditor( false );
+	}, [] );
+	// once the ticket editor closes, unset the cached ticket,
+	// which will get picked up by the main ticket editor list
+	// and free up the cache for creating another new ticket
+	const onEditorClose = useCallback( () => {
```
nicolad: Do we need a memoized callback here?

assets/src/editor/events/tickets/editor-ticket/edit-form/new-ticket-form-modal.js
tn3rb: my understanding is that when you pass callbacks down as props that they can trigger rerenders because a new instance of the callback is generated each time and therefore any comparisons will fail, so even though it may not be an expensive function, it could cause unnecessary renders for a component that IS expensive. 

If I'm misinterpreting something here then I can change this.

Issue #1739, opened 2019-10-23
Avoid mysterious fatal errors during 4.10 migrations


mnelson4:
## Issue Overview

There are occasional reports of errors during the 4.10 DMS. Eg,
![Screen Shot 2019-10-23 at 2 39 14 PM](https://user-images.githubusercontent.com/1766382/67425261-878e7e80-f58c-11e9-8d6b-be0c21c15c42.png)
Information is a little sparse, but I'm suspicious that the database queries are timing out (especially if the tables are large, like the above is).
The table being updated has 2 indexes: on the `EVT_ID` column, and on the `QSG_ID` column. The update and delete queries being ran use the `EQG_primary`, `EVT_ID` and `QSG_ID` columns during the query. 
They should be able to make use of at least one of those indexes (although not both, because once we've filtered by one, eg by `EVT_ID`, the remaining items are not organized by `QSG_ID`, so it's fastest to ignore the second index at that point).
BUT, because the database query is first querying `EQG_primary`, I think the remaining results are no longer indexed like they were originally; so NEITHER index gets used.
And if there are over a million database records, that's bad news bears.

Possibly, reordering the database query's WHERE conditions could mean we're able to use the query.
Using MySQL's `describe query` may help to confirm what indexes are in use(if I'm able to understand it's output :confounded:) and if the change makes any different to the query's efficiency.

## Bug report or feature request?

* [x] Bug
* [ ] Feature
* [ ] Neither

## Environment Data:

Version of EE: 
4.10.2

Version of WordPress:
?
PHP Version:  
?
Browser used:  
Shouldn't matter.

## Steps to Reproduce (for bugs)

Not sure, but possibly: install EE 4.9. Create over a million event-question-group relations. Upgrade to 4.10. Run the migrations on a weak-sauce server.

## Expected Behaviour

They should all work fine.

## Current Behaviour

Random fatal error somewhere along the way.

mnelson4:
I ran `EXPLAIN UPDATE wp_esp_event_question_group SET EQG_additional = 1 WHERE QSG_ID = 0 and EVT_ID = 0`, which is one of the queries we're using,  got this output:

```
"id","select_type","table","partitions","type","possible_keys","key","key_len","ref","rows","filtered","Extra"
"1","UPDATE","wp_esp_event_question_group",\N,"range","EVT_ID,QSG_ID","EVT_ID","8","const","1","100.00","Using where"
```

[...]

PR #1805, opened 2019-11-05
Shorter Table locking queries during 4.10 migration

mnelson4:

An attempt to fix https://github.com/eventespresso/event-espresso-core/issues/1739 by having the 4.10 migration not doing UPDATE queries with lots of conditions (which can make the queries take a while, and locks the table the whole time) and instead do a SELECT query with all the conditions (which is about the same speed, but doesn't lock the table) to get all the IDs we want to update, then do a really quick UPDATE query (which locks the table for much less time).
Here's my benchmarking (time in milliseconds, my wp_esp_event_question_group table had 2000 records):

```
       | master run 1 | master run 2 | this branch run 1 | this branch run 2
update1| 03.001       | 02.000       | 01.000            | 01.000
update2| 10.000       | 04.000       | 04.000            | 04.001
```

So not only is the table locked for less time, it seems to consistently be about an order of magnitude faster (I'm suspicious of that- I don't have a controlled environment by any means, but certainly it's not significantly slower than before)

## How has this been tested

* [ ] Run the 4.10 DMS. There should be no errors

## Checklist

* [x] My code is tested.

### DIFF
--- a/core/data_migration_scripts/4_10_0_stages/EE_DMS_4_10_0_Event_Question_Group.dmsstage.php
+++ b/core/data_migration_scripts/4_10_0_stages/EE_DMS_4_10_0_Event_Question_Group.dmsstage.php
@@ -40,7 +40,7 @@ class EE_DMS_4_10_0_Event_Question_Group extends EE_Data_Migration_Script_Stage_
             global $wpdb;
             // If the question group was also for primary attendees, we should just update that row.
             // And we delete this row.
-            // Updating all the rows could be slow on massive DBs, so do the slow selection first, then a quick delete
+            // Updating all the rows could be slow on massive DBs, so do the slow selection first, then a quick update
             // in order to avoid locking the table for too long.
             $ids_to_update = $wpdb->get_col(
                 $wpdb->prepare(
--- a/core/data_migration_scripts/4_10_0_stages/EE_DMS_4_10_0_Event_Question_Group.dmsstage.php
+++ b/core/data_migration_scripts/4_10_0_stages/EE_DMS_4_10_0_Event_Question_Group.dmsstage.php
@@ -45,7 +45,7 @@ class EE_DMS_4_10_0_Event_Question_Group extends EE_Data_Migration_Script_Stage_
             $ids_to_update = $wpdb->get_col(
                 $wpdb->prepare(
                     'SELECT EQG_ID FROM ' . $this->_old_table . ' WHERE EQG_primary=1 AND EVT_ID=%d AND QSG_ID=%d',
-                    $event_question_group['EQG_ID'],
+                    $event_question_group['EVT_ID'],
                     $event_question_group['QSG_ID']
                 )
             );
--- a/core/data_migration_scripts/4_10_0_stages/EE_DMS_4_10_0_Event_Question_Group.dmsstage.php
+++ b/core/data_migration_scripts/4_10_0_stages/EE_DMS_4_10_0_Event_Question_Group.dmsstage.php
@@ -40,6 +40,8 @@ class EE_DMS_4_10_0_Event_Question_Group extends EE_Data_Migration_Script_Stage_
             global $wpdb;
             // If the question group was also for primary attendees, we should just update that row.
             // And we delete this row.
+            // Updating all the rows could be slow on massive DBs, so do the slow selection first, then a quick delete
+            // in order to avoid locking the table for too long.
             $ids_to_update = $wpdb->get_col(
                 $wpdb->prepare(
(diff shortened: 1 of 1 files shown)


core/data_migration_scripts/4_10_0_stages/EE_DMS_4_10_0_Event_Question_Group.dmsstage.php
```diff
@@ -40,17 +40,24 @@ protected function _migrate_old_row($event_question_group)
...
-                ['EQG_additional' => true],  // data
-                [
-                    'EQG_primary' => true,
-                    'EVT_ID' => $event_question_group['EVT_ID'],
-                    'QSG_ID' => $event_question_group['QSG_ID']
-                ],  // where
-                array( '%d' ),   // data format
-                array( '%d', '%d', '%d' )  // where format
+            // Updating all the rows could be slow on massive DBs, so do the slow selection first, then a quick delete
+            // in order to avoid locking the table for too long.
```
tn3rb: not that it matters *that* much but i think you meant to say "update" and not "delete"

ie:

```
// Updating all the rows could be slow on massive DBs, so do the slow selection first,
// then a quick update in order to avoid locking the table for too long.
```

mnelson4:
Thanks @tn3rb, I fixed the typo in the comment. Passing off for testing

joshfeck:
I tested this and found that question groups assigned to events for Additional attendees were no longer assigned after running the migration. So this may need some fixing.

mnelson4:
Thanks for catching that Josh! I did exactly what the testing step I put was: just check there were no errors. I forgot to check the DMS actually did what it was supposed to do!
The latest commit resolved the issue for me.

mnelson4:
Oh and I re-ran the benchmarking, and updated the results (because the queries will certainly run super fast if they error out before actually doing anything!). The results show the queries take about the same time as they used to.
If this doesn't resolve the issue for the user who's reproducing the issue, it would probably be very helpful to get a copy of their DB, if possible.

joshfeck:
It's still broken with the latest commits to this branch. Before the migration, I set one question group for Additional attendees. After the migration, that question group's box is unchecked when I go in to edit the events.

mnelson4:
Bah! I don't know what I was on last time. It retried it and it wasn't working (I remember specifically checking it, but I must have looked at the primary attendees' question groups by accident). Its (again) working for me.

joshfeck:
OK this works now.

Do the unit tests needs fixing before we merge?

mnelson4:
oh nuts, yes that unit test failure needs to be resolved.

mnelson4:
Ok its now passing the test (it turned out to be a legitimate failure- it wasn't handling question groups that were ONLY for additional attendee correctly.)

Would you like to re-test? I've re-tested but according to this ticket that means just about nothing! 😓

mnelson4:
Oh also, @joshfeck do you want to try this on the client's site who's experiencing the issue? If it doesn't resolve the issue for them (and it very well might not, it's a stab in the dark), we may choose to not bother releasing these changes (they're not significantly faster, and there's always a possibility of regressions with any changes.)

joshfeck:
> @joshfeck do you want to try this on the client's site who's experiencing the issue?

I do not have access to that site, but I can send them another reply to see where they're at.

mnelson4:
Josh shared the following screenshot from a site owner experiencing a problem with the 4.10 migration:
0 items migrated
The DMS thought there was more items to migrate, but everytime it queried for more, it didn't find any more rows needing migrating, and this got it stuck in a loop. (One way to make that happen would be to start a large migration, then manually delete some of the rows from `wp_esp_event_question_group` during the migration. Although I kinda doubt that's what the customer did.)
So I added a commit that handles that situation better, by assuming "if there's nothing more to migrate, I must be done!"
So this is off for review again. (Note: I think this same change as https://github.com/eventespresso/event-espresso-core/pull/1805/commits/ed4e6ff755bb096738c3387e294be1f703c18047 would make sense to apply to `\EE_Data_Migration_Script_Stage_Table::_migration_step()`, but that would affect nearly all our other migrations and would require a lot more testing to verify it didn't break any of them, so I didn't make that change here. But I'm open to it if anyone else thinks it should happen.)

mnelson4:
Ok rebased on master and forced push.

garthkoyle:
Does this need rebased before I try testing the migrations?

garthkoyle:
I tested this by: 
- Restting all my EE data
- installing 4.9.25.p
- creating an event
- creating a new question
- creating a new question group
- checking out master
- running the migrations

I verified the questions for additional attendees were still selected.

What do we do with this now, @tn3rb ?

tn3rb:
i just need to fix the linting issues, and then I'll merge it
thnx for testing

PR #1809, opened 2019-11-06
Fet/add action to transactions actions

raphaelkross:

Related to https://github.com/eventespresso/EE4-Promotions/issues/20 - adding a new filter where will allow us to add more fields to Transactions actions area, therefore allowing to add a new Apply Discount input.

## How has this been tested

It's just a new action, no need to be tested. Required by => https://github.com/eventespresso/EE4-Promotions/pull/24

## Checklist

* [x] My code is tested.


admin_pages/transactions/templates/txn_admin_details_main_meta_box_txn_details.template.php
```diff
@@ -302,6 +302,12 @@ class="admin-primary-mbox-total-tr<?php echo $pay_totals_class; ?>">
                         <?php esc_html_e('Apply Refund', 'event_espresso'); ?>
                     </a>
                 </li>
+                <?php
+                /**
+                 * Allows extend the fields at actions area.
+                 */
+                do_action('AHEE__txn_admin_details__after_actions_buttons');
```
tn3rb: this should use the full template name in the filter name, so:

```suggestion
                do_action('AHEE__txn_admin_details_main_meta_box_txn_details__after_actions_buttons');
```

raphaelkross: Awesome! Thanks for the suggestion 👍

joshfeck:
Per the discussion in chat, we're going to move the action hook so it's outside the conditional, pass in the $can_edit_payments arg. The promotions add-on will have some logic to check for $can_edit_payments before it will show the promo form.

PR #1812, opened 2019-11-06
Make direct POST request upon date event creation.

nicolad:
##  Problem this Pull Request solves
This approach is aimed toward simplification of the existing complexity.
 ***
This is just a draft.
A more optimal solution would be to use some patterns like:
* https://redux.js.org/advanced/example-reddit-api
or
* https://github.com/coinbase/rest-hooks

### DIFF
--- a/assets/src/editor/events/ticket-assignments-manager/submit-ticket-assignments-button.js
+++ b/assets/src/editor/events/ticket-assignments-manager/submit-ticket-assignments-button.js
@@ -3,8 +3,10 @@
  */
 import PropTypes from 'prop-types';
 import { useMemo, useState } from '@wordpress/element';
+import apiFetch from '@wordpress/api-fetch';
 import { twoColumnAdminFormLayout } from '@eventespresso/components';
 import { __ } from '@eventespresso/i18n';
+import { applyQueryString } from '@eventespresso/model';
 
 /**
  * Internal imports
@@ -24,10 +26,11 @@ const nullFunc = () => true;
  */
 const SubmitTicketAssignmentsButton = ( {
 	assignedState,
-	hasNoAssignments,
 	beforeEditorClose = nullFunc,
 	closeEditor = nullFunc,
+	dateEntity,
 	disabled = false,
+	hasNoAssignments,
 } ) => {
 	const [ submitting, setSubmitting ] = useState( false );
 	const processChanges = useProcessTicketAssignments( {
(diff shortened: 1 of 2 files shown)


assets/src/editor/events/ticket-assignments-manager/submit-ticket-assignments-button.js
```diff
@@ -37,11 +40,31 @@ const SubmitTicketAssignmentsButton = ( {
...
+		const path = applyQueryString( modelName );
+		const request = {
+			path: isNew ? path : path + '/' + id,
+			method: isNew ? 'POST' : 'PUT',
+			data: forPersist,
+		};
+
+		try {
+			processChanges();
+			await apiFetch( request );
```
tn3rb: this won't add the date to the store though ?

I don't like the idea of mixing persistence and API logic directly in with UI components.
I'm not opposed to having updates being made immediately (as opposed to waiting until the user clicks the WP Post `Update` button) but I think that should be managed by a different layer so to speak.

As well, the "Instant Updates" (changes saved immediately like Google Docs) versus "Deferred Updates" (changes not saved until you click button like current WP admin) paradigm  should be discussed with  support and management first.

assets/src/editor/events/ticket-assignments-manager/submit-ticket-assignments-button.js
```diff
@@ -37,11 +40,31 @@ const SubmitTicketAssignmentsButton = ( {
...
+		}
+	};
+
 	return useMemo( () => (
 		<FormSubmitButton
-			onClick={ () => processChanges() }
+			onClick={ onClick }
 			buttonText={ __(
-				'Update Ticket Assignments',
+				'Updating Ticket Assignments AND CREATE EVENT DATE NOW !!!',
```
tn3rb: srsly? way too much text for a button and what's with the all caps?

manzoorwanijk:
It actually needs a thorough discussion before going ahead with this approach.

assets/src/editor/events/ticket-assignments-manager/submit-ticket-assignments-button.js
nicolad: :)
it is just a draft PR and it will get closed, it wasn't intended to be merged

nicolad:
> What is the actual problem that we are trying to solve here?
I was trying to find a way to bypass the complexity created by JS models and just use `forPersist` data from entity.
But it seems that things are too tightly coupled and I can't get away such simple.
Also, this PR has been created before the initiative to rethink/ refactor the JS models.

PR #1815, opened 2019-11-07
Tweak Filter Bar Styles

tn3rb:
In #1774 the layout and structure of the entity filter bar filters changed which resulted in some of the CSS styles no longer getting applied. 

The fixes in this branch include changing the colour for the main filter icon buttons when they are active, as well as fixing some margins that were out of whack.

before:
filter bar styles - before

after:
filter bar styles - after

### DIFF
--- a/assets/src/components/ui/entity-list-filter-bar/collapsible.js
+++ b/assets/src/components/ui/entity-list-filter-bar/collapsible.js
@@ -33,7 +33,7 @@ const Collapsible = ( {
 	perPage,
 	searchText,
 	setPerPage,
-	setSearchText,	
+	setSearchText,
 	showEntityFilters = false
 } ) => {
 	const ref = useRef();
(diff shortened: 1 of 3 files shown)


nicolad:
Thanks for the tweaks, I should've applied them in the PR related to filters.
***
As a side note and as medium priority focus we should be considering a more maintainable approach regarding CSS.
I like the approach `wordpress-seo` had applied, please let me know your thoughts:
https://github.com/Yoast/wordpress-seo/blob/trunk/js/src/components/modals/Container.js
https://github.com/Yoast/javascript/blob/develop/packages/helpers/src/styled-components.js
https://github.com/Yoast/wordpress-seo/blob/trunk/js/src/components/contentAnalysis/HelpLink.js
https://github.com/Yoast/wordpress-seo/blob/trunk/js/src/components/AnalysisUpsell.js
https://github.com/Yoast/wordpress-seo/blob/trunk/js/src/components/LocalSEOUpsell.js

Issue #1784, opened 2019-11-01
Setup pre-commit hooks with husky and lint-staged


nicolad:
This setup will mirror what has been configured in GB:
https://github.com/WordPress/gutenberg/blob/master/package.json#L208

More info about the setup in this [post](https://codeburst.io/continuous-integration-lint-staged-husky-pre-commit-hook-test-setup-47f8172924fc)

PR #1818, opened 2019-11-07
Added and configured husky and lint-staged.

nicolad:
Implements: https://github.com/eventespresso/event-espresso-core/issues/1784
This setup helps us run some specific checks before commit is being made. If checks are failing commit is not completed.

## How has this been tested
Peek 2019-11-07 14-27

## Checklist

* [ ] My code is tested.


package.json
```diff
@@ -14,33 +16,43 @@
 		}
 	},
 	"scripts": {
```
nicolad: Here I've ordered scripts alphabetically.

nicolad: And added:
```
"lint-js": "wp-scripts lint-js",
"lint-js:fix": "npm run lint-js -- --fix",
```

manzoorwanijk: Yes, got that... It's good to be sorted 👍

package.json
```diff
@@ -1,6 +1,8 @@
 {
 	"name": "@eventespresso/core",
 	"version": "1.0.0",
+	"author": "Event Espresso",
+	"license": "GPL-2.0",
```
tn3rb: we're bumping up to GPLv3 now

package.json
```diff
@@ -14,33 +16,43 @@
...
 		"test-unit": "wp-scripts test-unit-js --config tests/javascript-config/unit/jest.config.json",
-		"test-unit:update": "npm run test-unit -- --updateSnapshot",
-		"test-unit:watch": "npm run test-unit -- --watch",
 		"test-unit:coverage": "npm run test-unit -- --coverage",
 		"test-unit:coverage-ci": "npm run test-unit -- --coverage --maxWorkers 1 && codecov",
-		"lc": "jsgl --local ./",
-		"build:css": "node ./bin/css-builder/css-builder.js"
+		"test-unit:update": "npm run test-unit -- --updateSnapshot",
+		"test-unit:watch": "npm run test-unit -- --watch",
+		"watch": "concurrently \"webpack --watch --config webpack.dev.js\" \"npm-watch build:css\"",
```
tn3rb: hey I've been thinking it would be handy to have a command for running a dev build but without the watch flag added.
ie: something like:

```js
		"dev": "concurrently \"webpack --config webpack.dev.js\" \"npm-watch build:css\"",
```

is there already a convention for naming a command like that other than "dev" ?

If so, we can use that...
although imho it would make more sense to have:

- `npm run dev`
- `npm run prod` 

rather than

- `npm run watch`
- `npm run build`

and for the watch and polling dev builds we could have:

- `npm run dev-w` for `webpack --watch --config webpack.dev.js`
- `npm run dev-p` for `webpack --watch --watch-poll --config webpack.dev.js`

what do you think?
I have no experience with these things outside of this project, so if my suggestions are comical, then plz feel free to laugh! I can take it !

Bottom line is that we should follow standard conventions as much as possible for the benefit of all

nicolad: In context of gatsby projects I've seen `dev` being used over the longer `develop`, so I am fine with `dev`, rather than `watch` or `develop`.
I've noticed `build` being more often used than `prod`, but don't want to debate over this, but rather find an agreement and move forward since this is a minor thing.

package.json
```diff
@@ -22,6 +22,9 @@
 		"ci:build": "npm run build",
 		"ci:lint": "npm run lint",
 		"ci:test": "npm run test-unit:coverage-ci",
+		"dev": "concurrently \"webpack --config webpack.dev.js\" \"npm-watch build:css\"",
+		"dev-w": "webpack --watch --config webpack.dev.js",
```
tn3rb: this would still need to be 

```suggestion
		"dev-w": "concurrently \"webpack --watch --config webpack.dev.js\" \"npm-watch build:css\"",
```

in order to build CSS as well no?

nicolad: I saw that it's different than the old `watch`, but I trusted and copy-pasted that from you :D

tn3rb: HA!!

ya don't blindly follow **MY** suggestions!!!

js-green-licenses.json
```diff
@@ -4,6 +4,7 @@
 		"GPL-2.0",
 		"GPL-2.0-or-later",
 		"GPL-2.0+",
+		"GPL-3.0",
```
nicolad: Interesting, I thought that `GPL-3.0` is included in `"GPL-2.0-or-later"`, in the `or-later` part :)

tn3rb: ya I thought the exact same thing, but 2+ must mean things like 2.1, 2.1, etc

PR #1820, opened 2019-11-07
Update related prices when the ticket price changes.

### LINKED ISSUE #1749: Fix ticket price editing in EDTR for existing ticket after switching from original editor
1) start with the original editor, set a ticket’s price to 10.00
2) save the event
3) Use the option to switch to new editor
4) go in and edit the event
Screen Shot 2019-10-28 at 7 14 08 PM
5) using the ticket Card UI, change the price to 20.00
Screen Shot 2019-10-28 at 7 14 20 PM
Screen Shot 2019-10-28 at 7 14 26 PM
6) Save the event
7) Go to the front end to register, the price will be 10.00
Screen Shot 2019-10-28 at 7 14 55 PM

Debug output in ticket price calculator:
Screen Shot 2019-10-28 at 7 16 02 PM
and in the ticket editor:
Screen Shot 2019-10-28 at 7 17 07 PM

URL to edit the above event on dev site:
https://brent.eventespresso.com/wp-admin/admin.php?page=espresso_events&action=edit&post=63

manzoorwanijk:
Fixes #1749 

## How has this been tested
Unit tests

## Checklist

* [x] My code is tested.

### DIFF
--- a/assets/src/editor/events/tickets/editor-ticket/grid-view/inline-edit-ticket-price.js
+++ b/assets/src/editor/events/tickets/editor-ticket/grid-view/inline-edit-ticket-price.js
@@ -14,6 +14,7 @@ import { Money, SiteCurrency } from '@eventespresso/value-objects';
  */
 import useTicketBasePriceCalculator from
 	'../price-calculator/hooks/use-ticket-base-price-calculator';
+import { useBasePrice } from '../price-calculator/hooks';
 
 /**
  * InlineEditTicketPrice inline-edit-ticket-price
@@ -37,11 +38,16 @@ const InlineEditTicketPrice = ( {
 	);
 	const ticketPriceAmount = ticket.price.amount.toNumber();
 
+	const basePrice = useBasePrice( prices );
 	const updateTicketPrice = useCallback(
 		( amount ) => {
 			amount = parseMoneyValue( amount );
 			ticket.price = new Money( amount, SiteCurrency );
-			calculateTicketBasePrice( amount );
+			const basePriceValue = calculateTicketBasePrice( amount );
+			basePrice.amount = new Money(
+				parseMoneyValue( basePriceValue ),
+				SiteCurrency
+			);
 			return amount;
 		},
 		[ calculateTicketBasePrice ]
--- a/assets/src/editor/events/tickets/editor-ticket/price-calculator/hooks/use-base-price.js
+++ b/assets/src/editor/events/tickets/editor-ticket/price-calculator/hooks/use-base-price.js
@@ -8,13 +8,11 @@ const { BASE_PRICE_TYPES } = priceTypeModel;
 
 /**
  * @function
- * @param {number|string} ticketId
- * @param {Object} formData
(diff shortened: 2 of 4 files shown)


assets/src/editor/events/tickets/editor-ticket/price-calculator/hooks/use-base-price.js
```diff
@@ -8,13 +8,11 @@ const { BASE_PRICE_TYPES } = priceTypeModel;
...
- * @param {BaseEntity[]} prices,
- * @return {Array} price modifiers
+ * @param {Array} prices
+ * @return {Object} A price
  */
-const useBasePrice = useMemo( ( prices ) => prices.find(
+const useBasePrice = ( prices ) => useMemo( () => prices.find(
 	( price ) => price.PRT_ID === BASE_PRICE_TYPES.BASE_PRICE
-) );
+), [ prices ] );
```
tn3rb: doesn't this run the risk of memoization issues?
previously the prices were always passed in and therefore were always "fresh"

nicolad:
@manzoorwanijk 
I see a repeating pattern here:
image

can we extract that into a selector?

assets/src/editor/events/tickets/editor-ticket/grid-view/inline-edit-ticket-price.js
```diff
@@ -37,11 +38,16 @@ const InlineEditTicketPrice = ( {
...
 	const updateTicketPrice = useCallback(
 		( amount ) => {
 			amount = parseMoneyValue( amount );
 			ticket.price = new Money( amount, SiteCurrency );
-			calculateTicketBasePrice( amount );
+			const basePriceValue = calculateTicketBasePrice( amount );
+			basePrice.amount = new Money(
+				parseMoneyValue( basePriceValue ),
+				SiteCurrency
+			);
```
tn3rb: I know this is a super small block of code,
but if we need to use this logic in more places
we should think about extracting it into a central location.
I'm thinking that mutations folder you added recently would be a good place for things like this,
but I also want to hold off until we talk about the whole model refactor thing because that could change everything

assets/src/editor/events/tickets/editor-ticket/price-calculator/hooks/use-base-price.js
manzoorwanijk: No, it doesn't. The reason for that is the base price entity for a ticket is always going to be the same. Technically it just needs to run once.

assets/src/editor/events/tickets/editor-ticket/grid-view/inline-edit-ticket-price.js
manzoorwanijk: Yes, we can do the same as we did for updating related tickets for a datetime. We can create a mutation function for such operation and we can use it wherever we need it. I think that can be done in another PR.

manzoorwanijk:
@nicolad yes we can extract the logic as @tn3rb mentioned but I think that should be another PR.

PR #1822, opened 2019-11-07
Added eslint-plugin-react and eslint-plugin-react-hooks.

nicolad:
## Checklist

* [ ] My code is tested.


nicolad:
I turned `off` `comma-dangle` rule otherwise we've got to deal with a lot of errors:

image

nicolad:
There are a number of warnings.
We can disable some of which we might think are not relevant, but also add new ones, like `no-magic-numbers`. That needs to be done in a separate PR otherwise, this would get messy and hard to follow.

image

nicolad:
This PR needs to be merged after this one will get merged: https://github.com/eventespresso/event-espresso-core/pull/1840

nicolad:
@tn3rb @manzoorwanijk I need your blessings here.

tn3rb:
> This PR needs to be merged after this one will get merged: #1840

can't do it now because that PR has already been merged, but in the future you can set another PR as a dependency and it will prevent your PR from getting merged prior to the other PR

PR #1823, opened 2019-11-08
Fix Required Fields Notice Display - closes #1703

### LINKED ISSUE #1703: Remove "All fields marked with an asterisk (*) are required" text from form information
I think this message at the top of the Event Date & Ticket editors, along with the close icon next to it, isn't needed:

Screen Shot 2019-10-14 at 5 31 14 PM

Most people know that an asterisk next to a form field will mean it's required. Also in this case the required fields are automatically populated with something, and you can't blank out the required fields anyway.

tn3rb:
plz see #1703

this branch does the following:

- separates required fields notices from other form info notices
- moves the required fields notice to the bottom of forms
- fixes dismiss buttons so they don't display when form info isn't dismissable
- tweaks the text "focus priority" (size and colour) for form info notices and help text
- uses classNames library for setting css classes

### DIFF
--- a/assets/src/components/form/base/form-info-base.js
+++ b/assets/src/components/form/base/form-info-base.js
@@ -1,6 +1,7 @@
 /**
  * External imports
  */
+import classNames from 'classnames';
 import PropTypes from 'prop-types';
 import { __ } from '@eventespresso/i18n';
 import { Dashicon, IconButton } from '@wordpress/components';
@@ -38,12 +39,14 @@ const FormInfoBase = ( {
 	onDismiss,
 	formInfoVars,
 } ) => {
-	let classes = htmlClass ?
-		`${ htmlClass } ee-form-info` :
-		'ee-form-info';
-	classes = dashicon ?
-		`${ classes } ee-form-info-type-${ dashicon } has-info-type` :
-		classes;
+	const classes = classNames(
+		htmlClass,
+		'ee-form-info',
+		{
+			[ `ee-form-info-type-${ dashicon }` ]: dashicon,
+			'has-info-type': dashicon,
+		}
+	);
 	const typeIcon = dashicon ? (
 		<Dashicon
 			className="ee-form-info-type"
--- a/assets/src/components/form/inputs/base/input-help-text.js
+++ b/assets/src/components/form/inputs/base/input-help-text.js
@@ -8,7 +8,10 @@ import PropTypes from 'prop-types';
  */
 const InputHelpText = ( { helpTextID, helpText } ) => {
 	return helpText ? (
-		<p id={ helpTextID } className="espresso-input-help-text" >
+		<p
+			id={ helpTextID }
+			className={ 'espresso-input-help-text ee-focus-priority-6' }
+		>
 			{ helpText }
 		</p>
 	) : null;
--- a/assets/src/components/form/layouts/two-column-admin/form-section.js
+++ b/assets/src/components/form/layouts/two-column-admin/form-section.js
@@ -42,6 +42,7 @@ const FormSection = ( {
 	...extraProps
 } ) => {
 	const {
+		formInfoNotice,
 		requiredNotice,
 		errorsNotice,
 		hasValidationErrors,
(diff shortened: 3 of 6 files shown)


assets/src/components/form/base/form-info-base.js
```diff
@@ -38,12 +39,14 @@ const FormInfoBase = ( {
...
-		`${ classes } ee-form-info-type-${ dashicon } has-info-type` :
-		classes;
+	const classes = classNames(
+		htmlClass,
+		'ee-form-info',
+		{
+			[ `ee-form-info-type-${ dashicon }` ]: dashicon,
+			'has-info-type': dashicon,
+		}
+	);
```
manzoorwanijk: 👍

assets/src/components/form/layouts/two-column-admin/form-section.js
```diff
@@ -51,20 +52,37 @@ const FormSection = ( {
...
-	), [ showRequiredNotice ] );
-	const formErrors = hasValidationErrors ? (
-		<FormInfo
-			formInfo={ errorsNotice }
-			dashicon={ 'warning' }
-			dismissable={ false }
-		/>
-	) : null;
+	const formInfo = useMemo(
+		() => formInfoNotice ? (
```
manzoorwanijk: why not `formInfoNotice && ( ...` or `!! formInfoNotice && ( ...`? to be safer?

tn3rb: I may have done something wrong but at some point in other components I had been doing something like:

```js
const thing = showThing && <Thing />;

return (
	<SomeComponent>
		{ thing }
	</SomeComponent>
);
```

and when `showThing` was `false` I would get `0` rendered where `{ thing }` was.

Again, it's totally possible I was doing something wrong, because that was earlier on in my budding React career, but that's why I don't use that shortcut.

I did just notice that I passed the wrong dependencies to `useMemo` so I'll be pushing changes to this in a minute

manzoorwanijk: `false` would not render anything. Just to make sure that you are making a boolean comparison, add `!!` before the expression.

nicolad: Yeah, or sometimes I do this:
```
const thing = Boolean(showThing) && <Thing />;
```
If I know that `showThing` might have value `0`.
In this example `0` will be coerced to `false`.

assets/src/components/form/inputs/base/input-help-text.js
```diff
@@ -8,7 +8,10 @@ import PropTypes from 'prop-types';
  */
 const InputHelpText = ( { helpTextID, helpText } ) => {
 	return helpText ? (
-		<p id={ helpTextID } className="espresso-input-help-text" >
+		<p
+			id={ helpTextID }
+			className={ 'espresso-input-help-text ee-focus-priority-6' }
```
nicolad: `ee-focus-priority-6` - magic class name :)

https://media.giphy.com/media/12NUbkX6p4xOO4/giphy.gif

tn3rb: ... not sure whether you are implying that is good ro bad { : \

here's an explanation of what all is rolled into the "focus priority" classes:
https://brent.eventespresso.com/wp-content/plugins/event-espresso-core/assets/src/components/ui/styles/themes/default/demo/index.html#font-sizes

assets/src/components/form/layouts/two-column-admin/form-section.js
tn3rb: right I was probably getting a non-boolean falsey value...

I'll update this to use `!! formInfoNotice && `

assets/src/components/form/inputs/base/input-help-text.js
nicolad: Makes sense, made a search through files and stumbled across: `--ee-font-size-extreme`, `--ee-font-size-bigger` and got distracted :)

nicolad: Saw this:
```
@media screen and ( max-width: 782px ) {
 .
.
.
    .ee-focus-priority-6 {
        font-size: var(--ee-font-size-smaller);
    }
.
.
.
}

```
and didn't understood at first, but after your comment got it

PR #1829, opened 2019-11-08
remove mijireh payment method, modules, and test files

mnelson4:

Removes Mijireh payment method and modules.

## How has this been tested

* [ ] On master, make sure Mijireh is active. Preferably use a site that already has some Mijireh payments (if not, go into the DB, find the Mijireh payment method's ID, and update a payment's `PMD_ID` to be that same ID).
* Update to this branch
* Visit the payment method page. No warning should come up saying Mijireh was unexpectedly deactivated or anything
* Visit a transaction that had a payment made with Mijireh. It should still say it was made by Mijireh
* [ ] Apply a manual payment to a transaction. Mijireh should not be a new option

## Checklist

* [ ] My code is tested.


mnelson4:
Josh and I discussed in a Slack DM, and we're going to try using a DB query to change active Mijireh payment methods to have `PMD_type='Admin_Only'`, and `PMD_scope={}` so they admins won't get the error message because Mijireh's files suddenly disappeared, old payment made with it won't appear as "Unknown", but it won't be usable from the admin either. Kinda the best of everything, theoretically.

core/EE_Payment_Processor.core.php
```diff
@@ -208,7 +208,7 @@ public function get_ipn_url_for_payment_method($transaction, $payment_method)
      * @param boolean            $update_txn           whether or not to call
      *                                                 EE_Transaction_Processor::update_transaction_and_registrations_after_checkout_or_payment()
      * @param bool               $separate_IPN_request whether the IPN uses a separate request ( true like PayPal )
-     *                                                 or is processed manually ( false like Mijireh )
+     *                                                 or is processed manually ( false like Mijireh used to )
```
knazart: I think we might be able to use TransFirst or Aut.Net Accept here for the reference, although those don't go bundled with EE core.

mnelson4: yeah good point, although each of those probably doesn't have a super long shelf-life now either...

joshfeck:
I checked out this branch and when I loaded a transactions page a warning about Mijireh being deactivated
Screen Shot 2019-11-14 at 4 14 26 PM

Also, any transactions that have a payment via Mijireh now say unknown

Screen Shot 2019-11-14 at 4 15 09 PM

These are the same transactions shown with master branch activated:
Screen Shot 2019-11-14 at 4 16 15 PM

mnelson4:
Sorry Josh! In order for the upgrade logic to be triggered (that's what switches the Mijireh payment method's PMD_type and PMD_scope to "Admin_Only" and to not usable anywhere), you needed to deactivate and reactivate EE. But on this latest commit I bumped the version, so the code will realize there was an upgrade so that's no longer necessary.

So, to re-test you should start with a fresh EE site that hasn't previously used this branch, then switch to branch and THIS time it should just work.

joshfeck:
re-tested

PR #1866, opened 2019-11-18
EDTR: Added Apollo Client.

nicolad:
## Checklist

* [ ] My code is tested.


assets/prototype/infrastructure/TopLevelProviders/Apollo.js
```diff
@@ -0,0 +1,27 @@
+import { ApolloClient } from 'apollo-client';
+import { InMemoryCache } from 'apollo-cache-inmemory';
+import { HttpLink } from 'apollo-link-http';
+import { ApolloProvider } from '@apollo/react-hooks';
+import get from 'lodash/get';
+
+const { graphqlEndpoint } = window;
+const nonce = get( window, [ 'eejsdata', 'data', 'eejs_api_nonce' ] );
```
nicolad: It will return `undefined`, in case object's nested shape might get out-of-shape.

https://lodash.com/docs/4.17.15#get

assets/prototype/infrastructure/TopLevelProviders/Apollo.js
```diff
@@ -0,0 +1,27 @@
...
+
+const cache = new InMemoryCache();
+const client = new ApolloClient( {
+	cache,
+	link: new HttpLink( {
+		uri: graphqlEndpoint || '/graphql',
```
tn3rb: 👍

assets/prototype/infrastructure/TopLevelProviders/index.js
```diff
@@ -0,0 +1,17 @@
+/* Internal dependencies */
+import ApolloProvider from './Apollo';
```
nicolad: W might add here other providers as well.

nicolad:
> not sure about the `TopLevelProviders` folder name, but everything else looks great

I've followed the pattern from [here](https://github.com/Yoast/wordpress-seo/blob/trunk/js/src/components/TopLevelProviders.js#L22), in case we might need some theming context(for example) to pass down at some point in future.
Let me know I should refactor this.

tn3rb:
head cold is making my brain run at half speed so I want to amend my previous comments...

the logic where we combine provider contexts should go into the application folder since that is clearly specific to the app itself. 

So the logic you currently have in `assets/prototype/infrastructure/TopLevelProviders/index.js`
should go somewhere like: `assets/prototype/application/services/context_providers/TopLevelProvider.js`


and then what you have in `assets/prototype/infrastructure/TopLevelProviders/Apollo.js`
should go somewhere like:  `assets/prototype/infrastructure/services/apollo/Apollo.js`

Issue #1865, opened 2019-11-18
Build `DateCard` UI Component


tn3rb:
Receives Date object in props

Uses:
- `DeleteDateButton`

- dispatches:
    - update date
    - delete date

Issue #1864, opened 2019-11-18
Build `TicketCard` UI Component


tn3rb:
- Reuse components from `DateCard`
- Add following components reusing existing ones:
  - `DeleteTicketButton`

- dispatch actions for:
    - update ticket
    - delete ticket

Issue #1863, opened 2019-11-18
Build `DatesListData` Container


tn3rb:
- Use `useReducer` to implement Redux pattern

what do we need `useReducer` for ?  - bc

- queries Apollo for list of Dates
- returns array of date IDs

nicolad:
@tn3rb to dispatch actions, here is an example: https://codepen.io/kinsomicrote/pen/zQzmqy

not something set in stone, will see what makes sense and isn't adding obsolete complexity

manzoorwanijk:
We won't need `useReducer` for this. `useReducer` can be used to store the changes in individual entities before mutating the data.

Issue #1861, opened 2019-11-18
Append AddTicketButton to TicketList


tn3rb:
UI component uses:
 - `AddTicketButton`

Data container uses:
 - `useCreateTicket`

PR #1877, opened 2019-11-21
DatesList and TicketsList UI Components

tn3rb:
plz see:
- #1858
- #1859
- #1860 
- #1861 
- #1864 
- #1865

built on top of branch `EDTR/DatesListData-container` from #1863 

wanted to try out this component lib: https://blueprintjs.com/docs/#blueprint

so installed that and used a few of its pieces just to see what they were like, does not necessarily mean we will use it.


assets/prototype/domain/eventEditor/datesList/DateCard.js
```diff
@@ -0,0 +1,96 @@
...
+
+const console = window.console;
+
+const endTwoYears = new Date( new Date().getFullYear() + 2, 11, 31 );
+
+const btnStyle = {
+	margin: '0 0 0 .5rem',
+};
+
+const cardStyle = {
```
nicolad: Collocation FTW :sunglasses:

Issue #1878, opened 2019-11-21
Create globally unique IDs for non CPT GQL types


manzoorwanijk:
We need to have globally unique string IDs for types like Datetime, Ticket etc. for automatic UI update after mutations.

PR #1879, opened 2019-11-21
Create globally unique IDs for non CPT GQL types

manzoorwanijk:
Completes #1878 

## Checklist

* [x] My code is tested.


core/domain/services/graphql/connection_resolvers/AbstractConnectionResolver.php
```diff
@@ -3,6 +3,7 @@
 namespace EventEspresso\core\domain\services\graphql\connection_resolvers;
 
 use WPGraphQL\Data\Connection\AbstractConnectionResolver as WPGraphQLConnectionResolver;
+use GraphQLRelay\Relay;
```
tn3rb: whaaaaaaaaaaaat?

isn't that redundant because of Apollo? or was this added because it has the logic needed for UUIDs?

manzoorwanijk: That package is bundled with WPGraphQL and it uses that package to generate those IDs, it's a better option than UUIDs as the Relay package is managed by Facebook and works well with GraphQL.

core/domain/services/graphql/types/Datetime.php
```diff
@@ -49,6 +49,12 @@ public function getFields()
         return [
             new GraphQLField(
                 'id',
+                ['non_null' => 'ID'],
+                null,
+                esc_html__('The globally unique ID for the object.', 'event_espresso')
+            ),
+            new GraphQLOutputField(
+                lcfirst($this->name()) . 'Id',
```
tn3rb: why not just hardcode `datetimeId` ?

manzoorwanijk: As mentioned in [this card](https://github.com/eventespresso/event-espresso-core/projects/6#card-29371659), we can change the type name. So, I set it dynamically so as not to miss those when changing the type names.

PR #1897, opened 2019-11-25
Ticket delete mutation + cache update.

nicolad:
## How has this been tested
Peek 2019-11-25 21-51

## Checklist

* [ ] My code is tested.

### DIFF
--- a/assets/prototype/domain/eventEditor/containers/mutations/useDeleteTicketMutation.js
+++ b/assets/prototype/domain/eventEditor/containers/mutations/useDeleteTicketMutation.js
@@ -1,9 +1,45 @@
+import get from 'lodash/get';
 import { useMutation } from '@apollo/react-hooks';
 import { DELETE_TICKET } from './tickets';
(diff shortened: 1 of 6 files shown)


assets/prototype/domain/eventEditor/containers/mutations/useDeleteTicketMutation.js
```diff
@@ -1,9 +1,45 @@
+import get from 'lodash/get';
 import { useMutation } from '@apollo/react-hooks';
 import { DELETE_TICKET } from './tickets';
+import { GET_TICKETS } from '../queries/tickets';
 
-const useDeleteTicketMutation = () => {
+const useDeleteTicketMutation = ({ datetimeIn, id }) => {
```
manzoorwanijk: It's better to generate `datetimeIn` right there in the hook instead of being passed down to it all the way from `TicketList`. It should only get the `id` and do the rest itself. What we could do later is to create another hook like `useDatetimeIds` to fetch the value to be used for `datetimeIn`.

PR #1899, opened 2019-11-26
Toasty Toaster Toast

tn3rb:
mmmmm toast...

adds toaster notifications during editor initialization and for feedback during mutations


assets/prototype/infrastructure/services/toaster/useLoadingToast.js
```diff
@@ -0,0 +1,64 @@
+import { Spinner } from '@blueprintjs/core';
+import { useCallback } from '@wordpress/element';
+import useDismissToast from './useDismissToast';
+
+const useLoadingToast = (toaster, hash) => {
+	const dismiss = useDismissToast(toaster, hash);
```
nicolad: @tn3rb ESLint was complaining:
image
 so, I moved that in outer scope and passed `dismiss` as a dependency of `useCallback`

nicolad:
Very nice abstraction :1st_place_medal:

assets/prototype/infrastructure/services/toaster/useLoadingToast.js
tn3rb: oh right... I know that too, but just forgot while doing that obviously, thnx

PR #1900, opened 2019-11-26
EDTR: DateRangeInput.

nicolad:
- Added description update.
- Added `DateRangeInput`.

## Checklist

* [ ] My code is tested.

### DIFF
--- a/assets/prototype/domain/eventEditor/datesList/dateCard/DateCard.js
+++ b/assets/prototype/domain/eventEditor/datesList/dateCard/DateCard.js
@@ -1,6 +1,7 @@
-import { Button, Card, EditableText, Elevation, H4, Popover } from '@blueprintjs/core/lib/esm';
+import { Button, Card, EditableText, Elevation, H4, H6, Popover } from '@blueprintjs/core/lib/esm';
 import { DatePicker, TimePrecision } from '@blueprintjs/datetime/lib/esm';
 import DeleteDateButton from './DeleteDateButton';
+import DateRangeInput from '../../../shared/dateRangeInput/DateRangeInput';
 import useDateItem from '../../containers/queries/useDateItem';
 import useUpdateDateMutation from '../../containers/mutations/useUpdateDateMutation';
 import { A_LONG_TIME_AGO, PLUS_ONE_MONTH, PLUS_TEN_YEARS } from '../../../shared/defaultDates';
@@ -27,13 +28,7 @@ const idStyle = {
 
 const DateCard = ({ eventId, id }) => {
 	const date = useDateItem({ id });
-	const { updateDate } = useUpdateDateMutation();
-
-	const onTitleConfirm = (name) => {
-		updateDate({
-			variables: { input: { clientMutationId: 'xyz', id, name } },
-		});
-	};
+	const onFieldUpdate = useUpdateDateMutation({ id });
 
 	const startDate = date.start ? new Date(date.start * 1000) : PLUS_ONE_MONTH;
 	// const endDate = date.end ?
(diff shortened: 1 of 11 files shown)


manzoorwanijk:
GQL server side is prepared to handle only the fields that were supposed to be worked upon in prototype viz. `name`, `description`, `startDate` and `endDate` for datetimes and likewise for tickets.

assets/prototype/domain/eventEditor/datesList/dateCard/DateCard.js
```diff
@@ -38,19 +39,30 @@ const idStyle = {
...
+	// 			onCancel={(value) => console.log(value)}
+	// 			onConfirm={(capacity) =>
+	// 				onFieldUpdate({ capacity: Number(capacity) })
+	// 			}
+	// 			minWidth={'320px'}
+	// 			selectAllOnFocus
+	// 		/>
+	// 	</div>
+	// );
+
```
nicolad: @manzoorwanijk as per your reminder, I've commented this out and will add back after proto phase is done.

nicolad: The other logic related to capacity field is not introducing any breaking changes.

assets/prototype/domain/shared/dateRangeInput/DateRangeInput.js
```diff
@@ -0,0 +1,63 @@
...
+import isFunction from 'lodash/isFunction';
+import moment from 'moment';
+import {
+	DateRangeInput as DefaultDateRangeInput,
+	TimePrecision
+} from '@blueprintjs/datetime';
+import { FORMATS } from './formatSelect';
+import { MomentDateRange } from './momentDate';
+
+import './style.css';
```
nicolad: @tn3rb did you tried to override some `blueprint`'s styles?
I tried here, but I am missing something.

assets/prototype/domain/shared/dateRangeInput/DateRangeInput.js
```diff
@@ -0,0 +1,53 @@
...
+	const format = FORMATS[0];
+	const defaultRangeValues = [startDate || null, endDate || null];
+	const [range, setRange] = useState(defaultRangeValues);
+
+	const handleRangeChange = (value) => {
+		const [start, end] = value;
+		const isUpdatable = isFunction(onFieldUpdate) && start && end;
+
+		if (isUpdatable) {
+			const format = 'YYYY-MM-DD HH:mm:ss';
```
tn3rb: would make a nice global constant...

I'm thinking we should add some kind of ConfigProvider to the top level app contexts,
which could include things like the various date and time formats we need, as well as timezone information, currency, etc

but that's for another PR :)

assets/prototype/domain/shared/dateRangeInput/formatSelect.js
```diff
@@ -0,0 +1,21 @@
...
+	momentFormatter('YYYY-MM-DD HH:mm:ss'),
+	{
+		formatDate: (date) => (date == null ? '' : date.toLocaleDateString()),
+		parseDate: (str) => new Date(Date.parse(str)),
+		placeholder: 'JS Date',
+	},
+	momentFormatter('MM/DD/YYYY'),
+	momentFormatter('YYYY-MM-DD'),
+	momentFormatter('YYYY-MM-DD HH:mm:ss'),
+];
```
tn3rb: IMHO this should be an object with named formats (ie: keys are string constants)

because seeing something like `const format = FORMATS[FORMAT.MYSQL];`

will be more clear than seeing `const format = FORMATS[0];`

assets/prototype/domain/shared/dateRangeInput/momentDate.js
```diff
@@ -0,0 +1,29 @@
+import { Icon, Intent, Tag } from '@blueprintjs/core/lib/esm';
+import classNames from 'classnames';
+import moment from 'moment';
+import * as React from 'react';
+
+const FORMAT = 'dddd, LL';
+const FORMAT_TIME = 'dddd, LL LT';
```
tn3rb: ya this should be globalized somewhere in a single file/variable else there's going to be an explosion of these constants all over the place

nicolad: small note here, credits go to the original example: https://github.com/palantir/blueprint/blob/develop/packages/docs-app/src/examples/datetime-examples/common/momentDate.tsx

tn3rb:
plz see:

- https://github.com/eventespresso/event-espresso-core/projects/6#card-29833602
- https://github.com/eventespresso/event-espresso-core/projects/6#card-29833585

assets/prototype/domain/shared/dateRangeInput/DateRangeInput.js
nicolad: Sure, great suggestion.
This variable is a temp solution for prototype.
But for the long-term I've borrowed `formatSelect.js` from blueprint and will centralize that into a context provider as you said. I wasn't sure about the context provider at first, but later I recalled that we might need to pull some user specific information from WP.
Great suggestion and thanks for tracking this into that project card.

assets/prototype/domain/shared/dateRangeInput/formatSelect.js
nicolad: Thx, updated: https://github.com/eventespresso/event-espresso-core/pull/1900/commits/3d4e154dce62d07bd04d8e40743ee668419af770

nicolad:
> plz see:
> 
> * https://github.com/eventespresso/event-espresso-core/projects/6#card-29833602
> * https://github.com/eventespresso/event-espresso-core/projects/6#card-29833585

Thanks for tracking these down, otherwise I can forget about these :)

PR #1908, opened 2019-11-27
Relations Selector

tn3rb:
Replaces check boxes in the New Ticket Form with a new multi-select component that uses way less screen real estate and also incorporates search functionality


assets/prototype/domain/shared/FormModalForm.js
```diff
@@ -0,0 +1,75 @@
...
+
+const dataStyle = {
+	borderRadius: '5px',
+	boxSizing: 'border-box',
+	padding: '1em 2em',
+	color: '#a9ce47',
+	backgroundColor: '#26203d',
+};
+
+const FormModalForm = ({ FormComponent, form, values, handleSubmit, submitting, pristine, onClose, ...formProps }) => {
```
nicolad: Why not just `ModalForm` ?

tn3rb: 🤷‍♂

PR #1909, opened 2019-11-27
Prepare prototype for relations

manzoorwanijk:
This PR exposes the ticket prices to GraphQL and handles the relations between them.

## Checklist

* [ ] My code is tested.

### DIFF
--- a/core/domain/services/graphql/connection_resolvers/PriceConnectionResolver.php
+++ b/core/domain/services/graphql/connection_resolvers/PriceConnectionResolver.php
@@ -1,0 +1,160 @@
+<?php
+
+namespace EventEspresso\core\domain\services\graphql\connection_resolvers;
+
+use EE_Ticket;
+use EE_Error;
+use EEM_Price;
+use EventEspresso\core\exceptions\InvalidDataTypeException;
+use EventEspresso\core\exceptions\InvalidInterfaceException;
+use InvalidArgumentException;
+use ReflectionException;
+
+/**
+ * Class PriceConnectionResolver
+ */
+class PriceConnectionResolver extends AbstractConnectionResolver
+{
+
+    /**
+     * @return EEM_Price
+     * @throws EE_Error
+     * @throws InvalidArgumentException
+     * @throws InvalidDataTypeException
+     * @throws InvalidInterfaceException
+     */
+    // phpcs:ignore PSR1.Methods.CamelCapsMethodName.NotCamelCaps
+    public function get_query()
+    {
+        return EEM_Price::instance();
+    }
+
+
+    /**
+     * Return an array of items from the query
+     *
+     * @return array
+     */
+    // phpcs:ignore PSR1.Methods.CamelCapsMethodName.NotCamelCaps
+    public function get_items()
+    {
+        $results = $this->query->get_col($this->query_args);
+
+        return ! empty($results) ? $results : [];
+    }
+
+
+    /**
+     * Determine whether the Query should execute. If it's determined that the query should
+     * not be run based on context such as, but not limited to, who the user is, where in the
+     * ResolveTree the Query is, the relation to the node the Query is connected to, etc
+     * Return false to prevent the query from executing.
+     *
+     * @return bool
+     */
+    // phpcs:ignore PSR1.Methods.CamelCapsMethodName.NotCamelCaps
+    public function should_execute()
+    {
+        if (false === $this->should_execute) {
+            return false;
+        }
+
+        return $this->should_execute;
+    }
+
+
+    /**
+     * Here, we map the args from the input, then we make sure that we're only querying
+     * for IDs. The IDs are then passed down the resolve tree, and deferred resolvers
+     * handle batch resolution of the posts.
+     *
+     * @return array
+     * @throws EE_Error
+     * @throws InvalidArgumentException
+     * @throws ReflectionException
+     * @throws InvalidDataTypeException
+     * @throws InvalidInterfaceException
+     */
+    // phpcs:ignore PSR1.Methods.CamelCapsMethodName.NotCamelCaps
+    public function get_query_args()
+    {
+        $where_params = [];
+        $query_args   = [];
+
+        $query_args['limit'] = $this->getLimit();
+
+        // Avoid multiple entries by join.
+        $query_args['group_by'] = 'PRC_ID';
+
+        /**
+         * Collect the input_fields and sanitize them to prepare them for sending to the Query
+         */
+        $input_fields = [];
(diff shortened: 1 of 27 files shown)


docs/GraphQL-API/mutations/price.md
```diff
@@ -0,0 +1,93 @@
...
+  createPrice(input: $input) {
+    price {
+      id
+      name
+    }
+  }
+}
+```
+
+### Query variables for `createPrice`
```
nicolad: Maybe mutation variables ?

manzoorwanijk: They are still called query variables but yes we can name them as mutation variables.

core/domain/services/graphql/data/mutations/TicketMutation.php
```diff
@@ -67,4 +71,27 @@ public static function setRelatedDatetimes($entity, array $datetimes)
...
+     * Sets the related prices for the given ticket.
+     *
+     * @param EE_Ticket $entity The Ticket instance.
+     * @param array     $prices Array of entity IDs to relate.
+     */
+    public static function setRelatedPrices($entity, array $prices)
+    {
+        $relationName = 'Price';
+        // Remove all the existing related entities
+        $entity->_remove_relations($relationName);
```
tn3rb: this does not appear to be a very efficient way to go about doing this, since it would potentially have to run multiple queries to do so (I'm guessing the underlying deletions running in `EEM_Base` are happening in a loop) so it might be far more efficient to query for existing relations, then find the difference between the incoming price entities, and then only add and delete records for the actual changes and leave everything else alone.

that said, let's not worry about that now, but can you plz add a `todo` notice here for making that change?

core/domain/services/graphql/mutators/PriceCreate.php
```diff
@@ -0,0 +1,79 @@
...
+         */
+        return static function ($input, AppContext $context, ResolveInfo $info) use ($model, $type) {
+
+            /**
+             * Stop now if a user isn't allowed to create an entity.
+             */
+            if (! current_user_can('ee_edit_events')) {
+                // translators: the %1$s is the name of the object being mutated
+                throw new UserError(
+                    sprintf(__('Sorry, you are not allowed to create %1$s', 'event_espresso'), $type->name())
```
tn3rb: we should be using `esc_html__()` everywhere instead of `__()`

core/domain/services/graphql/connection_resolvers/PriceConnectionResolver.php
```diff
@@ -0,0 +1,160 @@
...
+
+    /**
+     * @return EEM_Price
+     * @throws EE_Error
+     * @throws InvalidArgumentException
+     * @throws InvalidDataTypeException
+     * @throws InvalidInterfaceException
+     */
+    // phpcs:ignore PSR1.Methods.CamelCapsMethodName.NotCamelCaps
+    public function get_query()
```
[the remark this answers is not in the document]
manzoorwanijk: We can't do that because that method is expected by WP GraphQL in `WPGraphQL\Data\Connection\AbstractConnectionResolver`. All the connection resolvers extend that class.

core/domain/services/graphql/connection_resolvers/PriceConnectionResolver.php
```diff
@@ -0,0 +1,160 @@
...
+     * not be run based on context such as, but not limited to, who the user is, where in the
+     * ResolveTree the Query is, the relation to the node the Query is connected to, etc
+     * Return false to prevent the query from executing.
+     *
+     * @return bool
+     */
+    // phpcs:ignore PSR1.Methods.CamelCapsMethodName.NotCamelCaps
+    public function should_execute()
+    {
+        if (false === $this->should_execute) {
```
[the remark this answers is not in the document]
manzoorwanijk: That's only for internal use just in case we need to check for user caps for example.

core/domain/services/graphql/connection_resolvers/PriceConnectionResolver.php
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+
+    /**
+     * Return an array of items from the query
+     *
+     * @return array
+     */
+    // phpcs:ignore PSR1.Methods.CamelCapsMethodName.NotCamelCaps
+    public function get_items()
```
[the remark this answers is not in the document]
manzoorwanijk: yes, same as above

core/domain/services/graphql/connection_resolvers/PriceTypeConnectionResolver.php
```diff
@@ -0,0 +1,93 @@
...
+     * not be run based on context such as, but not limited to, who the user is, where in the
+     * ResolveTree the Query is, the relation to the node the Query is connected to, etc
+     * Return false to prevent the query from executing.
+     *
+     * @return bool
+     */
+    // phpcs:ignore PSR1.Methods.CamelCapsMethodName.NotCamelCaps
+    public function should_execute()
+    {
+        if (false === $this->should_execute) {
```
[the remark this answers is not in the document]
manzoorwanijk: 🤦‍♂

core/domain/services/graphql/data/mutations/TicketMutation.php
manzoorwanijk: Sure, I did look for a single query solutions but the model system does not seem to have one.

core/domain/services/graphql/mutators/PriceCreate.php
manzoorwanijk: Oops!

core/domain/services/graphql/connection_resolvers/PriceConnectionResolver.php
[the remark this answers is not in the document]
tn3rb: oh oops... sry, just realized what I wrote above is a bit wonky...

here's a better commnet:

similar to how `EventEspresso\core\domain\services\graphql\abstracts\AbstractRootQueryConnection` has a $model property and getter for it, you could do the same in `EventEspresso\core\domain\services\graphql\connection_resolvers\AbstractConnectionResolver` (which extends `WPGraphQL\Data\Connection\AbstractConnectionResolver` and then add:

```php
public function get_query()
{
	return $this->model;
}
```

to that class and then all of its children would have this method covered

core/domain/services/graphql/data/mutations/TicketMutation.php
tn3rb: right... can you plz add a todo note to this though?

PR #1910, opened 2019-11-27
EDTR: Ticket: added description field and fixed update mutation.

nicolad:
## Checklist

* [ ] My code is tested.

### DIFF
--- a/assets/prototype/domain/eventEditor/containers/mutations/useUpdateTicketMutation.js
+++ b/assets/prototype/domain/eventEditor/containers/mutations/useUpdateTicketMutation.js
@@ -2,9 +2,8 @@ import { useMutation } from '@apollo/react-hooks';
 import { UPDATE_TICKET } from './tickets';
 import useToaster from '../../../../infrastructure/services/toaster/useToaster';
 
-const useUpdateTicketMutation = () => {
+const useUpdateTicketMutation = ({ id = 0 }) => {
 	const toaster = useToaster();
-	const id = 0;
 	const toasterMessage = `updating ticket ${id}`;
 	const [updateTicket, { loading, error }] = useMutation(UPDATE_TICKET, {
 		onCompleted: () => {
@@ -20,7 +19,29 @@ const useUpdateTicketMutation = () => {
 	toaster.loading(loading, toasterMessage);
 	toaster.error(error);
 
-	return { updateTicket, loading, error };
+	const updateHandler = ({ description, name, price }) => {
+		if (description) {
+			return updateTicket({
+				variables: { input: { clientMutationId: 'xyz', id, description } },
+			});
+		}
+
+		if (name) {
+			return updateTicket({
+				variables: { input: { clientMutationId: 'xyz', id, name } },
+			});
+		}
+
+		if (price) {
+			return updateTicket({
+				variables: { input: { clientMutationId: 'xyz', id, price } },
+			});
+		}
+
+		return null;
+	};
+
+	return updateHandler;
 };
 
 export default useUpdateTicketMutation;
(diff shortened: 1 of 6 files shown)


assets/prototype/domain/eventEditor/containers/mutations/useUpdateTicketMutation.js
```diff
@@ -2,9 +2,8 @@ import { useMutation } from '@apollo/react-hooks';
 import { UPDATE_TICKET } from './tickets';
 import useToaster from '../../../../infrastructure/services/toaster/useToaster';
 
-const useUpdateTicketMutation = () => {
+const useUpdateTicketMutation = ({ id = 0 }) => {
 	const toaster = useToaster();
-	const id = 0;
```
tn3rb: should we do this same thing for `assets/prototype/domain/eventEditor/containers/mutations/useCreateTicketMutation.js` ?

assets/prototype/domain/eventEditor/containers/mutations/useUpdateTicketMutation.js
```diff
@@ -20,7 +19,29 @@ const useUpdateTicketMutation = () => {
...
+			return updateTicket({
+				variables: { input: { clientMutationId: 'xyz', id, name } },
+			});
+		}
+
+		if (price) {
+			return updateTicket({
+				variables: { input: { clientMutationId: 'xyz', id, price } },
+			});
+		}
```
tn3rb: what about something **like** :

```js
let field = '';
let fieldName = ''
// set description 
field = description ? description : field;
fieldName = description ? 'description' : fieldName ;
// set name
field = name? name: field;
fieldName = name? 'name' : fieldName ;
// set price
field = price? price: field;
fieldName = price? 'price' : fieldName ;
if (field) {
    const data = { [fieldName]: field };
    return updateTicket({
        variables: { input: { clientMutationId: 'xyz', id, ...data } },
    });
}
```

just a suggestion,
but this only requires an extra two lines of code to add a new field, which we will eventually be doing.
As well, this logic could likely be extracted into a helper function so that you just feed in all of the incoming field props, and any that have values will get set in the returned data object.
Again, just a suggestion (ie: not a requirement) that could make this logic easier to manage as we add more things

assets/prototype/domain/eventEditor/ticketsList/AddNewTicketModal.js
```diff
@@ -1,82 +1,15 @@
...
-			});
-		};
-
-		createTicket({
-			variables,
-			optimisticResponse,
-			update,
-		});
-	};
+	const createTicket = useCreateTicketMutation({ datetimes });
```
tn3rb: 👍

assets/prototype/domain/eventEditor/containers/mutations/useUpdateTicketMutation.js
nicolad: Ticket id is generated on the server-side, so on `useCreateTicketMutation` init we don't have that available yet, only after the mutation has fired and we received a response from the server that id is pushed into Apollo cache.

assets/prototype/domain/eventEditor/containers/mutations/useUpdateTicketMutation.js
nicolad: My initial implementation started with 2 fields and I thought that was acceptable, but it's not working well with more than 2 fields.
I liked this suggestion:
```
As well, this logic could likely be extracted into a helper function so that you just feed in all of the incoming field props, and any that have values will get set in the returned data object.
```

nicolad: I've added conditionally keys to that `variables` object as described here: https://medium.com/@mikeh91/conditionally-adding-keys-to-javascript-objects-using-spread-operators-and-short-circuit-evaluation-acf157488ede
***
but this will needs to be simplified somehow with ramda or lodash to enhance readability

assets/prototype/domain/eventEditor/containers/mutations/useUpdateTicketMutation.js
tn3rb: oh duh.. right of course!

Issue #1903, opened 2019-11-26
REST API Edit relation between Event and Person seems broken


joshfeck:
Following the documentation here:

https://github.com/eventespresso/event-espresso-core/blob/master/docs/C--REST-API/ee4-rest-api-editing-relations.md

Someone was trying to update the relationship between an event and an EE Person CPT (added by the People add-on). You can send an auth'd POST request like this:

`wp-json/ee/v4.8.36/events/521/people/606/`

or even like this:

`wp-json/ee/v4.8.36/people/606/events/521`

The JSON response will be:
```
"code": "error_occurred",
"message": "There is no model named &#039;EVT_ID&#039; related to EEM_Person_Post. Query param type is 0 and original query param is EVT_ID",
"data": null
```

Despite that error, there is a new row that gets added to the `_esp_people_to_post` table. Also, if you try to send additional params to specify values for `P2P_Order` or `PT_ID`, a 0 gets written to those columns in the db.

Can you look into this Mike?

mnelson4:
Thanks Josh, I see the issue: `\EventEspresso\core\libraries\rest_api\controllers\model\Write::addRelation` adds the relation then tries to return the new join-table row. It assumes the foreign key columns on the join table are the same as the primary keys on the joined models. That's correct in core, but not in `\EEM_Person_Post`, whose foreign key to the event model is `OBJ_ID`

PR #1911, opened 2019-11-27
Fix error adding relation between people and events

mnelson4:
…eign keys to them on the join model

Fixes an error in the REST API when adding relations between people and events. See https://github.com/eventespresso/event-espresso-core/issues/1903
The problem was the code assumed that join table's foreign keys's names matched the primary keys on the joined models, which is usually right, but not for people-event relations.

## How has this been tested

* [ ] Send a request like `POST wp-json/ee/v4.8.36/events/{event-id}/people/{person-id}`. It shouldn't return an error

## Checklist

* [x] My code is tested.

### DIFF
--- a/core/libraries/rest_api/controllers/model/Write.php
+++ b/core/libraries/rest_api/controllers/model/Write.php
@@ -442,8 +442,8 @@ class Write extends Base
             $join_model_obj = $relation->get_join_model()->get_one(
                 array(
                     array(
-                        $model->primary_key_name() => $model_obj->ID(),
-                        $relation->get_other_model()->primary_key_name() => $related_obj->ID()
+                        $relation->get_join_model()->get_foreign_key_to($model->get_this_model_name())->get_name() => $model_obj->ID(),
+                        $relation->get_join_model()->get_foreign_key_to($relation->get_other_model()->get_this_model_name())->get_name() => $related_obj->ID()
                     )
                 )
             );


mnelson4:
Ya it's quite the epic method chain isn't it? Ah well, I think its ok.

joshfeck:
I tested this and it does fix the error. 

One question though, is there a way to also set the other values when adding a relation? For example if I'm adding a Person to Event relation, can I also specify the `PER_OBJ_order` and `PT_ID`?

mnelson4:
> For example if I'm adding a Person to Event relation, can I also specify the PER_OBJ_order and PT_ID?

Ya you should be able to pass those exact keys in the request body just like a POST.

Anyways, merging...

joshfeck:
When I pass those exact keys in the request body, the corresponding values don't get set in the database.

mnelson4:
> When I pass those exact keys in the request body, the corresponding values don't get set in the database.
Lemme see... that sounds like a bug. If so, I'll make a new ticket for it.

mnelson4:
Hey Josh, I reproduced the issue when [I used "params" from postman, which puts the parameters into the querystring](https://drive.google.com/file/d/1T1oggYwmBkB2K8dch-WoPplShAPBCMZ0/view), but the code expects BODY params. 
So if you instead [open up the "body" tab, and set it to "form-data" and add the parameters there,](https://drive.google.com/file/d/1crx4qrK5YDPNNXZ_0yuDRqI963XTNf4o/view) it worked for me.
Reviewing [our documentation on adding relations via the REST API](https://github.com/eventespresso/event-espresso-core/blob/master/docs/C--REST-API/ee4-rest-api-editing-relations.md), I don't think how to send these extra parameter is mentioned at all, much less that they need to be in the body (not the querystring). 
So what to do? 
1) Update documentation 
2) Accept either querystring parameters OR body parameters,
3) Throw an error if querystring parameters are used.

Thoughts?

joshfeck:
Updating the documentation there should be sufficient.

mnelson4:
OK thanks; I made a pull request to document that here: https://github.com/eventespresso/event-espresso-core/pull/1931

Issue #1901, opened 2019-11-26
Create useDatetimeIds hook.


nicolad:
More info here: https://github.com/eventespresso/event-espresso-core/pull/1897#discussion_r350642515

tn3rb:
As standard practice, let's include the names of both entities involved in the relations in the hook name.

So this hook would be named `useDatetimeIdsForTicket` since it would be retrieving the IDs for the datetimes related to the provided ticket.

Similarly, we might end up with hooks like:

- `useDatetimeIdsForEvent`
- `useTicketIdsForDatetime`
- `usePriceIdsForTicket`

(but these will need to be done via additional issues/PRs)

I've estimated 8 hours for this (plz give feedback if that does not seem to be enough)

PR #1954, opened 2019-12-03
Created useDatetimeIds hook.

nicolad:
Implements: https://github.com/eventespresso/event-espresso-core/issues/1901
Additional info: https://github.com/eventespresso/event-espresso-core/pull/1897#discussion_r350642515

## Checklist

* [ ] My code is tested.

### DIFF
--- a/assets/prototype/domain/eventEditor/containers/mutations/useDeleteTicketMutation.js
+++ b/assets/prototype/domain/eventEditor/containers/mutations/useDeleteTicketMutation.js
@@ -1,4 +1,4 @@
-import get from 'lodash/get';
+import { path } from 'ramda';
 import { useMutation } from '@apollo/react-hooks';
 import { DELETE_TICKET } from './tickets';
 import { GET_TICKETS } from '../queries/tickets';
@@ -28,7 +28,7 @@ const useDeleteTicketMutation = ({ id }) => {
 	const variables = { input: { clientMutationId: 'xyz', id } };
 
 	const update = (proxy, { data }) => {
-		const ticket = get(data, ['deleteTicket', 'ticket']);
+		const ticket = path(['deleteTicket', 'ticket'], data);
 
 		const options = {
 			query: GET_TICKETS,
--- a/assets/prototype/domain/eventEditor/containers/mutations/useDeleteTicketMutation.js
+++ b/assets/prototype/domain/eventEditor/containers/mutations/useDeleteTicketMutation.js
@@ -2,10 +2,12 @@ import get from 'lodash/get';
 import { useMutation } from '@apollo/react-hooks';
 import { DELETE_TICKET } from './tickets';
 import { GET_TICKETS } from '../queries/tickets';
+import useDatetimeIds from '../queries/useDatetimeIds';
 import useToaster from '../../../../infrastructure/services/toaster/useToaster';
 import useRelations from '../../../../infrastructure/services/relations/useRelations';
 
-const useDeleteTicketMutation = ({ datetimeIn, id }) => {
+const useDeleteTicketMutation = ({ id }) => {
+	const datetimeIn = useDatetimeIds();
 	const toaster = useToaster();
 	const { removeRelation, dropRelations } = useRelations();
 	const toasterMessage = `deleting ticket ${id}`;
--- a/assets/prototype/domain/eventEditor/containers/mutations/useDeleteTicketMutation.js
+++ b/assets/prototype/domain/eventEditor/containers/mutations/useDeleteTicketMutation.js
@@ -1,4 +1,4 @@
-import { path } from 'ramda';
+import path from 'ramda/src/path';
 import { useMutation } from '@apollo/react-hooks';
 import { DELETE_TICKET } from './tickets';
 import { GET_TICKETS } from '../queries/tickets';
--- a/assets/prototype/domain/eventEditor/containers/queries/useDatetimeIds.js
+++ b/assets/prototype/domain/eventEditor/containers/queries/useDatetimeIds.js
@@ -20,10 +20,7 @@ const useDatetimeIds = () => {
 		data = {};
 	}
 
-	const getDatetimes = (data) => R.pathOr([], ['datetimes', 'nodes'], data);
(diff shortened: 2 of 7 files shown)


assets/prototype/domain/eventEditor/containers/queries/useDatetimeIds.js
```diff
@@ -0,0 +1,30 @@
+import * as R from 'ramda';
+import { useApolloClient } from '@apollo/react-hooks';
+import { GET_DATETIMES } from './dates';
+
+const useDatetimeIds = () => {
+	const eventId = R.pathOr(null, ['eeEditorEventData', 'eventId'], window);
```
nicolad: I am not :100: happy with this.
Looking/ open for a better approach.

assets/prototype/domain/eventEditor/containers/queries/useDatetimeIds.js
```diff
@@ -0,0 +1,30 @@
+import * as R from 'ramda';
+import { useApolloClient } from '@apollo/react-hooks';
+import { GET_DATETIMES } from './dates';
+
+const useDatetimeIds = () => {
```
nicolad: @tn3rb you wrote that it should be named `useDatetimeIdsForTicket`, but as of what `useDeleteTicketMutation` is expecting this hook is not aware of any ticket information.

assets/prototype/domain/eventEditor/containers/queries/useDatetimeIds.js
```diff
@@ -0,0 +1,30 @@
...
+				},
+			},
+		});
+	} catch (error) {
+		throw new Error(error);
+	}
+
+	const getDatetimes = R.pathOr([], ['datetimes', 'nodes']);
+	const getIds = R.map(({ id }) => id);
+	const datetimeIds = R.pipe(getDatetimes, getIds)(data);
```
nicolad: @tn3rb does this piece make sense?
Is it grokkable?

assets/prototype/domain/eventEditor/containers/queries/useDatetimeIds.js
tn3rb: ok great, that's even better then since there's less coupling

assets/prototype/domain/eventEditor/containers/mutations/useDeleteTicketMutation.js
```diff
@@ -1,11 +1,13 @@
-import get from 'lodash/get';
+import { path } from 'ramda';
```
tn3rb: if you only need the one function you can import directly too:

```suggestion
import path from 'ramda/src/path';
```

assets/prototype/domain/eventEditor/containers/queries/useDatetimeIds.js
tn3rb: GitHub (or Chrome) is being janky and refuses to update the code shown here, but what you displayed in chat looked 💯 👌 👍 🏅

assets/prototype/domain/eventEditor/containers/queries/useDatetimeIds.js
nicolad: Yeah, great suggestion.
TBD here: https://github.com/eventespresso/event-espresso-core/issues/1955

assets/prototype/domain/eventEditor/containers/queries/useDatetimeIds.js
```diff
@@ -0,0 +1,30 @@
+import * as R from 'ramda';
```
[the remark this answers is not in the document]
nicolad: As noted in ramda's docs:
`Manually cherry picking methods is cumbersome`
so I've added tree-shaking support here: https://github.com/eventespresso/event-espresso-core/pull/1954/commits/7dd61099ba85066ec354586c79311a2df44b633c
As described in docs:
image

assets/prototype/domain/eventEditor/containers/queries/useDatetimeIds.js
```diff
@@ -0,0 +1,31 @@
...
+		data = client.readQuery({
+			query: GET_DATETIMES,
+			variables: {
+				where: {
+					eventId,
+				},
+			},
+		});
+	} catch (error) {
+		throw new Error(error);
```
manzoorwanijk: I don't think throwing an error is enough here. It's possible that there may not be any datetimes in cache, so `readQuery` will throw an error, in that case we should return an empty array.

assets/prototype/domain/eventEditor/containers/queries/useDatetimeIds.js
```diff
@@ -0,0 +1,31 @@
...
+	}
+
+	const getDatetimes = (data) => R.pathOr([], ['datetimes', 'nodes'], data);
+	const getIds = (datetimes) => R.map(({ id }) => id, datetimes);
+	const getDatetimeIds = R.pipe(getDatetimes, getIds);
+	const datetimeIds = getDatetimeIds(data);
+
+	return datetimeIds;
```
manzoorwanijk: Why not simply this?
```js
return R.pathOr([], ['datetimes', 'nodes'], data).map(({ id }) => id);
```
Also, `data` in `getDatetimes()` is already defined in the upper scope.

nicolad: That is a great question.
When error will be thrown we will have `data = {}`
and in this case the `.map(({ id }) => id)` will throw an error saying that there is no `id` property, because it will try to map over an empty array.
This is why I've updated this:
`R.map((datetime) => R.prop('id', datetime), datetimes)`
in this case when we will have empty array `R.prop('id', datetime)` will result in undefined, not with an unexpecte error.

manzoorwanijk: Even if `data` is `{}`, `R.pathOr([], ['datetimes', 'nodes'], data)` should take care of that and return an empty array.

manzoorwanijk: image

PR #1968, opened 2019-12-06
EDTR: Ticket Price Calculator Logic

tn3rb:
This is still a big pile of mess that needs cleaning up, but the underlying logic is working for the most part.

to do:

- [x] fetch ticket from ACC (Apollo Client Cache) either directly or via Hook and/or Provider
- [x] fetch prices for ticket from ACC
- [x] fetch all price types from ACC
- [x] figure out how to sort form `FieldArray` rows after calculations are made
- [x] create onSubmit handler for processing form submission and triggering mutations accordingly
- [ ] extract all utility functions into appropriate locations
- [ ] add types and interfaces 
- [ ] additional code cleanup and refactoring to make code SOLID


assets/prototype/domain/eventEditor/ticketsList/ticketPriceCalculator/TicketPriceCalculatorForm.js
```diff
@@ -1,201 +1,66 @@
...
-			</tr>
-		);
-	};
-	const renderPriceRowWithDelete = (price) => renderPriceRow(price, 'trash');
+// just temporary
+import styles from './inlineStyles';
 
+const TicketPriceCalculatorForm = ({ form, values: { ticket, prices } }) => {
+	const calcDir = !!ticket.reverseCalculate;
+	const toggleCalcDir = () => form.mutators.toggleCalcDir('ticket.reverseCalculate');
```
nicolad: This might be ok for proto, but I would suggest that we'd rather make this less nested or make another approach.
Also, we can create an issue just for that and take care later.

assets/prototype/domain/eventEditor/ticketsList/ticketPriceCalculator/TicketPriceCalculatorModal.js
```diff
@@ -1,80 +1,141 @@
...
+	'id',
+	'isBasePrice',
+	'isDiscount',
+	'isPercent',
+	'name',
+	'order',
+	'priceType',
+];
+const NO_INDEX = -1;
+const isPriceFormField = (val, key) => indexOf(key, priceFormFields) > NO_INDEX;
```
nicolad: IMO `includes` is cleaner: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes

nicolad: Unless we need `indexOf` for some specific reasons.

nicolad: Also, with `includes` we won't need that `NO_INDEX`

assets/prototype/domain/eventEditor/ticketsList/ticketPriceCalculator/TicketPriceModifierRow.js
```diff
@@ -0,0 +1,156 @@
...
+	let afterPrice = sign;
+	// isPercent T  currencySignB4 T  sign appears  afterPrice
+	// isPercent T  currencySignB4 F  sign appears  afterPrice
+	// isPercent F  currencySignB4 F  sign appears  afterPrice
+	// isPercent F  currencySignB4 T  sign appears  b4Price
+	if (currencySignB4 && !price.isPercent) {
+		b4Price = sign;
+		afterPrice = '';
+	}
+	const actions = [];
```
nicolad: What about this:
```suggestion
	const Actions = () => (
	<>
		{price.id === 'NEW_PRICE' && (
			<Button
				key={'add'}
				icon={'add'}
				onClick={() => {
					const newPrice = clone(price);
					newPrice.id = null;
					const baseType = getBaseType(newPrice.priceType);
					newPrice.order = baseType.order;
					newPrice.isDiscount = baseType.isDiscount;
					newPrice.isPercent = baseType.isPercent;
					push(newPrice);
					reset(name);
					sort();
				}}
				minimal
			/>
		)}
		{!price.isBasePrice && (
			<Button
				key={'trash'}
				icon={'trash'}
				onClick={() => {
					remove(index);
				}}
				minimal
			/>
		)}
	</>
);


```

?

assets/prototype/domain/eventEditor/ticketsList/ticketPriceCalculator/TicketPriceCalculatorForm.js
tn3rb: good point
ya I can explicitly pass in the mutators as separate props.

tn3rb:
going to merge this because a bunch of logic that affects queries and mutations as well as other UI components was added in this branch and we don't want that being held hostage by it

assets/prototype/domain/eventEditor/ticketsList/ticketPriceCalculator/TicketPriceModifierRow.js
```diff
@@ -0,0 +1,156 @@
...
+			<Button
+				key={'trash'}
+				icon={'trash'}
+				onClick={() => {
+					remove(index);
+				}}
+				minimal
+			/>
+		);
+	}
```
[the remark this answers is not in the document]
tn3rb: I should likely just extract those into their own components, then what you suggested above would work even better

assets/prototype/domain/eventEditor/ticketsList/ticketPriceCalculator/TicketPriceTotalRow.js
```diff
@@ -0,0 +1,62 @@
...
+import styles from './inlineStyles';
+
+// need to change these based on site i18n config
+const currencySign = '$';
+const currencySignB4 = true;
+const decimalPlaces = 2;
+// const decimalMark = '.';
+// const thousandsSep = ',';
+
+const  formatPriceAmount = (amount) => amount === undefined ? '' : Number.parseFloat(amount).toFixed(decimalPlaces);
```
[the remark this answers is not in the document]
tn3rb: > Should we check also if that string is a number?

which string? `amount` ? It's being parsed to a float

> We might extract something into some utils later.

oh ya this will absolutely go into some kind of utility file somewhere

assets/prototype/domain/eventEditor/ticketsList/ticketPriceCalculator/useTicketPriceCalculatorFormDecorator.js
```diff
@@ -0,0 +1,128 @@
...
+		return { ['prices[0].amount']: newBasePrice };
+	}, [calculator]);
+	const updateTicketTotal = useCallback((formData) => {
+		const result = calculator(formData, { type: 'CALCULATE_TOTAL' })
+		const newTicketTotal = getValue('ticket.price', result);
+		return { ['ticket.price']: newTicketTotal };
+	}, [calculator]);
+	return createDecorator(
+		{
+			field: /^prices\[\d+\]\.amount$/,
```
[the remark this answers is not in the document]
tn3rb: that's what RFF (React Final Form) expects for that... I just use online tools like https://jex.im/regulex/#!flags=&re=%5E(a%7Cb)*%3F for figuring things out.

Issue #1957, opened 2019-12-04
Investigate Console Warning


tn3rb:
Getting this warning in the console after the datetimes are initialized but before the tickets are initialized:

```
backend.js:6 Warning: Cannot update during an existing state transition (such as within `render`). Render methods should be a pure function of props and state. 
    in EventEditor
```

nicolad:
The problem is partially solved, I will need to get back after we are done with hydration.
I didn't want to spend too much time on this in PROTO phase.
cc @tn3rb @manzoorwanijk

PR #1972, opened 2019-12-06
Closes #1928 and improves #1957

### LINKED ISSUE #1928: DateRangePicker default state/ values
- Currently picker isn't picking default values passed through `dateRange`, need to check why this is not accepting that.

nicolad:
- For issue #1928 it's fixing the missed default state.
- For #1957 I am attempting to find the cause of the bug and fix that.

## Checklist

* [ ] My code is tested.


assets/prototype/domain/eventEditor/containers/queries/useInitQueries.js
```diff
@@ -27,24 +25,18 @@ const useInitQueries = ({ eventId }) => {
...
-	}
-
 	const datetimes = pathOr([], ['datetimes', 'nodes'], datetimesData);
 	const datetimeIn = datetimes ? datetimes.map(({ id }) => id) : [];
-
 	const toasterTicketsMessage = 'initializing event editor tickets';
-	const [
-		fetchRelatedTickets,
-		{ data: ticketsData, error: ticketError, loading: loadingTickets, called },
-	] = useLazyQuery(GET_TICKETS, {
```
nicolad: From what I understand [from this Apollo post ](https://blog.apollographql.com/apollo-client-now-with-react-hooks-676d116eeae2)is intended to be used upon some user action, check `Lazy evaluation for queries` paragraph.
***
And based on this discussion: https://github.com/apollographql/apollo-feature-requests/issues/119
`skip` option seem more suitable for cases when query needs to be executed conditionally.
I might be missing something, please correct me, if so.

assets/prototype/domain/eventEditor/containers/queries/useInitQueries.js
```diff
@@ -27,24 +25,18 @@ const useInitQueries = ({ eventId }) => {
...
+	const skipTicketsQuery = datetimeIn.length === 0;
+
+	const { data: ticketsData, error: ticketError, loading: loadingTickets } = useQuery(GET_TICKETS, {
 		variables: {
 			first: 50,
 			where: {
 				datetimeIn,
 			},
+			skip: skipTicketsQuery,
```
nicolad: On the other hand, passing `skip` true gives me this result:
image
@manzoorwanijk is something that needs to be done on serverside ?

nicolad: Also `skip` option is a nice declarative way of doing this chaining of queries.

nicolad:
@tn3rb @manzoorwanijk 
in context of #1957:
image
***
I see that we have the data we need in `useInitQueries` available in DOM JSON dump:
image

Maybe should I go and do Apollo rehydration and query from there, instead of making these HTTP requests.

assets/prototype/domain/shared/dateRangeInput/DateRangePicker.js
```diff
@@ -27,7 +27,7 @@ const DateRangePicker = ({ onFieldUpdate, range, setRange }) => {
 	const appliedProps = {
 		allowSingleDayRange: true,
 		contiguousCalendarMonths: false,
-		dateRange: range,
+		value: range,
```
nicolad: This is a 1-line fix for #1928, that's why I didn't created a separate PR just for that.
Here is the fix:

Peek 2019-12-06 14-19

assets/prototype/domain/eventEditor/containers/queries/useInitQueries.js
manzoorwanijk: `skip` is not to be added to variables, it should be added to options in `useMutation`.

assets/prototype/domain/shared/dateRangeInput/DateRangePicker.js
tn3rb: HA!  so much for my 8 hour estimate :\

assets/prototype/domain/eventEditor/containers/queries/useInitQueries.js
nicolad: :man_facepalming:

nicolad: thx :)

assets/prototype/domain/shared/dateRangeInput/DateRangePicker.js
nicolad: I got misled before by the example set by blueprint: https://github.com/palantir/blueprint/blob/develop/packages/docs-app/src/examples/datetime-examples/dateRangePickerExample.tsx#L36

PR #1975, opened 2019-12-06
Created ConfigProvider. 

nicolad:
## Problem this Pull Request solves
Solves this issue: https://github.com/eventespresso/event-espresso-core/issues/1941

## Checklist

* [ ] I have read the documentation relating to systems affected by this pull request, see https://github.com/eventespresso/event-espresso-core/tree/master/docs
* [ ] User input is adequately validated and sanitized
* [ ] all publicly displayed strings are internationalized (usually using `esc_html__()`, see https://codex.wordpress.org/I18n_for_WordPress_Developers)
* [ ] My code is tested.
* [ ] My code follows the Event Espresso code style.
* [ ] My code has proper inline documentation.
* [ ] My code accounts for when the site is in Maintenance Mode (MM2 especially disallows usage of models)

PR #1976, opened 2019-12-06
Create new EventProvider closes #1955

### LINKED ISSUE #1955: Create EventProvider
Considering we are working on the **Event** editor, having access to the event dataz is going to be a common requirement, especially after we start converting other sections of the editor over to React. 

Therefore we should create a Context Provider for making the Event ID available, which can then be used to obtain the Event data from the Apollo client (via another hook?).

The Event ID can be set in `/assets/prototype/domain/eventEditor/index.js` when the provider is applied, and the file should go in `assets/prototype/infrastructure/services/contextProviders`.

raphaelkross:

https://github.com/eventespresso/event-espresso-core/issues/1955

## Checklist

* [ ] My code is tested.


assets/prototype/infrastructure/services/contextProviders/ContextProviders.js
```diff
@@ -13,7 +14,9 @@ import { RelationsProvider } from './RelationsProvider';
 const ContextProviders = ({ children, eventId }) => (
 	<ToastProvider>
 		<ApolloProvider>
-			<RelationsProvider eventId={eventId}>{children}</RelationsProvider>
+			<RelationsProvider eventId={eventId}>
+				<EventProvider eventId={eventId}>{children}</EventProvider>
+			</RelationsProvider>
 		</ApolloProvider>
```
tn3rb: This is conflicting with `EDTR/master` because #1975 was just merged which added a Create ConfigProvider here as well
(smart of me to have two different people work on separate issues that both affect the same block of code eh? What could go wrong?)

Now on top of that conflict, I think we want to have the `EventProvider` on the outside of this stack and then have the `RelationsProvider` retrieve the `eventId` from the Apollo cache since that is going to be our "single source of truth". As well, after looking at this code and thinking about things a bit, I think we want to also rename this context provider to `EventEditorEventIdProvider` because we will very likely need an `EventProvider` for providing details about any event (for use in other admin pages and/or the site frontend)

But before you do that, you want to pull the latest changes to `EDTR/master` and then rebase this branch onto that one...

**BUT**

We want to avoid the conflict so here's what you could do (although you are welcome to handle things differently as long as the final result is the same):

1. rename your local branch for this PR to something like `FET/create-event-provider-2`
2. pull latest changes for `EDTR/master` 
3. cut a new branch from `EDTR/master` and give it the current name for this branch: `FET/create-event-provider`
4. copy the code over from `FET/create-event-provider-2` into the new branch
5. moving `EventProvider` to the outside of the stack, ie: have it wrap `RelationsProvider`:

```jsx
<EventProvider eventId={eventId}>
    <ConfigProvider>
        <RelationsProvider>{children}</RelationsProvider>
    </ConfigProvider>
</EventProvider>
```

6. then you can change `RelationsProvider` so that it is not getting the Event ID directly from the DOM, but is instead getting it from the EventProvider, but to do that...
7. create a custom hook called `useEventEditorEventId` that imports the EventContext, implements `useContext` to get the event id, and then returns it
8. then import this new hook into `assets/prototype/domain/eventEditor/containers/relations/useRelationsManager.js` and use it to get the event id
9. update `useRelationsManager` and `RelationsProvider` accordingly
10. force push your changes for this branch (set the origin to the same one this PR)

if the above is too much of a pain in the but, then you can also just rebase this onto master and just deal with the conflict since it won't be difficult to resolve

My apologies for the run around with this.

Now one more thing, after the above is all done, you will need to address the issue addressed in this comment: https://github.com/eventespresso/event-espresso-core/issues/1955#issuecomment-562208734

raphaelkross:
@tn3rb did the requested changes. Quick question => seems like RelationsProvider do not use eventId anymore, so didn't work on that.

About the hook to get the eventId, I've used one that was already done.

tn3rb:
> seems like RelationsProvider do not use eventId anymore, so didn't work on that.
> About the hook to get the eventId, I've used one that was already done.

ya that's fine but there really isn't any point in using an `EventEditorEventIdProvider` if it is going to pull the event ID from the `useEventId` hook right? Since most, if not all components will use the hook, the context provider would be doing nothing.

The primary point of #1955 was to avoid accessing `eventId` directly from the DOM and instead have the `EventEditorEventIdContext` handle that and then provide that data to the rest of the app. We don't want to be constantly going to the DOM to get the `eventId`.

So you need to:

- extract the logic in `useEventId` that pulls the `eventId` from the DOM and move it into `EventEditorEventIdProvider`
- remove the use of `useEventId` in `EventEditorEventIdProvider`
- throw an error in `EventEditorEventIdProvider` if `eventId` does not exist and/or pass the error message to the toaster

As well, I think we want to rearrange the order of these providers:

```js
<EventEditorEventIdProvider>
	<StatusProvider>
		<ConfigProvider>
			<RelationsProvider>{children}</RelationsProvider>
		</ConfigProvider>
	</StatusProvider>
</EventEditorEventIdProvider>
```

and move the `ConfigProvider` to the outside of that stack


```js
<ConfigProvider>
	<EventEditorEventIdProvider>
		<StatusProvider>
			<RelationsProvider>{children}</RelationsProvider>
		</StatusProvider>
	</EventEditorEventIdProvider>
</ConfigProvider>
```

because it's possible it may have some i18n settings that we may need

assets/prototype/application/services/context/EventEditorEventIdProvider.js
```diff
@@ -0,0 +1,20 @@
...
+import useToaster from '../toaster/useToaster';
+
+const EventEditorEventIdContext = createContext();
+
+const EventEditorEventIdProvider = ({ children }) => {
+	const toaster = useToaster();
+	const { eeEditorEventData: { eventId = 0 } = {} } = window;
+
+	if (eventId == 0 || eventId == null) {
+		toaster.error('Event ID is empty or not valid.');
```
tn3rb: great! 👍

Issue #1906, opened 2019-11-26
Execute the deletion


mnelson4:
Take the gathered info from https://github.com/eventespresso/event-espresso-core/issues/1904 and delete it. Maybe we can organize the data into flat arrays of IDs for quicker deletion (or possibly batch them, so we don't have any epically big queries).

* [x] reorganize the tree of data into flat lists of items (eg all the registrations, all the answers)
* [x] execute the deletion
* [x] record the deletion
* [ ] make the deletions use the command bus

from https://github.com/eventespresso/event-espresso-core/issues/349#issuecomment-557718323

mnelson4:
Since https://github.com/eventespresso/event-espresso-core/pull/2110 was merged into `CASC/base`, that branch now performs the cascading deletion vi a batch process, and we can start testing that aspect.
First note: so far, the UI isn't complete. The preview page is just a placeholder. But the code that performs the actual deletion could benefit from testing. Here's what I'd suggest:

* [ ] Create an event with no registrations. Trash it then permanently delete it. It should be deleted fine (you may want to check the database that its datetimes, tickets, message-group relations, etc, were removed properly.)
* [ ] Create an event with some registrations and do the same. Those registrations should also be deleted.
* Visit the transaction pages for one of those registrations. It won't have any registrations, but it should still be a valid record of a real transaction and payment. (Deleting transactions and payments is another job for another day.)
* [ ] Create an event that uses some custom questions, payments, and registrations and do the same. Check those registrations' answer rows are also deleted. 
* [ ] Create an event with 1000 registrations and do the same. It will take a few minutes to generate the preview page, and a minute to delete, but it should work. Preferably do this test from a cheap shared host (I have one on Dreamhost if that helps). Note though the event can't be infinitely big, I know it doesn't work deleting an event with over 15,000 registrations (in which I outlined a solution on https://github.com/eventespresso/event-espresso-core/pull/2110#issuecomment-570462637)

mnelson4:
Created https://github.com/eventespresso/event-espresso-core/issues/2194 just for tracking testing of the new cascading deletion of events. I consider this done (https://github.com/eventespresso/event-espresso-core/issues/2006 has some discussion around using the command bus for executing deletion)

Issue #1905, opened 2019-11-26
Preview deletions


mnelson4:
Show the user what data will be deleted before deleting it. This should use https://github.com/eventespresso/event-espresso-core/issues/1904 to figure out what information that is.
It shouldn't crash when there are thousands of registrations.
Users should confirm they have a database backup and confirm they are deleting the correct event.

* [x] generate the preview page via a batch job 
* * [x] count the records- tricky because what we're doing is discovering
* * [x] keep track of where we left off
* * [x] store the built tree (even when it has thousands of entries)
From https://github.com/eventespresso/event-espresso-core/issues/349#issuecomment-557718323
* [x] display the preview page
* [x] add a link to it
* [x] Clean out old "built trees" (database options) that weren't ever used for deletion

mnelson4:
With #2110 merged into `CASC/base`, the next thing I think to do is actually make the preview page. 

My current plan is something like this:

`
-------------------------------------
| Preview Data for deletion
|
| X Events:
| Event x
| Event Y
|
| Y datetimes:
| datetime x
| datetime y
| 
| X tickets:
| ticket x
| ticket y
|
| ... registrations, prices, custom message templates, and I think that's all I'd mention (relation rows are assumed)
| 
| Confirm you made a database backup []
|
| You are about to delete Event X (whose permalink is y). Please confirm the permalink of the event [______]
| You are about to delete Event Y (whose permalink is x). Please confirm the permalink of the event [______]
| ...etc for each event
| (confirm and begin deletion)
`

In prose, we'll show the number of events to delete, and their names; then the number of datetimes to delete and their names (or time range etc); then tickets, registrations, prices, and message templates. (But not other data that's more removed from users, like extra meta, post meta, join-table entries, etc.)

Then we'll have a checkbox asking them to confirm they have made a database backup.

And lastly, for each event being deleted, we'll tell them they're about to delete the event's name, say its permalink, and then ask them to re-type the permalink (probably just the slug portion, the rest is should be autofilled).

@tn3rb @joshfeck does that make sense and do you have feedback on it?

joshfeck:
I don't think it should make a mention of deleting custom message templates. I know it's only deleting the relationship between a custom message template and the event, but that could be confusing to the end user and they may think their custom message template will also be deleted, which they may not want.

mnelson4:
[...]

Issue #349, opened 2018-04-25
Allow trashed events to be deleted and at the same time prevent them from orphaning other data records GDPR


sethshoultes:
See original Codebase ticket:
https://events.codebasehq.com/projects/event-espresso/tickets/9009

This has come up in the forums and I think there's some room for improvement when it comes to making things easier for event managers.

Steps to reproduce: 
1) Set up an event and assign some people CPTs to the event, save the event.
2) Trash the event
3) Go the trashed events view in the wp admin list table
4) check the box next to the event
5) select the Delete Permanently action and click Apply

Result is you'll see some errors that say you can't delete the event. So now you have to go and restore the event from the trash, then you have to edit the event, and remove all of the relationships to things like People CPts. Then trash, then you can permanently delete.

Can we change the above so the relationships automatically get removed so all they'll have to do is trash then delete the event?

sethshoultes:
Oops. Didn't mean to do that. I updated the category to GDPR.

mnelson4:
thanks Stale bot, but I'm pretty sure we still intend to work on this

joshfeck:
see also:
https://events.codebasehq.com/projects/event-espresso/tickets/7290

joshfeck:
Would this also allow for deleting events that have registrations?

mnelson4:
I don't think it's been specified. What would customers probably want to happen when deleting an event with registrations? orphan them? block deletion of the event until registrations are deleted? automatically delete all the registrations too?

joshfeck:
The current behavior is it blocks deletion of the event until registrations are deleted.

Would it be OK to instead allow the event to be deleted and orphan the registrations?

mnelson4:
> Would it be OK to instead allow the event to be deleted and orphan the registrations?

It would certainly test how defensive our code- the code should generally just say "oups, can't find the event for this registration" and then keep working. But there may be the odd fatal error that would happen because we got careless.

Although I've reviewed this ticket some more: I think this work would only help with GDPR and user privacy etc if it actually facilitated deleting registrations/contacts. Just deleting the event and orphaning the registrations doesn't help with GDPR because the personal information (on the registration and contact) is still in the system.

I'm actually not sure what use-case facilitating just deleting the event (and orphaning its related data) would benefit.

So I would think this is really only useful if it will also delete the personal information contained on registrations and contacts etc.

joshfeck:
[...]

PR #1977, opened 2019-12-06
Add classes to identify dependent model objects. Merge to CASC/base

mnelson4:
Original PR: https://github.com/eventespresso/event-espresso-core/pull/1888 

Adds logic so we can provide a starter model object, then identify all of its dependent model objects (ie, ones model objects who have a foreign key to this model object), and all of their dependent model objects, recursively.
This can be done in batches, although this doesn't add the logic to store anything to the database. I'll do that in https://github.com/eventespresso/event-espresso-core/issues/1905. 
This also does not perform the deletion. For that see https://github.com/eventespresso/event-espresso-core/issues/1906

See https://github.com/eventespresso/event-espresso-core/issues/349

## How has this been tested

Unit tested only

## Checklist

* [x] My code is tested.


mnelson4:
Same PR as before, but into a the cascading deletion base branch

PR #1985, opened 2019-12-09
Created useConfigProvider hook. closes #1979

### LINKED ISSUE #1979: Add `useConfigProvider` Hook
I missed that this wasn't part of #1975 but we need a custom hook that imports the ContextProvider and then retrieves the various config objects and returns them to the component using the hook.

nicolad:
Implements: https://github.com/eventespresso/event-espresso-core/issues/1979

## Checklist

* [ ] My code is tested.

### DIFF
--- a/assets/prototype/domain/eventEditor/containers/useConfigProvider.ts
+++ b/assets/prototype/domain/eventEditor/containers/useConfigProvider.ts
@@ -1,0 +1,30 @@
+import { useContext } from 'react';
+import { pick } from 'ramda';
+import { ConfigContext } from '../../../infrastructure/services/contextProviders/ConfigProvider';
+
+type useConfigProviderProps = {
+	dateFormat: boolean;
+	email: boolean;
+	timezone: boolean;
+};
+
+/**
+ *
+ * useConfigProvider is used to cherry-pick the exact configs we might need from ConfigProvider.
+ */
+
+const useConfigProvider = (props: useConfigProviderProps) => {
+	const config = useContext(ConfigContext);
+
+	const configArray = Object.entries(props).reduce((configs, [configName, configBool]) => {
+		if (configBool) {
+			configs.push(configName);
+		}
+
+		return configs;
+	}, []);
+
+	return pick(configArray, config);
+};
+
+export default useConfigProvider;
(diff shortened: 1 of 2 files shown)


assets/prototype/domain/eventEditor/containers/useConfigProvider.ts
```diff
@@ -0,0 +1,25 @@
+import { useContext } from 'react';
+import { pick } from 'ramda';
+import { ConfigContext } from '../../../infrastructure/services/contextProviders/ConfigProvider';
+
+type Props = {
+	dateFormat: boolean;
+	email: boolean;
+	timezone: boolean;
```
nicolad: We will add/remove here more as we will need.

manzoorwanijk: `boolean`??

nicolad: renamed the type name to make more clear the intention

assets/prototype/domain/eventEditor/containers/useConfigProvider.ts
```diff
@@ -0,0 +1,30 @@
+import { useContext } from 'react';
+import { pick } from 'ramda';
+import { ConfigContext } from '../../../infrastructure/services/contextProviders/ConfigProvider';
+
+type configsToBeRetrieved = {
```
nicolad: Have used `type` here, because this isn't yet a case for interface, based on:
image

nicolad: from: https://github.com/typescript-cheatsheets/react-typescript-cheatsheet#types-or-interfaces

assets/prototype/domain/eventEditor/containers/useConfigProvider.ts
```diff
@@ -0,0 +1,30 @@
+import { useContext } from 'react';
+import { pick } from 'ramda';
+import { ConfigContext } from '../../../infrastructure/services/contextProviders/ConfigProvider';
+
+type ConfigsToBeRetrieved = {
```
nicolad: CNO @tn3rb I might need a better naming here, open to suggestions.

tn3rb: I think we should standardize the naming for all type declarations and interfaces.

Since we are type casting (not sure if that is the correct term when using TS?) the props for a component/hook, then the naming schema should be:

```js
`${componentOrHookName}Props`
```

So in this case, the prop types for `useConfigProvider` should simply be named `useConfigProviderProps`

This way, no one will ever have to guess or think about what the corresponding props type is.

assets/prototype/domain/eventEditor/containers/useConfigProvider.ts
tn3rb: 👍

PR #1987, opened 2019-12-09
Get IDs of related model objects

mnelson4:

In #1977, we can create a tree of dependent model objects. With this ticket, we convert that tree of data into a 2D array, where the top-level keys are model names, and the value is an array of model object Ids. 
Using this, we can delete all the events in one query, all the datetimes in one query, etc. Which should be much more efficient than deleting them all as we traverse the tree again.

This addresses the first part of #1906 

## How has this been tested

Unit tested. There's no UI for it yet.

## Checklist

* [x] My code is tested.


core/services/orm/tree_traversal/ModelObjNode.php
```diff
@@ -127,6 +127,28 @@ public function toArray()
...
+     * @throws InvalidDataTypeException
+     * @throws InvalidInterfaceException
+     * @throws ReflectionException
+     * @throws \EE_Error
+     */
+    public function getIds()
+    {
+        $ids = [
+            $this->model_obj->get_model()->get_this_model_name() => [
+                $this->model_obj->ID() => $this->model_obj->ID()
```
tn3rb: Will using the ID for the value cause issues if the `relation_nodes` array does not have a corresponding key for this model (which would result in this value being left as is)?

Would it be better to maybe use an empty array for the value because then any consuming code does not have to handle different types? 
ie: consuming code merely has to check if the array is empty or not

core/services/orm/tree_traversal/RelationNode.php
```diff
@@ -187,11 +190,22 @@ protected function discover()
...
+        } catch (EE_Error $e) {
+            // This will happen for has-and-belongs-to-many relations, when this node's related model is that join table
+            // which hasn't been explicitly declared in the main model object's model's relations.
+            $relation_settings = null;
+        }
+        if ($relation_settings instanceof EE_Has_Many_Any_Relation) {
+            $where_params[ $this->related_model->get_field_containing_related_model_name()->get_name() ] = $this->main_model_obj->get_model()->get_this_model_name();
+        }
+        return $where_params;
     }
```
mnelson4: I made this change because I realized has-many-any relations (like from event to event meta) could select the wrong things if we weren't also adding a WHERE condition to use the right model name in the model name field (eg settings `'EXM_type'='Event'`)

core/services/orm/tree_traversal/RelationNode.php
```diff
@@ -209,6 +223,28 @@ public function toArray()
...
+    public function getIds()
+    {
+        if (empty($this->model_obj_nodes)) {
+            return [];
+        }
+        $ids = [
+            $this->related_model->get_this_model_name() => array_combine(
+                array_keys($this->model_obj_nodes),
+                array_keys($this->model_obj_nodes)
+            )
```
tn3rb: same thing here... are IDs the best thing to use for the value?

You could populate the array using IDs for keys and an empty array for the values like:

```php
$this->related_model->get_this_model_name() => array_fill_keys(
    array_keys($this->model_obj_nodes),
    []
)
```

mnelson4:
Ready for review

core/services/orm/tree_traversal/ModelObjNode.php
mnelson4: I'm not sure if I totally grok but here goes...
The values are basically unused; what matters is that the top-most key is a model name; and then the 2nd level keys are the model object IDs.
I don't think consuming code needs to handle different types for the values because (so far) the 2nd-level values are always IDs.

mnelson4:
I'm going to merge to the main branch but that doesn't set this in stone. I'm really trying to race towards getting something that works and does the bare minimum and then improve it from there.

PR #1995, opened 2019-12-10
Added Datetime format config support; closes #1942.

### LINKED ISSUE #1942: Global Date & Time Format Constants Config
- create centralized file for defining the various date and time formats required throughout the system
 - make config available via ConfigContext
 - import into any and all classes that are currently setting date related constants (formats, etc) and update logic to use new config constants 

I've estimated 4 hours for this (plz give feedback if that does not seem to be enough)

nicolad:

Implements: https://github.com/eventespresso/event-espresso-core/issues/1942

## Checklist

* [ ] My code is tested.


assets/prototype/domain/shared/dateRangeInput/DateRangePicker.js
```diff
@@ -1,21 +1,21 @@
+import moment from 'moment';
```
nicolad: Ideally, we should have our own `format` function somewhere in datetime utils and in that file we can swap in the future `moment` with whatever we want.

tn3rb: yay that gets tricky because the different libs use different formats, but I guess that's the point of adapters eh?

assets/prototype/application/value-objects/date-time/getDateTimeFormat.ts
```diff
@@ -0,0 +1,14 @@
+import { defaultDateFormat, defaultTimeFormat } from './defaults';
+
+type getDateTimeFormatProps = {
+	dateFormat?: string;
+	timeFormat?: string;
+};
+
+const getDateTimeFormat = ({ dateFormat, timeFormat }: getDateTimeFormatProps) => {
+	const format = dateFormat || defaultDateFormat + ' ' + timeFormat || defaultTimeFormat;
```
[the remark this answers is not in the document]
nicolad: yes, that piece can and need to be improved, thx for the comment.

assets/prototype/domain/shared/dateRangeInput/DateRangePicker.js
```diff
@@ -1,21 +1,21 @@
...
 	const handleRangeChange = (value) => {
 		const [start, end] = value;
 		const isUpdatable = isFunction(onFieldUpdate) && start && end;
 
 		if (isUpdatable) {
 			const formattedDates = {
-				startDate: format.formatDate(start),
-				endDate: format.formatDate(end),
+				startDate: moment(start).format(format),
+				endDate: moment(end).format(format),
```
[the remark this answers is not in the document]
nicolad: yes, true story and also if we'd omit this part from the original issue:
`import into any and all classes that are currently setting date related constants (formats, etc) and update logic to use new config constants`
then we can keep rolling with this PR (don't want to have conflicts)

assets/prototype/domain/shared/dateRangeInput/DateRangePicker.js
nicolad: yeah and we don't want to paint yourself in the corner with a lib.

assets/prototype/application/value-objects/date-time/getDateTimeFormat.ts
```diff
@@ -0,0 +1,14 @@
...
+
+	return format;
```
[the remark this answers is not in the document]
nicolad: done: https://prettier.io/docs/en/precommit.html
thx

assets/prototype/application/value-objects/date-time/getDateTimeFormat.ts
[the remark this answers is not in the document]
nicolad: updated, please check.

nicolad:
> looks good but had a couple of comments/suggestions
> 
> as well, I noticed that some of the files were in
> 
> `/application/value-objects/date-time/`
> 
> which is fine, but we had decided to use camelCase for folder names right?
> 
> so this should be changed to:
> 
> `/application/valueObjects/dateTime/`

updated, made the same approach as here without thinking:
image

assets/prototype/application/valueObjects/dateTime/getDateTimeFormat.ts
```diff
@@ -0,0 +1,20 @@
...
+
+type getDateTimeFormatProps = {
+	dateFormat?: string;
+	timeFormat?: string;
+};
+
+const getDateTimeFormat = (props: getDateTimeFormatProps) => {
+	const dateFormat = props.dateFormat || defaultDateFormat;
+	const timeFormat = props.timeFormat || defaultTimeFormat;
+	const dateTimeFormat = dateFormat + ' ' + timeFormat;
```
tn3rb: cool ! even better than what I suggested!!! 💯

assets/prototype/domain/eventEditor/containers/useConfigProvider.ts
```diff
@@ -1,30 +1,34 @@
...
+const useConfigProvider = ({ dateTimeFormat }: useConfigProviderProps) => {
+	const { dateFormat, timeFormat } = useContext(ConfigContext);
+	let config: any = {};
 
-	const configArray = Object.entries(props).reduce((configs, [configName, configBool]) => {
-		if (configBool) {
-			configs.push(configName);
-		}
+	if (dateTimeFormat) {
+		config.dateTimeFormat = getDateTimeFormat({ dateFormat, timeFormat });
```
tn3rb: does this need to be updated as well to account for the combined `dateTimeFormat` being added?

assets/prototype/infrastructure/services/contextProviders/ConfigProvider.js
```diff
@@ -41,7 +42,7 @@ const ConfigProvider = ({ children }) => {
 		],
 		currentUser
 	);
-	const value = { ...generalSettingsProps, ...currentUserProps };
+	const value = { dateFormat, timeFormat, ...currentUserProps };
```
tn3rb: same thing here... do we need `dateTimeFormat` too ?

assets/prototype/domain/eventEditor/containers/useConfigProvider.ts
nicolad: yes, my bad here, didn't fully finish all

assets/prototype/domain/eventEditor/containers/useConfigProvider.ts
```diff
@@ -1,30 +1,25 @@
...
+const useConfigProvider = ({ dateTimeFormat }: useConfigProviderProps) => {
+	const { dateFormat, timeFormat } = useContext(ConfigContext);
+	let config: any = {};
 
-const useConfigProvider = (props: useConfigProviderProps) => {
-	const config = useContext(ConfigContext);
+	if (dateTimeFormat) {
+		const dateTimeFormat = getDateTimeFormat({ dateFormat, timeFormat });
+		config = { ...config, ...dateTimeFormat };
+	}
```
nicolad: @tn3rb this is the result, maybe I'll refactor the approach a bit as new information will come in

nicolad: if nothing will be passed before `if (dateTimeFormat) {`
the `config` will be:
```
config = {
        dateFormat,
		dateTimeFormat,
		timeFormat,
}
```
again, I am not married to this approach, this is just the simplest way to pass data around

assets/prototype/infrastructure/services/contextProviders/ConfigProvider.js
nicolad: no this is the data we get from `GET_GENERAL_SETTINGS` query, WPGraphQL isn't exposing `dateTimeFormat`:

image

assets/prototype/domain/eventEditor/containers/useConfigProvider.ts
```diff
@@ -1,30 +1,25 @@
...
-	timezone: boolean;
+	dateTimeFormat: boolean;
 };
 
 /**
  *
  * useConfigProvider is used to cherry-pick the exact configs we might need from ConfigProvider.
  */
+const useConfigProvider = ({ dateTimeFormat }: useConfigProviderProps) => {
+	const { dateFormat, timeFormat } = useContext(ConfigContext);
```
manzoorwanijk: Instead of `useContext(ConfigContext)`, we could create a hook like `useConfiguration()` or `useConfig()` to return that context.

assets/prototype/domain/eventEditor/containers/useConfig.js
```diff
@@ -0,0 +1,6 @@
+import { useContext } from 'react';
```
nicolad: Didn't made it a `.ts` file cause it would require a chain of other changes.
Will loop back later when will have a better TS coverage.

nicolad: Also in other PR we might unite these kind of hooks in a special folder.

PR #2007, opened 2019-12-11
Fix GraphQL Errors When Updating Entity Relations fixes #1993

### LINKED ISSUE #1993: Look for the server-side GQL error
When updating datetime by only changing the related tickets or vice versa, it results in an unknown error. I think the error is caused because there are no changes in the actual entity, rather only the related items change. The source of error can be found by dumping he result of `$result = $entity->save($args);` in `DatetimeUpdate` and `TicketUpdate` under `EventEspresso\core\domain\services\graphql\mutators`.

tn3rb:
plz see #1993

There were a couple of issues happening that resulted in errors appearing in the GQL responses:

 - `DatetimeUpdate`, `TicketUpdate`, and `PriceUpdate` we were using the return value from `EE_Base_Class::save()` to determine whether the update was successful, but entities don't save if there were no changes made to them and that method returns `0`. The solution was simply to wrap the calls to `save()` in a try catch block and handle any actual errors that occur.

 - in `useOnSubmitPrices` values were getting submitted that were not of the correct type, such as `"false"` (string)  instead of `false` (boolean), so the solution was to normalize any values we know could cause problems.

- also fixed a couple other places where values were being set to ensure they were the correct data type

- also fixed another hook related problem in the ticket price calculator

### DIFF
--- a/assets/prototype/domain/eventEditor/ticketsList/ticketPriceCalculator/hooks/useOnSubmitPrices.js
+++ b/assets/prototype/domain/eventEditor/ticketsList/ticketPriceCalculator/hooks/useOnSubmitPrices.js
@@ -14,6 +14,22 @@ const PRICE_INPUT_FIELDS = [
 	'wpUser',
 ];
 
+const parseBooleanField = (value) => {
+	value = typeof value === 'string' ? value.toLowerCase().trim() : value;
+	switch (value) {
+		case 'true':
+		case 'yes':
+		case '1':
+			return true;
+		case 'false':
+		case 'no':
+		case '0':
+			return false;
+		default:
+			return Boolean(value);
+	}
+};
+
 const useOnSubmitPrices = (existingPrices) => {
 	const { createEntity, updateEntity, deleteEntity } = useEntityMutator('Price');
 	const { updateEntity: updateTicket } = useEntityMutator('Ticket');
@@ -35,10 +51,8 @@ const useOnSubmitPrices = (existingPrices) => {
 			const normalizedPriceFields = {
 				...input,
 				amount: parseFloat(price.amount || 0),
-				isDefault: !! price.isDefault,
+				isDefault: parseBooleanField(price.isDefault),
 				order: parseInt(price.order, 10),
-				overrides: parseInt(price.overrides, 10),
-				parent: parseInt(price.parent, 10),
 			};
 			// if it's a newly added price
 			if (!id) {
@@ -59,7 +73,7 @@ const useOnSubmitPrices = (existingPrices) => {
 		const normalizedTicketFields = {
 			...ticket,
 			price: parseFloat(ticket.price || 0),
-			reverseCalculate: !! ticket.reverseCalculate,
+			reverseCalculate: parseBooleanField(ticket.reverseCalculate),
 		};
 		// Finally update the ticket price relation
 		updateTicket({ ...normalizedTicketFields, prices: updatedPrices });
--- a/assets/prototype/domain/eventEditor/ticketsList/ticketPriceCalculator/hooks/useOnSubmitPrices.js
+++ b/assets/prototype/domain/eventEditor/ticketsList/ticketPriceCalculator/hooks/useOnSubmitPrices.js
@@ -31,11 +31,20 @@ const useOnSubmitPrices = (existingPrices) => {
 				return;
 			}
 			const { id, ...input } = pick(PRICE_INPUT_FIELDS, price);
+
+			const normalizedPriceFields = {
+				...input,
+				amount: parseFloat(price.amount || 0),
+				isDefault: !! price.isDefault,
+				order: parseInt(price.order, 10),
+				overrides: parseInt(price.overrides, 10),
+				parent: parseInt(price.parent, 10),
+			};
 			// if it's a newly added price
 			if (!id) {
-				createEntity({ ...input, ticketId: ticket.id });
+				createEntity({ ...normalizedPriceFields, ticketId: ticket.id });
 			} else {
-				updateEntity({ id, ...input });
+				updateEntity({ id, ...normalizedPriceFields });
 			}
 			updatedPrices.push(id);
 		});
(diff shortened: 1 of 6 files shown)


assets/prototype/domain/eventEditor/ticketsList/ticketPriceCalculator/hooks/useOnSubmitPrices.js
```diff
@@ -31,11 +31,20 @@ const useOnSubmitPrices = (existingPrices) => {
...
 			}
 			const { id, ...input } = pick(PRICE_INPUT_FIELDS, price);
+
+			const normalizedPriceFields = {
+				...input,
+				amount: parseFloat(price.amount || 0),
+				isDefault: !! price.isDefault,
+				order: parseInt(price.order, 10),
+				overrides: parseInt(price.overrides, 10),
+				parent: parseInt(price.parent, 10),
```
manzoorwanijk: Many of these fields may not be present in `input`. Also, `parent` is `id` (GUID) of another price.

assets/prototype/domain/eventEditor/ticketsList/ticketPriceCalculator/hooks/useOnSubmitPrices.js
```diff
@@ -47,8 +56,13 @@ const useOnSubmitPrices = (existingPrices) => {
 			deleteEntity({ id });
 		});
 
+		const normalizedTicketFields = {
+			...ticket,
+			price: parseFloat(ticket.price || 0),
+			reverseCalculate: !! ticket.reverseCalculate,
```
manzoorwanijk: `ticket.reverseCalculate` is observed to be a string e.g. `"false"`, `!! ticket.reverseCalculate` will always be true.

assets/prototype/domain/eventEditor/ticketsList/ticketPriceCalculator/hooks/useTicketPriceCalculator.js
```diff
@@ -1,18 +1,27 @@
...
-		default:
-			return state;
-	}
-};
+const useTicketPriceCalculator = () => useCallback(
+	(state, action) => {
+		switch (action.type) {
+			case 'CALCULATE_BASE_PRICE':
+				return calculateBasePrice(state);
+			case 'CALCULATE_TICKET_TOTAL/':
```
manzoorwanijk: May be that `/` is a typo?

core/domain/services/graphql/mutators/DatetimeUpdate.php
```diff
@@ -83,13 +87,24 @@ public static function mutateAndGetPayload(EEM_Datetime $model, Datetime $type)
...
+                $entity->save($args);
+            } catch (Exception $exception) {
+                new ExceptionStackTraceDisplay(
+                    new RuntimeException(
+                        sprintf(
+                            esc_html__(
+                                'The Datetime failed to update because of the following error(s):%1$s%2$s',
+                                'event_espresso'
+                            ),
+                            '<br/>',
```
manzoorwanijk: GraphQL clients usually treat those messages as escaped text, HTML is no rendered.

core/domain/services/graphql/mutators/PriceUpdate.php
```diff
@@ -76,10 +79,21 @@ public static function mutateAndGetPayload(EEM_Price $model, Price $type)
...
+                $entity->save($args);
+            } catch (Exception $exception) {
+                new ExceptionStackTraceDisplay(
+                    new RuntimeException(
+                        sprintf(
+                            esc_html__(
+                                'The Price failed to update because of the following error(s):%1$s%2$s',
+                                'event_espresso'
+                            ),
+                            '<br/>',
```
manzoorwanijk: Same here

assets/prototype/domain/eventEditor/ticketsList/ticketPriceCalculator/hooks/useOnSubmitPrices.js
nicolad: Let's avoid using double negation trick.
More info on why:
- https://www.sitepoint.com/javascript-double-negation-trick-trouble/
- https://riptutorial.com/javascript/example/3047/double-negation----x-

assets/prototype/domain/eventEditor/ticketsList/ticketPriceCalculator/hooks/useOnSubmitPrices.js
tn3rb: oh right... forgot those are guids client side

assets/prototype/domain/eventEditor/ticketsList/ticketPriceCalculator/hooks/useTicketPriceCalculator.js
tn3rb: DOH! yes it is...  thnx

assets/prototype/domain/eventEditor/ticketsList/ticketPriceCalculator/hooks/useTicketPriceCalculatorFormDecorator.js
```diff
@@ -42,15 +42,15 @@ const useTicketPriceCalculatorFormDecorator = () => {
...
 			const result = calculator(formData, { type: 'CALCULATE_BASE_PRICE' });
 			const newBasePrice = getValue('prices[0].amount', result);
-			return { ['prices[0].amount']: newBasePrice };
+			return { ['prices[0].amount']: parseFloat(newBasePrice || 0) };
 		},
 		[calculator]
 	);
 	const updateTicketTotal = useCallback(
 		(formData) => {
 			const result = calculator(formData, { type: 'CALCULATE_TICKET_TOTAL/' });
```
manzoorwanijk: `/` here a well

tn3rb: OH DANG!

PR #2019, opened 2019-12-12
Bug/restore ticket id attendee template

joshfeck:

In #1723 there were some changes to the templating and this little html class got overlook. This PR makes an attempt at restoring the CSS class that includes the ticket ID which is in the original attendee_info_main.template.php template.
## How has this been tested

To test do a registration with one ticket and check the source. There should be a CSS classname like `spco-attendee-ticket-{ticket ID}`, which can be used to uniquely style a reg form's div for a specific ticket.

## Checklist

* [x] My code is tested.


modules/single_page_checkout/reg_steps/attendee_information/attendee_info_single.template.php
```diff
@@ -28,28 +28,28 @@
     ?>
 
     <div id="spco-attendee-panel-dv-single"
-        class="spco-attendee-panel-dv spco-attendee-ticket-single">
+        class="spco-attendee-panel-dv spco-attendee-ticket-<?php echo key($template_args['ticket_count']); ?>">
```
tn3rb: `ticket_count` is not the ticket ID, it is literally the number of tickets, which in this case will always be one.

if you want to get the actual ticket ID, you could do something like the following:

```php
if (count($registrations) > 0) {
    // rewind registrations array
    reset($registrations);
    // grab first element without modifying array
    $registration = current($registrations);
    // set default for ticket ID (same as what ticket count is now)
    $ticketID = 1;
    if ($registration instanceof EE_Registration && $registration->ticket() instanceof EE_Ticket) {
        $ticketID = $registration->ticket()->ID();
    }
    ?>  

    <div id="spco-attendee-panel-dv-<?php echo $ticketID; ?>"
        class="spco-attendee-panel-dv spco-attendee-ticket-<?php echo $ticketID; ?>">
```

joshfeck: Can you take another look at the commit? You may have missed that I wasn't actually using the ticket count value, but instead the array key for the ticket count value. Which is the ticket ID.

From EE_SPCO_Reg_Step_Attendee_Information.class.php:202

```
$template_args['ticket_count'][ $registration->ticket()->ID() ] = isset(
                        $template_args['ticket_count'][ $registration->ticket()->ID() ]
                    )
                        ? $template_args['ticket_count'][ $registration->ticket()->ID() ] + 1
                        : 1;
```

It seems to me that we've already got the ticket ID passed in via template args so couldn't we use that?

sethshoultes:
I can see the class name was added here:
image

Issue #2016, opened 2019-12-11
Data Hydration


tn3rb:
Investigate and implement logic for hydrating the Apollo Client cache using the data that is already being dumped into the DOM.

The relations data can be dumped separately as it is not stored in Apollo cache. -mw

PR #2024, opened 2019-12-12
Store rehydration

manzoorwanijk:
Completes #2016 

## Checklist

* [ ] My code is tested.


assets/prototype/domain/eventEditor/containers/queries/useCacheRehydration.js
```diff
@@ -0,0 +1,84 @@
...
+				},
+				data: {
+					datetimes,
+				},
+			});
+		}
+		setIsLoaded('datetimes', true);
+
+		const datetimeIn = nodes.map(({ id }) => id);
+		({ nodes = [] } = tickets);
```
nicolad: not sure what it does

manzoorwanijk: destructuring by using an existing variable i.e. without `let` or `const` when destructuring. You need to wrap it in `()` to make it an expression.

nicolad: too much magic involved, can this be accomplished in other ways?

nicolad: from what you describe it sounds like the logic might be split into 2 or more small functions

assets/prototype/domain/eventEditor/containers/queries/useCacheRehydration.js
```diff
@@ -0,0 +1,84 @@
...
+import { useApolloClient } from '@apollo/react-hooks';
+import useCacheRehydrationData from './useCacheRehydrationData';
+import useRelations from '../../../../infrastructure/services/relations/useRelations';
+import useStatus from '../../../../infrastructure/services/status/useStatus';
+import useEventId from './useEventId';
+import { queries } from './';
+
+const { GET_TICKETS, GET_DATETIMES, GET_PRICE_TYPES, GET_PRICES } = queries;
+
+const useCacheRehydration = () => {
```
nicolad: @manzoorwanijk this isn't exactly the approach I saw here:
https://www.apollographql.com/docs/react/performance/server-side-rendering/#store-rehydration

nicolad: my understanding is that this is updating cache on `EventEditor` component mount

nicolad: I want to understand why we didn't go through this path: `cache: new InMemoryCache().restore(window.__APOLLO_STATE__)` ?

assets/prototype/domain/eventEditor/containers/queries/useCacheRehydration.js
manzoorwanijk:
```js
let { name } = person;
```
is same as
```js
let name;
({ name} = person);
```
That way you can reuse an already declared variable

assets/prototype/domain/eventEditor/containers/queries/useTicketIds.js
```diff
@@ -1,35 +1,9 @@
-import * as R from 'ramda';
-import { useApolloClient } from '@apollo/react-hooks';
-
-import useDatetimeIds from './useDatetimeIds';
-import { GET_TICKETS } from './tickets';
-import useStatus from '../../../../infrastructure/services/status/useStatus';
+import useTickets from './useTickets';
 
 const useTicketIds = () => {
```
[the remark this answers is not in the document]
manzoorwanijk: Good idea

assets/prototype/domain/eventEditor/containers/queries/useCacheRehydration.js
manzoorwanijk: The reason for that is we don't use Apollo Server on the backend. So we need to write to the cache using `client.writeQuery` as mentioned [here](https://www.apollographql.com/docs/react/data/local-state/#initializing-the-cache).

Issue #2035, opened 2019-12-13
Review queries related to "clean up" scheduled tasks


joshfeck:
There were some "out of memory" errors on ES and we tracked one of the causes down to a site that had over 10,000 failed transactions over a week old (or junk transactions) where it was attempting to delete them all in one request.

There was a hot fix applied directly to ES in 
https://github.com/eventespresso/eventsmart.com-website/pull/644

So we'll need to do something similar in EE4 core so a limit set there.

In addition to that, all other similar queries that run with scheduled tasks should be reviewed and a reasonable LIMIT added. 

It might help to imagine a worse-case scenario of wp-cron getting shut down for a long time, then a lot of registrations happen, then wp-cron getting turned back on again. What will the query results look like the first time the scheduled tasks fire?

Some of the other methods that include queries needing review include:
`EEM_Message::instance('')->delete_old_messages()`
EEM_Transaction::instance('')->delete_junk_transactions();
EEM_Registration::instance('')->delete_registrations_with_no_transaction();
EEM_Line_Item::instance('')->delete_line_items_with_no_transaction();

mnelson4:
On a related note, `\EED_Ticket_Sales_Monitor::reset_reservation_counts` seems to sometimes be a major time (and I think memory) sink. See this call stack:
![screenshot-rpm newrelic com-2019 12 13-10_33_20](https://user-images.githubusercontent.com/1766382/70826647-4badb380-1d9c-11ea-92eb-9998fdca99c5.png)

I think the method doesn't actually need to build all the transactions (especially when there can be thousands of them on ES), and it could probably skip fetching the transaction's main line item, and just fetch the ticket line items directly. 

@tn3rb are you aware of any impediments to optimizing that method? (It looks like we had assumed it would only be dealing with a few at a time; but some ES sites are dealing with hundreds at a time, maybe thousands.)

mnelson4:
https://github.com/eventespresso/event-espresso-core/pull/2060 is off for testing in EE core. It might be feasible to cherry-pick into ES, (although it was a sizeable PR, the essential changes are just located on about 3 lines).
It may significantly help with load because it changes `EED_Ticket_Sales_Monitor::reset_reservation_counts();` from looping over a week's worth of failed transactions to instead just an hour's worth.

mnelson4:
related PR #2060 merged

PR #2037, opened 2019-12-13
Bug/add limits to cron queries

mnelson4:

Helps with timeouts when there are sites with lots of expired, incomplete transactions. See https://github.com/eventespresso/event-espresso-core/issues/2035

## How has this been tested

* [x] View an event and submit a ticket purchase. But don't enter any info. In the DB, change its transaction's `TXN_timestamp` to be a year ago, then run the cron task `AHEE__EE_Cron_Tasks__clean_up_junk_transactions` (see previous item`). The transaction should get deleted
* [x] Verify old messages are still deleted as they were before (after a month by default, although you can use the filter `FHEE__EEM_Message__delete_old_messages__time_to_leave_alone` to set a different amount of time to wait, in seconds)

## Checklist

* [x] My code is tested.

### DIFF
--- a/core/db_models/EEM_Line_Item.model.php
+++ b/core/db_models/EEM_Line_Item.model.php
@@ -308,7 +308,8 @@ class EEM_Line_Item extends EEM_Base
     /**
      * Deletes line items with no transaction who have passed the transaction cutoff time.
      * This needs to be very efficient
-     * because if there are spam bots afoot there will be LOTS of line items
+     * because if there are spam bots afoot there will be LOTS of line items. Also MySQL doesn't allow a limit when
+     * deleting and joining tables like this.
      *
      * @return int count of how many deleted
      * @throws EE_Error
--- a/core/db_models/EEM_Message.model.php
+++ b/core/db_models/EEM_Message.model.php
@@ -604,8 +604,8 @@ class EEM_Message extends EEM_Base implements EEI_Query_Filter
                     'MSG_modified' => array('<', time() - $time_to_leave_alone)
                 ),
                 'limit' => apply_filters(
-                    2000,
                     'EEM_Message__delete_old_messages__limit',
+                    2000,
                     $delete_threshold
                 )
             )
--- a/core/db_models/EEM_Message.model.php
+++ b/core/db_models/EEM_Message.model.php
@@ -602,6 +602,11 @@ class EEM_Message extends EEM_Base implements EEI_Query_Filter
                 0 => array(
                     'STS_ID' => array('NOT_IN', $message_stati_to_keep),
                     'MSG_modified' => array('<', time() - $time_to_leave_alone)
+                ),
+                'limit' => apply_filters(
+                    2000,
+                    'EEM_Message__delete_old_messages__limit',
+                    $delete_threshold
                 )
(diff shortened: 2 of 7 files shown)


tn3rb:
> is it a problem if $transactions_not_in_progress is only a subset of all the incomplete transactions whose session has expired, instead of the complete list?
> Is it necessary for EED_Ticket_Sales_Monitor::release_reservations_for_tickets) that its list of $valid_reserved_tickets (which is actually line items for ticket purchases) be generated from the complete list of incomplete transactions whose session expired?

yes and yes

why? Because it won't ever reset reserved tickets for anything not returned in that initial query now.

I think in order for this to work properly, we will need to do the following:

- add an offset value to the query in `EEM_Transaction::_get_transactions_in_progress()`
- allow the offset and limit values to be passed in to the above method as well as `get_transactions_not_in_progress()`
- refactor `EED_Ticket_Sales_Monitor::reset_reservation_counts()` so that it runs recursively, passing values for limit and offset to the above until no more records are returned
- likely have to run those queries and reserved count resets in additional requests to avoid timeouts

mnelson4:
Thanks Brent. So as I understand it, if we limited `EEM_Transaction::_get_transactions_in_progress()`, we could only release tickets for the latest 1000 expired transactions, at least until they're deleted (which happens after 1 week.)
So that change wouldn't really work.
If we want me to continue this, I think making `\EED_Ticket_Sales_Monitor::reset_reservation_counts` more efficient (avoiding grabbing whole transaction objects, instead just grab IDs or something; and avoiding retrieving the grand total line items when all we want are the ticket line items) would be the best next step, before trying to do batches etc.
Also, maybe making the cron task run more frequently for high-traffic sites (like every half hour instead of every hour) should halve the amount of work we need to do each time it runs.
Anyways, I've reverted the part of this commit that wasn't ready, but the stuff that is ready I left in here and am passing off for testing

joshfeck:
I tested the AHEE__EE_Cron_Tasks__clean_up_junk_transactions and everything went well. If you wanted to merge what we have so far that will be fine.

Could you also look into the `EEM_Message::instance()->delete_old_messages()` query, maybe that one can have a limit?

mnelson4:
> Could you also look into the EEM_Message::instance()->delete_old_messages() query, maybe that one can have a limit?

Oh sure, I didn't notice that one because it's not scheduled from `EE_Cron_Task` like the others. I set the limit to 2000 because it only runs daily and we also should avoid falling behind the number of messages that are actually sent (as is, I think if there's a big event which sells thousands of tickets in a day, it should just take a few days to catch up and delete all the messages that got generated.)

mnelson4:
On a related ticket, Nazar and I were discussing basically the exact same issue for ES, where I advocating to fetch the OLDEST first... So it makes sense to delete the oldest junk transaction first too.
Sorry, this idea should probably be verified too (even though it's just a couple characters difference in the commit.

mnelson4:
This should get Josh's blessing again before merging (since you last approved it, I added your suggestion of limiting the messages cleanup task, and tweaked the other code too)

joshfeck:
tested message activity log cleanup, it still works

PR #2044, opened 2019-12-16
Created utility to Convert To & From UTC; closes #1943.

### LINKED ISSUE #1943: Use Global Timezone Data Config to Convert To & From UTC
- create centralized file for defining the various Timezone related constants required throughout the system
 - make config available via ConfigContext
 - import into any and all classes that are currently handling date related logic and update to use new config constants as needed.

#### **IMPORTANT**
We want **ALL** of our logic to be operating on dates that are in UTC so that there are never any issues with offsets when performing calculations. Therefore we want to **ALWAYS** do the following two things:
 - convert from UTC to localized time at the absolute last moment prior to display
 - convert from localized time to UTC immediately upon receiving user input and prior to performing any kind of other logic in dates

I've estimated 4 hours for this (plz give feedback if that does not seem to be enough)

nicolad:
Solves: https://github.com/eventespresso/event-espresso-core/issues/1943
With few adjustments:
- It is pulls `timezone` from `ConfigContext`, but I didn't wanted to add to much responsability on that context and extracted the logic into a custom hook.
- Based on [date-fn docs](https://date-fns.org/v2.8.1/docs/Time-Zones) I've used `date-fns-tz` which might make into the `date-fns` core. Check [here](https://github.com/date-fns/date-fns/pull/707).
- As mentioned in the first bullet point I've created custom hook, the intention being to abstract away external dependency and to pull `timezone` from the config, this will make the UI component less agnostic of underlying details.

## Checklist

* [ ] My code is tested.

### DIFF
--- a/assets/prototype/application/hooks/useZonedTime.ts
+++ b/assets/prototype/application/hooks/useZonedTime.ts
@@ -1,0 +1,21 @@
+import { zonedTimeToUtc, utcToZonedTime } from 'date-fns-tz';
+import useConfig from '../../application/services/config/useConfig';
+
+const useZonedTime = () => {
+	const { timezone } = useConfig();
+
+	const zonedTimeToUtcCb = (date: Date | string | number) => {
+		return zonedTimeToUtc(date, timezone);
+	};
+
+	const utcToZonedTimeCb = (isoDate: Date | string | number) => {
+		return utcToZonedTime(isoDate, timezone);
+	};
+
+	return {
+		zonedTimeToUtc: zonedTimeToUtcCb,
+		utcToZonedTime: utcToZonedTimeCb,
+	};
+};
+
+export default useZonedTime;
--- a/assets/prototype/application/services/config/useConfig.ts
+++ b/assets/prototype/application/services/config/useConfig.ts
@@ -1,12 +1,12 @@
 import { useContext } from 'react';
 import { ConfigContext } from '../context/ConfigProvider';
 
-type useConfigReturnType = {
+type ConfigReturnType = {
 	dateFormat: string;
 	timeFormat: string;
 	timezone: string;
 };
 
-const useConfig = (): useConfigReturnType => useContext(ConfigContext);
+const useConfig = (): ConfigReturnType => useContext(ConfigContext);
 
 export default useConfig;
--- a/assets/prototype/application/services/config/useConfig.ts
+++ b/assets/prototype/application/services/config/useConfig.ts
@@ -1,0 +1,12 @@
+import { useContext } from 'react';
+import { ConfigContext } from '../context/ConfigProvider';
+
+type useConfigReturnType = {
+	dateFormat: string;
+	timeFormat: string;
+	timezone: string;
+};
+
+const useConfig = (): useConfigReturnType => useContext(ConfigContext);
+
+export default useConfig;
--- a/assets/prototype/application/hooks/useTimeZoneTime.ts
+++ b/assets/prototype/application/hooks/useTimeZoneTime.ts
@@ -1,7 +1,7 @@
 import { zonedTimeToUtc, utcToZonedTime } from 'date-fns-tz';
-import useConfig from '../../application/services/config/useConfig';
+import useConfig from '../services/config/useConfig';
 
-const useZonedTime = () => {
+const useTimeZoneTime = () => {
 	const { timezone } = useConfig();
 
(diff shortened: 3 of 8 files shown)


assets/prototype/application/hooks/useZonedTime.ts
```diff
@@ -0,0 +1,21 @@
...
+import useConfig from '../../application/services/config/useConfig';
+
+const useZonedTime = () => {
+	const { timezone } = useConfig();
+
+	const zonedTimeToUtcCb = (date: Date | string | number) => {
+		return zonedTimeToUtc(date, timezone);
+	};
+
+	const utcToZonedTimeCb = (isoDate: Date | string | number) => {
```
nicolad: This parameter name is used as seen here:
image
***
[link](https://github.com/marnusw/date-fns-tz#utctozonedtime)

assets/prototype/application/services/config/useConfig.ts
```diff
@@ -0,0 +1,12 @@
+import { useContext } from 'react';
+import { ConfigContext } from '../context/ConfigProvider';
+
+type returnType = {
+	dateFormat: string;
+	timeFormat: string;
+	timezone: string;
+};
```
nicolad: This is a bit of safety for destructuring.

nicolad: image

assets/prototype/application/services/config/useConfig.ts
```diff
@@ -0,0 +1,12 @@
+import { useContext } from 'react';
+import { ConfigContext } from '../context/ConfigProvider';
+
+type useConfigReturnType = {
```
nicolad: Open for a better naming here.

assets/prototype/application/hooks/useZonedTime.ts
```diff
@@ -0,0 +1,21 @@
+import { zonedTimeToUtc, utcToZonedTime } from 'date-fns-tz';
+import useConfig from '../../application/services/config/useConfig';
+
+const useZonedTime = () => {
```
nicolad: The name of the hook is inspired by the `date-fns-tz` function names that are used here.

assets/prototype/application/services/config/useConfig.ts
manzoorwanijk: I think we should avoid using `use` at the beginning of TS types to avoid confusion with React Hooks

nicolad:
The item from the issue requirements:
```
import into any and all classes that are currently handling date related logic and update to use new config constants as needed.
```
needs to be done in a separate PR.

assets/prototype/application/services/config/useConfig.ts
nicolad: Updated, thx.

assets/prototype/application/hooks/useZonedTime.ts
```diff
@@ -0,0 +1,21 @@
...
+		return zonedTimeToUtc(date, timezone);
+	};
+
+	const utcToZonedTimeCb = (isoDate: Date | string | number) => {
+		return utcToZonedTime(isoDate, timezone);
+	};
+
+	return {
+		zonedTimeToUtc: zonedTimeToUtcCb,
+		utcToZonedTime: utcToZonedTimeCb,
```
tn3rb: CNO says:

> how about `localTimeToUtc` && `utcToLocalTime` instead of `zonedTimeToUtc` && `utcToZonedTime` ?
> then you also don't have to append the callbacks with `Cb` nor rename them within the returned object

assets/prototype/application/hooks/useZonedTime.ts
tn3rb: CNO says:

> how about `useTimeZoneTime` ?

only because `useZonedTime` is so awkward to say since `use` ends in a zzzz sound followed immediately by an actual Z, then `Zoned` ends with the plosive `d` sound followed immediately by the plosive `t` sound of Time. Hard to say that without either pausing to get those transitions in properly or having the name come out sounding like 'useOwnTime`

nicolad: Thx, that is a better name. updated :bowing_man:

assets/prototype/application/hooks/useZonedTime.ts
nicolad: I don't know who is this CNO, but he has some great suggestions.
Updated.

PR #2048, opened 2019-12-16
Add Date Comparison & Utility Predicates; closes #2041.

### LINKED ISSUE #2041: Add Date Comparison & Utility Predicates
We are going to need entity specific (datetime & ticket) variations of the following:

- add(d, interval)
- subtract(d, interval)
- difference(d1, d2)
- datesAreEqual(d1, d2)
- sortDates([d1, d2, d3, ...], sort)

I'm open to alternate name suggestions.

These can be located in:
`assets/prototype/domain/shared/predicates/datetimes` && `assets/prototype/domain/shared/predicates/tickets` and be composed of "entity-non-specific" building blocks located either in `assets/prototype/domain/shared/predicates/shared` or a new `assets/prototype/domain/shared/predicates/dateUtilities` if we think we will have enough of these.

The `date-fns` library can be installed as part of this PR

nicolad:
Implements this: https://github.com/eventespresso/event-espresso-core/issues/2041

## Checklist

* [ ] My code is tested.


assets/prototype/domain/shared/predicates/dateUtilities/add-subtract.ts
```diff
@@ -0,0 +1,53 @@
+import {
```
nicolad: Instead of creating 2 files for both `add` and `subtract`, I've decided to take the same approach as in `moment.js`.
The reason being that there are some duplication.
More info: https://github.com/moment/moment/blob/develop/src/lib/moment/add-subtract.js

assets/prototype/application/utilities/converters/number/toInteger.js
```diff
@@ -0,0 +1,13 @@
+export default function toInteger(dirtyNumber) {
```
nicolad: This is the same function as in `date-fns`: https://github.com/date-fns/date-fns/blob/master/src/_lib/toInteger/index.js

assets/prototype/domain/shared/predicates/dateUtilities/add-subtract.ts
```diff
@@ -0,0 +1,53 @@
...
+	const func = addMapping[interval];
+	return func(date, amount);
+};
+
+export const sub = (interval: IntervalType, dirtyDate: Date, dirtyAmount: number) => {
+	const func = addMapping[interval];
+	const amount = toInteger(dirtyAmount);
+	const date = toDate(dirtyDate);
+
+	return func(date, -amount);
```
nicolad: I've used the same approach as in `date-fns`' sub* op: https://github.com/date-fns/date-fns/blob/master/src/subMinutes/index.js#L34

assets/prototype/domain/shared/predicates/dateUtilities/diff.ts
```diff
@@ -0,0 +1,69 @@
...
+	differenceInDays,
+	differenceInHours,
+	differenceInISOWeekYears,
+	differenceInMilliseconds,
+	differenceInMinutes,
+	differenceInMonths,
+	differenceInQuarters,
+	differenceInSeconds,
+	differenceInWeeks,
+	differenceInYears,
```
nicolad: I've added all functions, but we need to decide what we will need.
This will reduce the bundle size:
image

assets/prototype/domain/shared/predicates/dateUtilities/datesAreEqual.ts
```diff
@@ -0,0 +1,7 @@
+import { isEqual } from 'date-fns';
+
+const datesAreEqual = (dateLeft: Date | number, dateRight: Date | number): boolean => {
```
nicolad: I was thinking that since this function lives under `dateUtilities` we might simplify it's name by a similar approach as is `date-fns`
image

nicolad: smth like `areEqual` or `isEqual`

assets/prototype/domain/shared/predicates/dateUtilities/addSub.ts
```diff
@@ -0,0 +1,53 @@
+import {
```
nicolad: The fact that the file is named `dateUtilities/addSub.ts` won't matter that much because we will be using functions from `dateUtilities`, like:
`import { add } from '../../.. ../dateUtilities'`

assets/prototype/domain/shared/predicates/dateUtilities/sortDates.ts
```diff
@@ -0,0 +1,18 @@
+import { compareAsc, compareDesc } from 'date-fns';
+
+type DatesAreEqualType = {
+	dates: Date[];
+	order: 'asc' | 'desc';
+};
+
+const sortDates = ({ dates, order }: DatesAreEqualType) => {
```
nicolad: The same proposal as in `datesAreEqual`, maybe we can use something less redundant?

assets/prototype/domain/shared/predicates/dateUtilities/sortDates.ts
```diff
@@ -0,0 +1,18 @@
...
+	if (order === 'asc') {
+		return dates.sort(compareAsc);
+	}
+
+	if (order === 'desc') {
+		return dates.sort(compareDesc);
+	}
```
nicolad: Here is the same approach as in `date-fns`' docs:
image
*** 
https://date-fns.org/v2.8.1/docs/compareAsc

assets/prototype/domain/shared/predicates/dateUtilities/datesAreEqual.ts
tn3rb: i like that idea... in this case `areEqual` makes the most sense imho
despite it running the risk of making you sound like a pirate when you say the name

assets/prototype/domain/shared/predicates/dateUtilities/sortDates.ts
tn3rb: ya that's fine with me.

assets/prototype/domain/shared/predicates/dateUtilities/sortDates.ts
tn3rb: why not:

```js
return order === 'asc' ? dates.sort(compareAsc) : dates.sort(compareDesc);
```

assets/prototype/domain/shared/predicates/dateUtilities/datesAreEqual.ts
nicolad: Updated.

nicolad: Agree, pronunciation is a tricky domain: https://www.youtube.com/watch?v=Dd7FixvoKBw

assets/prototype/domain/shared/predicates/dateUtilities/sortDates.ts
nicolad: Renamed to just `sort`.
I am thinking if there might be any cases when we have sort dates and tickets at the same time. But as of now, I think that this kind of function is a bit low-level and even if there will be some mixture between dates and tickets we will use some services or other high-level ways.

assets/prototype/application/utilities/converters/number/toInteger.js
nicolad: Updated here: https://github.com/eventespresso/event-espresso-core/pull/2048/commits/f241f5249131d0697de8705454e6e4b2b20571c1

nicolad: And made a sanity check for these `if (dirtyNumber === null || dirtyNumber === true || dirtyNumber === false)`:
image
*** 
all good

nicolad: back to your question: I didn't find an exact or similar equivalent of `toInteger` in Ramda.

assets/prototype/domain/shared/predicates/dateUtilities/addSub.ts
```diff
@@ -0,0 +1,53 @@
...
+};
+
+export const add = (interval: IntervalType, date: Date, amount: number) => {
+	const func = addMapping[interval];
+	return func(date, amount);
+};
+
+export const sub = (interval: IntervalType, dirtyDate: Date, dirtyAmount: number) => {
+	const func = addMapping[interval];
+	const amount = toInteger(dirtyAmount);
```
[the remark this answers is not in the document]
nicolad: This might be the answer: https://github.com/date-fns/date-fns/blob/master/src/subDays/test.js#L19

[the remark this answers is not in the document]
nicolad: `add` function is already doing that internally

assets/prototype/domain/shared/predicates/dateUtilities/datesAreEqual.ts
tn3rb: > Video unavailable
> The uploader has not made this video available in your country.

😢

assets/prototype/domain/shared/predicates/dateUtilities/addSub.ts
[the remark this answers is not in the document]
nicolad: My intention was to avoid importing the equivalent of these:
```
        addDays,
	addHours,
	addISOWeekYears,
	addMilliseconds,
	addMinutes,
	addMonths,
	addQuarters,
	addSeconds,
	addWeeks,
	addYears,
```
***
And when I took a look what [`sub*` functions](https://github.com/date-fns/date-fns/blob/master/src/subDays/index.js#L33) are I realized that we can just swap the addition functions and get what we need.

assets/prototype/domain/shared/predicates/dateUtilities/sortDates.ts
tn3rb: ya i totally agree. tickets and datetimes don't share the same property names for their date fields, so they are going to require different predicates anyways. That said, it's possible that we may see some duplication within those functions and end up extracting some common logic to its own file, but we don't need to be concerned with that now

assets/prototype/domain/shared/predicates/dateUtilities/datesAreEqual.ts
nicolad: That's unacceptable.
Here is the query: https://www.youtube.com/results?search_query=Substitute+Teacher

assets/prototype/domain/shared/predicates/dateUtilities/sortDates.ts
nicolad: yes, that is simpler, updated.

assets/prototype/application/utilities/date/sort.ts
```diff
@@ -0,0 +1,12 @@
+import { compareAsc, compareDesc } from 'date-fns';
+
+type SortProps = {
+	dates: Date[];
```
tn3rb: YEEEEESSSSSSSSsssssss 👍

PR #2058, opened 2019-12-17
Moved EntityTextFieldSearch; closes #2038.

### LINKED ISSUE #2038: Extract Search Text Logic from EDTR v1
EDTR v1 used `useFilterEntitiesBySearchText` in `assets/ZZZ/components/ui/entity-list-filter-bar/use-filter-entities-by-search-text.js` for executing text field searches.

This logic needs to be moved over to the EDTR prototype but should be moved to a more generic non-domain location such as `assets/prototype/application/services` and renamed to something like `EntityTextFieldSearch` (other ideas???).

As well, it would be beneficial to refactor the logic in that function as follows:

- receive an array of entity fields to search instead of just one
- detect the field value type and react accordingly, for example: detecting date fields and comparing `toString()` versions of the date against search terms
- adding TS interface for props
- adding a tests folder and unit test to validate logic using an array of varied search terms run in a loop

nicolad:
Partially implements: https://github.com/eventespresso/event-espresso-core/issues/2038

## Checklist

* [ ] My code is tested.


assets/prototype/application/services/entityTextFieldSearch/index.ts
```diff
@@ -0,0 +1,28 @@
...
+ * matches the current search text (fully or partially)
+ *
+ * @param {BaseEntity[]} entities
+ * @param {string} searchText
+ * @param {string} field
+ * @return {BaseEntity[]} entities matching search criteria
+ */
+const entityTextFieldSearch = ({ entities, searchText }: EntityTextFieldSearchProps) => {
+	const isSearchable = searchText && Array.isArray(entities);
+	const searchResult = entities.filter(({ name: title }) => {
```
nicolad: As discussed with @tn3rb it will search only through titles on the first iteration.
After proto will add more searchable fields.

assets/prototype/application/services/entityTextFieldSearch/index.ts
```diff
@@ -0,0 +1,28 @@
...
+ * searches through provided list of entities and returns an array
+ * of entities where the set field (defaults to entity name)
+ * matches the current search text (fully or partially)
+ *
+ * @param {BaseEntity[]} entities
+ * @param {string} searchText
+ * @param {string} field
+ * @return {BaseEntity[]} entities matching search criteria
+ */
+const entityTextFieldSearch = ({ entities, searchText }: EntityTextFieldSearchProps) => {
```
nicolad: Made it camelCase since it's not a component.

assets/prototype/application/services/entityTextFieldSearch/index.ts
```diff
@@ -0,0 +1,28 @@
+type EntityTextFieldSearchProps = {
+	entities: any[];
+	searchText: string;
+	field?: string;
```
nicolad: left `field` for future iterations

nicolad:
For future references, if we'd have to use memoization, we might use the same approach as here: 
```
// ...
import { createSelector } from 'reselect';

const todoSelector = state => state.todo.todos;
const searchTermSelector = state => state.todo.searchTerm;

export const filteredTodos = createSelector(
  [todoSelector, searchTermSelector],
  (todos, searchTerm) => {
    return todos.filter(todo => todo.title.match(new RegExp(searchTerm, 'i')));
  }
);

// ...
```
from: https://alligator.io/redux/reselect/

assets/prototype/application/services/entityTextFieldSearch/index.ts
```diff
@@ -0,0 +1,28 @@
...
+ *
+ * @param {BaseEntity[]} entities
+ * @param {string} searchText
+ * @param {string} field
+ * @return {BaseEntity[]} entities matching search criteria
+ */
+const entityTextFieldSearch = ({ entities, searchText }: EntityTextFieldSearchProps) => {
+	const isSearchable = searchText && Array.isArray(entities);
+	const searchResult = entities.filter(({ name: title }) => {
+		const textToSearch = searchText.toLowerCase();
```
tn3rb: this should be moved outside of the filter callback since it would remain the same for all entities and therefore does not have to be recalculated on every iteration

nicolad: yes, I was too focused on micro-optimizations and lost the big picture :D
thx :bowing_man:

PR #2059, opened 2019-12-17
Migrations: Don't check for EE shortcodes in attachments or menu items too

joshfeck:

There was a support request where the user was trying to migrate from EE3 to EE4 and this was the error in the migration:

```
[_feedback_message] => Migrated 50 records successfully during Shortcodes
[_priority] => 10
[_records_to_migrate] => 83926
[_records_migrated] => 105
[_status] => status_fatal_error
[_pretty_name] => Data Migration from Event Espresso 3 to Event Espresso 4.1.0
```

The site in question has no WP posts, and has about 30 WP pages (along with some custom post types). The bulk of the shortcode records to migrate were attachment posts. This PR will help make the shortcode part of the migration be more efficient because we know EE shortcodes will not be in attachment posts or nav menu items. 
## How has this been tested

I tested an EE3 to EE4 migration locally using the site's _posts table imported onto a local site. The shortcode part of the migration took much less time with this branch.
## Checklist

* [x] My code is tested.

### DIFF
--- a/core/data_migration_scripts/4_1_0_stages/EE_DMS_4_1_0_shortcodes.dmsstage.php
+++ b/core/data_migration_scripts/4_1_0_stages/EE_DMS_4_1_0_shortcodes.dmsstage.php
@@ -54,6 +54,6 @@ class EE_DMS_4_1_0_shortcodes extends EE_Data_Migration_Script_Stage
 
     private function _sql_to_only_select_non_drafts()
     {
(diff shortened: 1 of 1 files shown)


joshfeck:
Yeah CPTs are probably an edge case, it's likely very few if any are using the original EE event list shortcode within a CPT. That said it's possible to have a post type that's essentially a "page" and use shortcodes on it.

Issue #2047, opened 2019-12-16
Optimize \EED_Ticket_Sales_Monitor::reset_reservation_counts


mnelson4:
## Issue Overview

On https://github.com/eventespresso/event-espresso-core/issues/2035#issuecomment-565579431 we noticed that `\EED_Ticket_Sales_Monitor::reset_reservation_counts` can be a major bottleneck (especially with a site with a spike in traffic). Aside from breaking it up into more bitesize chunks, it's probably worthwhile to try to make the function a bit more efficient. (Probably by avoiding fetching unnecessary intermediate data from the DB, and possibly by avoiding instantiating model objects if possible.)

## Bug report or feature request?

* [ ] Bug
* [ ] Feature
* [ ] Neither
* [x] Optimization

mnelson4:
The New Relic data is helpful: https://rpm.newrelic.com/accounts/1628382/applications/62114495/transactions#id=5b225765625472616e73616374696f6e2f5572692f68747470646f63732f77702d63726f6e2e706870222c22225d&app_trace_id=e8ba1c40-1f9e-11ea-bd0d-0242ac110008_0_9286

It shows that the slow cron task requests are primarily being eaten up by `EEH_Line_Item::_get_descendants_by_type_and_object_type` and `EED_Ticket_Sales_Monitor::reset_reservation_counts`...

mnelson4:
Oh man, I may have just struck bug-finding gold: I am successfully overselling tickets :money_mouth_face: !
Here's a video: https://drive.google.com/file/d/1emSMmhy1sSPhvFaL7w6cvDT4U4MqaJnM/view
Here's the steps:
1. Create an event, set a limit to 10 on its ticket
2. Register 10 spots for the event in Firefox, but don't actually answer any reg details. Keep the tab open though. Notice the ticket's "Rsvd" count is now at 10.
3. Execute the cron task `AHEE__EE_Cron_Tasks__clean_up_junk_transactions` (I use the WP CLI command `wp cron event run AHEE__EE_Cron_Tasks__clean_up_junk_transactions`)
4. Notice that the event's "Rsvd" count is back to 0 (even though there is an active transaction for 10 of them!!!)
5. In Google Chrome, go to the event and register another 10 spots. Complete the registration.
6. Complete the registration you started in Firefox. 
7. Notice the event now has 20 tickets sold, despite having a limit of 10 on the ticket ☠️ ☠️ ☠️

tn3rb:
[...]

PR #2060, opened 2019-12-17
Don't release ticket reservations for transactions that are in…

mnelson4:

If the cron task `AHEE__EE_Cron_Tasks__clean_up_junk_transactions` ran while a transaction was in progress and had tickets reserved, those reservations were getting released, would could lead to overselling. See https://github.com/eventespresso/event-espresso-core/issues/2047#issuecomment-566655969

I vaguely remember that over the past few years we've had some overselling issues. The problem this is fixing may have been the cause of them. (Specifically, that there were reservations on the ticket for registrations in progress, then the cron task would run and release the ticket reservations, and before the registrations finished up, someone else would start to register.)

As far as internal details of what changed, it's a bit confusing...
The logic in `\EED_Ticket_Sales_Monitor::reset_reservation_counts` seems to have gotten confused at some point.
There were two problems:

### 1. PHP array union overwrites based on array keys
The list of `$valid_reserved_tickets` was being built using `$valid_reserved_tickets += EED_Ticket_Sales_Monitor::get_ticket_line_items_for_grand_total($total_line_item            );`. That operator `+=` union'd `$valid_reserved_tickets` with `EED_Ticket_Sales_Monitor::get_ticket_line_items_for_grand_total()`'s results based on array **keys**. And the array was indexed **numerically**. So if `$valid_reserved_tickets` had one item, its array key was `0`. If `EED_Ticket_Sales_Monitor::get_ticket_line_items_for_grand_total()` returned an array which also had a single item, its key was also `0`. So when they were union'd, the second array's `0` was ignored.
> The + operator returns the right-hand array appended to the left-hand array; for keys that exist in both arrays, the elements from the left-hand array will be used, and the matching elements from the right-hand array will be ignored. 
from https://www.php.net/manual/en/language.operators.array.php

So, even if there were 400 `$transactions_not_in_progress`, the `$valid_reserved_tickets` derived from them was usually just a single item.

The fix was to instead index the array using the line item's ID.

### 2. Fetching Expired Transactions instead of In-Progress Ones
`\EED_Ticket_Sales_Monitor::reset_reservation_counts` was fetching the list of expired transactions (which could never complete, and so their reservations on tickets could be released) but then it considered their tickets' reservations as "valid" by adding them to `$valid_reserved_tickets` 😵. That seems to be the opposite of what we want: the transactions are **expired**, and so their ticket reservations are **no longer** valid. 

The fix was to do the opposite: fetch the list of transactions that were still **in-progress** (ie, their session hadn't yet expired) and could be completed. Those are the transactions whose ticket reservations are still valid because they could still complete their registrations.

So in summary, on this branch, we are fetching all the in-progress transactions, and building the list of `$valid_reserved_tickets` from that, indexing the items by the ticket line item's ID in the array (rather than numerically) in order for the array `+=` operator to work as expected.

### Regarding efficiency
I originally dug into this code to see if there was a way to optimize it further. I got sidetracked fixing this bug, but I think this may also help significantly with efficiency too.
Before we were fetching the list all expired transactions (which hadn't yet been deleted). So that could have been a week's worth of transactions being fetched on every execution of the cron job. On some sites, there were hundreds or even thousands of these.
With these changes, we're instead only grabbing all the in-progress transactions. While on some sites this number can also be high, it's at least only an hour's worth of transactions instead of a week's.
So I'm guessing this means that where before we were fetching hundreds or thousands of transactions everytime the cron ran, we'll now instead be fetching dozens, so the load will be alleviated.

## How has this been tested

Besides the new unit test, 
* [x] Create an event, set a limit to 10 on its ticket
* Register 10 spots for the event in Firefox, but don't actually answer any reg details. Keep the tab open though. Notice the ticket's "Rsvd" count is now at 10.
* Execute the cron task AHEE__EE_Cron_Tasks__clean_up_junk_transactions (I use the WP CLI command wp cron event run AHEE__EE_Cron_Tasks__clean_up_junk_transactions)
* Notice that the event's "Rsvd" count is back to 0 (even though there is an active transaction for 10 of them!!!)
* In Google Chrome, go to the event and register another 10 spots. Complete the registration.
* Complete the registration you started in Firefox.
* Notice the event now has 20 tickets sold, despite having a limit of 10 on the ticket skull_and_crossbones skull_and_crossbones skull_and_crossbones. Here's a video of that on master: drive.google.com/file/d/1emSMmhy1sSPhvFaL7w6cvDT4U4MqaJnM/view
* [ ] Create another event (or two) with various tickets. Make a few registrations and have a few registrations in progress, so some of the tickets have tickets reserved. 
* Go to the maintenance -> "Reset/Delete Data" and click "Reset Ticket and Datetime Reserved Counts". The tickets' reserved counts should NOT change (on master, they'd all get set back to 0.)
* [x] Create a ticket with a limit of 10. Begin 10 registrations for that ticket, but don't answer any reg questions. So, the ticket's fully reserved.
* In another browser, try to register for that ticket. You should be prevented from doing so.
* In the DB, manually change that first transaction's `TXN_timestamp` to anything older than a week, and then run the cron task again.
* The ticket's reserved count should be set back to 0

## Checklist

* [x] My code is tested.

### DIFF
--- a/modules/ticket_sales_monitor/EED_Ticket_Sales_Monitor.module.php
+++ b/modules/ticket_sales_monitor/EED_Ticket_Sales_Monitor.module.php
@@ -990,9 +990,6 @@ class EED_Ticket_Sales_Monitor extends EED_Module
         array $valid_reserved_ticket_line_items = array(),
         $source
     ) {
-        if (self::debug) {
-            echo self::$nl . self::$nl . __LINE__ . ') ' . __METHOD__ . '()';
-        }
         $total_tickets_released = 0;
         $sold_out_events = array();
         foreach ($tickets_with_reservations as $ticket_with_reservations) {
@@ -1005,19 +1002,11 @@ class EED_Ticket_Sales_Monitor extends EED_Module
             // the number to RELEASE. It's the same end result, just different path.
             // Begin by assuming we're going to release all the reservations on this ticket.
             $num_tix_for_expired_reservations = $ticket_with_reservations->reserved();
-            if (self::debug) {
-                echo self::$nl . ' . $ticket_with_reservations->ID(): ' . $ticket_with_reservations->ID();
-                echo self::$nl . ' . $reserved_qty: ' . $num_tix_for_expired_reservations;
-            }
             // Now reduce that number using the list of current valid reservations.
             foreach ($valid_reserved_ticket_line_items as $valid_reserved_ticket_line_item) {
                 if ($valid_reserved_ticket_line_item instanceof EE_Line_Item
                     && $valid_reserved_ticket_line_item->OBJ_ID() === $ticket_with_reservations->ID()
                 ) {
-                    if (self::debug) {
-                        echo self::$nl . ' . $valid_reserved_ticket_line_item->quantity(): '
-                             . $valid_reserved_ticket_line_item->quantity();
-                    }
                     $num_tix_for_expired_reservations -= $valid_reserved_ticket_line_item->quantity();
                 }
             }
@@ -1036,9 +1025,6 @@ class EED_Ticket_Sales_Monitor extends EED_Module
                     $sold_out_events[] = $event;
                 }
             }
-        }
-        if (self::debug) {
-            echo self::$nl . ' . $total_tickets_released: ' . $total_tickets_released;
         }
         // Double check whether sold out events should remain sold out after releasing tickets
         if ($sold_out_events !== array()) {
--- a/modules/ticket_sales_monitor/EED_Ticket_Sales_Monitor.module.php
+++ b/modules/ticket_sales_monitor/EED_Ticket_Sales_Monitor.module.php
@@ -978,8 +978,12 @@ class EED_Ticket_Sales_Monitor extends EED_Module
      * whose transactions aren't complete and also aren't yet expired (ie, they're incomplete and younger than the
      * session's expiry time) to update the ticket (and their datetimes') reserved counts.
      *
-     * @param EE_Ticket[]    $tickets_with_reservations
-     * @param EE_Line_Item[] $valid_reserved_ticket_line_items
+     * @param EE_Ticket[]    $tickets_with_reservations all tickets with TKT_reserved > 0
+     * @param EE_Line_Item[] $valid_reserved_ticket_line_items all line items for tickets and incomplete transactions
+     *                       whose session has NOT expired. We will use these to determine the number of ticket
+     *                       reservations are now invalid. We don't use the list of invalid ticket line items because
+     *                       we don't know which of those have already been taken into account when reducing ticket
+     *                       reservation counts, and which haven't.
      * @return int
      * @throws UnexpectedEntityException
      * @throws DomainException
--- a/modules/ticket_sales_monitor/EED_Ticket_Sales_Monitor.module.php
+++ b/modules/ticket_sales_monitor/EED_Ticket_Sales_Monitor.module.php
@@ -985,7 +985,7 @@ class EED_Ticket_Sales_Monitor extends EED_Module
      * @throws DomainException
      * @throws EE_Error
      */
-    private static function release_reservations_for_tickets(
+    protected static function release_reservations_for_tickets(
         array $tickets_with_reservations,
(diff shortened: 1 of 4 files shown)


mnelson4:
Update: In Slack, Brent and I discussed a little. Brent said
> i think i probably should not have named that var `$valid_reserved_tickets` because that gets subtracted from the total of reserved tickets
> ie: those get released (edited) 
> look at `\EED_Ticket_Sales_Monitor::release_reservations_for_tickets()`
> you've got this:
> `$reserved_qty = $ticket_with_reservations->reserved();`
> and then this:
> `$reserved_qty -= $valid_reserved_ticket_line_item->quantity();`
> ya it was bad naming on my behalf

Which led to me think the above changes weren't correct...

BUT, when I investigated more, I saw

```
$ticket_with_reservations->decreaseReserved($reserved_qty,...);
```

SO, `$reserved_qty` isn't the number of valid reserved spots we're setting `TKT_reserved` TO, it's the number of reserved spots WE'RE RELEASING. The opposite.
So there's a double-negative here: `$valid_reserved_ticket_line_item->quantity()` is reducing the number to release.
So, "`$valid_reserved_tickets` gets subtracted from the total of reserved tickets" isn't correct; if there are no `$valid_reserved_tickets`, then all that ticket's reservations get released. But if there are `$valid_reserved_tickets`, we reduce the number to release (or if there are enough valid reserved tickets, we don't release any).

SO, I'm really sorry, I know this is as frustrating as it is confusing, these changes again appear to be correct. I added a test to verify `\EED_Ticket_Sales_Monitor::release_reservations_for_tickets` worked as I expected (namely, that `$valid_reserved_ticket_line_items`'s name is actually true to its purpose, as all other tickets reservations get released unless they have valid reserved ticket line items), and some comments to try to justify why the code works.

Lastly, I noticed an unnecessary save happening in `\EED_Ticket_Sales_Monitor::release_reservations_for_tickets` (we just want to update the ticket's `TKT_reserved` count, which `\EE_Ticket::decreaseReserved` already does) so I removed it.

modules/ticket_sales_monitor/EED_Ticket_Sales_Monitor.module.php
```diff
@@ -960,70 +964,72 @@ private static function get_ticket_line_items_for_grand_total(EE_Line_Item $tota
...
-            if (self::debug) {
-                echo self::$nl . ' . $ticket_with_reservations->ID(): ' . $ticket_with_reservations->ID();
-                echo self::$nl . ' . $reserved_qty: ' . $reserved_qty;
-            }
+            // The $valid_reserved_ticket_line_items tells us what the reserved count on their tickets (and datetimes)
+            // SHOULD be. Instead of just directly updating the list, we're going to use EE_Ticket::decreaseReserved()
+            // to try to avoid race conditions, so instead of just finding the number to update TO, we're going to find
+            // the number to RELEASE. It's the same end result, just different path.
+            // Begin by assuming we're going to release all the reservations on this ticket.
+            $num_tix_for_expired_reservations = $ticket_with_reservations->reserved();
```
tn3rb: how about just `$expired_reservations` or `$expired_reservations_count` or `$expired_reserved_qty`???

mnelson4: Yeah good suggestions IMO

joshfeck:
I tested this and it looks good to go. 

I also tested what would happen after one hour when the transaction's AHEE__EE_Cron_Tasks__expired_transaction_check was fired and found that the reserved counts were cleared, but that I could still complete the transaction as it seemed to still have an open session. Which could lead to overselling if someone else happened to register in between the time the AHEE__EE_Cron_Tasks__expired_transaction_check was run and the other transaction was completed. This is also an issue on master, so not a regression due to the work on this branch.

mnelson4:
> I tested this and it looks good to go.

Thanks, merged!

> I also tested what would happen after one hour when the transaction's AHEE__EE_Cron_Tasks__expired_transaction_check was fired and found that the reserved counts were cleared, but that I could still complete the transaction as it seemed to still have an open session

Hmmmm ya I'll make an issue for that...

Issue #1959, opened 2019-12-04
Add type safety to useRelationsManager actions


nicolad:
By adding TS types we need to ensure that we are not passing wrong data and we have all the required data for the following actions:
- addRelation
- updateRelations
- removeRelation
- dropRelations
***
We can go as simple as I did in this PR:
https://github.com/eventespresso/event-espresso-core/pull/1958
by just verifying the function arguments' types.

nicolad:
I've added there as `type`, but it can also be a TS `interface`.
For more info: https://pawelgrzybek.com/typescript-interface-vs-type/
![image](https://user-images.githubusercontent.com/1477453/70234814-23093680-176a-11ea-8bcc-bf7106cb9d80.png)

PR #2061, opened 2019-12-17
Typefy relations manager

manzoorwanijk:
I worked on this PR offline :)
Completes #1959 

## Checklist

* [ ] My code is tested.

### DIFF
--- a/assets/prototype/application/services/apollo/relations/types.ts
+++ b/assets/prototype/application/services/apollo/relations/types.ts
@@ -1,0 +1,38 @@
+interface CommonProps {
+	relation?: string;
+	relationId?: string;
+	relationIds?: string[];
+}
+export interface RelationFunctionProps extends CommonProps {
+	entity: string;
+	entityId: string;
+}
+export interface RelationAction extends CommonProps {
+	type: string;
+	entity?: string;
+	entityId?: string;
+	data?: any;
+}
+type PossibleRelation = {
+	datetimes?: string[];
+	tickets?: string[];
+	prices?: string[];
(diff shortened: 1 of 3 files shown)


assets/prototype/application/services/apollo/relations/useRelationsManager.ts
```diff
@@ -1,22 +1,22 @@
 import { useReducer, useEffect } from '@wordpress/element';
-import get from 'lodash/get';
-import set from 'lodash/set';
-import cloneDeep from 'lodash/cloneDeep';
+import { pathOr, assocPath, dissocPath, clone } from 'ramda';
```
tn3rb: 👍

assets/prototype/application/services/apollo/relations/types.ts
```diff
@@ -0,0 +1,38 @@
...
+	entity?: string;
+	entityId?: string;
+	data?: any;
+}
+type PossibleRelation = {
+	datetimes?: string[];
+	tickets?: string[];
+	prices?: string[];
+	priceTypes?: string[];
+};
```
tn3rb: not sure what to think about this... but it seems like it would be tricky to solve in a more exact manner and I certainly can not think of a solution

nicolad: this sounds like we need to track this specific case via a tech debt issue, I've created that:
https://github.com/eventespresso/event-espresso-core/issues/2064
***
if you think that is obsolete or too much than that issue can be removed

PR #2065, opened 2019-12-18
Moved filterTicketEntities; closes #2042.

### LINKED ISSUE #2042: Extract TicketsList Filter Logic from EDTR v1
EDTR v1 used a variety of functions founds in `assets/ZZZ/editor/events/tickets/editor-ticket/filter-bar/tickets-list-entity-filter-utils.js` for filtering the tickets list.

This logic needs to be moved over to the EDTR prototype but should have as much of its logic as possible extracted and moved to `assets/prototype/domain/shared/predicates/tickets` and broken down as granularly as possible and then recomposed into logical blocks that can be shared throughout the application.

As well, plz:

- add TS interface for props for any components
- add a tests folder and unit test to validate filter logic using an array of varied datums run in a loop

nicolad:
Implements: https://github.com/eventespresso/event-espresso-core/issues/2042

## Checklist

* [ ] My code is tested.


assets/src/application/entities/ticket/isOnSale.ts
```diff
@@ -0,0 +1,17 @@
...
+ */
+import { differenceInMinutes } from 'date-fns';
+
+interface IisOnSale {
+	startDate: Date;
+	endDate: Date;
+}
+
+const isOnSale = ({ startDate, endDate }: IisOnSale): boolean => {
+	const now = new Date();
```
tn3rb: this could be moved outside of this function since this could potentially be called on every ticket in a list, but we don't need to recalculate "now" for every iteration

assets/src/application/entities/ticket/isPending.ts
```diff
@@ -0,0 +1,17 @@
...
+
+/**
+ * @function
+ * @param {Object} ticketEntity model object
+ * @param {boolean} includeArchived if true will not filter out archived entities
+ * @return {boolean} 	true if ticket is not yet available for purchase,
+ * 						but will be at some date in the future
+ */
+const isPending = ({ startDate }) => {
+	return differenceInMinutes(startDate, new Date()) > 0;
```
tn3rb: same deal, "now" can be calculated once outside of the function

assets/src/domain/shared/predicates/tickets/filterTicketEntities/index.ts
```diff
@@ -0,0 +1,73 @@
...
+/**
+ * Internal dependencies
+ */
+import { isOnSale, isPending } from '../../../../../application/entities/ticket';
+import nextOnSaleOrPendingOnly from './nextOnSaleOrPendingOnly';
+import percentSoldAtOrAbove from './percentSoldAtOrAbove';
+import soldOutOnly from './soldOutOnly';
+
+interface FilterTicketEntities {
+	ticketEntities: any[];
```
tn3rb: plz remove "Entity" or "Entities" from any prop names like this.
That was done in EDTRv1 because I initially wanted to differentiate between EE datetimes and JS Dates via some kind of namespace / naming convention.
My preference was to use `eventDate` or something like that to **only** mark EE Datetime objects but the other primary dev at the time came up with the idea of appending "Entity" or "Entities" to **ALL** of our object types and would not ever back down because their ideas were always superior. So to just move forwards, I acquiesced and we were stuck with having to add those suffixes to the names for ALL of our props. In these kinds of cases I think it's more sensible to add the extra namespacing to the item that is used the least in order to reduce on typing, unnecessary code, and unneeded "noise" in the code. But logic doesn't always overrule self gratification with some people. Can you tell I'm still bitter about things??? 😆 

Further to the above, although I like `eventdate` because it is clear as to what the item is (a date that is specific to an event), it causes confusion when you start trying to apply some kind of naming convention to relations involving events and datetimes (ie: `eventEventDate` ???) so I'm ok with using `Datetime` to denote an EE Datetime object and just `Date` to indicate a JS Date object, although I also often consider using `JsDate` to namespace those more clearly.

assets/src/domain/shared/predicates/tickets/filterTicketEntities/validInfiniteQuantity.ts
```diff
@@ -0,0 +1,16 @@
...
+ */
+import { differenceInMinutes } from 'date-fns';
+
+/**
+ * @param {Object} ticketEntity    event ticket object
+ * @return {boolean} true if qty property is valid and unlimited
+ */
+const validInfiniteQuantity = (ticketEntity) => {
+	return (
+		isModelEntityOfModel(ticketEntity, 'ticket') && (ticketEntity.qty === 'INF' || ticketEntity.qty === Infinity)
```
tn3rb: TS will remove the need to be doing this all over the place:  `isModelEntityOfModel(ticketEntity, 'ticket')`

nicolad:
> Looking great so far!!! 
> 
> #### My comments/observations:
> Everything in `assets/src/application/entities/ticket/` needs to be moved into the `domain` folder since anything related to tickets is EE business domain logic. The application folder is for logic that could be shared between multiple applications and does not connect directly with anything external to the app (like twitter). So for example, "ticket" logic would not make any sense in an application that did not involve tickets in any way.
> 
> As well, since this logic pertains to tickets regardless of the scope (ie, admin vs event editor vs frontend, etc) it should go into the "shared" folder:
> 
> ```
> assets/prototype/domain/shared/tickets/
> ```
> 
> or maybe
> 
> ```
> assets/prototype/domain/shared/entities/tickets/
> ```
> 
> or maybe even
> 
> ```
> assets/prototype/domain/shared/predicates/tickets/
> ```
> 
> since that last folder already exists, think I like that last option the most since it locates this code with the other predicates you have added.
> 
> Further on in the review I noticed your adding files to this folder:
> 
> ```
> assets/src/domain/shared/predicates/tickets/filterTicketEntities/
> ```
> 
> but similar to your comments on #2048 regarding using the folder names as part of the namespacing for files, we could maybe change this to just:
> 
> ```
> assets/src/domain/shared/predicates/tickets/filters/
> ```

Yup, great suggestion, updated.

assets/src/application/entities/ticket/isOnSale.ts
nicolad: moved out: https://github.com/eventespresso/event-espresso-core/pull/2065/commits/93c6687ba8294909fa36694945cdf4f3ee8841d2

assets/prototype/domain/shared/predicates/tickets/filters/index.ts
```diff
@@ -0,0 +1,56 @@
...
+import expiredOnly from './expiredOnly';
+import nextOnSaleOrPendingOnly from './nextOnSaleOrPendingOnly';
+import onSaleAndPending from './onSaleAndPending';
+import onSaleOnly from './onSaleOnly';
+import pendingOnly from './pendingOnly';
+import percentSoldAtOrAbove from './percentSoldAtOrAbove';
+import percentSoldBelow from './percentSoldBelow';
+import soldOutOnly from './soldOutOnly';
+
+export const now = new Date();
```
nicolad: When coupled with UI will check if this isn't lifted too high in the hierarchy.

assets/prototype/domain/shared/predicates/tickets/filters/nextOnSaleOrPendingOnly.ts
```diff
@@ -0,0 +1,20 @@
...
+ */
+import { isOnSale, isPending } from '../index';
+import sortTicketEntitiesList from './sortTicketEntitiesList';
+
+const nextOnSaleOrPendingOnly = (ticketEntities: any[]) => {
+	const selectedPredicates = [isOnSale, isPending];
+	const onSaleAndPending = filter(allPass(selectedPredicates))(ticketEntities);
+	const sortedItems = sortTicketEntitiesList(onSaleAndPending);
+
+	return [head(sortedItems)];
```
nicolad: I am not very happy with the readability here.
On the first iteration, I've just mimicked the old implementation.
On the next iteration, when coupled with UI, this needs to be improved/ simplified.

assets/prototype/domain/shared/predicates/tickets/filters/percentSoldAtOrAbove.ts
```diff
@@ -0,0 +1,24 @@
...
+/**
+ * @param {Object} ticketEntity    event ticket object
+ * @param {number} maxQuantity
+ * @return {boolean} true if sold/qty >= maxQuantity
+ */
+const percentSoldAtOrAbove = ({ maxQuantity, ticketEntities }: PercentSoldAtOrAboveProps) => {
+	const filterFn = (ticketEntity) => {
+		return (
+			validFiniteQuantity(ticketEntity) &&
+			parseInt(ticketEntity.sold, 10) / parseInt(ticketEntity.qty, 10) >= maxQuantity / 100
```
nicolad: Same here, not :100: happy with these.
I need to come back on the next iteration and rethink this.

assets/prototype/domain/shared/predicates/tickets/filters/index.ts
tn3rb: ya I initially thought we should out that somewhere global, like maybe on the config, 
but don't want to have to drill down for something so basic

assets/prototype/domain/shared/predicates/tickets/filters/archivedOnly.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ * Internal dependencies
+ */
+import isArchived from '../isArchived';
+
+const archivedOnly = (ticketEntities: any[]) => {
+	return ticketEntities.filter((ticketEntity) => isArchived(ticketEntity));
```
tn3rb:
```suggestion
	return tickets.filter((ticket) => isArchived(ticket));
```

would need to adjust the function declaration too

assets/prototype/domain/shared/predicates/tickets/filters/expiredOnly.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ * Internal dependencies
+ */
+import isExpired from '../isExpired';
+
+const expiredOnly = (ticketEntities: any[]) => {
+	return ticketEntities.filter((ticketEntity) => isExpired(ticketEntity));
```
tn3rb:
```suggestion
	return tickets.filter((ticket) => isExpired(ticket));
```

would need to adjust the function declaration too

assets/src/domain/shared/predicates/tickets/filterTicketEntities/index.ts
nicolad: `would not ever back down because their ideas were always superior` :man_facepalming:

nicolad: >  Can you tell I'm still bitter about things???
I think what you say makes sense. And those rules aren't very helpful, with the same success tickets would have been named `ticketItems` which is the same mouthful and useless `ticketEntities` which isn't enhancing understanding in any way.

assets/prototype/domain/shared/predicates/tickets/isPending.ts
```diff
@@ -0,0 +1,18 @@
+/**
+ * External dependencies
+ */
+import { differenceInMinutes } from 'date-fns';
+import { now } from './filters';
+
+/**
+ * @function
+ * @param {Object} ticket model object
+ * @param {boolean} includeArchived if true will not filter out archived entities
```
tn3rb: this jsDoc can be removed since the parameter does not exist

assets/prototype/domain/shared/predicates/tickets/filters/sortTicketEntitiesList.ts
```diff
@@ -0,0 +1,19 @@
...
+		case 'by-name':
+			return sortWith([ascend(prop('name'))], tickets);
+		case 'by-id':
+			return sortWith([ascend(prop('id'))], tickets);
+		case 'by-order':
+			return sortWith([ascend(prop('order'))], tickets);
+	}
+};
+
+export default sortTicketsList;
```
tn3rb: this file should be renamed to match the function

nicolad: renamed, thx

assets/prototype/domain/shared/predicates/tickets/isPending.ts
nicolad: updated, thx

Issue #2018, opened 2019-12-11
Storing and loading model object traversal state


mnelson4:
## Issue Overview

On https://github.com/eventespresso/event-espresso-core/pull/1977 I add `\EventEspresso\core\services\orm\tree_traversal\ModelObjNode`, which takes a model object, and then recursively discovers its dependent model objects (ie, model objects that have a foreign key to it). It can take a break mid-search and then pickup where it left off (its parent class has a method `\EventEspresso\core\services\orm\tree_traversal\BaseNode::visit` which accepts `$model_objects_to_identify`, which is a count of how many objects to identify in this batch). So far, that's been shown to work, provided we do all the discovering during a SINGLE HTTP request, so the thing can be stored in-memory.
In order to complete https://github.com/eventespresso/event-espresso-core/issues/1904, we need to be able to do that across MULTIPLE HTTP requests, in order to support doing it via a batch process. So the question is: how to persist the state of how far along we are in the model object tree traversal/discovery across HTTP requests?

The easiest way would be to just serialize the object and dump it into a WP option. But that has blown up for us in the past, and it that might not be efficient enough to work...

The trouble is we're going to inevitably be dealing with some events with thousands of registrations, which, if we don't store the data compactly, could easily get too big for the WP options table (yes, according to MySQL, you should be able to store GBs of data in each cell; but shared hosts are limited to sending about 1 MB of data in each insertion/update query, so the effective limit is actually around 1 MB; we've encountered this limit while doing migrations which is why migrations use multiple options, one for each `EE_Data_Migration_Script_Stage`.)

## The current approach: Recursive ModelObjNodes

I had assumed I'd use the `\EventEspresso\core\services\orm\tree_traversal\ModelObjNode::toArray` to convert the `ModelObjNode`s into an array, which can be JSON-encoded into a WordPress option; then have a static method (or a factory) for generating the objects from that JSON string on the subsequent requests.

When we're keeping track of thousands of model objects, converting the tree of `ModelObjNode`s into an array then into a JSON string, sending it to the DB; then on the next request, turning that JSON string into an array, then that array into a plethora of objects (one `ModelObjNode` and several `RelationNode`s for each model object), I'm worried that might use up a lot of memory and CPU.

So really, the problem isn't just "how to persist state" between requests, but how to do so **efficiently** enough to handle an event with thousands of registrations.

## Alternative: a single ModelObjDiscoverer class
[...]

PR #2101, opened 2019-12-27
Make model object tree traversal progress saveable to the DB

mnelson4:

Allows us to store the model tree traversal's state to the DB so it can span multiple HTTP requests for the batch system.

* [x] Create a DTO which is just a wrapper around a very easily JSON-encodeable PHP array
* [x] Create a tree traversal managager class (see https://github.com/eventespresso/event-espresso-core/issues/2018#issuecomment-564728286)
* [ ] keep discovering, don't just stop at 1 (basically, convert recursive algorithm to iterative one)
* [ ] factor in the number-to-discover
* [ ] Use business classes that know how to plan and execute tree traversal (corresponding to the now-old relations and model object nodes)
* [ ] re-test that we can traverse a model object tree in a single request still
* [ ] test that we can fully serialize the tree traversal state and then continue traversal on another request

## How has this been tested

* [ ] The existing functionality should still work: you should be able to fully delete an event and its dependent data (eg registrations) in a single request. Implementing the batch job will be another ticket.

## Checklist

* [ ] My code is tested.

### DIFF
--- a/core/services/orm/tree_traversal/ModelObjNodeTreeDTO.php
+++ b/core/services/orm/tree_traversal/ModelObjNodeTreeDTO.php
@@ -1,0 +1,166 @@
+<?php
+
+namespace EventEspresso\core\services\orm\tree_traversal;
+
+use EE_Base_Class;
+use Exception;
+
+/**
+ * Class ModelObjNodeTreeDTO
+ *
+ * Class for storing the discovered tree of model object dependencies. It is stored in a format that can be easily and
+ * efficiently serialized for persistence, and unserialized with minimal memory usage, and so we can pickup where
+ * we left off.
+ * A tree of objects would be more convenient (my original implementation used that) but we'd need to either unserialize
+ * to a PHP array and then ALSO build that data structure (which would be trouble when there could be thousands of
+ * entries); or we'd need to write our own parser (which was fun excercise in school, but no thank you.)
+ *
+ * @package     Event Espresso
+ * @author         Mike Nelson
+ * @since         $VID:$
+ *
+ */
+class ModelObjNodeTreeDTO
+{
+    const INDEX_PATH = 'ip';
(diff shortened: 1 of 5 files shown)


mnelson4:
I've been analyzing how to be handle the algorithm etc for the cascading deletion etc etc. Maybe my head is full of cheese, but it's pretty tricky (both to build, and I'm afraid I think it will tricky to maintain in the future). Also, I realized: is it really necessary to serialize the transitory data to JSON? This data should only be stored to the database for at most a few minutes (after they click "delete event" and before the deletion job is finished). If I instead just used PHP's serialize function, I could serialize the objects themselves super easily and I think the efficiency would be comparable. I could also use the classes I've already built to implement the algorithm (instead of building them all over again) which would make it much more likely to meet the project's deadline.

I ping'd Brent with the above, asking if he thought serializing the PHP objects would be ok in this instance. (Remember this data won't be stored for long, and is easily re-creatable if it somehow gets corrupted.)

He remarked 
> technically it's considered a security issue to store objects directly in the db
> because apparently someone can modify the data and include malicious code that can be run via wakeup()

So it hinges on the user being able to modify the serialized string that's stored to the DB. Because the string is generated by our EE code, without any direct user input, we can't see a way for that to be exploited. 

But because we're going to try serializing objects in this instance (the data is transitory and re-creatable and there's no user input) doesn't mean it's ok to do in other situations. We still generally would prefer storing data in JSON in the DB. 

Also, I want to verify using `serialize` will still work for very large datasets. So once I have something working, I'd like to test a big bulk deletion on a low-end host by

1. Install this version of EE on a low-end host (eg BlueHost)
2. Use the attendee importer to import thousands of registrations to an event
3. Use the cascading deletion feature to delete that event.

But for now, it appears the work in this branch may be unnecessary.

mnelson4:
I think we're just going with the simpler option

PR #2161, opened 2020-01-01
Extract data handling from UI components; closes #2152.

### LINKED ISSUE #2152: (maybe) Extract ALL Data Handling from UI Components
As the last step before moving over the UI components from EDTR v1, we _may_ want to consider adopting the `Data Container => UI Component` pattern where all data handling logic is completely decoupled from UI components and moved into a separate file that then wraps the UI component and passes it the required props.

This will likely benefit us when we get to writing tests for our UI components because we will be able to separate the different scopes and use regular tests for the data containers and snapshots for the UI components.

Typically this pattern is achieved by appending `Container` onto the end of the UI component's name for the wrapper component, for example:

- DatesList (UI only)
- DatesListContainer (data handling)

The thing I don't like about this is that your application files need to import the container component, so everything in your JSX tree has `Container` on the end of the name:

```jsx
const DatesList = ({ datetimes }) => (
	<>
		<DatesListFilterBarContainer />
		{datetimes.map((datetime) => (
			<DateCardContainer datetime={datetime} />
		))}
		<AddNewDateButtonContainer />
	</>
);
```

So I think I would prefer to do the reverse instead and append `UI` onto the end of the JSX only components:

```jsx
const DatesListUI = ({ datetimes }) => (
	<>
		<DatesListFilterBar />
		{datetimes.map((datetime) => (
			<DateCard datetime={datetime} />
		))}
		<AddNewDateButton />
	</>
);
```

Which introduces a lot less noise into the code, but I'm open for suggestions for something other than `UI` since that looks a bit funny (as does `Ui`).
Maybe `View` would be better (ex: `DatesListView`)?

Another option would be to use HOCs (which I normally don't like but they are somewhat appropriate for this)

```jsx
import withDatesListData from './withDatesListData';

export const DatesList = ({ datetimes }) => (
	<>
		<DatesListFilterBar />
		{datetimes.map((datetime) => (
			<DateCard datetime={datetime} />
		))}
		<AddNewDateButton />
	</>
);

export default withDatesListData( DateList );
```

The benefit of this approach is that you can use the default export for the actual application, but still use the direct export of `DatesList` for testing purposes, and for testing the HOC, you can pass a mock component that renders the passed props in a simple and direct way allowing you to mount the HOC and examine the rendered output to verify the results. That said, NOT using a **HOC** might be slightly easier to test because the Data Container simply wraps the UI component as opposed to being a function that returns an enhanced component.

What are your guys thoughts/opinions/ideas regarding this?

Any better suggestions?

nicolad:
Closes: https://github.com/eventespresso/event-espresso-core/issues/2152

## Checklist

* [ ] My code is tested.


nicolad:
I am not :100: happy with files/ folder structure:
image
***
open to suggestions and improvements

assets/prototype/application/ui/components/display/ErrorIndicator.js
```diff
@@ -1,14 +1,15 @@
+/**
```
nicolad: TS is very smart.
***
I was getting this error and did not understood at first what was the issue.
image

nicolad: Because of TS I found out, that these are returning boolean:
image

nicolad: So, error isn't an object with `code` and `message` as being destructured here: 
image

nicolad: Also checked blueprint, title and description aren't required:
https://github.com/palantir/blueprint/blob/develop/packages/core/src/components/non-ideal-state/nonIdealState.tsx#L42

tn3rb: oh I think originally that code may have been utilizing `ApolloError` objects, but ya TS is quickly becoming very loved

assets/prototype/domain/eventEditor/EventEditor.tsx
```diff
@@ -0,0 +1,24 @@
...
+import DatesListData from './datetimes/datesList/DatesListData';
+import TicketsListData from './tickets/ticketsList/TicketsListData';
+
+const EventEditor = () => {
+	useInitQueries();
+
+	return (
+		<>
+			<DatesListData />
+			<TicketsListData />
```
manzoorwanijk: IMHO `<DatesList />` and `<TicketsList />` are better

assets/prototype/domain/eventEditor/datetimes/datesList/DatesList.tsx
```diff
@@ -0,0 +1,49 @@
...
+	flexFlow: 'row wrap',
+	justifyContent: 'space-between',
+	width: '100%',
+};
+
+interface DatesListProps {
+	datetimes: Datetime[];
+}
+
+const DatesList = ({ datetimes }: DatesListProps) => {
```
manzoorwanijk: Also, as discussed earlier, we should prefer `DatetimeList` instead of just `DatesList`

assets/prototype/domain/eventEditor/tickets/ticketsList/AddNewTicketButton.js
```diff
@@ -1,9 +1,12 @@
-import { useState } from '@wordpress/element';
+/**
+ * External dependencies
+ */
+import * as React from 'react';
```
manzoorwanijk: That could be better, no?
```suggestion
import React, { useState } from 'react';
```

nicolad: Yes, updated, thx.

assets/prototype/application/ui/components/display/ErrorIndicator.js
nicolad: @manzoorwanijk 's suggestion fixed this: https://github.com/eventespresso/event-espresso-core/pull/2161/commits/a891d00bcbba5240a44747e94e9ba54d0de17477

assets/prototype/domain/eventEditor/datetimes/datetimeForm/DatetimeForm.js
```diff
@@ -59,7 +59,7 @@ const formatSecondaryField = (ticketPrice, toString = false) => {
 	return toString ? renderToString(<Currency quantity={ticketPrice} />, null) : <Currency quantity={ticketPrice} />;
 };
 
-const DateForm = ({ formReset, relatedTickets, tickets = [], title }) => {
+const DatetimeForm = ({ formReset, relatedTickets, tickets = [], title }) => {
```
manzoorwanijk: 👍

assets/prototype/domain/eventEditor/datetimes/datetimeForm/DatetimeForm.js
```diff
@@ -61,7 +65,9 @@ const formatSecondaryField = (ticketPrice, toString = false) => {
 
 const DatetimeForm = ({ formReset, relatedTickets, tickets = [], title }) => {
 	const { id } = useContext(DateTimeContext);
-	const { description = '', name = '' } = useDatetimeItem({ id });
```
manzoorwanijk: It could be solved by `useDatetimeItem({ id }) || {}` 😄

nicolad: Removed that commit, that is out of scope of original PR.

PR #2168, opened 2020-01-02
Check-ins page: Add error proofing and display error message if no valid REG ID is passed

joshfeck:

This PR doesn't solve a problem. What it does do is it will display an error message instead of a fatal error on the check ins page if an invalid REG ID is passed.

## How has this been tested

This can be tested by going to Event Espresso > Registrations > Event Check-In, then select an event. Then check in an attendee. Then refresh the page. After the page loads, you'll see timestamp of the check-in, then click on the timestamp. You've now tested the functionality of viewing the check in records. Now, to test the error message, change the `_REG_ID` in URL so it's an invalid reg ID. The page should load an error message (and not a fatal error like it would with master branch).

## Checklist

* [x] My code is tested.


sethshoultes:
Looks like this is working. Here's the error message:
image

PR #2175, opened 2020-01-04
Preview Data before deletion and ask for confirmation

mnelson4:
On the event deletion preview page, show the user a summary of what will be deleted, and ask for them to confirm they're deleting the correct events and that they have a database backup.
If WP_DEBUG is on, shows ALL models with items to delete, even for join tables (eg https://nimb.ws/GYQhSN). When WP_DEBUG is off, only shows the ones most users probably care to see (eg https://nimb.ws/p24F3D). I'm not positive this is the best way to do this, we might want another constant to enable showing absolutely everything like this.
For https://github.com/eventespresso/event-espresso-core/issues/1905

## How has this been tested

* [ ] Try permanently deleting an event. The preview page should now summarize what data you're going to delete (if you have WP_DEBUG on, it shows all the join table entries too).
* [ ] It will ask you to confirm you have a database backup and to enter the slug of all the events you're able to delete. If you don't enter them correctly, there should be an error and the events won't be deleted.

## Checklist

* [x] My code is tested.


mnelson4:
@joshfeck here's a screenshot of what this PR makes the deletion preview page look like:
screenshot-laragonee test-2020 01 03-21_36_41

Do you think that will make sense to users, and if not what would you change to make it more understandable?

Does asking for users to enter the events' slugs do a good enough job of ensuring they're not deleting the wrong event? (Also I just realized: the user already put the event in the trash and then clicked again to permanently delete it; adding yet another verification might be unnecessary?)

joshfeck:
I think some of what's shown isn't needed, and we could make this page more user-friendly by removing the following:

under Datetimes: remove the IDs

Instead of the first 20 registrations, only show 10 and in the following format:
First Name Last Name, (1 of 2)
First Name Last Name, (2 of 2)
the reg codes and IDs aren't needed here and add clutter

Probably no need to mention that tickets will be deleted as they generally don't help identify the event and can't be accessed in other places.


Also, it appears that only one event is permanently deleted if I select more than one event from the trash to be deleted. Should we allow only one event to be perm. deleted at a time?

mnelson4:
Great feedback Josh, I'll implement that right away.

It should support permanently delete more than one at a time, but TBH I haven't gotten around to testing it. I'm hoping fixing it will take less than an hour.

mnelson4:
Ok deleting multiple events simultaneously is working, and I updated the template. This is how it looks:
screenshot-laragonee test-2020 01 07-10_57_05

joshfeck:
One thing I thought of and forgot to mention would be a "Get out!" button to cancel the deletion. The back button could also work of course, but since we're going through the trouble of making sure the end user knows what they're about to permanently delete they may realize it's not what they wanted and may want a clear path out and back to the trashed events list table.

mnelson4:
Oh yeah good idea.
Brent I think this is ready for reviewing. ALTHOUGH I apparently epically borked the history on this branch. I must have rebased onto master from this branch (ie, did `git checkout CASC/preview-page; git rebase master`) or something. I'm sorry I don't know how that happened. I suppose it's not the biggest deal given I'll just do a squash merge into CASC/base` and GitHub has the "Files Changed" tab. 
One oddity though is the "Files Changed" tab shows changes that are already in `master` and `CASC/base`. I almost wonder if it's a bug in GitHub... Anyways, the changes on this PR really just affect the Events admin page code, added a new form class, modified the deletion preview page, and I added `EE_REgistration::attendeeName()`.

tn3rb:
>  I must have rebased onto master from this branch

ya it should have been rebased onto `CASC/base`.

This could potentially cause a big mess when merged into master because it will replay those commits again. 
Is there any chance you can simply cherry pick the commits you made for this PR into a new branch cut off of `CASC/base`?

If you don't have the time to do it, then I can do it for you, because I really don't want to risk having something screwed up in master.

I really don't mind doing that, although it would be better for you to do so just in case there are any conflicts while rebasing.

admin_pages/events/Events_Admin_Page.core.php
```diff
@@ -2131,58 +2134,165 @@ protected function generateDeletionPreview($event_ids)
         );
     }
 
+    /**
+     * Checks for a POST submission
+     * @since $VID:$
+     */
+    protected function confirmDeletion()
+    {
+        $deletion_job_code = isset($this->_req_data['deletion_job_code']) ? $this->_req_data['deletion_job_code'] : '';
```
tn3rb: 👽 

```suggestion
        $deletion_job_code = isset($this->_req_data['deletion_job_code']) ? sanitize_key( $this->_req_data['deletion_job_code'] ) : '';
```

admin_pages/events/Events_Admin_Page.core.php
```diff
@@ -2131,58 +2134,165 @@ protected function generateDeletionPreview($event_ids)
...
+                                    'status' => 'trash'
+                                ],
+                                EVENTS_ADMIN_URL
+                            )
+                        )
+                    ),
+                    admin_url()
+                )
+            );
+        } else {
```
tn3rb: i think this entire else block (not the code within it, but just the actual `else` structure)
might be unnecessary because `EEH_URL::safeRedirectAndExit()` calls `exit()` after the redirect,
so the else block will not ever get called if the form is valid.

It is essentially the same as:

```
if ( valid ) {
	// do stuff
	exit();
}
// don't need an else here because of the above exit
// if the form is not valid, it won't redirect and exit,
// and will end up here anyways
```

admin_pages/events/Events_Admin_Page.core.php
```diff
@@ -2131,58 +2134,165 @@ protected function generateDeletionPreview($event_ids)
...
+
     /**
      * A page for users to preview what exactly will be deleted, and confirm they want to delete it.
      * @since $VID:$
+     * @throws EE_Error
      */
     protected function previewDeletion()
     {
-        $EVT_IDs = isset($this->_req_data['EVT_IDs']) ? (array) $this->_req_data['EVT_IDs'] : array();
+        $deletion_job_code = isset($this->_req_data['deletion_job_code']) ? $this->_req_data['deletion_job_code'] : '';
```
tn3rb: plz sanitize

caffeinated/admin/extend/registrations/Extend_Registrations_Admin_Page.core.php
```diff
@@ -796,6 +796,14 @@ protected function _registration_checkin_list_table()
         $reg_id = isset($this->_req_data['_REG_ID']) ? $this->_req_data['_REG_ID'] : null;
         /** @var EE_Registration $registration */
         $registration = EEM_Registration::instance()->get_one_by_ID($reg_id);
```
tn3rb: plz sanitize with `absint()` if `$reg_id` is not sanitized by the models

core/db_classes/EE_Registration.class.php
```diff
@@ -2082,4 +2104,24 @@ public function get_related_primary_datetime()
...
+            sprintf(
+                // translators: 1: registration code, 2: number, 3: group size, 4: attendee name.
+                __('Registration with code "%1$s", %2$d of %3$d for %4$s', 'event_espresso'),
+                $this->reg_code(),
+                $this->count(),
+                $this->group_size(),
+                $attendee_name
+            )
+        );
+    }
```
tn3rb: Where is this used?
If this is totally new (looks it) then maybe we can come up with a better name for the method than just `name` ?

It certainly appears to be providing more info than just the registrant's name.

how about:

- description
- regDescription
- registrationDescription
- something else ?

i obviously couldn't come up with something more descriptive and/or appropriate than something related to "description"

modules/single_page_checkout/reg_steps/attendee_information/attendee_info_single.template.php
```diff
@@ -25,31 +25,32 @@
...
 
-    <div id="spco-attendee-panel-dv-single"
-        class="spco-attendee-panel-dv spco-attendee-ticket-single">
+    <div id="spco-attendee-panel-dv-<?php echo $ticketID; ?>"
+        class="spco-attendee-panel-dv spco-attendee-ticket-<?php echo $ticketID; ?>">
 
         <div class="spco-ticket-info-dv small-text">
             <?php if (!is_admin()) : ?>
-                <h5><?php _e('Details', 'event_espresso'); ?></h5>
+                <h5><?php esc_html_e('Details', 'event_espresso'); ?></h5>
```
tn3rb: thnx 👍

mnelson4:
Created https://github.com/eventespresso/event-espresso-core/pull/2187

admin_pages/events/Events_Admin_Page.core.php
mnelson4: thanks! yeah somehow I totally overlooked all sanitizing :disappointed_relieved:

mnelson4: applying to new PR

admin_pages/events/Events_Admin_Page.core.php
mnelson4: oh thanks! yes

admin_pages/events/Events_Admin_Page.core.php
mnelson4: :taco:

caffeinated/admin/extend/registrations/Extend_Registrations_Admin_Page.core.php
mnelson4: that's actually not part of this PR, it turns out... and yes the models do sanitize data also BUT I've heard it's best to sanitize ASAP and layers of security is best anyway.

core/db_classes/EE_Registration.class.php
mnelson4: It's actually overriding a method in `EE_Base_Class` which is meant to provide a pretty name we can display to users when there's dynamic code. I was using it earlier in this branch but no more. 
I've replaced this method's body to instead just return `attendeeName()` as it sounds like that's what you were expecting a method by this name to return.

modules/single_page_checkout/reg_steps/attendee_information/attendee_info_single.template.php
mnelson4: sorry this wasn't actually part of this PR again! But I'll accept your goodwill regardless.

mnelson4:
Merged the other PR. This one got borked because I must have rebased this branch on master (I don't remember doing it, but that's what the log suggests.)

PR #2177, opened 2020-01-07
extract money utilities

tn3rb:
extracts generic money related logic from ticket price calculator

### DIFF
--- a/assets/prototype/application/utilities/money/amountsMatch.ts
+++ b/assets/prototype/application/utilities/money/amountsMatch.ts
@@ -1,0 +1,14 @@
+import parsedAmount from './parsedAmount';
+
+/**
+ * returns true if the two supplied values are equal after being parsed as floats
+ *
+ * @param {number|string} amount1
+ * @param {number|string} amount2
+ * @return {boolean}
+ */
+const amountsMatch = (amount1: number | string, amount2: number | string): boolean => {
+	return parsedAmount(amount1) === parsedAmount(amount2);
+};
+
+export default amountsMatch;
--- a/assets/prototype/application/utilities/money/formatAmount.ts
+++ b/assets/prototype/application/utilities/money/formatAmount.ts
@@ -1,0 +1,16 @@
+import parsedAmount from './parsedAmount';
+
+/**
+ * returns a function that when supplied a value for the number of decimal places used by a currency,
+ * returns a second function that can be passed an amount which will then be appropriately formatted
+ *
+ * @param {number} decimalPlaces
+ * @return {Function}
+ */
+const formatAmount = (decimalPlaces: number): ((amount: number | string) => string) => (
+	amount: number | string
+): string => {
+	return parsedAmount(amount).toFixed(decimalPlaces);
+};
+
+export default formatAmount;
--- a/assets/prototype/application/utilities/money/formatAmount.ts
+++ b/assets/prototype/application/utilities/money/formatAmount.ts
@@ -1,5 +1,7 @@
 import parsedAmount from './parsedAmount';
 
+export type FormatAmountFunction = (amount: number | string) => string;
+
 /**
  * returns a function that when supplied a value for the number of decimal places used by a currency,
  * returns a second function that can be passed an amount which will then be appropriately formatted
@@ -7,9 +9,7 @@ import parsedAmount from './parsedAmount';
  * @param {number} decimalPlaces
  * @return {Function}
  */
-const formatAmount = (decimalPlaces: number): ((amount: number | string) => string) => (
-	amount: number | string
-): string => {
+const formatAmount = (decimalPlaces: number): FormatAmountFunction => (amount: number | string): string => {
 	return parsedAmount(amount).toFixed(decimalPlaces);
 };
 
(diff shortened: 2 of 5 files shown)


assets/prototype/application/utilities/money/parsedAmount.ts
```diff
@@ -0,0 +1,11 @@
+/**
+ * returns amount parsed as a float (if not already a number)
+ *
+ * @param {number|string} amount
+ * @return {number}
+ */
+const parsedAmount = (amount: number | string): number => {
```
manzoorwanijk: When can an amount be a `string`?

tn3rb: when it comes from react final form

assets/prototype/application/utilities/money/amountsMatch.ts
```diff
@@ -0,0 +1,14 @@
+import parsedAmount from './parsedAmount';
+
+/**
+ * returns true if the two supplied values are equal after being parsed as floats
+ *
+ * @param {number|string} amount1
+ * @param {number|string} amount2
+ * @return {boolean}
+ */
+const amountsMatch = (amount1: number | string, amount2: number | string): boolean => {
```
nicolad: I noticed in `date-fns`, the use instead of arg1/ arg2 the `dateLeft, dateRight` approach: https://date-fns.org/v2.9.0/docs/compareAsc
may not be relevant here, but worth mentioning

assets/prototype/application/utilities/money/formatAmount.ts
```diff
@@ -0,0 +1,16 @@
+import parsedAmount from './parsedAmount';
+
+/**
+ * returns a function that when supplied a value for the number of decimal places used by a currency,
+ * returns a second function that can be passed an amount which will then be appropriately formatted
+ *
+ * @param {number} decimalPlaces
+ * @return {Function}
+ */
+const formatAmount = (decimalPlaces: number): ((amount: number | string) => string) => (
```
nicolad: Maybe return type can be extracted somehow?

assets/prototype/application/utilities/money/parsedAmount.ts
nicolad: I like the `strict typing` idea, I think it will ease the understanding and removes away the mental overload related to coercion.

assets/prototype/application/utilities/money/formatAmount.ts
nicolad: Maybe return can be used from `MoneyDisplay.formatAmount` from `useMoneyDisplay.ts`?

assets/prototype/application/utilities/money/useMoneyDisplay.ts
```diff
@@ -0,0 +1,29 @@
...
+	currency: CurrencyProps;
+	// function for formatting the amount using the correct number of decimal places for the currency
+	formatAmount: (amount: number | string) => string;
+};
+
+const useMoneyDisplay = (): MoneyDisplay => {
+	const { config } = useConfig();
+	const afterAmount = config.currency.signB4 ? '' : config.currency.sign;
+	const beforeAmount = config.currency.signB4 ? config.currency.sign : '';
+	const formatMoney = formatAmount(config.currency.decimalPlaces);
```
nicolad: I like how this function ended up using `decimalPlaces` from config.
And the fact that it is curried it's double win, since it is acting as micro-factory.
So much noise will be moved away from the components that will use this hook.
Amazing job :bow_and_arrow:

assets/prototype/application/utilities/money/amountsMatch.ts
tn3rb: ya I think that makes more sense in the context of a sorting algorithm like they are using that function for, where you are stepping through a list and comparing a list item against the one before it. This function is just for comparing any two amounts that don't even have to be from the same "container", so deciding which is the "left" item and which is the "right" item doesn't really make sense imho.

Thank you for bringing that up though because it's always good to be looking at how things are done in other places. We don't want to work in a bubble.

assets/prototype/application/utilities/money/formatAmount.ts
tn3rb: do you mean into a named type/interface?

something like:

```ts
const formatAmount = (decimalPlaces: number): AmountFormatter => (
```

nicolad: Yes, the example from above is more grokkable.

assets/prototype/application/utilities/money/useMoneyDisplay.ts
tn3rb: thank you...
although I'm not sure if I'm totally happy with this yet.

I'm thinking that `formatAmount` isn't entirely clear as to what formatting is occuring, ie:

- is this only controlling the number of decimal places displayed?
- is it adding thousands separators?
- is it adding currency signs?
- is it adding the currency name/abbreviation?

So I may refactor this to add functions for applying all of those other formatting capabilities and/or a single formatter with some kind of options and/or config parameters for controlling the formatting. But that said, it would still ideally keep all of that noise away from the components.

assets/prototype/application/utilities/money/formatAmount.ts
tn3rb: ok I can do that... expect another review request in a bit

PR #2180, opened 2020-01-07
Cascading Deletions of Events (CASC/base)

mnelson4:

Merges in the cascading deletion work.
### Summary of UX Change
Events, and all their related dependent data (as defined by the database through foreign keys, like datetimes and registrations; as well as implied by the UI, such as tickets and transactions with no registrations for other events) can now be permanently deleted without needing to delete all the individual payments and registrations etc.
To delete an event, trash it as normal, then permanently delete it. You can do this for multiple events at once.
Because this could be deleted thousands of records at once, there are two batch jobs for this:
* `\EventEspressoBatchRequest\JobHandlers\PreviewEventDeletion` assesses all the data that will be deleted, after which the user is presented with a confirmation screen,
* after which `\EventEspressoBatchRequest\JobHandlers\ExecuteBatchDeletion` performs the deletion queries

### Implementation Details
Traversing the tree of model objects/database rows is undertaken mainly by classes `\EventEspresso\core\services\orm\tree_traversal\ModelObjNode`, which wraps a model object and recursively contains sub-`ModelObjNode`s for all its dependent model object, and `\EventEspresso\core\services\orm\tree_traversal\NodeGroupDao` which takes care of persistence between HTTP requests. Persistence is done just by serializing the `ModelObjNode`s, which are designed to have a small serialized representation. In this way, we should be able to delete registrations with up to 1000 registrations even on cheap shared hosting. For events with more registrations than that, the database option `NodeGroupDao` stores the data in can get very large, like over 16MB, at which point unserializing and serializing that much data starts to take up many seconds during each request, and database queries start failing.

### Dealing with Crazy Big Events
Deleting more registrations at once may require increasing their `max_allowed_packet` in MySQL (see https://wordpress.stackexchange.com/a/355527/52760), making the webserver not time out, and temporarily set the constant `EE_BATCHRUNNER_BATCH_SIZE` to `INF` so the entire process happens in a single request (because just serializing and unserializing the model object tree is what takes the longest.) Long term fix would be to make `ModelObjNode`s "serialize" to something even smaller (currently they're about 200 characters serialized, about half of which is the class' name) and/or storing the data across multiple database options, or a different table, or adding a WP CLI command to perform all the cascading deletions in a single process without needing to serialize the data to the database.

### Model Object Tree Traversal Classes and Their Uses
A new document was added explaining how to use the new model object tree traversal classes (eg `ModelObjNode`) in `docs/G--Model-System/tree-traversal.md`. These classes could be useful when you want to permanently delete another model object and all its dependent model objects (eg transactions) or otherwise gather all the model objects that directly depend on another.

## How has this been tested

* [x] Create an event with no registrations. 
* Trash the event then permanently delete it. 
* It should be deleted fine (you may want to check the database that its datetimes, tickets, message-group relations, etc, were removed properly.) 
* [x] Create an event with some registrations.
* Take note of those registrations' transaction IDs. 
* Trash the event then delete it. 
* Those registrations and transactions should all be deleted
* [x] Create an event that uses some custom questions, payments, and registrations and do the same. Check those registrations' answer rows are also deleted.
* Create a few events with registrations, and bulk trash them then bulk permanently delete them. They should all get deleted fine. 
* [x] Repeat the above but with some add-ons active that add data related to events or registrations. Like promotions (which has `Promotion_Object` model which relates promotions to events; or Infusionsoft which records extra meta about events that are sync'd; or MailChimp which has other tables pointing to events; or payment methods.) It may not be a hard requirement that this always delete those add-ons' data, but it certainly shouldn't have any errors, and it would be great if this PR did clean up their data.
* [x] Create an event with 1000 registrations and do the same. It will take a few minutes to generate the preview page, and a minute to delete, but it should work. Preferably do this test from a cheap shared host (I have one on Dreamhost if that helps). Note though the event can't be infinitely big, I know it doesn't work deleting an event with over 15,000 registrations (in which I outlined a solution on https://github.com/eventespresso/event-espresso-core/pull/2110#issuecomment-570462637)

## Checklist

* [x] My code is tested.


mnelson4:
I did not know that! 😵

Ok passing off for testing.

joshfeck:
If you’re on the Preview Deletion page, and you click on the “Preview Deletion” tab, the page reloads with a fatal error. We might disable that link.

PR #2187, opened 2020-01-08
Preview Data before deletion and ask for confirmation

mnelson4:

On the event deletion preview page, show the user a summary of what will be deleted, and ask for them to confirm they're deleting the correct events and that they have a database backup.
Cherry-picking from https://github.com/eventespresso/event-espresso-core/pull/2175, just the good stuff that's meant to be in there.

## How has this been tested

* [ ] Try permanently deleting an event. The preview page should now summarize what data you're going to delete (if you have WP_DEBUG on, it shows all the join table entries too).
* [ ] It will ask you to confirm you have a database backup and to enter the slug of all the events you're able to delete. If you don't enter them correctly, there should be an error and the events won't be deleted.
## Checklist

* [x] My code is tested.


mnelson4:
> The only big thing (well not that big really) is that I would like the logic within the following methods on Events_Admin_Page:

>     confirmDeletion()
>     getModelsAndIdsToDelete()
>     previewDeletion()

> to be moved into a separate class (or classes), because Events_Admin_Page is already way too big (several THOUSAND LoC). 

I moved `getModelsAndIdsToDelete()` into a new class `ModelObjNodeGroupPersister`, which has the job of saving, getting, and deleting the groups of `ModelObjNode`s used during these cascading deletion. (I'm hoping you would have a suggestion on a better name; it's pretty descriptive currently but awfully verbose.)
And was about to move the logic from `confirmDeletion` and `previewDeletion` into their own class(es?), but was hesitant as the logic in there looks appropriate to have in a controller class like this.
`previewDeletion` just fetches some of the model objects for deletion then uses them to render a template; `confirmDeletion` just receives a form submission and redirects.
BUT if you just want to put the bodies of those two methods in a new class just to make that file smaller, I can just roll with it. Just what classname(s) and method(s) would you like? I don't they'd really fit very well in the `core/services/orm` namespace as I don't expect them to be reused anywhere else and seem much more like code built on top of the ORM (like pretty well everything else). BUT I don't really have a better suggestion, besides leaving the methods in the events controller where they are now :stuck_out_tongue_closed_eyes: .

mnelson4:
Oh I also tacked onto this PR a feature Garth requested: logging who deleted the events and when.

tn3rb:
> I moved getModelsAndIdsToDelete() into a new class ModelObjNodeGroupPersister, which has the job of saving, getting, and deleting the groups of ModelObjNodes used during these cascading deletion.

awesome!

> I'm hoping you would have a suggestion on a better name; it's pretty descriptive currently but awfully verbose.

well that sounds to me like a [data access object (DAO)](https://en.wikipedia.org/wiki/Data_access_object) so I would be inclined to change "Persister" (which is kinda awkward imho) to "Dao", and assuming that the **only** kinds of "Node Groups" are of the "Model Object" type, then it's not necessary to include that extra qualifier (unless you want to do so as a folder, but again, not needed if there are no other types of Node Groups). So if there are never going to be other types of Node Groups, then I would drop "Model Object" from the name, otherwise it's like appending "Dog" onto the name of every dog breed (ChihuahuaDog, HuskyDog, etc).

So ultimately, I think you could probably get away with `NodeGroupDao` and that wouldn't get confused with anything else unless there were other kinds of Node Groups.

re: 

> was about to move the logic from confirmDeletion and previewDeletion into their own class(es?)...
> BUT if you just want to put the bodies of those two methods in a new class just to make that file smaller, I can just roll with it.

oh yes plz, that is one of the primary reasons for doing so, although I have many reasons that (I at least believe) are valid (better testing, more SOLID, etc). Sorry for causing the extra work but I really want to avoid adding anything more to the admin page classes than is absolutely needed. In this case that would just be some logic for loading and running the appropriate classes for performing the required business logic.

> Just what classname(s) and method(s) would you like? I don't they'd really fit very well in the core/services/orm namespace

how about something like:

```
├── core
|	├── domain
|	|	├── services
|	|	|	├── admin
|	|	|	|	├── events
|	|	|	|	|	├── data
|	|	|	|	|	|	├── ConfirmDeletion.php
|	|	|	|	|	|	├── PreviewDeletion.php
```

or as I mentioned previously, it would be absolutely ideal to build these as commands:

```
├── core
|	├── domain
|	|	├── services
|	|	|	├── commands
|	|	|	|	├── events
|	|	|	|	|	├── data
|	|	|	|	|	|	├── ConfirmDeletionCommand.php 
|	|	|	|	|	|	├── ConfirmDeletionCommandHandler.php
|	|	|	|	|	|	├── PreviewDeletionCommand.php
|	|	|	|	|	|	├── PreviewDeletionCommandHandler.php
```

where `ConfirmDeletionCommand` is created in `Events_Admin_Page` and supplied the necessary data to perform the deletion, which is then passed to the command bus:

```php
    try {
        EE_Registry::instance()->BUS->execute(
            new ConfirmDeletionCommand(
                $this->_req_data['EVT_IDs'],
                $this->_req_data['deletion_job_code']
            )
        );
    } catch ($e) {
        // deal with exception
    }
```

that class then would deal with verifying the above data (as well as anything else that is needed to perform the command) exists and is valid, and would then set on some internal properties, or throw an exception(s) if the data is not valid, or missing, or whatever.

Then `ConfirmDeletionCommandHandler` would receive the above command knowing that it has all of the necessary data required to perform its duties, and would basically do everything else that is currently in `Events_Admin_Page::confirmDeletion()`.

Ideally the command bus would standardise how all of our business logic was executed, which also makes it easier to trigger all of the same logic from other types of requests in the future if need be, like the REST API, or React/GraphQL, or CLI, or whatever. Using the command bus would also make it easy for us to perform additional functionality when commands are run, such as:

- logging ex: https://tactician.thephpleague.com/plugins/logger/
- transactional database queries ex: https://matthiasnoback.nl/2015/01/responsibilities-of-the-command-bus/#an-example%3A-database-transactions
- error handling ie: one try catch block to rule them all (or at least as a fallback)

If you are strapped for time, or unsure about anything, I could make those changes for you if you want.

mnelson4:
Thanks ya that new classname sounds much better.
I extracted `Events_Admin_Page::previewDeletion` and `confirmDeletion` into their own classes.
Brent and I discussed, and because those two methods rendered a template and performed a redirect, respectively (they don't actually perform business logic) they didn't actually seem appropriate for a command.

core/domain/services/admin/events/data/PreviewDeletion.php
```diff
@@ -0,0 +1,141 @@
...
+        EEM_Registration $registration_model
+    )
+    {
+        $this->dao = $dao;
+        $this->event_model = $event_model;
+        $this->datetime_model = $datetime_model;
+        $this->registration_model = $registration_model;
+    }
+
+    public function handle(Events_Admin_Page $admin_page)
```
mnelson4: I'm not injecting the admin page because it ended up instantiating a SECOND admin page, which caused warnings from the constants already being defined the first time. I avoided that can of worms by just passing it as a variable to this method.

core/domain/services/admin/events/data/ConfirmDeletion.php
```diff
@@ -0,0 +1,105 @@
...
+
+    /**
+     * @since $VID:$
+     * @param \Events_Admin_Page $admin_page
+     * @throws EE_Error
+     * @throws \EventEspresso\core\exceptions\InvalidDataTypeException
+     * @throws \EventEspresso\core\exceptions\InvalidInterfaceException
+     * @throws \EventEspresso\core\exceptions\UnexpectedEntityException
+     * @throws \InvalidArgumentException
+     * @throws \ReflectionException
```
tn3rb: plz import all classes (ie use `use` statements) so that all dependencies are listed at the top of the file

core/domain/services/admin/events/data/PreviewDeletion.php
tn3rb: ya I would have just passed the `$request_data` array and the URL returned from `$admin_page->admin_base_url()` as parameters and then have this class return an array of template args and let `Events_Admin_Page` handle setting them and loading the actual view.

ie something like:

```php
    protected function previewDeletion()
    {
        $preview_deletion = $this->getLoader()->getShared(
			'\EventEspresso\core\domain\services\admin\events\data\PreviewDeletion'
		);
        $template_args = $preview_deletion->handle(
			$this>get_request_data(),
			$this->admin_base_url()
		);
		$this->set_template_args( $template_args );
		$this->display_admin_page_with_no_sidebar();
    }
```

But this works as is and I don't want to hold it up any more

core/domain/services/admin/events/data/ConfirmDeletion.php
mnelson4: good catch thanks

core/domain/services/admin/events/data/ConfirmDeletion.php
```diff
@@ -0,0 +1,105 @@
...
+     * ConfirmDeletion constructor.
+     * @param NodeGroupDao $dao
+     */
+    public function __construct(
+        NodeGroupDao $dao
+    )
+    {
+
+        $this->dao = $dao;
+    }
```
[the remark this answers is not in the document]
mnelson4: ya I forgot to run it!

mnelson4:
Oh yeah that's a good suggestion. It was weird passing in the whole event object.

PR #2192, opened 2020-01-09
Fix Editor Controls

tn3rb:
Editor controls for opening and closing modals got messed up in all of the refactoring, but this PR fixes things.


assets/prototype/domain/eventEditor/context/DateTimeProvider/useDateEditorId.ts
```diff
@@ -2,13 +2,13 @@
...
+import { EditorId } from '../../../../application/ui/components/layout/editor-modal/useEditorModalState/types';
 
 /**
  * Function that returns editor id for a given event date
  */
-const useDateEditorId = (dateId: EntityId, prefix: string = ''): string => {
-	const editorId: string = `event-date-editor-${dateId}`;
-	return prefix !== '' ? `${prefix}-${editorId}` : editorId;
+const useDateEditorId = (prefix: string = 'datetime-editor', datetimeId: EntityId): EditorId => {
+	return prefix !== '' ? `${prefix}-${datetimeId}` : `datetime-editor-${datetimeId}`;
```
nicolad: Just realized that since we have the default value for `prefix`,  the ``datetime-editor-${datetimeId}`` result never will occur :smiley:

tn3rb: i thought that if someone passes an empty string it would negate the default, so I added it back in as a fallback.
that was my thought process... maybe that's not accurate and it's not necessary... haven't confirmed with tests yet 😆

nicolad: In continuation of your previous suggestion to have a more strict type system - I think that if we'd somehow enforce non-empty string type or a string utility helper.
That could make life easier :)
***
out of the scope of current PR :point_up:

tn3rb: >  if we'd somehow enforce non-empty string type

can you do that when there is a default?

ie: if the type is set to string, then `''` is valid

nicolad: > can you do that when there is a default?
yeah, with a default value, that isn't possible
my comment was a general hope that TS would help to prevent some boilerplate code, like this(maybe):
```
if(variable !== undefined) {
  return doThis();
}

if(isString(variable) && isEmpty(variable)) {
  return doThat();  
}

if(isString(variable) && isNotEmpty(variable)) {
  return doWhatever();  
}
```
just thoughts out loud, not smth related to `useDateEditorId`

tn3rb:
so I realized that we had an extra layer of logic in the `useEditorModalState()` hook that was completely unnecessary.

for example, we were wrapping some of the functions within another function that just passed the same props along:

```js
	const getIsOpen = (editorId: EditorId) => isEditorOpen(editorId);
	const onClose = (editorId: EditorId) => () => closeEditor(editorId);
	const setIsOpen = (editorId: EditorId) => () => openEditor(editorId);
```

those were initially using `id + modalId` instead of just `editorId` but that also wasn't necessary and got removed, because the editor ID should already incorporate the entity ID, but that is handled by other hooks like `useDateEditorId()`. So after switching to just the editor ID as the only parameter for those functions, it became completely unnecessary to have the extra function wrap.

tn3rb:
not grokking why these tests are failing:

```shell
    ERROR in /assets/prototype/domain/eventEditor/tickets/ticketCard/EditTicketModal.tsx
    [tsl] ERROR in /assets/prototype/domain/eventEditor/tickets/ticketCard/EditTicketModal.tsx(8,13)
          TS2339: Property 'editors' does not exist on type '{}'.
    
    ERROR in /assets/prototype/domain/eventEditor/tickets/ticketCard/EditTicketModal.tsx
    [tsl] ERROR in /assets/prototype/domain/eventEditor/tickets/ticketCard/EditTicketModal.tsx(8,22)
          TS2339: Property 'editorState' does not exist on type '{}'.
    
    ERROR in /assets/prototype/domain/eventEditor/tickets/ticketCard/EditTicketModal.tsx
    [tsl] ERROR in /assets/prototype/domain/eventEditor/tickets/ticketCard/EditTicketModal.tsx(8,35)
          TS2339: Property 'id' does not exist on type '{}'.
```

here: https://travis-ci.com/eventespresso/event-espresso-core/jobs/274398695

the referenced line of code in `EditTicketModal` is:

```jsx
const { editors, editorState, id } = useContext<TicketContextProps>(TicketContext);
```

That component is loaded by `EditTicket` which in turn is loaded by `TicketCard` and is used right below the `TicketProvider`:

```jsx
		<TicketProvider id={ticket.id}>
			<Card elevation={Elevation.ONE} style={cardStyle}>
				<EditTicket position='top' relatedDates={relatedDates} />
```

really puzzled...

PR #2193, opened 2020-01-10
Close the div at the correct location (after the reg form not before it)

joshfeck:
This fixes a mistake in the newish attendee_info_single.template.php template where the div that's supposed to wrap the reg form is instead closed before the reg form. This change just follows how it's done in the original attendee_info_main.template.php

## How has this been tested

Register 1 ticket, view the source of the reg form or use the inspector. The div should close after the form now.

## Checklist

* [x] My code is tested.

### DIFF
--- a/modules/single_page_checkout/reg_steps/attendee_information/attendee_info_single.template.php
+++ b/modules/single_page_checkout/reg_steps/attendee_information/attendee_info_single.template.php
@@ -74,7 +74,6 @@ if (count($registrations) > 0) {
                 </tbody>
             </table>
         </div>
-    </div>
 
     <?php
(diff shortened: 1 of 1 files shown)


sethshoultes:
Verified this fixes the broken div:
Developer Tools - Registration Checkout – Event Espresso Testing - httpsespresso-4-9 dev ccpage_id=5 uts=1579017114#checko… 2020-01-14 09-18-17(1)
Developer Tools - Registration Checkout – Event Espresso Testing - httpsespresso-4-9 dev ccpage_id=5 uts=1579017114#checko… 2020-01-14 09-19-26(1)

PR #2196, opened 2020-01-10
Add Tests for Price Type Predicates; closes #2118.

### LINKED ISSUE #2118: Add Tests for Price Type Predicates
- assets\prototype\domain\shared\predicates\priceTypes
- convert to TS
- add unit tests

nicolad:
- Converts `selectionPredicates` to a folder with `index.ts` and `index.test.ts`
- Adds tests for `getDefaultPriceType`
- Closes https://github.com/eventespresso/event-espresso-core/issues/2118

### DIFF
--- a/assets/prototype/domain/shared/predicates/priceTypes/index.ts
+++ b/assets/prototype/domain/shared/predicates/priceTypes/index.ts
@@ -1,0 +1,1 @@
+export * from './selectionPredicates';
--- a/assets/prototype/domain/shared/predicates/priceTypes/selectionPredicates/index.test.ts
+++ b/assets/prototype/domain/shared/predicates/priceTypes/selectionPredicates/index.test.ts
@@ -2,17 +2,17 @@
  * ?Internal dependencies
  */
 import { PriceBasetype } from '../../../../../domain/eventEditor/data/types';
-import { getDefaultPriceType } from './index';
+import { getDefaultPriceModifierType } from './index';
 import { nodes as priceTypes } from '../../../../../domain/eventEditor/data/queries/priceTypes/test/data';
 
 describe('getDefaultPriceType', () => {
 	it('should get default price type - SURCHARGE', () => {
-		const defaultPriceType = getDefaultPriceType(priceTypes);
+		const defaultPriceType = getDefaultPriceModifierType(priceTypes);
 		expect(defaultPriceType.baseType).toBe(PriceBasetype.SURCHARGE);
 	});
 
 	it('should return undefined if we pass an empty array', () => {
-		const defaultPriceType = getDefaultPriceType([]);
+		const defaultPriceType = getDefaultPriceModifierType([]);
 		expect(defaultPriceType).toBeUndefined();
 	});
 });
--- a/assets/prototype/domain/shared/predicates/priceTypes/selectionPredicates/index.test.ts
+++ b/assets/prototype/domain/shared/predicates/priceTypes/selectionPredicates/index.test.ts
@@ -87,7 +87,9 @@ describe('isFlatFeeSurcharge', () => {
 			const { isBasePrice, isDiscount, isPercent } = priceType;
 			const basePrice = isBasePrice === true && isDiscount === false && isPercent === false;
 
-			expect(isFlatFeeSurcharge(priceType)).toBe(basePrice);
+			if (basePrice === true) {
+				expect(isFlatFeeSurcharge(priceType)).toBe(false);
+			}
 		});
 	});
 
@@ -96,7 +98,9 @@ describe('isFlatFeeSurcharge', () => {
 			const { isBasePrice, isDiscount, isPercent } = priceType;
 			const dollarDiscount = isBasePrice === false && isDiscount === true && isPercent === false;
 
-			expect(isFlatFeeSurcharge(priceType)).toBe(dollarDiscount);
+			if (dollarDiscount === true) {
+				expect(isFlatFeeSurcharge(priceType)).toBe(false);
+			}
 		});
 	});
 
@@ -105,7 +109,9 @@ describe('isFlatFeeSurcharge', () => {
 			const { isBasePrice, isDiscount, isPercent } = priceType;
 			const percentageSurcharge = isBasePrice === false && isDiscount === false && isPercent === true;
 
-			expect(isFlatFeeSurcharge(priceType)).toBe(percentageSurcharge);
+			if (percentageSurcharge === true) {
+				expect(isFlatFeeSurcharge(priceType)).toBe(false);
+			}
 		});
 	});
 
@@ -114,7 +120,9 @@ describe('isFlatFeeSurcharge', () => {
 			const { isBasePrice, isDiscount, isPercent } = priceType;
 			const precentageDiscount = isBasePrice === false && isDiscount === true && isPercent === true;
 
-			expect(isFlatFeeSurcharge(priceType)).toBe(precentageDiscount);
+			if (precentageDiscount === true) {
+				expect(isFlatFeeSurcharge(priceType)).toBe(false);
+			}
 		});
 	});
 });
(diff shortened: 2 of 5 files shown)


assets/prototype/domain/shared/predicates/priceTypes/index.ts
```diff
@@ -0,0 +1 @@
+export * from './selectionPredicates';
```
nicolad: @tn3rb should `selectionPredicates` be exposed at this level ?

nicolad:
@tn3rb @manzoorwanijk I noticed that only `getDefaultPriceType` is being used, so I added tests only for that function.

assets/prototype/domain/shared/predicates/priceTypes/selectionPredicates/index.test.ts
```diff
@@ -0,0 +1,18 @@
...
+ * ?Internal dependencies
+ */
+import { PriceBasetype } from '../../../../../domain/eventEditor/data/types';
+import { getDefaultPriceType } from './index';
+import { nodes as priceTypes } from '../../../../../domain/eventEditor/data/queries/priceTypes/test/data';
+
+describe('getDefaultPriceType', () => {
+	it('should get default price type - SURCHARGE', () => {
+		const defaultPriceType = getDefaultPriceType(priceTypes);
+		expect(defaultPriceType.baseType).toBe(PriceBasetype.SURCHARGE);
```
manzoorwanijk: `defaultPriceType` doesn't always need to be a `SURCHARGE`

manzoorwanijk: Am I right @tn3rb ?

nicolad: Since `getDefaultPriceType` returns `find(isFlatFeeSurcharge)(priceTypes)` I was expecting `SURCHARGE` :smiley:

tn3rb: Well first off, a better name for "default price type" would have been something like "default price modifier", since that is basically the default option for the price type selector in the ticket price calculator when you go to add a new price modifier.

Currently we will be using surcharge for the default price type(modifier) so what Vadim has is correct.
We could decide at some point in the future to change that, but if that happens we'll need to change the tests too.

What would be ideal is to have an option where the user could decide what they want to use for default price modifiers, then set that value in the config, and then read that value whenever we have to add a new price modifier. 

But what is here will be fine for the foreseeable future.

assets/prototype/domain/shared/predicates/priceTypes/index.ts
tn3rb: i think so, the intent is for these functions to be the building blocks (ie: Lego) for more complex business logic.

tn3rb:
> I noticed that only getDefaultPriceType is being used, so I added tests only for that function.

can you plz add tests for all of them now?

I understand that only that one function is being utilized at the moment, but as I said in another comment, these predicate functions are intended to be building blocks for more complex logic, so I think we want to verify that they operate as expected sooner rather than later. Otherwise, someone could waste time trying to use a faulty lego block in something they are trying to build and not be able to grok why things are not working. If we first confirm that these blocks all work as expected, then they can be used with confidence.

while you are at it, can you plz rename `getDefaultPriceType` to `getDefaultPriceModifierType` so that it is a little more clear as to what its purpose is.

thnx

assets/prototype/domain/shared/predicates/priceTypes/selectionPredicates/index.test.ts
```diff
@@ -0,0 +1,127 @@
...
+	isBasePrice,
+	isNotBasePrice,
+	isDiscount,
+	isNotDiscount,
+	isPercent,
+	isNotPercent,
+	isTax,
+	isNotTax,
+	isFlatFeeSurcharge,
+	getDefaultPriceModifierType,
```
nicolad: Kept the same order in the other file.

assets/prototype/domain/shared/predicates/priceTypes/selectionPredicates/index.test.ts
```diff
@@ -0,0 +1,127 @@
...
+} from './index';
+import { nodes as priceTypes } from '../../../../../domain/eventEditor/data/queries/priceTypes/test/data';
+
+describe('isBasePrice & isNotBasePrice', () => {
+	it('should return true if priceType is base price', () => {
+		priceTypes.forEach((priceType) => {
+			if (priceType.isBasePrice === true) {
+				expect(isBasePrice(priceType)).toBe(true);
+			}
```
tn3rb: should maybe add an else clause to confirm that the function is not sending false positives

ie:

```js
	if (priceType.isBasePrice === true) {
		expect(isBasePrice(priceType)).toBe(true);
	} else {
		expect(isBasePrice(priceType)).toBe(false);
	}
```

or you could just compare directly against `priceType.isBasePrice`:


```js
	expect(isBasePrice(priceType)).toBe(priceType.isBasePrice);
```

maybe do the same for all of the tests using this pattern

assets/prototype/domain/shared/predicates/priceTypes/selectionPredicates/index.test.ts
```diff
@@ -0,0 +1,127 @@
...
+		});
+	});
+});
+
+describe('isFlatFeeSurcharge', () => {
+	it('should return true if price type is flat fee surcharge', () => {
+		priceTypes.forEach((priceType) => {
+			if (priceType.isBasePrice === false && priceType.isDiscount === false && priceType.isPercent === false) {
+				expect(isFlatFeeSurcharge(priceType)).toBe(true);
+			}
```
tn3rb: this test has at least three potential points of failure since it is testing against three properties as well as the logic that combines those into one comparison, so if any of the tests used internally by `isFlatFeeSurcharge()` fails, then it could be giving false positives, but this unit test doesn't check for that.

this one could capture the result of the conditional into a var and then compare against that

ie:

```js
	const dollarSurcharge = priceType.isBasePrice === false && priceType.isDiscount === false && priceType.isPercent === false;
	expect(isFlatFeeSurcharge(priceType)).toBe(dollarSurcharge);
```

[EDIT]

oops... just saw below that you are also testing all of the property combinations separately, so my above comment is not entirely accurate at all since you are testing other possibilities.

that said, I still don't think it's wise to apply tests conditionally like this since you are potentially missing out on an opportunity to catch some weird edge case failure.

assets/prototype/domain/shared/predicates/priceTypes/selectionPredicates/index.test.ts
nicolad: I like this suggestion:
`expect(isBasePrice(priceType)).toBe(priceType.isBasePrice);`
it looks a bit cleaner
thx :bowing_man:

nicolad: Updated: https://github.com/eventespresso/event-espresso-core/pull/2196/commits/07c28fb1b222c22f2271c489d1ed88be17491490

assets/prototype/domain/shared/predicates/priceTypes/selectionPredicates/index.test.ts
```diff
@@ -0,0 +1,140 @@
...
+	});
+
+	it('should return false if it is basePrice', () => {
+		priceTypes.forEach((priceType) => {
+			const { isBasePrice, isDiscount, isPercent } = priceType;
+			const basePrice = isBasePrice === true && isDiscount === false && isPercent === false;
+
+			if (basePrice === true) {
+				expect(isFlatFeeSurcharge(priceType)).toBe(false);
+			}
```
nicolad: I've made these a bit explicit

nicolad: `if (basePrice === true)` then `.toBe(false)`

PR #2199, opened 2020-01-11
Run do_shortcode() on text in EVENT_META_* for Messages

dcdigit:
In case there are shortcodes in the post_meta, running do_shortcode() ensures the shortcode output is correctly parsed. Previously, shortcodes in post_meta (such as ACF wysiwyg fields) were ignored.

THis is a very simple patch to fix an issue I discovered while trying to use EVENT_META_* shortcodes to include text that itself has a shortcode in it. 

I maintain HTML text in a custom field using ACF. I use EVENT_META_* shortcode in EMAIL message templates to render this text. Previous to this patch, shortcodes in the post_meta custom field were not being parsed. After this patch, the shortcode is correctly parsed and the post_meta field text is included with the correctly parsed shortcode.

I don't believe this will have any adverse effects because it is just a small change to allow shortcodes to be used in post_meta data.

## How has this been tested

I modified the core and generated an email. I looked at the email - it contained no errors and carried the properly parsed shortcode output.

## Checklist

* [X] I have read the documentation relating to systems affected by this pull request, see https://github.com/eventespresso/event-espresso-core/tree/master/docs
* [X] User input is adequately validated and sanitized
* [X] all publicly displayed strings are internationalized (usually using `esc_html__()`, see https://codex.wordpress.org/I18n_for_WordPress_Developers)
* [X] My code is tested.
* [X] My code follows the Event Espresso code style.
* [X] My code has proper inline documentation.
* [X] My code accounts for when the site is in Maintenance Mode (MM2 especially disallows usage of models)

### DIFF
--- a/espresso.php
+++ b/espresso.php
@@ -3,7 +3,7 @@
   Plugin Name:Event Espresso
   Plugin URI: http://eventespresso.com/pricing/?ee_ver=ee4&utm_source=ee4_plugin_admin&utm_medium=link&utm_campaign=wordpress_plugins_page&utm_content=support_link
   Description: Manage events, sell tickets, and receive payments from your WordPress website. Reduce event administration time, cut-out ticketing fees, and own your customer data. | <a href="https://eventespresso.com/add-ons/?utm_source=plugin_activation_screen&utm_medium=link&utm_campaign=plugin_description">Extensions</a> | <a href="https://eventespresso.com/pricing/?utm_source=plugin_activation_screen&utm_medium=link&utm_campaign=plugin_description">Sales</a> | <a href="admin.php?page=espresso_support">Support</a>
-  Version: 4.10.3.rc.025
+  Version: 4.10.3.rc.026
   Author: Event Espresso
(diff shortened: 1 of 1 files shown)


Pebblo:
Hi Daniel,

Firstly, thank you for the pull request but unfortunately, I can't merge this in as-is.

It's a great idea and whilst this is a relatively small change it's actually a rather expensive one as `do_shortcode()` invokes a fairly intensive regex each time its called which is simply a waste for anyone that doesn't use shortcodes in event meta values or those that use them very rarely.

Rather than enabling this on every shortcode by default I think a better solution is to:

1. Add a do_shortcode attribute to the EVENT_META_* shortcode to enable this functionality on a per shortcode basis.
2. Add a filter to that same function that allows users to enable do_shortcode on every call should they wish.

That way there's no wasted resources for most users and devs such as yourself wouldn't need to go through and set the above attribute on each instance of the shortcode. Rather than passing this back requesting the above changes, I've created a PR to do it here:

#2209

joshfeck:
closing because we're going with #2209

PR #2202, opened 2020-01-13
Add Espresso Button and Icon Components

tn3rb:
This PR does the following:

- copies over the `EspressoButton` and `EspressoIcon` components from EDTRv1
- converts those components to TS
- extracts logic into separate files
- updates paths to components
- adds click event and typing to `AddNewDateButton` and `AddNewTicketButton`


assets/prototype/application/ui/components/display/espressoIcon/EspressoIcon.tsx
```diff
@@ -0,0 +1,60 @@
...
+			path = PATHS.REM;
+			break;
+
+		case ICONS.ROTATE:
+			path = PATHS.ROTATE;
+			break;
+
+		case ICONS.SAVE:
+			path = PATHS.SAVE;
+			break;
```
nicolad: It looks like a great candidate for an enum.

tn3rb: oh yeah I guess that would have been a better idea
i used enums for the button too, but didn't think to do so here!

assets/prototype/application/ui/components/input/espressoButton/EspressoButton.tsx
```diff
@@ -0,0 +1,64 @@
...
+		case EspressoButtonSize.SMALL:
+			iconSize = 21;
+			break;
+		case EspressoButtonSize.BIG:
+			iconSize = 27;
+			break;
+		case EspressoButtonSize.HUGE:
+			iconSize = 30;
+			break;
+	}
```
nicolad: I've used enums here as well:
image

tn3rb: good catch. I missed that. will fix

assets/prototype/application/ui/components/display/espressoIcon/isEspressoIcon.ts
```diff
@@ -0,0 +1,19 @@
+import { Icon } from './types';
+
+/**
+ * returns true if icon is an EspressoIcon
+ */
+const isEspressoIcon = (icon: string): boolean => {
```
nicolad: I am thinking if we can use just enums.

tn3rb: right... will do...
the only catch is that I still have to allow `string` as a default type, because the intent was to allow `EspressoIcon` to also be a wrapper for WP Dashicons where we could add additional styles or add other tweaks if need be

assets/prototype/application/ui/components/input/espressoButton/EspressoButton.tsx
```diff
@@ -28,26 +28,11 @@ const EspressoButton: React.FunctionComponent<EspressoButtonProps> = ({
...
-		case EspressoButtonSize.SMALL:
-			iconSize = 21;
-			break;
-		case EspressoButtonSize.BIG:
-			iconSize = 27;
-			break;
-		case EspressoButtonSize.HUGE:
-			iconSize = 30;
-			break;
-	}
```
nicolad: I love when I see code being removed :heart_eyes:

assets/prototype/application/ui/components/display/espressoIcon/isEspressoIcon.ts
nicolad: Good point, wasn't aware of that :+1:

assets/prototype/domain/eventEditor/datetimes/DatesList/AddNewDateButton.tsx
```diff
@@ -17,18 +17,14 @@ const btnRowStyle = {
...
-	const onClick = (e) => {
-		e.preventDefault();
-		e.stopPropagation();
+	const onClick = (): void => {
 		editorState.openEditor(editorId);
 	};
-	const onClose = (e) => {
-		e.preventDefault();
-		e.stopPropagation();
+	const onClose = (): void => {
```
nicolad: @tn3rb if this change makes sense, then I will need to go and apply this on `AddNewTicketButton`

tn3rb: done... pushed change

assets/prototype/application/ui/components/input/espressoButton/EspressoButton.tsx
```diff
@@ -39,9 +39,9 @@ const EspressoButton: React.FunctionComponent<EspressoButtonProps> = ({
 		classes += ' ee-noIcon';
 	}
 
-	const onClickHandler = (e) => {
-		e.preventDefault();
-		e.stopPropagation();
+	const onClickHandler: ClickEvent = (click) => {
```
nicolad: Thanks, I was just poking around in that commit.

PR #2209, opened 2020-01-14
Run EVENT_META_* values through do_shortcode

Pebblo:
See Daniel's initial PR for this to get an idea of why #2199 

This PR adds the ability to run your custom meta fields through do_shortcode using either an attribute set on `EVENT_META_*` or a filter:

`add_filter('FHEE__EventEspresso_core_libraries_shortcodes_EE_Event_Shortcodes___parser__event_meta_do_shortcode', '__return_true');`

~The filter name seems a little different from the norm and I'm more than happy to change it.~

## How has this been tested
Create a shortcode on your site:

```
function tw_ee_my_shortcode(){
	return 'This is some shortcode text';
}
add_shortcode( 'my_shortcode', 'tw_ee_my_shortcode' );
```

Or use whatever shortcode you already have if preferred.

Set that on a custom field within your event.

Example - https://monosnap.com/file/pW63fDb5ggVnW2DFYew5W07jAqu8Bv

Now edit a message template (Reg Approved) and add a call to EVENT_META_* anywhere that it is valid, obviously calling your custom meta field.

Example - https://monosnap.com/file/5YuiBTegWepIBTkQ7vLNIBQ52tMshw

Trigger a registration approved message however you see fit, I use Event Espresso -> Events -> {hover over event name} -> Registrations.

Click 'Resend Registration Details', generate and then preview the message, it should NOT parse the shortcode as is.

Add this to the site:

`add_filter('FHEE__EventEspresso_core_libraries_shortcodes_EE_Event_Shortcodes___parser__event_meta_do_shortcode', '__return_true');`

Retrigger, generate and preview the message again, the shortcode should parse.

Remove the above, edit the shortcode to have `do_shortcode=true`

Re-test, the shortcode should parse.

Any 'truthy' value on do_shortcode should work, 'falsey' values should not.

Confirm that `[EVENT_META_*something]` and `[EVENT_META_* something]` (with and without the  do_shortcode attribute) work as expected (it makes a difference with how the shortcode attributes parsed which is why the shortcode is stripped first)

----

Confirm the Help text (top tab on the right) shows the update text for EVENT_META_* and that it makes sense.

## Checklist

* [x] My code is tested.

### DIFF
--- a/core/libraries/shortcodes/EE_Event_Shortcodes.lib.php
+++ b/core/libraries/shortcodes/EE_Event_Shortcodes.lib.php
@@ -236,6 +236,7 @@ class EE_Event_Shortcodes extends EE_Shortcodes
 
         if (strpos($shortcode, '[EVENT_META_*') !== false) {
             // Strip the shortcode itself from $shortcode leaving any attributes set.
+            // Removing the * is correct here as _* is used to indiciate a dynamic shortcode.
             $shortcode = str_replace('[EVENT_META_*', '', $shortcode);
             $shortcode = trim(str_replace(']', '', $shortcode));
             // Get any attributes set on this shortcode.
--- a/core/libraries/shortcodes/EE_Event_Shortcodes.lib.php
+++ b/core/libraries/shortcodes/EE_Event_Shortcodes.lib.php
@@ -101,9 +101,13 @@ class EE_Event_Shortcodes extends EE_Shortcodes
                 'This will return the Twitter URL for the event if you have it set via custom field in your event, otherwise it will use the Twitter URL set in "Your Organization Settings". To set the facebook url in your event, add a custom field with the key as <code>event_twitter</code> and the value as your facebook url',
                 'event_espresso'
             ),
-            '[EVENT_META_*]'                          => __(
-                'This is a special dynamic shortcode. After the "*", add the exact name for your custom field, if there is a value set for that custom field within the event then it will be output in place of this shortcode.',
-                'event_espresso'
+            '[EVENT_META_*]'                          => sprintf(
+                __(
+                    'This is a special dynamic shortcode. After the "*", add the exact name for your custom field, if there is a value set for that custom field within the event then it will be output in place of this shortcode. If you use shortcodes within your custom fields set %1$sdo_shortcode=true%2$s at the end of the shortcode to run the value through the do_shortcode function. ',
+                    'event_espresso'
+                ),
+                '<code>',
+                '</code>'
             ),
             '[REGISTRATION_LIST_TABLE_FOR_EVENT_URL]' => __(
                 'This parses to the url for the registration list table filtered by registrations for this event.',
--- a/core/libraries/shortcodes/EE_Event_Shortcodes.lib.php
+++ b/core/libraries/shortcodes/EE_Event_Shortcodes.lib.php
@@ -250,7 +250,7 @@ class EE_Event_Shortcodes extends EE_Shortcodes
             }
             // Add a filter to allow all instances of EVENT_META_* to run through do_shortcode, default to false.
             // Check if a do_shortcode attribute was set to true and if so run $event_meta through that function.
-            if (apply_filters('FHEE__EE_Event_Shortcodes___parser__event_meta_do_shortcode', false)
+            if (apply_filters('FHEE__EventEspresso_core_libraries_shortcodes_EE_Event_Shortcodes___parser__event_meta_do_shortcode', false)
                 || !empty($attrs['do_shortcode']) && filter_var($attrs['do_shortcode'], FILTER_VALIDATE_BOOLEAN)
             ) {
                 return do_shortcode($event_meta);
--- a/core/libraries/shortcodes/EE_Event_Shortcodes.lib.php
+++ b/core/libraries/shortcodes/EE_Event_Shortcodes.lib.php
@@ -231,13 +231,28 @@ class EE_Event_Shortcodes extends EE_Shortcodes
         }
 
         if (strpos($shortcode, '[EVENT_META_*') !== false) {
+            // Strip the shortcode itself from $shortcode leaving any attributes set.
             $shortcode = str_replace('[EVENT_META_*', '', $shortcode);
             $shortcode = trim(str_replace(']', '', $shortcode));
-
-            // pull the meta value from the event post
-            $event_meta = $this->_event->get_post_meta($shortcode, true);
-
-            return ! empty($event_meta) ? $this->_event->get_post_meta($shortcode, true) : '';
+            // Get any attributes set on this shortcode.
+            $attrs = $this->_get_shortcode_attrs($shortcode);
+            // The meta_key set on the shortcode should always be the first value in the array.
+            $meta_key = $attrs[0];
+            // Pull the meta value from the event post.
+            $event_meta = $this->_event->get_post_meta($meta_key, true);
+            // If we have no event_meta, just return an empty string.
+            if (empty($event_meta)) {
+                return '';
+            }
+            // Add a filter to allow all instances of EVENT_META_* to run through do_shortcode, default to false.
+            // Check if a do_shortcode attribute was set to true and if so run $event_meta through that function.
+            if (apply_filters('FHEE__EE_Event_Shortcodes___parser__event_meta_do_shortcode', false)
+                || !empty($attrs['do_shortcode']) && filter_var($attrs['do_shortcode'], FILTER_VALIDATE_BOOLEAN)
+            ) {
+                return do_shortcode($event_meta);
+            }
+            // Still here? We just need to return the event_meta value as is.
+            return $event_meta;
         }
 
         if (strpos($shortcode, '[EVENT_TOTAL_AVAILABLE_SPACES_*') !== false) {


core/libraries/shortcodes/EE_Event_Shortcodes.lib.php
```diff
@@ -231,13 +235,28 @@ protected function _parser($shortcode)
         }
 
         if (strpos($shortcode, '[EVENT_META_*') !== false) {
+            // Strip the shortcode itself from $shortcode leaving any attributes set.
```
tn3rb: might want to add a comment that the `*` **IS** supposed to be kept and is not replaced like a wildcard normally would be

dcdigit:
Thanks for the note Tony. 

I'll try this method - seems optimal for most people's installations - I realize I am a bit of an outlier in this need but our installation requires some pretty intensive customizations using ACF and shortcodes.

I prefer adding the attribute to the EVENT_META_* shortcode and as a dev I would have been happy to discover this feature in the docs. 

I think layering in a filter is not needed because all this is doing is making sure that shortcodes are parsed when they are included in the ACF. I don't see any usage where someone would want to route the entire payload of the ACF and do custom programming. But, that's just my use case. There may be other devs out there in the future who will be happy to see EVENT_META_* being completely filterable.

Cheers!

core/libraries/shortcodes/EE_Event_Shortcodes.lib.php
```diff
@@ -231,13 +235,28 @@ protected function _parser($shortcode)
...
+            $meta_key = $attrs[0];
+            // Pull the meta value from the event post.
+            $event_meta = $this->_event->get_post_meta($meta_key, true);
+            // If we have no event_meta, just return an empty string.
+            if (empty($event_meta)) {
+                return '';
+            }
+            // Add a filter to allow all instances of EVENT_META_* to run through do_shortcode, default to false.
+            // Check if a do_shortcode attribute was set to true and if so run $event_meta through that function.
+            if (apply_filters('FHEE__EE_Event_Shortcodes___parser__event_meta_do_shortcode', false)
```
tn3rb: ya the full naming convention we use for hooks includes the FQCN namespace, so this should be:

```
FHEE__EventEspresso_core_libraries_shortcodes_EE_Event_Shortcodes___parser__event_meta_do_shortcode
```

that way anybody seeing the filter name in use anywhere will know exactly where to look for it

core/libraries/shortcodes/EE_Event_Shortcodes.lib.php
```diff
@@ -231,13 +235,28 @@ protected function _parser($shortcode)
...
+                || !empty($attrs['do_shortcode']) && filter_var($attrs['do_shortcode'], FILTER_VALIDATE_BOOLEAN)
```
tn3rb: great job 👍 

thnx for using `filter_var()`

core/libraries/shortcodes/EE_Event_Shortcodes.lib.php
dcdigit: Totally agree with this comment about the need for a comment on this. My first instinct (in the absence of explicit instructions was to replace the star with the field name. Then, I added the name without a space, then I finally got it with the space. It took just a few minutes,, but it would benice to spare someone the trial and error.

Pebblo: The shortcode help text already explicitly states to add your custom field **after** the * so I'm not sure what else I should do here as all of the _* shortcodes work in exactly the same way.

`[EXAMPLE_SHORTCODE_*{value} {attribute} {attribute}]`

(Some shortcodes don't have a 'value' to pass and some don't have attributes but the format for them all remains the same)

The help text is available for all shortcodes valid in the current template in the Message Template Editor -> Help (tab at top right) -> Message Shortcodes.

See the updated (adds some info about the do_shortcode parameter) help text here: https://github.com/eventespresso/event-espresso-core/pull/2209#pullrequestreview-342235261

Doc: https://eventespresso.com/wiki/messages-system-shortcodes/

More than happy to add a comment in the code but the docs and help text already show how to use the dynamic shortcodes (shortcodes with `_*`) and instructions for each specific shortcode so if those are already being missed then this comment is only going to help the select few who go through the code.

Also, a space before the custom field name shouldn't make any difference at all so I'm not sure what is happening there, the shortcode trims the value before using it for this exact reason. Here is a POC using the original code the parser ran before the changes here:

https://monosnap.com/file/bTEjn47YLLqtU6aE4myl6OzFFGHAjR

Both shortcodes return exactly the same value to parse.

The new code returns exactly the same values for both of those shortcodes.

Pebblo:
> I'll try this method - seems optimal for most people's installations - I realize I am a bit of an outlier in this need but our installation requires some pretty intensive customizations using ACF and shortcodes.

Yeah, that's why I think this approach is better here, no point invoking regex when the majority of users may not use it but having the option available is a nice addition.

> I prefer adding the attribute to the EVENT_META_* shortcode and as a dev I would have been happy to discover this feature in the docs.

Sorry, I'm not sure I follow this, which feature are you referring to?

Passing the above shortcode an additional attribute other than the meta key value wasn't available until the changes in this branch so we wouldn't have that in the docs already.

> I think layering in a filter is not needed because all this is doing is making sure that shortcodes are parsed when they are included in the ACF. I don't see any usage where someone would want to route the entire payload of the ACF and do custom programming. But, that's just my use case. There may be other devs out there in the future who will be happy to see EVENT_META_* being completely filterable.

If I understand your comment correctly then the filter I've added doesn't do any of the above, it is basically a filtered default value for the `do_shortcode` attribute on EVENT_META_*

Meaning rather than adding `do_shortcode=true` to each instance of the above shortcode, if you know that the majority of your custom fields should be passed through do_shortcode() you can add a callback return true and just use the shortcode as is and each instance will then parse shortcodes (as it was in your original PR).

core/libraries/shortcodes/EE_Event_Shortcodes.lib.php
```diff
@@ -101,9 +101,13 @@ protected function _init_props()
...
                 'event_espresso'
             ),
-            '[EVENT_META_*]'                          => __(
-                'This is a special dynamic shortcode. After the "*", add the exact name for your custom field, if there is a value set for that custom field within the event then it will be output in place of this shortcode.',
-                'event_espresso'
+            '[EVENT_META_*]'                          => sprintf(
+                __(
+                    'This is a special dynamic shortcode. After the "*", add the exact name for your custom field, if there is a value set for that custom field within the event then it will be output in place of this shortcode. If you use shortcodes within your custom fields set %1$sdo_shortcode=true%2$s at the end of the shortcode to run the value through the do_shortcode function. ',
+                    'event_espresso'
+                ),
```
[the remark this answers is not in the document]
Pebblo: Nope, none of the help text strings for message shortcodes use `esc_html__()`

I can go through and change them all if preferred?

dcdigit:
> Sorry, I'm not sure I follow this, which feature are you referring to?

Just saying thank you for building this in the EE way! Now, future devs will be able to discover how to handle shortcodes in the EVENT_META_* message shortcode.

core/libraries/shortcodes/EE_Event_Shortcodes.lib.php
Pebblo: Thanks, I pushed a fix for this.

core/libraries/shortcodes/EE_Event_Shortcodes.lib.php
Pebblo: Thank you, good sir.

Pebblo:
Passing this back to Brent so he can see my replies above.

core/libraries/shortcodes/EE_Event_Shortcodes.lib.php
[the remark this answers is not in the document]
tn3rb: ya it's preferable to change any uses of `__()` to `esc_html_()`. I just wasn't sure if there was any reason for that not being the case with the messages code, ie: was there anything weird in the messages logic that prevented the use of `esc_html_()`.

core/libraries/shortcodes/EE_Event_Shortcodes.lib.php
tn3rb: the purpose of an additional comment at that point in the code would be for the benefit of future developers who are looking at the code and may not have ever looked at the help text or documentation. Basically it's to tell them that the presence of the `*` in that `str_replace()` is correct and not a mistake like both myself and @dcdigit initially thought.

Even something like:

```php
// yes the * is correct !!!
if (strpos($shortcode, '[EVENT_META_*') !== false) {
```

would let anyone reading the code that that is intentional and not a mistake.

tn3rb:
responded to comments directly

core/libraries/shortcodes/EE_Event_Shortcodes.lib.php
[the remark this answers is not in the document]
Pebblo: kk, I'll do that in another pr

core/libraries/shortcodes/EE_Event_Shortcodes.lib.php
Pebblo: Ah, kk. Will do, thanks.

joshfeck:
All testing items pass. Nice work @Pebblo

Issue #2194, opened 2020-01-10
Testing Cascading Deletion of Events


mnelson4:
## Issue Overview

Although I've been getting some feedback during development, I haven't been putting work on hold for it too much either.
Anyways, with all the pieces in place, this needs to be officially tested.

## Testing Checklist
* [ ] Create an event with no registrations. Trash it then permanently delete it. It should be deleted fine (you may want to check the database that its datetimes, tickets, message-group relations, etc, were removed properly.)
* [ ] Create an event with some registrations and do the same. Those registrations should also be deleted.
* Visit the transaction pages for one of those registrations. It won't have any registrations, but it should still be a valid record of a real transaction and payment. (Deleting transactions and payments is another job for another day.)
* [ ] Create an event that uses some custom questions, payments, and registrations and do the same. Check those registrations' answer rows are also deleted.
* Create a few events with registrations, and bulk trash them then bulk permanently delete them. They should all get deleted fine. 
* [ ] Repeat the above but with some add-ons active that add data related to events or registrations. Like promotions (which has `Promotion_Object` model which relates promotions to events; or Infusionsoft which records extra meta about events that are sync'd; or MailChimp which has other tables pointing to events; or payment methods.) It may not be a hard requirement that this always delete those add-ons' data, but it certainly shouldn't have any errors, and it would be great if this PR did clean up their data.
* [ ] Create an event with 1000 registrations and do the same. It will take a few minutes to generate the preview page, and a minute to delete, but it should work. Preferably do this test from a cheap shared host (I have one on Dreamhost if that helps). Note though the event can't be infinitely big, I know it doesn't work deleting an event with over 15,000 registrations (in which I outlined a solution on https://github.com/eventespresso/event-espresso-core/pull/2110#issuecomment-570462637)

joshfeck:
Has there been consideration for how the orphaned transactions will be deleted?

The current functionality when you trash, then permanently delete a registration is it will also delete the transaction too. It seems like the transaction should also be deleted too. Otherwise how will they be deleted without going in and directly running deleted commands on the database?

mnelson4:
> Has there been consideration for how the orphaned transactions will be deleted?

Good thing to bring up. Looking back over the earlier discussions [you mentioned it](https://github.com/eventespresso/event-espresso-core/issues/349#issuecomment-549498120), I must have glossed over the fact there's no UI to delete transactions. 

[...]

PR #2225, opened 2020-01-15
delete transactions too if theres no other valid registrations

mnelson4:
## Note to Brent
This first commit is ugly. I just wanted to first focus on getting the logic right. Next, I'll point it out in the code, I think there's definitely a clear opportunity for using the command pattern, but you may see other worthwhile code improvements.

When we're deleting an event and its registrations, also deletes the registration's transaction IF there are no other registrations for other events. Also deletes the transaction's related data too (like line items, payments, messages, and extra meta.) If there are other registrations on the transaction that are for other events, the transaction is not deleted, nor are those other registrations (but the registrations for the event being deleted ARE also deleted.)
See the related discussion from https://github.com/eventespresso/event-espresso-core/issues/2194#issuecomment-573218561

## How has this been tested

* [ ] Register for a paid event. Look directly in the DB for the registration's transaction, its payment, messages, and extra metas.
* Delete that event.
* Check that the transaction, payment, messages and extra metas are also deleted.
* [ ] Use MER. Simultaneously register for two events. 
* Delete one of the events
* Double-check the transaction you just created wasn't deleted; although the registration for the deleted event should be deleted.

## Checklist

* [ ] My code is tested.

### DIFF
--- a/core/libraries/batch/JobHandlers/ExecuteBatchDeletion.php
+++ b/core/libraries/batch/JobHandlers/ExecuteBatchDeletion.php
@@ -4,6 +4,7 @@ namespace EventEspressoBatchRequest\JobHandlers;
 
 use EE_Change_Log;
 use EE_Registry;
+use EE_Transaction;
 use EEM_Event;
 use EEM_Price;
 use EEM_Ticket;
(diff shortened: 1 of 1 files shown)


core/libraries/batch/JobHandlers/ExecuteBatchDeletion.php
```diff
@@ -147,16 +161,117 @@ public function continue_job(JobParameters $job_parameters, $batch_size = 50)
...
+        foreach ($reg_ids as $reg_id) {
+            $count_deleted += $this->deleteRegistrationAndTransaction($reg_id, $job_parameters);
+            if ($count_deleted >= $batch_size) {
+                return $count_deleted;
+            }
+        }
+        return $count_deleted;
+    }
+
+    /**
```
mnelson4: I bet the contents of this method would be appropriate for a command

core/libraries/batch/JobHandlers/ExecuteBatchDeletion.php
```diff
@@ -147,16 +161,117 @@ public function continue_job(JobParameters $job_parameters, $batch_size = 50)
...
+                        'REG_deleted' => false,
+                        'Event.status' => ['!=', EEM_Event::post_status_trashed]
+                    ]
+                ]
+            );
+            if (! $valid_regs) {
+                // delete the transaction
+                // and its dependent data
+                $transaction_as_root_of_tree = new ModelObjNode($TXN->ID(), $TXN->get_model());
+                $transaction_as_root_of_tree->visit(1000);
```
mnelson4: I thought it was reasonable to assume a transaction won't have more than 1000 dependent database rows (eg registrations, extra metas, payments, line items, etc).

core/libraries/batch/JobHandlers/ExecuteBatchDeletion.php
```diff
@@ -147,16 +161,117 @@ public function continue_job(JobParameters $job_parameters, $batch_size = 50)
...
+                $ids_to_delete = $transaction_as_root_of_tree->getIds();
+                foreach ($ids_to_delete as $model_name => $ids) {
+                    // just grab the IDs the easy way, there's no Term_Relationship model related to transactions
+                    $model_to_delete = EE_Registry::instance()->load_model($model_name);
```
tn3rb: confused by this comment... where are we grabbing the IDs other than from the foreach variable?

core/libraries/batch/JobHandlers/ExecuteBatchDeletion.php
```diff
@@ -98,37 +107,42 @@ public function continue_job(JobParameters $job_parameters, $batch_size = 50)
...
+                            $where_conditions
+                        ],
+                        false
+                    );
+                    $units_processed += count($ids_to_delete_this_query);
+                    $remaining_ids = array_diff_key($ids_to_delete, $ids_to_delete_this_query);
+                    // If there's any more from this model, we'll do them next time.
+                    if (count($remaining_ids) > 0) {
+                        $models_and_ids_remaining[ $model_name ] = $remaining_ids;
                     }
```
tn3rb: would increase readability if this was extracted into a new method and then simply called here in this `else` block
but that can be done later

core/libraries/batch/JobHandlers/ExecuteBatchDeletion.php
tn3rb: eee... ya... I would hope so, but I guess you never know... does that included answers as well?

core/libraries/batch/JobHandlers/ExecuteBatchDeletion.php
mnelson4: Ya that doesn't make a lot of sense. I meant we can rely on the IDs actually just being simple int or string primary keys. They're not going to be those weird strings of key-value pairs we need to use for the `EEM_Term_Relationship` model (like we need to use for https://github.com/eventespresso/event-espresso-core/blob/CASC/base/core/libraries/batch/JobHandlers/ExecuteBatchDeletion.php#L113
If that makes sense I will update the comment (if not let's discuss some more.)

core/libraries/batch/JobHandlers/ExecuteBatchDeletion.php
mnelson4: oh yeah it does include answers... right some people sometimes have hundreds of answers (but I seriously doubt they'll make a hundred of registrations on a single transaction, right?)
If this isn't a reasonable assumption, I think we may need to try to discover what transaction we want to delete when generating the preview page instead. Plus then we wouldn't have this weird thing where the batch deletion job's "job size" can grow...

mnelson4: I created https://github.com/eventespresso/event-espresso-core/pull/2247 which can help this ticket with that issue; but I also created https://github.com/eventespresso/event-espresso-core/pull/2248 based off it which is an alternative to this PR

mnelson4:
Note: this PR looks like it will be replaced by https://github.com/eventespresso/event-espresso-core/pull/2247 and https://github.com/eventespresso/event-espresso-core/pull/2248
but I've kept it around in case we realize to that implementation has problems and want this hackier one.

mnelson4:
leaving branch for now until the [two alternative PRs](https://github.com/eventespresso/event-espresso-core/pull/2225#issuecomment-578853360) are merged into `CASC/base`

PR #2245, opened 2020-01-16
Reduce PUE load 

Pebblo:
PUE calls home separately for EE core and each add-on individually.

That means that if you have 10/20/30 EE plugins active on your site, when PUE initiates a call home it sends 10/20/30 GET requests home, the majority of which technically do little more than check for a new version of the plugin which doesn't sound that bad, right?

The problem is that PUE checks for an active license on each of those requests and its hammering our servers.

The change here basically adds a 'latestVersion' value to the pue state for each add-on which PUE then compares against the installed version **before** calling home for add-ons but will **always** call home for core. This branch basically does almost nothing at all without some changes to PUE itself (to include the plugin versions in the response sent back to EE when it calls home).

The changes to EE_Addon and EE_Register_Addon.lib.php set up the PUE slugs on EE_Add-ons so they can be used later.

The additional code in [PluginUpdateEngineChecker::requestInfo()](https://github.com/eventespresso/event-espresso-core/compare/ENH/reduce-pue-load?expand=1#diff-91de7c437e6013f92801e60bff591f47R1035-R1053) grabs the new 'plugins' data that will be included in the PUE response which will just be associative array with `{pue_plugin_slug} => {latestVersion}`.

If that additional data is included in the response that code loops over any add-ons you have installed on the site and saves the latestVersion locally for that add-on.

The changes to [PluginUpdateEngineChecker::checkForUpdates()](https://github.com/eventespresso/event-espresso-core/compare/ENH/reduce-pue-load?expand=1#diff-91de7c437e6013f92801e60bff591f47R1665-R1671) check for a `latestVersion` value right before calling home, if a value is set and the currently installed version number is greater than or equal to the latestVersion value we saved earlier we return out of the function, there's no update so we don't need to call home for this add-on.

----

Incase this isn't clear above, this means that in order for EE to show an update is available for any add-ons, EE core has to call home and update the local values of latestVersion on each of the installed add-ons first. Then each add-on will compare those values before they call home themsevles (if needed).

The nice thing about this is everything else works exactly how it did, when there's an update available EE calls home for that add-on and will display the No valid license notices if they don't have one etc.

---

Worst case scenario, if this breaks something with updates is that core may need to manually updated to fix it (I've gone over this many times locally and I don't think there will be an issue) but because there is only a single instance of pue-client (EE core handles PUE for all add-ons) any issues with add-ons can be fixed by updating core (for example if add-ons stop updating and update for core can be released to remove the latestVersion check and go back to the current method). I don't expected this at all but just pointing it out.

--- 

I'll go through some testing notes on this shortly but I'd like Brent to have a look over the code first incase I'm missing something.

## Checklist

* [x] My code is tested.

### DIFF
--- a/core/EE_Addon.core.php
+++ b/core/EE_Addon.core.php
@@ -88,6 +88,13 @@ abstract class EE_Addon extends EE_Configurable implements RequiresDependencyMap
      */
     protected $_main_plugin_file;
 
+    /**
+     *    This is the slug used to identify this add-on within the plugin update engine.
+     *
+     * @type string
+     */
+    protected $pue_slug;
+
 
     /**
      * @var EE_Dependency_Map $dependency_map
@@ -883,5 +890,20 @@ abstract class EE_Addon extends EE_Configurable implements RequiresDependencyMap
     public function after_registration()
     {
         // cricket chirp... cricket chirp...
+    }
+
+    /**
+     * @return mixed
+     */
+    public function getPueSlug()
+    {
+        return $this->pue_slug;
+    }
+    /**
+     * @param mixed $pue_slug
+     */
+    public function setPueSlug($pue_slug)
+    {
+        $this->pue_slug = $pue_slug;
     }
 }
--- a/core/EE_Addon.core.php
+++ b/core/EE_Addon.core.php
@@ -893,7 +893,7 @@ abstract class EE_Addon extends EE_Configurable implements RequiresDependencyMap
     }
 
     /**
-     * @return mixed
+     * @return string
      */
     public function getPueSlug()
     {
--- a/core/EE_Addon.core.php
+++ b/core/EE_Addon.core.php
@@ -900,7 +900,7 @@ abstract class EE_Addon extends EE_Configurable implements RequiresDependencyMap
         return $this->pue_slug;
     }
     /**
-     * @param mixed $pue_slug
+     * @param string $pue_slug
      */
     public function setPueSlug($pue_slug)
     {
--- a/core/third_party_libs/pue/pue-client.php
+++ b/core/third_party_libs/pue/pue-client.php
@@ -1663,7 +1663,7 @@ if (! class_exists('PluginUpdateEngineChecker')):
             $state = get_site_option($this->optionName);
 
             // If this is an add-on, only call home if there is an update available for it.
-            if( $this->slug !== 'event-espresso-core-reg' && isset($state->latestVersion) ) {
+            if( $this->slug !== 'event-espresso-core-reg' && !empty($state->latestVersion) ) {
                 if( version_compare($this->_installed_version, $state->latestVersion, '>=') ) {
                     return;
                 }
--- a/core/third_party_libs/pue/pue-client.php
+++ b/core/third_party_libs/pue/pue-client.php
@@ -1033,8 +1033,10 @@ if (! class_exists('PluginUpdateEngineChecker')):
                     $this->add_persistent_notice($response->extra_notices);
                 }
                 // This instance is for EE Core, we have a response from PUE so lets check if it contains PUE Plugin data.
-                if ($this->slug == 'event-espresso-core-reg' && isset($response->extra_data) && !empty($response->extra_data->plugins) ) {
-                    // Pull PUE pugin data from 'extra_data'.
+                if ( stripos($this->slug, 'event-espresso-core') !== false
+                    && isset($response->extra_data)
+                    && !empty($response->extra_data->plugins)
+                ) {                    // Pull PUE pugin data from 'extra_data'.
                     $plugins_array = json_decode($result['body'], true);
                     $plugins = $plugins_array['extra_data']['plugins'];
                     // Pull all of the add-ons EE has active and update the local latestVersion value of each of them.
--- a/core/third_party_libs/pue/pue-client.php
+++ b/core/third_party_libs/pue/pue-client.php
@@ -1048,7 +1048,7 @@ if (! class_exists('PluginUpdateEngineChecker')):
                                 continue;
                             }
                             // Have an addon state? Set the latestVersion value!
-                            $addon_state->latestVersion = $response->extra_data->plugins->{ $addon_slug };
+                            $addon_state->latestVersion = $response->extra_data->plugins->{ $addon_slug }->version;
                             update_option('external_updates-' . $addon_slug, $addon_state);
                         }
                     }
--- a/core/third_party_libs/pue/pue-client.php
+++ b/core/third_party_libs/pue/pue-client.php
@@ -1657,17 +1657,17 @@ if (! class_exists('PluginUpdateEngineChecker')):
         {
             $state = get_site_option($this->optionName);
 
+            // If this is an add-on, only call home if there is an update available for it.
+            if( $this->slug !== 'event-espresso-core-reg' && isset($state->latestVersion) ) {
+                if( version_compare($this->_installed_version, $state->latestVersion, '>=') ) {
+                    return;
+                }
+            }
+
             if (empty($state) || ! is_object($state)) {
                 $state = new StdClass;
                 $state->checkedVersion = '';
                 $state->update = null;
-            }
-
-            // If this is an add-on, only ping ee.com if there is an update actually available for it.
-            if( $this->slug !== 'event-espresso-core-reg' && isset($state->latestVersion) ) {
-                if( version_compare($state->latestVersion, $this->_installed_version, '<=') ) {
-                    return;
-                }
             }
 
             $state->lastCheck = time();
--- a/core/third_party_libs/pue/pue-client.php
+++ b/core/third_party_libs/pue/pue-client.php
@@ -1032,6 +1032,20 @@ if (! class_exists('PluginUpdateEngineChecker')):
                 if (isset($response->extra_notices)) {
                     $this->add_persistent_notice($response->extra_notices);
                 }
+                if ($this->slug == 'event-espresso-core-reg' && isset($response->extra_data) && !empty($response->extra_data->plugins) ) {
+                    $plugins_array = json_decode($result['body'], true);
+                    $plugins = $plugins_array['extra_data']['plugins'];
+                    foreach (EE_Registry::instance()->addons as $addon) {
+                        if( isset($plugins[ $addon->getPueSlug() ]) ) {
+                            $addon_state = get_option('external_updates-' . $addon->getPueSlug());
+                            if( empty($addon_state)) {
+                                continue;
+                            }
+                            $addon_state->latestVersion = $plugins[$addon->getPueSlug()];
+                            update_option('external_updates-' . $addon->getPueSlug(), $addon_state);
+                        }
+                    }
+                }
             }
 
 
--- a/core/third_party_libs/pue/pue-client.php
+++ b/core/third_party_libs/pue/pue-client.php
@@ -1040,15 +1040,16 @@ if (! class_exists('PluginUpdateEngineChecker')):
                     $plugins = $plugins_array['extra_data']['plugins'];
                     // Pull all of the add-ons EE has active and update the local latestVersion value of each of them.
                     foreach (EE_Registry::instance()->addons as $addon) {
-                        if( isset($plugins[ $addon->getPueSlug() ]) ) {
-                            $addon_state = get_option('external_updates-' . $addon->getPueSlug());
+                        $addon_slug = $addon->getPueSlug();
+                        if( isset($response->extra_data->plugins->{ $addon_slug }) ) {
+                            $addon_state = get_option('external_updates-' . $addon_slug);
                             // If we don't have an addon state, get out we'll update it next time.
                             if( empty($addon_state)) {
                                 continue;
                             }
                             // Have an addon state? Set the latestVersion value!
-                            $addon_state->latestVersion = $plugins[$addon->getPueSlug()];
-                            update_option('external_updates-' . $addon->getPueSlug(), $addon_state);
+                            $addon_state->latestVersion = $response->extra_data->plugins->{ $addon_slug };
+                            update_option('external_updates-' . $addon_slug, $addon_state);
                         }
                     }
                 }
--- a/core/third_party_libs/pue/pue-client.php
+++ b/core/third_party_libs/pue/pue-client.php
@@ -1032,15 +1032,20 @@ if (! class_exists('PluginUpdateEngineChecker')):
                 if (isset($response->extra_notices)) {
                     $this->add_persistent_notice($response->extra_notices);
                 }
+                // This instance is for EE Core, we have a response from PUE so lets check if it contains PUE Plugin data.
                 if ($this->slug == 'event-espresso-core-reg' && isset($response->extra_data) && !empty($response->extra_data->plugins) ) {
+                    // Pull PUE pugin data from 'extra_data'.
                     $plugins_array = json_decode($result['body'], true);
                     $plugins = $plugins_array['extra_data']['plugins'];
+                    // Pull all of the add-ons EE has active and update the local latestVersion value of each of them.
                     foreach (EE_Registry::instance()->addons as $addon) {
                         if( isset($plugins[ $addon->getPueSlug() ]) ) {
                             $addon_state = get_option('external_updates-' . $addon->getPueSlug());
+                            // If we don't have an addon state, get out we'll update it next time.
                             if( empty($addon_state)) {
                                 continue;
                             }
+                            // Have an addon state? Set the latestVersion value!
                             $addon_state->latestVersion = $plugins[$addon->getPueSlug()];
                             update_option('external_updates-' . $addon->getPueSlug(), $addon_state);
                         }
--- a/core/third_party_libs/pue/pue-client.php
+++ b/core/third_party_libs/pue/pue-client.php
@@ -1662,6 +1662,14 @@ if (! class_exists('PluginUpdateEngineChecker')):
                 $state->checkedVersion = '';
                 $state->update = null;
             }
+
+            // If this is an add-on, only ping ee.com if there is an update actually available for it.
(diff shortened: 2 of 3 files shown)


core/EE_Addon.core.php
```diff
@@ -884,4 +891,19 @@ public function after_registration()
     {
         // cricket chirp... cricket chirp...
     }
+
+    /**
+     * @return mixed
```
tn3rb: i didn't add a type to the `$pue_slug` when I whipped up this code example, so my IDE used `mixed` for the type here since it didn't know this should be a string

```suggestion
     * @return string
```

core/EE_Addon.core.php
```diff
@@ -884,4 +891,19 @@ public function after_registration()
...
+     */
+    public function getPueSlug()
+    {
+        return $this->pue_slug;
+    }
+    /**
+     * @param mixed $pue_slug
```
tn3rb:
```suggestion
     * @param string $pue_slug
```

core/third_party_libs/pue/pue-client.php
```diff
@@ -1032,6 +1032,25 @@ private function requestInfo(array $queryArgs = array())
                 if (isset($response->extra_notices)) {
                     $this->add_persistent_notice($response->extra_notices);
                 }
+                // This instance is for EE Core, we have a response from PUE so lets check if it contains PUE Plugin data.
+                if ($this->slug == 'event-espresso-core-reg' && isset($response->extra_data) && !empty($response->extra_data->plugins) ) {
+                    // Pull PUE pugin data from 'extra_data'.
+                    $plugins_array = json_decode($result['body'], true);
+                    $plugins = $plugins_array['extra_data']['plugins'];
```
tn3rb: can you not use `$response->extra_data->plugins` here? why decode again?

you've already checked that it exists... or does it not contain the same data that's in raw response?

core/third_party_libs/pue/pue-client.php
```diff
@@ -1032,6 +1032,25 @@ private function requestInfo(array $queryArgs = array())
...
+                    foreach (EE_Registry::instance()->addons as $addon) {
+                        if( isset($plugins[ $addon->getPueSlug() ]) ) {
+                            $addon_state = get_option('external_updates-' . $addon->getPueSlug());
+                            // If we don't have an addon state, get out we'll update it next time.
+                            if( empty($addon_state)) {
+                                continue;
+                            }
+                            // Have an addon state? Set the latestVersion value!
+                            $addon_state->latestVersion = $plugins[$addon->getPueSlug()];
+                            update_option('external_updates-' . $addon->getPueSlug(), $addon_state);
```
tn3rb: how does this ever get set the very first time?

initially this option will not exist in the db right? or was this something that already existed in PUE?

core/third_party_libs/pue/pue-client.php
```diff
@@ -1643,11 +1662,19 @@ public function checkForUpdates()
         {
             $state = get_site_option($this->optionName);
 
+            // If this is an add-on, only call home if there is an update available for it.
+            if( $this->slug !== 'event-espresso-core-reg' && !empty($state->latestVersion) ) {
+                if( version_compare($this->_installed_version, $state->latestVersion, '>=') ) {
+                    return;
+                }
+            }
```
tn3rb: ultimately comes down to a simple yet brilliant fix... nice job 🥇

core/third_party_libs/pue/pue-client.php
Pebblo: `json_decode($result['body']);` returns stdClass objects.

`json_decode($result['body'], true)` returns assoc arrays.

Soooo, the short answer is I couldn't figure out how to use the stdClass objects without converting them to arrays anyway so just did the lot as I knew how to work with those.

However, I only need `$plugins_array['extra_data']['plugins']` so maybe I should just decode just that with:

`$plugins_array = json_decode($result['body'], true);` 

as `$response->extra_data->plugins` tells me if its empty anyway?

Or am I missing something obvious with the stdClass objects?

core/third_party_libs/pue/pue-client.php
Pebblo: It already exists within PUE, it's the option for each plugin that stores the update checks book-keeping ('checkPerion', 'update' details like the PUE download url for add-on etc), so it will already be set on any add-ons installed on the site, if it's not then on the next call home it will be.

core/third_party_libs/pue/pue-client.php
Pebblo: Thanks :)

Pebblo:
> there's a couple little fixes needed that were my fault cuz the code I told you to add wasn't 100% correct (just phpdoc changes, so not important)

No worries, I actually noticed those and meant to change them before committing.

> overall though is that this looks like a really brilliant fix for a fugly problem.
> Good Job

Thanks :)

Eventually, I'd like to refactor PUE and do pretty much everything you mentioned in slack if possible.

Pebblo:
Sorry brent, Josh found an issue with RC versions of core, which highlighted an issue with how Decaf updates worked.

The 'lastestVersions' of add-ons would only update when the slug matched the regular version of EE which meant that anyone running decaf and add-ons would still ping ee multiple times, this prevents that.

Note that RC versions of core are pretty much treated like decaf (wrt to PUE).

Also switch our some preg_match() calls to just use stripos(), theres no fancy processing just looking for 'p', 'decaf' and 'beta' in the currently installed version so no need for the big guns.

core/third_party_libs/pue/pue-client.php
tn3rb: > It already exists within PUE

so this same property gets set via some other code in PUE?


> it's the option for each plugin that stores the update checks book-keeping

...mmm... sry... not understanding this sentence... what are the "update checks book-keeping" ???

> so it will already be set on any add-ons installed on the site

... ok... but when/where does that happen? Again, is there some other code that saves this option? If so then won't this conflict with that other logic? ie: two blocks of code updating an option that's intended to store the current state of affairs sounds potentially fragile.

> if it's not then on the next call home it will be.

But isn't THIS the code that does the calling home? if the option isn't set, then on the next call home, won't it return as soon as it hits this block:

```php
if( empty($addon_state)) {
	continue;
}
```

sry not trying to be a PITA, just trying to ensure that everything is as solid as it can be

core/third_party_libs/pue/pue-client.php
tn3rb: it looks like you only need to access

```php
$plugins[ $addon->getPueSlug() ]
```

which you should be able to access with a `stdClass` object like this:

```php
$plugins->{ $addon->getPueSlug() }
```

core/third_party_libs/pue/pue-client.php
Pebblo: >so this same property gets set via some other code in PUE?

By property do you mean `$state->latestVersion` or `$state` as a whole?

`$state->latestVersion` is only set within this code and I chose this location because it is after PUE has set up `$state` for the current plugin, note that the code in question here only runs on EE core requests.

The only place that creates the 'state' object is in [checkForUpdates()](https://github.com/eventespresso/event-espresso-core/blob/ENH/reduce-pue-load/core/third_party_libs/pue/pue-client.php#L1662-L1686), which calls [requestUpdate(), which calls requestInfo()](https://github.com/eventespresso/event-espresso-core/blob/ENH/reduce-pue-load/core/third_party_libs/pue/pue-client.php#L1336-L1343) (which is where this code is)

That code to set `latestVersion` on each add-on `$state` can only run if we have a `$state` for that add-on, right? The call home happens if `$state->latestVerion` is empty or it is greater than the currently installed version. So I'm confirming we have a state object for the add-ons before setting `latestVersion` on them. The add-ons need to call home at least once and will do so when installed, that sets up `$state`, and it's set on the current request even before this code runs in checkForUpdates().

So, when we activate an add-on (let's use the calendar as an example), it runs through PUE and gets to `checkForUpdates()`, at that point it doesn't have a `$state` set yet.

Because it doesn't have a `$state`, my new version number check is skipped. `$state` is created and saved to the database (at that point is holds the current time in (`$state->lastCheck) and the currently installed version (`$state->checkedVersion`) and has a `update` property set to null (that property holds the update values WP needs to run an update on the plugin and is used only when an update is actually available).

The above is also what I meant about the 'book-keeping' regarding updates, `$state` stores when the last check was done, the update values etc, everything concerning this current plugins 'PUE State' is stored in this option.

> .. ok... but when/where does that happen? Again, is there some other code that saves this option? If so then won't this conflict with that other logic? ie: two blocks of code updating an option that's intended to store the current state of affairs sounds potentially fragile.

This is the only location that saves the `latestVersion` property on `$state` for each add-on that has a state.

I'm not sure I follow how this is fragile, should I not be updating an option after its been created?

Nowhere else uses `latestVersion`, they do use state but don't loop over the properties of it unsetting them etc.

> But isn't THIS the code that does the calling home? if the option isn't set, then on the next call home, won't it exit the loop as soon as it hits this block

The code to save `latestVersion` on each add-on only runs on a call home from core, so no.

Using the EE calendar add-on example from above, when you install that add-on it will call home and when it does it creates its `$state` and saves it, this code in question doesn't not run on that call as its not EE core. So then when core calls home it updates all add-ons states with their latest versions, if we don't have a state for EE Calendar something is up and I don't want to create one as PUE should be doing it.

The idea behind this method is to leave PUE to do whatever it does already unless `latestVersion` has been set and its value is equal to the current version (or less than but that's unlikely without people hacking stuff up so that's on them). `latestVersion` is only set on an add-on when we have a state for it (PUE did its thing as normal for that add-on) and EE4 core called home to pull the latestVersions and set them on the add-ons states.

> sry if I'm totally missing something, not trying to be a PITA, just trying to ensure that everything is as solid as it can be

I have 0 issues with you asking for more details on anything I've written, I learn a huge amount from you guys questioning what and why :)

core/third_party_libs/pue/pue-client.php
Pebblo: Ahhh! Ok... this make sense.

I'll check this out now, thank you!

Pebblo:
Going through the `stdClass` stuff next but passing this back in case there something that needs changing from your questions above.

Sorry if it's not really clear, I can go through it more in Slack if that helps?

core/third_party_libs/pue/pue-client.php
tn3rb: > By property do you mean $state->latestVersion or $state as a whole?

oh dang... no... sry that was a bad word choice... 
I meant **option** not property, as in the option that is saved using the key `'external_updates-' . $addon->getPueSlug()` (which I'll simply refer to as the "external_updates" option from here on in)

ok I shovelled through the PUE code some more after reading the above and think I grok what's going on...

So when each individual addon is registered, `EE_Register_Addon::load_pue_update()` instantiates an instance of `PluginUpdateEngineChecker` which at some point calls `checkForUpdates()`. That method will save the "external_updates" option for the add-on. Eventually the code gets to the new logic that you added in `requestInfo()`, and since it's not EE core, it will exit and NOT phone home. When EE core eventually gets around to `requestInfo()` it won't exit, and instead will loop through all of the registered addons, pull their "external_updates" option, which _should_ have already been set, and if needed, will phone home for updates.

Initially, when I saw `checkForUpdates()` doing this:

```php
update_site_option($this->optionName, $state);
```

and noticed that:

```php
$this->optionName = 'external_updates-' . $this->slug;
```

I was erringly thinking that the above only got set for EE core and NOT for addons because you had to add the `EE_Addon::getPueSlug()` method in order to get the slug for addons... so `'external_updates-' . $this->slug` couldn't be for addons because otherwise you would have used that for getting the addon slug. So that **HAD** to be just for EE core... see my confusion? Of course, `$this->slug` is ultimately the same thing as `EE_Addon::getPueSlug()` when the instance of `PluginUpdateEngineChecker` is for the same addon. So I was having a blonde moment :confused: 

So bottom line, I erringly thought that the only place "external_updates" option was getting saved for addons, was in your new code, but you **ONLY** did that if the option already existed, which is why I was wondering how it got created in the first place. Your statement that

> if it's not (set) then on the next call home it will be.

kinda reaffirmed my mistaken assumption the "external_updates" option was ONLY saved from within the new "call home" logic, whereas it is **actually** being saved in `checkForUpdates()` when the addon is registered... NOT when EE core phones home. I don't think you meant to imply how I interpreted things but it's all good now and i think I fully grok what's going on.

> I'm not sure I follow how this is fragile, should I not be updating an option after its been created?

it's not fragile to update an option after it has been created, else you could never update an option. My concern is because that option is being updated in two different places using potentially different information. If the information wasn't different, then there would be no reason to update the "external_updates" option for an addon from EE core, as the info that had already been saved by the addon would be the same. 

So this is my concern:

- on every request, `EE_Register_Addon::register()` is called for each addon, which in turn calls `_parse_pue_options()`
- the 'AHEE__EE_System__brew_espresso__after_pue_init' hook is set which is triggered in `EE_System::brew_espresso()`
- `EE_Register_Addon::load_pue_update()` runs for addon XYZ during its registration process
- `PluginUpdateEngineChecker::checkForUpdates()` is called 
- the `installed_version` is compared against the `latestVersion` and if we are already up to date, then we do nothing
- otherwise the "external_updates" option is saved with what we'll call "version A" for that info
- after all addons finish registering, EE core runs `PluginUpdateEngineChecker` for itself
- `PluginUpdateEngineChecker::requestInfo()` runs, calls home, and updates the `latestVersion` value for addon XYZ (cuz a new version is out)
- "external_updates" option is saved with what we'll call "version B" for that info

now here's where I'm imagining a potential issue...

- on the next request... if the user has not updated addon XYZ, the above process happens again...  which means that addon XYZ is going to overwrite the "version B" of the "external_updates" option because:

```php
            if( $this->slug !== 'event-espresso-core-reg' && !empty($state->latestVersion) ) {
                if( version_compare($this->_installed_version, $state->latestVersion, '>=') ) {
                    return;
                }
            }
```
(side note: the above conditionals can be combined into one big IF statement)

now because addon XYZ was not updated, the installed version is going to be LESS than the `latestVersion`, meaning we do NOT return from the above... and will continue on to resave the "external_updates" option, and essentially overwrite version B with version A.

So ultimately it seems like that the "external_updates" option is potentially going to get repeatedly saved and overwritten back n forth with different data , assuming the data is different during those two updates which I'm assuming is the case considering the amount of code that's in PUE, and also assuming that there isn't something else that I'm missing that happens somewhere else in the PUE code to prevent this.

Anyways... big ramble to explain what my concern was, but if the above is completely unwarranted worrying, then just disregard all of that : \

Pebblo: Thanks for digging into it, I knida breezed over a little too much above forgetting that not everyone has been going over PUE recently :)

> now because addon XYZ was not updated, the installed version is going to be LESS than the latestVersion, meaning we do NOT return from the above... and will continue on to resave the "external_updates" option, and essentially overwrite version B with version A.

Ah, ok. I follow.

So my understanding of this is that at that point "external_updates" (`$state` in code) is going to include:

```
$state->lastCheck
$state->checkedVersion
$state->latestVersion
$state->update
```

Because there is an update available, `update` is going to be populated with various data but the actual contents are unimportant for this so just noting that its there.

`$state->latestVersion` was _only_ set in in the previous request with from core.

`$state->update` is set by the call to `requestUpdate()` in the `checkForUpdates()` method, however, right before that the option is saved.

Which means `checkForUpdates()` is going to pull the option (which again, in this request includes `lastestVersion`) update `lastCheck`, update `checkedVersion` and save the option back again (it does this before `requestUpdate()` and has a comment stating it does this just in case something goes wrong with the `requestUpdate()` call).

It then pulls an update from ee.com and saves/updates `$state->update` with whatever is returned here, then again, saves the option.

So I don't think that `latestVersion` (which is the important part of this new code, right?) gets overwritten here as it's only actually set on the EE core request. It'll be read and then saved back to the option in the above.

That all make sense or am I looking at it wrong?

There is an important note on this in that I have to continue let PUE ping ee.com if the installed version is less than the current version, even if `update' has been set with all of the various data we need for the update. This is because the download has a nonce check so if/when someone takes a while to update their download it will fail otherwise, we could rewrite it so that isn't needed but for now I think this gets us most of the way there without rewriting too much of PUE.

joshfeck:
Do the unit test failures need to be addressed?

core/third_party_libs/pue/pue-client.php
tn3rb: ya that all sounds fine, I just wanted to make sure that everything is as solid as possible considering this is fairly critical code.

I'm satisfied now that everything you've done is correct and won't result in any foreseeable issues. Great fix 👍

tn3rb: ya that all sounds fine, I just wanted to make sure that everything is as solid as possible considering this is fairly critical code.

I'm satisfied now that everything you've done is correct and won't result in any foreseeable issues. Great fix 👍

Pebblo:
Unit tests were failing because `EE_Register_Addon_Test` wasn't setting up `pue_options` and the change [HERE](https://github.com/eventespresso/event-espresso-core/pull/2245/commits/ecc5ea02438d770c8b1d3586bfec79cd0422f670) meant that `$setup_args['pue_options']['pue_plugin_slug']` was required..... not so great for anyone who created a 3rd party add-on :|

Also switched over to use stdClasses, thanks for the help with that Brent!

Pebblo:
Apologies Brent, this is hopefully the last change! :)

I know you're already aware of this change so I wont go into it other than to link to the change in PUE itself that goes with it:

https://github.com/eventespresso/eventespresso.com-website/pull/404/commits/b67e7a1bd9b6d564e7459a07fc8a7c464849367a

tn3rb:
> Apologies Brent, this is hopefully the last change! :)

no worries, don't ever hesitate to send something back for review or confirmation or whatever.

change looks good

garthkoyle:
testing notes: https://github.com/eventespresso/eventespresso.com-website/pull/404

joshfeck:
This was retested again today and it's ready for merging

PR #2247, opened 2020-01-16
Casc/dont traverse models

mnelson4:

Makes the cascading deletion of events more efficient by avoiding repeatedly traversing registrations. 
Registrations depend (ie, have a foreign key to) both events and tickets, and `PreviewEventDeletion` used both an event and its indirectly-related tickets are root nodes, registrations used to get traversed twice (which took up extra CPU and disk space when they got stored.)
This can make the individual `ModelObjNode`s serialize about a tenth bigger, but it can avoid unnecessarily traversing almost half the number of model objects.
This should also help with https://github.com/eventespresso/event-espresso-core/pull/2225#discussion_r367101270 because there is no need to re-traverse registrations (and their answers) when deleting a transaction.

## How has this been tested

* [ ] On master, trash an event with registrations.
* Click to "permanently delete" it
* Run the DB query `select option_name, char_length(option_value) as c from wp_options WHERE option_name like 'ee_deletion_%' order by option_id desc limit 1;` to see the size of the most-recent option storing information about what will be deleted
* DON'T actually delete the event
* [ ] Repeat the the above steps, but with this branch. Note that the size of the wp_option should be significantly smaller.
* Now verify when you delete the event, its registrations were correctly deleted too.

## Checklist

* [x] My code is tested.

### DIFF
--- a/core/libraries/batch/JobHandlers/PreviewEventDeletion.php
+++ b/core/libraries/batch/JobHandlers/PreviewEventDeletion.php
@@ -2,10 +2,17 @@
 
 namespace EventEspressoBatchRequest\JobHandlers;
 
+use EE_Base_Class;
+use EE_Error;
 use EEM_Event;
 use EEM_Price;
+use EEM_Registration;
 use EEM_Ticket;
+use EEM_Transaction;
+use EETests\bootstrap\CoreLoader;
 use EventEspresso\core\exceptions\InvalidClassException;
+use EventEspresso\core\exceptions\InvalidDataTypeException;
+use EventEspresso\core\exceptions\InvalidInterfaceException;
 use EventEspresso\core\services\loaders\LoaderFactory;
 use EventEspresso\core\services\orm\tree_traversal\ModelObjNode;
 use EventEspresso\core\services\orm\tree_traversal\NodeGroupDao;
@@ -13,6 +20,8 @@ use EventEspressoBatchRequest\Helpers\BatchRequestException;
 use EventEspressoBatchRequest\Helpers\JobParameters;
 use EventEspressoBatchRequest\Helpers\JobStepResponse;
 use EventEspressoBatchRequest\JobHandlerBaseClasses\JobHandler;
+use InvalidArgumentException;
+use ReflectionException;
 
 /**
  * Class EventDeletion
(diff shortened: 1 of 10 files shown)


core/services/orm/tree_traversal/ModelObjNode.php
```diff
@@ -42,10 +42,11 @@ class ModelObjNode extends BaseNode
      * @param $model_obj_id
      * @param EEM_Base $model
      */
-    public function __construct($model_obj_id, EEM_Base $model)
+    public function __construct($model_obj_id, EEM_Base $model, $dont_traverse_models = array())
```
tn3rb:
```suggestion
    public function __construct($model_obj_id, EEM_Base $model, array $dont_traverse_models = []])
```

core/libraries/batch/JobHandlers/PreviewEventDeletion.php
```diff
@@ -93,6 +94,24 @@ public function create_job(JobParameters $job_parameters)
         );
     }
 
+    /**
+     * @since $VID:$
+     * @param EE_Base_Class[] $model_objs
+     * @param array $dont_traverse_models
+     */
+    protected function createModelObjNodes($model_objs, $dont_traverse_models = [])
```
tn3rb:
```suggestion
    protected function createModelObjNodes($model_objs, array $dont_traverse_models = [])
```

core/services/orm/tree_traversal/RelationNode.php
```diff
@@ -53,12 +53,13 @@ class RelationNode extends BaseNode
      */
     protected $nodes;
 
-    public function __construct($main_model_obj_id, EEM_Base $main_model, EEM_Base $related_model)
+    public function __construct($main_model_obj_id, EEM_Base $main_model, EEM_Base $related_model, $dont_traverse_models = [])
```
tn3rb:
```suggestion
    public function __construct($main_model_obj_id, EEM_Base $main_model, EEM_Base $related_model, array $dont_traverse_models = [])
```

mnelson4:
Thanks, committed your suggestions and passed back for review

mnelson4:
Cleaning up this project's branches by merging all the completed work into CASC/base

PR #2248, opened 2020-01-16
Delete transactions too (assess while generating preview)

mnelson4:

Same objective as https://github.com/eventespresso/event-espresso-core/pull/2225 but figures out which transactions to delete when generating the preview page. This seems like a less hacky approach (although it does require doing a direct DB query because it uses a subquery).

## How has this been tested

* [ ] Create two events, named "event A" and "event B"
* Register for just "event A". Notice the new transaction's ID in the `wp_esp_transaction` table
* After completing that registration, use MER to register for both. Notice the new transaction's ID in the `wp_esp_transaction` table again
* Trash and then permanently delete "event a"
* Notice that the older transaction (the one with registrations for only "event A", none ofr "event B") was deleted; but the newer transaction (the one with registrations for both "event A" and "event B") was left intact.
* [ ] It's also good to test this with a very large dataset. I used it on a cheap shared host to delete an event with about 1500 registrations and it worked ok

## Checklist

* [x] My code is tested.

### DIFF
--- a/core/libraries/batch/JobHandlers/PreviewEventDeletion.php
+++ b/core/libraries/batch/JobHandlers/PreviewEventDeletion.php
@@ -2,11 +2,16 @@
 
 namespace EventEspressoBatchRequest\JobHandlers;
 
+use EE_Base_Class;
+use EE_Error;
 use EEM_Event;
 use EEM_Price;
 use EEM_Registration;
 use EEM_Ticket;
+use EEM_Transaction;
 use EventEspresso\core\exceptions\InvalidClassException;
+use EventEspresso\core\exceptions\InvalidDataTypeException;
+use EventEspresso\core\exceptions\InvalidInterfaceException;
 use EventEspresso\core\services\loaders\LoaderFactory;
 use EventEspresso\core\services\orm\tree_traversal\ModelObjNode;
 use EventEspresso\core\services\orm\tree_traversal\NodeGroupDao;
@@ -14,6 +19,8 @@ use EventEspressoBatchRequest\Helpers\BatchRequestException;
 use EventEspressoBatchRequest\Helpers\JobParameters;
 use EventEspressoBatchRequest\Helpers\JobStepResponse;
 use EventEspressoBatchRequest\JobHandlerBaseClasses\JobHandler;
+use InvalidArgumentException;
+use ReflectionException;
 
 /**
  * Class EventDeletion
@@ -44,8 +51,12 @@ class PreviewEventDeletion extends JobHandler
     /**
      *
      * @param JobParameters $job_parameters
-     * @throws BatchRequestException
      * @return JobStepResponse
+     * @throws EE_Error
+     * @throws InvalidDataTypeException
+     * @throws InvalidInterfaceException
+     * @throws InvalidArgumentException
+     * @throws ReflectionException
      */
     public function create_job(JobParameters $job_parameters)
     {
@@ -54,7 +65,6 @@ class PreviewEventDeletion extends JobHandler
         // Find all the root nodes to delete (this isn't just events, because there's other data, like related tickets,
         // prices, message templates, etc, whose model definition doesn't make them dependent on events. But,
         // we have no UI to access them independent of events, so they may as well get deleted too.)
-        $model_objects_to_delete = [];
         $roots = [];
         foreach ($event_ids as $event_id) {
             $roots[] = new ModelObjNode(
@@ -96,7 +106,7 @@ class PreviewEventDeletion extends JobHandler
         foreach ($transactions_ids as $transaction_id) {
             $roots[] = new ModelObjNode(
                 $transaction_id,
-                \EEM_Transaction::instance(),
+                EEM_Transaction::instance(),
                 ['Registration']
             );
         }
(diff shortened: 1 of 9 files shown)


core/libraries/batch/JobHandlers/PreviewEventDeletion.php
```diff
@@ -75,19 +85,33 @@ public function create_job(JobParameters $job_parameters)
...
+                    $price->ID(),
+                    $price->get_model()
+                );
+            }
+        }
+        $transactions_ids = $this->getTransactionsToDelete($event_ids);
+        foreach ($transactions_ids as $transaction_id) {
+            $roots[] = new ModelObjNode(
+                $transaction_id,
+                \EEM_Transaction::instance(),
```
tn3rb: plz import all classes

core/libraries/batch/JobHandlers/PreviewEventDeletion.php
```diff
@@ -112,6 +136,51 @@ protected function createModelObjNodes($model_objs, $dont_traverse_models = [])
...
+                          t.TXN_ID
+                        FROM 
+                          {$wpdb->prefix}esp_transaction ti INNER JOIN 
+                          {$wpdb->prefix}esp_registration ri ON ti.TXN_ID=ri.TXN_ID
+                        WHERE
+                          ti.TXN_ID=t.TXN_ID AND
+                          ri.EVT_ID NOT IN ({$imploded_sanitized_event_ids})
+                       )"
+            )
+        );
```
tn3rb: not grokking this query at all...

it looks like you are running a near identical copy of the query as a subquery, then excluding the results of the inner from the outer results... ?!?!?!

> get TXN IDs where the Registration Event ID is in this list and not in this list ( get TXN IDs for the same TXN in the outer query where the Registration Event ID is in that same list )

so if the list of event IDs was `[ 1, 2, 3 ]` then it would be:

> get TXN IDs where the Registration Event ID is in [ 1, 2, 3 ] and not in ( get TXN IDs for the same TXN in the outer query where the Registration Event ID is in [ 1, 2, 3 ] )

it looks like the inner query will completely nullify the outer query since it is going to return a list of the same IDs that the outer query would find...

what am i missing?!?!


as well... the inner query does this:

```sql
SELECT t.TXN_ID ... ... WHERE ti.TXN_ID=t.TXN_ID AND ...
```

which looks to me like it is just going to return the TXN ID for whatever row the outer query happens to be looping over...

I'm soooo confused....

mnelson4: no worries, ya that `NOT` is obviously easy to miss. It could use another comment I think...

core/libraries/batch/JobHandlers/PreviewEventDeletion.php
mnelson4: d'oh!

core/libraries/batch/JobHandlers/PreviewEventDeletion.php
```diff
@@ -51,7 +65,6 @@ public function create_job(JobParameters $job_parameters)
         // Find all the root nodes to delete (this isn't just events, because there's other data, like related tickets,
         // prices, message templates, etc, whose model definition doesn't make them dependent on events. But,
         // we have no UI to access them independent of events, so they may as well get deleted too.)
-        $model_objects_to_delete = [];
```
mnelson4: this was leftover; this variable is no longer used

core/libraries/batch/JobHandlers/PreviewEventDeletion.php
```diff
@@ -112,6 +152,54 @@ protected function createModelObjNodes($model_objs, array $dont_traverse_models
...
+        $event_ids = array_map(
+            'intval',
+            $event_ids
+        );
+        $imploded_sanitized_event_ids = implode(',', $event_ids);
+        // Select transactions with registrations for the events $event_ids which also don't have registrations
+        // for any events NOT in $event_ids.
+        // Notice the outer query searched for transactions whose registrations ARE in $event_ids,
+        // whereas the inner query checks if the outer query's transaction has any registrations that are
+        // NOT IN $event_ids (ie, don't have registrations for events we're not just about to delete.)
```
mnelson4: I hope the above explanation will help down the line...

mnelson4:
Ok back at you for review again @tn3rb. Please note I rebased onto the base branch, and then did a force push (so if you've checked out this branch, you'll need to trash your local branch.)

core/libraries/batch/JobHandlers/PreviewEventDeletion.php
tn3rb: oh ya that's great thnx... 
I just had a brain fart moment yesterday where I initially didn't see the `NOT` and then my brain just refused to see it from that point on... or maybe I was just focusing too much on other aspects of the query.

core/libraries/batch/JobHandlers/PreviewEventDeletion.php
```diff
@@ -32,17 +40,23 @@ class PreviewEventDeletion extends JobHandler
...
     /**
      *
      * @param JobParameters $job_parameters
-     * @throws BatchRequestException
      * @return JobStepResponse
+     * @throws EE_Error
+     * @throws InvalidDataTypeException
+     * @throws InvalidInterfaceException
+     * @throws InvalidArgumentException
+     * @throws ReflectionException
```
tn3rb: thnx for declaring all of these 👍

mnelson4:
Because `CASC/base` is stable-ish, I think this would be good to test before merging into it (just in case this totally bombs and we decide the other PR, or some other approach, would be better.)

joshfeck:
I did run into some fatal errors on a transaction admin page when the deleted registration was also the primary registration. The last commit fixes those issues.

mnelson4:
My most recent commits address the requests from Josh on https://github.com/eventespresso/event-espresso-core/issues/2194#issuecomment-575819308
Passing to Brent for review

mnelson4:
Tidying up this project by merging all the work into CASC/base

PR #2258, opened 2020-01-17
Fix Types for Calculator PriceType Input

tn3rb:
Beautiful case of TypeScript identifying a potential issue and forcing you to deal with it better.

So both our Price and PriceType entities have an `isBasePrice` property (as well as a bunch of other shared properties) and I was using a selection predicate for Prices to filter an array of PriceTypes and remove those where `isBasePrice` was false. Code worked fine, but only because the selection predicates for Prices had not been converted to TS yet, but TS was smart enough to infer that I was not passing the correct types through some of my code. Very impressive.

The solution was to add a similar predicate for PriceTypes, and also to change some of the shared predicates to use generics so that the exact type can be set in client code.


assets/prototype/domain/eventEditor/tickets/ticketPriceCalculator/AddPriceModifierButtonData.tsx
```diff
@@ -17,8 +17,8 @@ const AddPriceModifierButtonData = ({
 	const getPriceType = findEntityByGuid(modifiers);
 	const addPriceModifier = useCallback(() => {
 		if (Number(price.amount)) {
-			const priceClone: Price = clone(price);
-			const baseType: Price = getPriceType(priceClone.priceType);
+			const priceClone = clone(price);
+			const baseType = getPriceType(priceClone.priceType);
```
tn3rb: the types are now properly inferred and therefore don't need to be hard coded

assets/prototype/domain/shared/predicates/priceTypes/selectionPredicates/index.ts
```diff
@@ -24,6 +24,8 @@ export const isNotTax = propEq('isTax', false);
 
 // returns true if supplied price type is a flat fee (dollar) surcharge
 export const isFlatFeeSurcharge = allPass([isNotBasePrice, isNotDiscount, isNotPercent]);
+// returns array of price types that are NOT base price types
+export const getPriceModifiers = (priceTypes: PriceType[]): PriceType[] => filter(isNotBasePrice, priceTypes);
```
tn3rb: this same function is in the selection predicates file for Prices, could maybe build an abstracted one that they could share, but it might start to get messy

assets/prototype/domain/eventEditor/data/types.ts
```diff
@@ -58,6 +58,7 @@ export interface Price extends Entity {
 	isTax: boolean;
 	order: number | string;
 	overrides: string;
+	priceType: string;
```
manzoorwanijk: This is the wrong assumption, it should be `PriceType` or we need to come up with an alternative:

GQL Schema for `Price`
image

PR #2270, opened 2020-01-21
Reorganize Entity Predicate Logic; closes #2232.

### LINKED ISSUE #2232: Reorganize Entity Predicate Logic
instead of `\domain\shared\predicates` with subfolders for `datetimes`, `prices`, `tickets`, etc... 

- [ ] create folders for those entities inside `\domain\shared\entities`

- [ ] add a `predicates` subfolder to each one of those new subfolders

- [ ] move predicate functions into appropriate subfolder. For ex: instead of `\domain\shared\predicates\datetimes\isTrashed` we would have `\domain\shared\entities\datetimes\predicates\isTrashed`. This way ALL shared logic pertaining to entities is in the same place.

- [ ] update **ALL** paths and references accordingly

- [ ] update checklist on #2224 when done

nicolad:
- Moved entities predicates from `\domain\shared\predicates` to `\domain\shared\entities`
- Created `predicates` subfolder for each entity.
- Closes: https://github.com/eventespresso/event-espresso-core/issues/2232


nicolad:
> I think that will keep things more organized and together

Totally agree, things starts to crystalize.

Issue #1569, opened 2019-09-03
Verify Authorize.net integration unaffected by infrastructure changes


mnelson4:
## Issue Overview

Authorize.net is moving off Akamai (they're switching the API endpoints to a public network and website and portal to Cloudflare). Our integration should not need any changes, but we should verify that. Here is their FAQ page: https://support.authorize.net/s/article/Authorize-Net-Network-Change-FAQ?utm_campaign=19Q4+AkamaiEOL%2FMigrationToPCN+Notice+Developer&utm_medium=email&utm_source=Eloqua

Importantly: between October 15 2019 and February 10 2020 we should retest all our the Authorize.net integrations on their sandbox servers to verify they still work

We don't specifically bundle a certificate in our integration, so merchants don't have to worry about re-uploading a new certificate or anything.

The only thing site owners will need to do is make sure that if they were previously whitelisting Authorize.net's IP addresses, they'll need to whitelist the new ones. Their FAQ page says

> I use a firewall that needs IP Addresses to white list outbound and/or inbound connections. What do I need to do?

> If your solution uses a firewall to filter outbound and/or inbound connections, make sure to white list IP Address provided in our support article: Authorize.Net Domains and IP Addresses

See https://support.authorize.net/s/article/Authorize-Net-Domains-and-IP-Addresses

We could probably make a blog post about this too, just to inform our users.

joshfeck:
We should remove the "Server" setting in the AIM gateway. (it's the setting that lets you toggle between Akamai and "deprecated" options)

joshfeck:
@mnelson4 can the "Server" setting in the AIM gateway's settings be removed?

PR #2287, opened 2020-01-27
remove server server option (so users could switch to or from akamai …

mnelson4:
…server). The setting hasnt applied since 2016

Removes the "server" option in Authorize.net as it hasn't been necessary since 2016 when they moved to Akamai (and now that[ they're moving OFF Akamai, it especially doesn't make sense](https://github.com/eventespresso/event-espresso-core/issues/1569)!)
This removes the option from the payment settings form, the help tab, and makes it so it's no longer used in the gateway class code.
All API calls are just sent to Authorize.net's normal API servers: 
SANDBOX: `https://test.authorize.net/gateway/transact.dll`, 
LIVE: `https://secure2.authorize.net/gateway/transact.dll`

## How has this been tested

* [x] Make a payment with AIM in debug mode. It should work fine as usual.

## Checklist

* [x] My code is tested.

### DIFF
--- a/caffeinated/payment_methods/Aim/EEG_Aim.gateway.php
+++ b/caffeinated/payment_methods/Aim/EEG_Aim.gateway.php
@@ -34,8 +34,6 @@ class EEG_Aim extends EE_Onsite_Gateway
 
     protected $_transaction_key;
 
-    protected $_server;
-
     protected $_currencies_supported = array(
         'AUD',
         'USD',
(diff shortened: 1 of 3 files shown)


mnelson4:
failing unit test appears to be from some unrelated JS build issue... I'll investigate but this is ready for testing.

mnelson4:
Yeah the same failure is happening on master: https://travis-ci.com/eventespresso/event-espresso-core/jobs/278544471 so I don't think it should hold this up

joshfeck:
good to go!

PR #2292, opened 2020-01-27
Fix travis ci config

tn3rb:
Unit tests have suddenly started failing with the following cryptic error message:

```bash
npm ERR! cb() never called!
```

as can be seen here: https://travis-ci.com/eventespresso/event-espresso-core/jobs/280638227

That error appears to have occured on and off over the years but offers little help as the issue is always some other underlying problem.
The current wave of this error which we have been caught up in has been documented here: 

https://github.com/npm/cli/issues/417

and here:

https://github.com/npm/npm/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc+%22cb%28%29+never+called%22+

I will be using this draft branch to experiment with various solutions I find in threads until the underlying problem is discovered and fixed


tn3rb:
So it appears that the solution to this problem was to manually specify q previous stable version of NodeJs since the latest appears to trigger this error in npm. NOt sure where exactly the bug is, but it is not within our code.

So the primary fix in this branch was to set the Node version used in our testing to `11.0.0` since that is the highest preinstalled version of Node on the Ubuntu Xenial build which our tests use. Using a preinstalled version also meant that we do not have to go through the install process during testing, so that was removed. I also activated a Travis beta feature that validates your YaML build file and fixed a few warnings that appeared as a result of that.

.travis.yml
```diff
@@ -1,4 +1,6 @@
-sudo: required
+dist: xenial
```
nicolad: Ok, makes sense, based on: https://blog.travis-ci.com/2019-04-15-xenial-default-build-environment

tn3rb: that doesn't really change anything as we would have been defaulted to using Xenial anyways, but the Tavis build validation feature I enabled pointed that out as an issue so I only added that to remove that warning.

PR #2295, opened 2020-01-28
Document new tree traversal classes

mnelson4:

Just adds some more higher-level documentation for the new tree traversal classes, because I hope they will be useful beyond just deleting events (at very least they could be reused for deleting transactions and other stuff.)

## How has this been tested

None needed

## Checklist

* [x] My code is tested.

### DIFF
--- a/docs/G--Model-System/tree-traversal.md
+++ b/docs/G--Model-System/tree-traversal.md
@@ -143,7 +143,7 @@ those same `ModelObjNode`s and continue using them, like so:
 $nodeGroupDao = LoaderFactory::getLoader()->getShared('\EventEspresso\core\services\orm\tree_traversal\NodeGroupDao');
 $some_obj_to_traverse = $nodeGroupDao->getModelObjNodesInGroup($code);
 
-// And now you cn continue using the ModelObjNodes
+// And now you can continue using the ModelObjNodes
 foreach($some_obj_to_traverse as $model_obj_node){
     $model_obj_node->visit(100);
 }
@@ -173,5 +173,4 @@ foreach($models_and_ids as $model_name => $ids){
 // And remove the option (which can be several MBs if it contains thousands of model objects.)
 $nodeGroupDao->deleteModelObjNodesInGroup($code);
 ```
-
 
--- a/docs/G--Model-System/tree-traversal.md
+++ b/docs/G--Model-System/tree-traversal.md
@@ -58,8 +58,9 @@ $t_serialized = serialize($transaction_node);
 // Unserialize it.
 $transaction_node_unserialized = unserialize($t_serialized);
 
-// Keep using it. It will pick up where it left off exploring the tree
-$transaction_node_unserialized->visit(20);
+// Keep using it to grab the next 10 rows (ie: 11-20).
+// It will pick up where it left off exploring the tree
+$transaction_node_unserialized->visit(10);
 ```
 
 `ModelObjNode::visit()` identifies all the model object's dependent model objects.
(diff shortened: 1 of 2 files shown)


docs/G--Model-System/tree-traversal.md
```diff
@@ -0,0 +1,90 @@
...
+$transaction_node->visit(10);
+
+// Serialize the transaction node.
+$t_serialized = serialize($transaction_node);
+
+// Unserialize it.
+$transaction_node_unserialized = unserialize($t_serialized);
+
+// Keep using it. It will pick up where it left off exploring the tree
+$transaction_node_unserialized->visit(20);
```
tn3rb: is this supposed to indicate that you are getting 20 additional rows or to fetch a **total** of 20 rows which includes the first 10?

If you mean the former (which I think you do), then it may be potentially less confusing to keep the number of rows visited the same, but change the comment to something like:

```php
// Keep using it to grab the next 10 rows (ie: 11-20).
// It will pick up where it left off exploring the tree
$transaction_node_unserialized->visit(10);
```

docs/G--Model-System/tree-traversal.md
```diff
@@ -0,0 +1,90 @@
...
+`ModelObjNode::getIds()` returns a 2D array, whose top-keys are the names of dependent model objects, and whose value is an array of their IDs.
+Useful if you want to organize the tree by model type.
+
+### A note about internal details of ModelObjNode::visit()
+`ModelObjNode::visit()` internally does the following:
+
+* Create a `RelationNode` for each of the model's `EE_Has_Many_Relation`s and the join models for `EE_HABTM_Relation`s
+(eg on `EEM_Transaction`, it would make one for `Registration`, `Payment`, `Line_Item`, `Message`, `Extra_Meta`, etc)
+* Call `visit()` on each of them, which
+* Finds all the model related model objects across that relation (eg when called on the `RelationNode` for `Registration`,
```
tn3rb: > model related model objects

??? is that correct? or maybe:

```suggestion
* Finds all of the model objects across that relation (eg when called on the `RelationNode` for `Registration`, 
```

docs/G--Model-System/tree-traversal.md
```diff
@@ -0,0 +1,90 @@
...
+So `RelationNode` is only expected to be used internally.
+
+## NodeGroupDao
+
+If there could be hundreds of dependent model objects, you'll probably need to call `ModelObjNode::visit()` multiple 
+times across multiple HTTP requests. This requires saving state between requests.
+The `ModelObjNode`s are designed to be quite small when serialized. 
+`NodeGroupDao` helps with saving `ModelObjNode`s across multiple requests by saving state to the WordPress Options table.
+It has methods for generating a unique code to be used in the option name, serializing a group of `ModelObjNode`s to the
+database, fetching a group of serialized `ModelObjNode`s from the database, and deleting the group of nodes from the database.
```
tn3rb: Are there going to be additional docs for this class with examples of the above functionality?

docs/G--Model-System/tree-traversal.md
mnelson4: ya, it's the former. I'll add your suggestion.

docs/G--Model-System/tree-traversal.md
mnelson4: ...there can be 😅...

Ya an example would be good wouldn't it?

mnelson4:
I made a few additions as you suggested and this is off for another review.

docs/G--Model-System/tree-traversal.md
```diff
@@ -0,0 +1,177 @@
...
+```
+
+Then, on a subsequent request, using the `$code` (which you can put in the session, querystring, etc) you can retrieve
+those same `ModelObjNode`s and continue using them, like so:
+
+```php
+$nodeGroupDao = LoaderFactory::getLoader()->getShared('\EventEspresso\core\services\orm\tree_traversal\NodeGroupDao');
+$some_obj_to_traverse = $nodeGroupDao->getModelObjNodesInGroup($code);
+
+// And now you cn continue using the ModelObjNodes
```
tn3rb:
```suggestion
// And now you can continue using the ModelObjNodes
```

PR #2322, opened 2020-02-03
Travis failing build fix.

nicolad:
- Fixes `cb() never called` issue
- Replaces `npm` with `yarn`


.travis.yml
```diff
@@ -1,146 +1,160 @@
...
-  - mysql
+    - mysql
 
 cache:
-  apt: true
-  directories:
-    - $HOME/.composer/cache
-    - $HOME/.npm
+    apt: true
+    yarn: true
```
nicolad: Travis will detect yarn based on `yarn.lock` usage: https://docs.travis-ci.com/user/languages/javascript-with-nodejs/#using-yarn

nicolad: Also: https://blog.travis-ci.com/2016-11-21-travis-ci-now-supports-yarn

.travis.yml
```diff
@@ -1,146 +1,160 @@
-dist: xenial
```
nicolad: Changes in this file does 2 things:
1. Reverts to the point before this [PR](https://github.com/eventespresso/event-espresso-core/pull/2292)
2. And adds `yarn: true` and `- $HOME/.yarn-cache` options to `cache`

.travis.yml
```diff
@@ -1,146 +1,160 @@
...
 
 services:
-  - mysql
+    - mysql
 
 cache:
-  apt: true
-  directories:
-    - $HOME/.composer/cache
-    - $HOME/.npm
```
nicolad: @tn3rb in case we will go with yarn, we will need to decide if we want to keep both cache directories.

nicolad:
@tn3rb @manzoorwanijk `package-lock.json` has been replaced by `yarn.lock`
Travis is detecting now that we are using `yarn` for deps management, so we will have to update `yarn.lock` file when we will add/ remove deps.

PR #2353, opened 2020-02-07
Refactor forms | Closes #2346

### LINKED ISSUE #2346: Forms Refactor
The existing form system in EDTRv1 is coupled tightly with the defunct model entities and also utilizes many WP GB components under the hood. Since we have switched to Apollo/GQL and the AntD component lib we need to refactor things accordingly.

### Primary Goals

- wrap all AntD components in adaptors in order to decouple our application from the underlying UI library. This should include:

	- maybe prefix all input component names with `Espresso` ie: `EspressoButton` (I'm open to alternate ideas)

	- standardizing prop names across all of our components (AntD is likely standardized, but we can changes prop names to our benefit)

	- allowing styles to be flexible as well as overridden and/or controllable via theming. Forms will be used in both the admin and public facing web pages, and potentially within dedicated mobile apps at some point in the future

	- applying generic styles to help us (and customers) with targeting our components and UI/DOM elements

	- apply the highest levels of a11y and i18n best practices

	- use dedicated input labels as opposed to placeholders (placeholders can be used for example content)

- integrate form elements with RFF (React Final Form) for managing validation, notices, and data submission

- integrate form elements with Apollo for managing data submission and user feedback (loading spinners, error notices, etc)

- attempt to maintain a clear path to implementing a dynamic form building UI at some point in the future

- attempt to maintain a clear path to implementing conditional form logic at some point in the future

- allow for complex yet clear and accessible form designs (see image below for example)

- please examine the previous form system in EDTRv1 for an idea of the scope of components we will need

form-design

manzoorwanijk:
This PR:
- Refactors forms to make them as dynamic as possible
- Closes #2346


assets/src/application/ui/forms/ee-form/FieldRenderer.tsx
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+
+import { FormItem, MappedField } from './adapters';
```
nicolad: Maybe just `Field`, not `MappedField`

nicolad: because it's clear that it's from `'./adapters'`, so it's not a lib-based field

manzoorwanijk: May be I wanted to be more expressive 😄

assets/src/application/ui/forms/ee-form/FieldRenderer.tsx
```diff
@@ -0,0 +1,26 @@
...
+	return (
+		<FormItem
+			label={label}
+			extra={desc}
+			validateStatus={validateStatus}
+			hasFeedback={meta.touched && !!(meta.error || meta.submitError)}
+			help={meta.touched && (meta.error || meta.submitError)}
+			className={`form-item form-item-${fieldType}`}
+		>
+			{before}
```
nicolad: `{before && before}` ?
the same with `after`?

manzoorwanijk: yes 👍

assets/src/application/ui/forms/ee-form/adapters/Button.tsx
```diff
@@ -0,0 +1,8 @@
+import React from 'react';
+import { Button as AntdButton } from 'antd';
```
nicolad: :+1:

assets/src/application/ui/forms/ee-form/config.tsx
```diff
@@ -0,0 +1,184 @@
+import { EspressoFormProps } from './types';
+
+export const formConfig: EspressoFormProps = {
+	onSubmit: (values) => console.log(values),
+	initialValues: {},
+	// validate: () => {},
```
nicolad: yeah, validation would involve some fun

assets/src/application/ui/forms/ee-form/Submit.tsx
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+import { __ } from '@wordpress/i18n';
+
+import { Button } from './adapters';
+import { SubmitProps } from './types';
+
+const Submit: React.FC<SubmitProps> = ({ submitting, label, ...rest }) => {
+	return (
+		<div className='submit-wrap'>
```
tn3rb: maybe we should also have: 

- a reset button for setting form data back to initial values

- a cancel button for closing form and NOT saving data

assets/src/application/ui/forms/ee-form/config.tsx
```diff
@@ -0,0 +1,233 @@
+import { EspressoFormProps } from './types';
+
+export const formConfig: EspressoFormProps = {
+	onSubmit: (values) => console.log(values),
+	initialValues: {},
```
tn3rb: so will this have to consist of the same structure as the "sections" field below?
if so, then would it not be easier to simply add an "initialValue" prop to each of the fields within "sections"?

assets/src/application/ui/forms/ee-form/Submit.tsx
manzoorwanijk: Yes, a reset button is already in the list but not sure what you mean by cancel button. The form will be submitted only when submit button is clicked. And why should we assume that the form will be rendered inside a modal to cancel it?

assets/src/application/ui/forms/ee-form/config.tsx
manzoorwanijk: Yes, if you pass `initialValues` prop to form, the value has to be in the same shape as the form fields but you can also pass `initialValue` prop to every individual field.

assets/src/application/ui/forms/ee-form/RenderField.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+const RenderField: React.FC<RenderFieldProps> = (props) => {
+	const { fieldType } = props;
+	if (!fieldType) {
+		return null;
+	}
+
+	const { repeatable, ...rest } = props;
+
+	if (repeatable) {
```
nicolad: Maybe `isRepeatable`?

assets/src/application/ui/forms/ee-form/RenderFields.tsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+
+import { RenderFieldsProps } from './types';
+import RenderField from './RenderField';
+
+const RenderFields: React.FC<RenderFieldsProps> = ({ fields, namespace }) => {
+	return (
+		<div>
+			{fields.map(({ name, ...fieldProps }, key) => {
+				const fieldName = namespace ? `${namespace}.${name}` : name;
```
nicolad: What is `namespace` in context of `RenderFields` ?
Maybe fields name or smth different ?

assets/src/application/ui/forms/ee-form/RenderSection.tsx
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+
+import { SectionProps } from './types';
+import RenderFields from './RenderFields';
+
+const RenderSection: React.FC<SectionProps> = ({ name, title, fields, namespaceFields }) => {
```
nicolad: Maybe smth different than `namespaceFields` ? :smiley:

assets/src/application/ui/forms/ee-form/RenderSections.tsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+
+import { RenderSectionsProps } from './types';
+import RenderSection from './RenderSection';
+
+const RenderSections: React.FC<RenderSectionsProps> = ({ sections }) => {
+	return (
+		<div className='sections-wrapper'>
+			{sections.map((section, key) => (
```
nicolad: :)
That's index, not quite `key` :)

nicolad: As per react docs:
`We don’t recommend using indexes for keys if the order of items may change.` from: https://reactjs.org/docs/lists-and-keys.html

nicolad: If order won't change I am ok with using indexes as keys.

assets/src/application/ui/forms/ee-form/RenderField.tsx
manzoorwanijk: may be...

assets/src/application/ui/forms/ee-form/RenderFields.tsx
manzoorwanijk: It's explained in types:
```ts
/**
 * If true, each field inside the section
 * will be saved as `${section.name}.{field.name}`
 */
namespaceFields?: boolean;
```

assets/src/application/ui/forms/ee-form/RenderSections.tsx
manzoorwanijk: Where it can change, I have added other values to the keys

assets/src/application/ui/forms/ee-form/RenderField.tsx
manzoorwanijk: Done

assets/src/application/ui/forms/ee-form/RenderSection.tsx
manzoorwanijk: May be our CNO (@tn3rb) has a better idea 😄

assets/src/application/ui/forms/ee-form/Submit.tsx
```diff
@@ -0,0 +1,39 @@
...
+				htmlType='submit'
+				disabled={submitting}
+				loading={submitting}
+				className='submit-button'
+				type='primary'
+				{...submitButton}
+			>
+				{submitButton.label || __('Submit')}
+			</Button>
+			&nbsp;
```
nicolad: `{ " " }`

assets/src/application/ui/forms/ee-form/Submit.tsx
tn3rb: ooo good point about the cancel button

assets/src/application/ui/forms/ee-form/Submit.tsx
manzoorwanijk: That is messed up by prettier and `&nbsp;` is perfectly supported by JSX.

assets/src/application/ui/forms/ee-form/fields/Group.tsx
```diff
@@ -0,0 +1,33 @@
...
+	const visible = useShouldBeVisible(conditions, groupName);
+	if (subFields.length && visible) {
+		return (
+			<div className='field-group-wrapper'>
+				<h5>{label}</h5>
+				{before}
+				<div className='field-group-items'>
+					{subFields.map(({ name: fieldname, fieldType, ...props }, i) => {
+						const name = `${groupName}.${fieldname}`;
+						const position = i % 2 === 0 ? 'left' : 'right';
```
tn3rb: ??? 

This appears to be an assumption about layout, with the position baked into the logic (ie: all groups will have two columns).

Why are even number items always on the right? What if a form design doesn't want two columns?
As well, what if a language is `rtl` ? (that can likely be handled via css, specifically flexbox)

Can layout decisions like this be abstracted away from the base form components? For example, in the EDTR v1, the `assets\ZZZ\components\form\layouts\two-column-admin` folder contains components that can be used to apply a two column design layout to your form, and although no other layout strategies exist, the idea was to have a similar set of components that would apply different layouts to a form. So the idea was that you could completely change the layout for a form simply by importing a different layout strategy, either manually:

```js
import { twoColumnAdminFormLayout } from '@eventespresso/components';
const { FormRow, FormSection, FormWrapper } = twoColumnAdminFormLayout;
```

or via props

```js
const { FormRow, FormSection, FormWrapper } = props.layoutStrategy;
```

with all of the other code in the component exactly the same.

This doesn't have to work exactly the same way that I did things in EDTR v1, as I will leave that up to your expertise to solve, but I can guarantee that there will be different requirements for form layouts in different areas of the plugin, so anything we build should have some ability to be versatile.

assets/src/application/ui/forms/ee-form/index.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+const EspressoForm: React.FC<EspressoFormProps> = ({ onSubmit, initialValues, validate, ...rest }) => {
+	return (
+		<div className='ee-form'>
+			<ReactFinalForm
+				initialValues={initialValues}
+				onSubmit={onSubmit}
+				validate={validate}
+				render={FormRenderer}
+				mutators={{ ...arrayMutators }}
```
tn3rb: this will need to be extendable via an incoming prop:

```suggestion
				mutators={{ ...arrayMutators, mutators }}
```

where `mutators` is a prop

assets/src/application/ui/forms/ee-form/utils.ts
```diff
@@ -0,0 +1,79 @@
+import { FieldMetaState } from 'react-final-form';
+import { getIn, AnyObject } from 'final-form';
+
+import { FieldConditions } from './types';
+
+export const getValidateStatus = ({ touched, error, submitError }: FieldMetaState<any>) => {
+	if (touched) {
+		if (!(error || submitError)) {
+			return 'success';
```
tn3rb: enum type maybe?

assets/src/application/ui/forms/ee-form/adapters/Form.tsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+import { Form as AntdForm } from 'antd';
+import { FormProps } from 'antd/es/form';
+
+const Form: React.FC<FormProps> = ({ children, onSubmit }) => {
+	return (
+		<AntdForm layout='vertical' onSubmit={onSubmit}>
```
tn3rb: maybe use `vertical` as the default value for a `layout` prop, that way we have the option if need be to use the other formats

assets/src/application/ui/forms/ee-form/Submit.tsx
```diff
@@ -0,0 +1,39 @@
...
+				<Button
+					htmlType='reset'
+					disabled={submitting}
+					className='reset-button'
+					onClick={() => form.reset()}
+					{...resetButton}
+				>
+					{resetButton.label || __('Reset')}
+				</Button>
+			) : null}
```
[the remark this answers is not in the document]
manzoorwanijk: May be put it in a different block?

assets/src/application/ui/forms/ee-form/adapters/Select.tsx
```diff
@@ -0,0 +1,35 @@
...
+		}
+		return (
+			<Select.Option {...optionProps} key={`${optionProps.value}${index}`}>
+				{label}
+			</Select.Option>
+		);
+	});
+
+	// make sure the value is an array when mode is "multiple"
+	const value = mode === 'multiple' ? input.value || [] : input.value;
```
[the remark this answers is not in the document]
manzoorwanijk: That's already a union type from AntD.

assets/src/application/ui/forms/ee-form/fields/Group.tsx
manzoorwanijk: Yeah, I wasn't sure about how to handle fields inside a group, so I thought to put 2 fields in a row. The reason for making it that way is again that JSON config oriented approach where you cannot import those layout components and can be a problem with those D&D form builders.

assets/src/application/ui/forms/ee-form/utils.ts
manzoorwanijk: Agreed

assets/src/application/ui/forms/ee-form/adapters/Form.tsx
manzoorwanijk: Sounds good

manzoorwanijk: Done

assets/src/application/ui/forms/ee-form/utils.ts
manzoorwanijk: Done

assets/src/application/ui/forms/ee-form/index.tsx
manzoorwanijk: Done

assets/src/application/ui/forms/ee-form/fields/Group.tsx
manzoorwanijk: I have removed that `left` and `right` class but would like to discuss with you about how we want to support the layout, especially for the form config based approach.

tn3rb: maybe the JSON config can contain a field for setting the layout?
and maybe allow each section to set a different layout (but default to the previously set layout),
that's how our PHP form system work actually.

assets/src/application/ui/forms/espressoForm/context.ts
```diff
@@ -0,0 +1,10 @@
+import { createContext } from 'react';
+import { FormContextProps } from './types';
+
+const Context = createContext<FormContextProps>({});
+
+const { Provider } = Context;
+
+export const FormProvider = Provider;
+
+export default Context;
```
tn3rb: cool... but can we plz move this to `assets\src\application\services\context` so that we never have to go searching for context providers? Or will this NEVER be used outside of the `application/ui/forms/espressoForm/` folder?

assets/src/application/ui/forms/espressoForm/utils.ts
```diff
@@ -3,7 +3,9 @@ import { getIn, AnyObject } from 'final-form';
 
 import { FieldConditions } from './types';
 
-export const getValidateStatus = ({ touched, error, submitError }: FieldMetaState<any>) => {
+type ValidateStatus = '' | 'error' | 'success' | 'warning' | 'validating';
```
tn3rb: 👍

assets/src/application/ui/forms/espressoForm/context.ts
manzoorwanijk: I didn't want to couple it tightly with EDTR and yes, it's never going to be used outside that.

tn3rb: > I didn't want to couple it tightly with EDTR

???

it wouldn't be tightly coupled to EDTR if it was in `application\services\context`

> it's never going to be used outside that.

ya ok that's fine then

assets/src/application/ui/forms/espressoForm/Submit.tsx
```diff
@@ -0,0 +1,43 @@
...
+						htmlType='reset'
+						disabled={submitting}
+						className='reset-button'
+						onClick={() => form.reset()}
+						{...resetButton}
+					>
+						{resetButton.label || __('Reset')}
+					</Button>
+				</div>
+			) : null}
```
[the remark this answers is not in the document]
manzoorwanijk: I don't think that will be the case, it's always going to be beside submit button.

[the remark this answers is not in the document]
manzoorwanijk: Done

PR #2361, opened 2020-02-07
Add Entity Paper Frame closes #2357

### LINKED ISSUE #2357: Migrate Entity Paper Frame
The EntityPaperFrame is primarily a cosmetic component for rendering the entity "cards" when in grid view:

- see `assets\ZZZ\components\ui\entity-paper-frame`

- can maybe use https://ant.design/components/card/ but only if that component does not interfere with the design. It's possible that the styles used by the existing `EntityPaperFrame` component can simply be applied to the AntD component.

- convert to TS

- move to `assets\src\application\ui\layout`

tn3rb:
- see #2357

- migrates EntityPaperFrame component from EDTR v1

- extracts entity ID display above cards to EntityIDs component

- converts all the things to TS

- adds some more aliases for import paths


tsconfig.json
```diff
@@ -12,7 +12,18 @@
...
+			"@appServices/*": ["assets/src/application/services/*"],
+			"@appDisplay/*": ["assets/src/application/ui/display/*"],
+			"@appForms/*": ["assets/src/application/ui/forms/*"],
+			"@appInputs/*": ["assets/src/application/ui/input/*"],
+			"@appLayout/*": ["assets/src/application/ui/layout/*"],
+			"@appConfig/*": ["assets/src/application/valueObjects/config/*"],
+			"@edtrHooks/*": ["assets/src/domain/eventEditor/hooks/*"],
+			"@edtrInterfaces/*": ["assets/src/domain/eventEditor/interfaces/*"],
+			"@edtrServices/*": ["assets/src/domain/eventEditor/services/*"],
+			"@edtrUI/*": ["assets/src/domain/eventEditor/ui/*"]
```
nicolad: YAY :+1:

assets/src/application/ui/display/EntityIDs/index.tsx
```diff
@@ -0,0 +1,36 @@
...
+}
+
+export interface EntityIdProps {
+	align?: EntityIdAlignment;
+	dbid: EntityDbId;
+	guid: EntityId;
+}
+
+const EntityIDs: React.FC<EntityIdProps> = ({ dbid, guid, align = 'left' }) => {
+	const htmlClass = classNames('ee-entity-ids', 'ee-focus-priority-9', {
```
nicolad: `ee-focus-priority-9` is interesting

nicolad: was thinking if we need to have these kind of classnames as enums, to have better control;
but this might be too much of typing; just thoughts out loud

Issue #2355, opened 2020-02-07
Migrate Entity List


tn3rb:
The EntityList component is an abstraction for controlling the toggle between the grid view

![EDTR - event dates.png](https://images.zenhubusercontent.com/55c8c0f14f7a08fbb7f431cd/81ff7be2-d7c6-4a35-b06d-4816401ded41)

and the list view

![EDTR - event dates list.png](https://images.zenhubusercontent.com/55c8c0f14f7a08fbb7f431cd/4c45a803-9b0d-4e69-813b-b6c9417160c0)

as well as displaying an empty state notice ie: 'no results found'

- see: `assets\ZZZ\components\ui\entity-list`

- move to `assets\src\application\ui\layout\entityList`

- convert to TS

PR #2366, opened 2020-02-10
Moved EntityList and converted to TS; closes 2355.

nicolad:
- see: assets\ZZZ\components\ui\entity-list

- move to assets\src\application\ui\layout\entityList

- convert to TS

- closes https://github.com/eventespresso/event-espresso-core/issues/2355


assets/ZZZ/components/ui/entity-list/use-reorder-entities.js
```diff
@@ -1,113 +0,0 @@
-/**
```
nicolad: @tn3rb , not sure if we will need this file since we have this: https://github.com/eventespresso/event-espresso-core/blob/EDTR/master/assets/src/domain/shared/entities/datetimes/predicates/sorters/index.ts

assets/src/application/ui/layout/EntityList/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+	view = 'grid',
+	noResultsText = '',
+	...otherProps
+}) => {
+	// verify array and remove undefined
+	const filteredEntities = Array.isArray(entities) ? without([undefined], entities) : [];
+
+	if (filteredEntities.length === 0) {
+		noResultsText = noResultsText !== '' ? noResultsText : __('no results found', 'event_espresso');
+		return <div className='ee-entity-list-no-results'>{noResultsText}</div>;
```
tn3rb: this could maybe use https://ant.design/components/empty/ although that would likely require the styles to be changed

nicolad: Updated: https://github.com/eventespresso/event-espresso-core/pull/2366/commits/dfe12e3f8bb818b1675c02d46a5c09e5e356b742

assets/ZZZ/components/ui/entity-list/use-reorder-entities.js
tn3rb: missed this comment during my first review...
ya if we can use shared predicates, then that would be preferred

assets/src/application/ui/display/EmptyState.tsx
```diff
@@ -2,31 +2,31 @@
...
 	return (
-		<>
+		<div className={className}>
 			<Empty
 				description={
 					<>
-						{title}
-						<span>{description}</span>
+						{title && __(title)}
+						<span>{__(description)}</span>
```
tn3rb: couple of points:

- I don't think you can run translations on variables like this (although I could be wrong), so let's just assume the incoming strings were translated at the source (which I think should be the case, otherwise it's a guessing game as to whether the receiving component has handled translations.)

- we should account for `description` being blank like you did with `title` no?

so this should be (imho) something like:

```js
{title ? title : '' }
{description ? <span>description</span> : '' }
```

I used ternarys because `{title && title}` just doesn't look right to me, but maybe it's ok?

assets/src/application/ui/display/EmptyState.tsx
```diff
@@ -2,31 +2,31 @@
...
 					</>
 				}
 				image={Empty.PRESENTED_IMAGE_SIMPLE}
 			/>
 
-			{children}
-		</>
+			{children && children}
```
tn3rb: ok I guess this IS ok...

how does this not just print that variable twice? or not at all?

like how come it will print the variable the second time it is encountered and not the first time?
I'm guessing this is just shorthand for a ternary right? Just not something I've ever seen documented anywhere so never fully grokked the underlying mechanics.

what would happen if you did `{children && children && children}` ???

not saying you need to change this but man... JS is crazy!!!

assets/src/application/ui/layout/EntityList/index.tsx
```diff
@@ -43,14 +43,15 @@ const EntityList = ({
 	const filteredEntities = Array.isArray(entities) ? without([undefined], entities) : [];
 
 	if (filteredEntities.length === 0) {
-		noResultsText = noResultsText !== '' ? noResultsText : __('no results found', 'event_espresso');
-		return <div className='ee-entity-list-no-results'>{noResultsText}</div>;
+		const description = noResultsText !== '' ? noResultsText : 'no results found';
```
tn3rb:
```suggestion
		const description = noResultsText !== '' ? noResultsText :  __('no results found')
```

assets/ZZZ/components/ui/entity-list/use-reorder-entities.js
nicolad: cool, it's has been deleted, wanted to double-check with you

assets/src/application/ui/display/EmptyState.tsx
nicolad: `{title && title}` is just short circuit evaluation I have often seen in react apps
updated: https://github.com/eventespresso/event-espresso-core/pull/2366/commits/85c7143fc0ec48d09b63c0b33da50cdd3161fbaf

assets/src/application/ui/display/EmptyState.tsx
nicolad: > how does this not just print that variable twice? or not at all?
cuz `{ undefined && smth }` is the same as `{ false && smth }` which won't return anything

nicolad: my understanding is that it's not a shorthand for ternary operator
this is from MDN:
image

*** 
it doesn't have direct link :(
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_Operators

nicolad: BTW, I wasn't aware of this part yet `if expr is a function call, the calling never takes place`

nicolad: last comment: before applying that logic there I had checked first things on antd side of things to make sure I am not goofing smth, so they have that prop optional:
https://github.com/ant-design/ant-design/blob/65eb713fe897f65bbf187e215c664e5f9b468252/components/empty/index.tsx#L22
***
and this is why they also applied that short-circuit eval as I did: https://github.com/ant-design/ant-design/blob/65eb713fe897f65bbf187e215c664e5f9b468252/components/empty/index.tsx#L68

assets/src/application/ui/layout/EntityList/index.tsx
nicolad: Updated: https://github.com/eventespresso/event-espresso-core/pull/2366/commits/f68c32c61e8943d20c0f09e7eb5b8bc389e27413

assets/src/application/ui/display/EmptyState.tsx
```diff
@@ -1,32 +1,28 @@
...
-
+const EmptyState = ({ className, children, description, title }: EmptyStateProps) => {
 	return (
-		<>
+		<div className={className}>
 			<Empty
 				description={
 					<>
-						{title}
+						{title && title}
```
manzoorwanijk: `{title && title}` is equivalent to just `{title}` if it's always going to be a either a valid React Node or (`null` | `undefined` | `false`). If you think it can be otherwise, then even `{title && title}` won't help because if `title` is a truthy value e.g. a function, it will cause an error.

package.json
```diff
@@ -102,6 +102,7 @@
 		"moment": "^2.24.0",
 		"moment-duration-format": "^2.3.2",
 		"moment-timezone": "^0.5.27",
+		"node-sass": "^4.13.1",
```
manzoorwanijk: I added that in my PR as well 😄

assets/src/application/ui/layout/entityList/EntityList/index.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+import { Entity } from '@edtrServices/apollo/types';
+import EmptyState from '@appDisplay/EmptyState';
+import './style.scss';
+
+interface EntityListProps {
+	entities: Entity[];
+	EntityGridView: React.ReactNode;
+	EntityListView: React.ReactNode;
+	htmlClass: string;
```
manzoorwanijk: I think we should decide about following a standard whether we want to use `htmlClass` or `className` or `classNames`.

assets/src/application/ui/layout/entityList/EntityList/index.tsx
```diff
@@ -0,0 +1,59 @@
...
+	view: string;
```
manzoorwanijk: `view` can be a union (`'grid' | 'list'`) or an enum.

assets/src/domain/eventEditor/ui/datetimes/datesList/index.tsx
```diff
@@ -5,9 +5,9 @@ import AddNewDateButton from './AddNewDateButton';
 import List from './List';
 import useDatetimes from '../../../services/apollo/queries/datetimes/useDatetimes';
 import { useStatus, TypeName } from '../../../../../application/services/apollo/status';
-import EmptyState from '../../../../../application/ui/display/EmptyState';
-import ErrorIndicator from '../../../../../application/ui/display/ErrorIndicator';
-import LoadingIndicator from '../../../../../application/ui/display/LoadingIndicator';
+import EmptyState from '@appDisplay/EmptyState';
+import ErrorIndicator from '@appDisplay/ErrorIndicator';
+import LoadingIndicator from '@appDisplay/LoadingIndicator';
```
manzoorwanijk: These could be exported/imported via an index file

assets/src/application/ui/display/EmptyState.tsx
nicolad: > {title && title} is equivalent to just {title}

`title` type is optional string, it can be `undefined` or a string

assets/src/application/ui/layout/entityList/EntityList/index.tsx
nicolad: agree, let's loop back on this after we will have that decision consolidated

assets/src/domain/eventEditor/ui/datetimes/datesList/index.tsx
nicolad: Yup, good point: https://github.com/eventespresso/event-espresso-core/pull/2366/commits/bf55ed1b86eb54dedce0ed1d189d9c15ab7545f2

nicolad: the only thing I don't like is the `index` from `@appDisplay/index`, but not a biggie

assets/src/application/ui/layout/entityList/EntityList/index.tsx
nicolad: thx, done

assets/src/application/ui/layout/entityList/EntityList/index.tsx
tn3rb: just discussed in chat that we will switch to using `className` but those can all be fixed in a later PR where we go through and polish things at the end

assets/src/application/ui/display/EmptyState.tsx
tn3rb: > cuz { undefined && smth } is the same as { false && smth } which won't return anything

right.. i get that part, but my brain wants to think that `{ smth && smth }` would return `smth smth` resulting in a duplicate, but that's obviously not the case.

PR #2372, opened 2020-02-11
Migrate Entity Pagination; closes #2356.

### LINKED ISSUE #2356: Migrate Entity Pagination
The EntityPagination component is an adapter component for controlling pagination (duh) and is controlled by the per page control from the entity list filter bar component

- see `assets\ZZZ\components\ui\entity-pagination`

- convert to use https://ant.design/components/pagination/

- convert to TS

- move to `assets\src\application\ui\layout`

nicolad:
- closes https://github.com/eventespresso/event-espresso-core/issues/2356
- see assets\ZZZ\components\ui\entity-pagination
- convert to use https://ant.design/components/pagination/
- convert to TS
- move to assets\src\application\ui\layout


assets/src/application/ui/layout/entityList/pagination/useEntityPagination.js
```diff
@@ -14,24 +14,22 @@ const EMPTY_ARRAY = [];
  *                  - setCurrentPage - callback for changing currentPage
```
nicolad: I kept this file around because there might be some logic extraction needed.

assets/src/application/ui/layout/entityList/pagination/usePaginationConfig.js
```diff
@@ -0,0 +1,26 @@
+/**
```
nicolad: Same as the file above :point_up:

nicolad: Also we might extract `onChange` logic here from pagination if it will became complex.

assets/src/application/ui/layout/entityList/pagination/EntityPagination.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * Adds pagination to an "EntityList" component
+ * @return EntityPagination
+ */
+const EntityPagination = ({
+	current,
+	onChange,
+	pageSize = 10,
+	pageSizeOptions = ['10', '20', '30', '40'],
```
tn3rb: I know it's unconventional, but can we use the same options as EDTR v1 plz? ie: 6, 12, 24, 48
those are all more divisable than multiples of 10.
for example, other than 6, the other numbers (12, 24, & 48) can all divide neatly into 2, 3, or 4 columns, whereas 10s can only fit into 2 or 5

assets/src/application/ui/layout/entityList/pagination/EntityPagination.tsx
```diff
@@ -0,0 +1,33 @@
...
+	return (
+		<div className='ee-entity-pagination'>
+			<Pagination
+				current={current}
+				onChange={onChange}
+				pageSize={pageSize}
+				pageSizeOptions={pageSizeOptions}
+				showTotal={showTotal}
+				total={total}
+			/>
```
tn3rb: should add a `{...rest}` prop to this so that we can utilize other props from here: https://ant.design/components/pagination/#API

assets/src/application/ui/layout/entityList/pagination/usePaginationConfig.js
```diff
@@ -0,0 +1,26 @@
...
+ * @return {Object} - pagination config
+ */
+const usePaginationConfig = (paginationConfig) =>
+	useMemo(() => {
+		const defaultConfig = {
+			labels: {
+				first: _x('First', 'First thing in list', 'event_espresso'),
+				last: _x('Last', 'Last thing in list', 'event_espresso'),
+				prev: _x('Prev', 'Previous thing in list', 'event_espresso'),
+				next: _x('Next', 'Next thing in list', 'event_espresso'),
```
tn3rb: Is this being used still?
More importantly, can this even be changed on the AntD pagination component?
I'm not seeing that in their API.

If we are not or can not use this, then might as well get rid of it

assets/src/application/ui/layout/entityList/pagination/useEntityPagination.js
tn3rb: we will need this for tracking state for paginated lists no?

assets/src/application/ui/layout/entityList/pagination/EntityPagination.tsx
nicolad: Thx for explanation, updated: https://github.com/eventespresso/event-espresso-core/pull/2372/commits/67d37e6a72c897ec6f13e5404fbf4d19951d525b

assets/src/application/ui/layout/entityList/pagination/EntityPagination.tsx
nicolad: The initial thought was to have more visibility/ control on what we need and do pass to pagination and add as needed, not a biggie; updated as required: https://github.com/eventespresso/event-espresso-core/pull/2372/commits/98dad297cd2b67e4ff76222a0a0ad7c319abb532

assets/src/application/ui/layout/entityList/pagination/usePaginationConfig.js
nicolad: Got rid of that.
Was hesitant on throwing away that, thought that maybe there is a possibility to extract some ~~juice~~ code

assets/src/application/ui/layout/entityList/pagination/useEntityPagination.js
nicolad: yup, there is `onChange` prop that we might need to use at some point, this is why I left this file;
for sure it will need to be refactored, so I didn't touch anything yet

Issue #2349, opened 2020-02-07
Migrate Calendar Components To EDTRv2


tn3rb:
see: `assets\ZZZ\components\ui\dates-and-times`

- move/copy and refactor the following components:
	- BiggieCalendarDate
	- CalendarDateRange
	- CalendarPageDate
	- MediumCalendarDate

- convert to TS
- extract any additional logic into separate files/utilities/predicates/etc accordingly
- write tests for any non-display logic
- change folder structure/names to something like:

```
├── calendars
|	├── dateDisplay
|	|	├── biggieCalendarDate
|	|	├── calendarDateRange
|	|	├── calendarPageDate
|	|	└── mediumCalendarDate
|	├── fullPageCalendar (future)
|	├── miniCalendar (future)
```

PR #2374, opened 2020-02-11
Move Biggie Calendar Date

tn3rb:
- see #2349
- migrates biggie calendar date component over from EDTR v1
- fixes some logic in EspressoButton


assets/src/application/ui/calendars/dateDisplay/BiggieCalendarDate/index.tsx
```diff
@@ -0,0 +1,78 @@
...
+}) => {
+	if (!isValid(date)) {
+		return null;
+	}
+	const classes = classNames('ee-biggie-calendar-date-bg', htmlClass, {
+		'ee-bcd-pos-left': position === Position.LEFT,
+		'ee-bcd-pos-right': position !== Position.LEFT,
+	});
+	const editDateButton =
+		typeof onEdit === 'function' ? (
```
nicolad: maybe a personal preference, but I like more short circuit evaluation in this situations then ternary op

assets/src/application/ui/calendars/dateDisplay/BiggieCalendarDate/index.tsx
```diff
@@ -0,0 +1,78 @@
...
+	footerText,
+	position = Position.LEFT,
+	onEdit = null,
+	editButton = {},
+	showTime = false,
+}) => {
+	if (!isValid(date)) {
+		return null;
+	}
+	const classes = classNames('ee-biggie-calendar-date-bg', htmlClass, {
```
nicolad: Although I do understand what is used for, maybe we should use just `className`/ `classNames`.
And for `classnames` package to have either the original name `classnames` or as we did in the past project smth like `cx`

assets/src/application/ui/calendars/dateDisplay/BiggieCalendarDate/index.tsx
```diff
@@ -0,0 +1,78 @@
...
+				onClick={(event: React.MouseEvent<HTMLButtonElement>) => onEdit(event)}
+				onKeyPress={(event: React.MouseEvent<HTMLButtonElement>) => onEdit(event)}
+				tooltip={editButton.tooltip}
+				labelPosition={editButton.tooltipPosition}
+				icon='calendar'
+			/>
+		) : null;
+	return (
+		<div className={classes}>
+			{headerText && <div className='ee-biggie-calendar-date-header'>{headerText}</div>}
```
nicolad: :+1:

assets/src/application/ui/calendars/dateDisplay/BiggieCalendarDate/style.css
```diff
@@ -0,0 +1,185 @@
...
+ * Event Espresso BiggieCalendarDate Stylesheet
+ * © 2020 Event Espresso
+ * ------------------------------------------------------------------------- */
+
+.ee-biggie-calendar-date-bg {
+	box-sizing: border-box;
+	color: var(--ee-color-white);
+	height: 100%;
+	padding: var(--ee-padding-tiny) var(--ee-padding-micro);
+	position: absolute;
```
nicolad: Which one is the relative parent for this?

assets/src/application/ui/calendars/dateDisplay/BiggieCalendarDate/style.css
```diff
@@ -0,0 +1,185 @@
...
+	letter-spacing: var(--ee-letter-spacing-font-size-tiny);
+}
+
+.ee-biggie-calendar-date .ee-bcd-month {
+	font-size: calc(var(--ee-font-size-tiny) * 2);
+	font-weight: bold;
+	line-height: calc(var(--ee-font-size-tiny) * 2);
+}
+
+.ee-bcd-month-day-sep {
```
nicolad: we might need a helper class at some point

assets/src/application/ui/calendars/dateDisplay/BiggieCalendarDate/style.css
```diff
@@ -0,0 +1,185 @@
...
+		width: 100% !important;
+	}
+
+	.ee-biggie-calendar-date-bg.ee-bcd-pos-left {
+		float: none;
+	}
+
+	.ee-biggie-calendar-date-bg.ee-bcd-pos-right {
+		float: none;
+	}
```
nicolad: Perhaps we can use `flex-direction: row | column`? and avoid floats and clearing.

assets/src/domain/shared/entities/datetimes/helpers/statusBgColor.ts
```diff
@@ -0,0 +1,19 @@
+import { Datetime } from '@edtrServices/apollo/types';
+
+const statusBgColor = (date: Datetime): string => {
```
nicolad: Is this returning class name ?
If so, `statusBgColorClassName` would be more explicit

assets/src/application/ui/calendars/dateDisplay/BiggieCalendarDate/style.css
tn3rb: Guess some of these styles are too presumptuous about where this is going to be used... 
should maybe move this kind of control into the parent component, or allow it to be overridden via props

assets/src/application/ui/calendars/dateDisplay/BiggieCalendarDate/style.css
tn3rb: ya OK I'll try that... I'm an older dog, but can still learn some new tricks

assets/src/domain/shared/entities/datetimes/helpers/statusBgColor.ts
tn3rb: agreed! will do

assets/src/application/ui/calendars/dateDisplay/BiggieCalendarDate/index.tsx
tn3rb: I can do that

assets/src/application/ui/calendars/dateDisplay/BiggieCalendarDate/index.tsx
tn3rb: me no grok ?!?!?!

assets/src/application/ui/calendars/dateDisplay/BiggieCalendarDate/style.css
tn3rb: for doing what? setting the line height? or adding separators?

nicolad: it uses only `display: none;`
so, maybe just `.hide` class instead

assets/src/application/ui/calendars/dateDisplay/BiggieCalendarDate/style.css
nicolad: I can't say I am early adopter, I was very late at this flexbox party

nicolad: and there is also the CSS grid, which I didn't completely understood: https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Grid_Layout
***
fun fact, this :point_up: has even polyfill for IE11: http://tomrothe.de/posts/css_grid_and_ie11.html

assets/src/application/ui/calendars/dateDisplay/BiggieCalendarDate/index.tsx
manzoorwanijk: I agree with @nicolad about using `classNames` as variable name instead of `classes`.

assets/src/application/ui/calendars/dateDisplay/BiggieCalendarDate/index.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+export enum Position {
+	LEFT = 'left',
+	RIGHT = 'right',
+}
+
+export interface BiggieCalendarDateProps {
+	date: Date;
+	editButton: EditButtonProps;
+	htmlClass: string;
```
manzoorwanijk: As mentioned in [this comment](https://github.com/eventespresso/event-espresso-core/pull/2366#discussion_r378202670), we should decide about a standard.

assets/src/application/ui/calendars/dateDisplay/BiggieCalendarDate/index.tsx
```diff
@@ -0,0 +1,78 @@
...
+			{headerText && <div className='ee-biggie-calendar-date-header'>{headerText}</div>}
+			<div className='ee-biggie-calendar-date'>
+				<div className='ee-bcd-weekday'>{format(date, 'eeee')}</div>
+				<div className='ee-bcd-month'>{format(date, 'MMMM')}</div>
+				<div className='ee-bcd-month-day-sep'></div>
+				<div className='ee-bcd-day'>{format(date, 'ee')}</div>
+				<div className='ee-bcd-year'>{format(date, 'yyyy')}</div>
+				{showTime && <div className='ee-bcd-time'>{format(date, 'h:mm a')}</div>}
+			</div>
+			{footerText && <div className='ee-biggie-calendar-date-footer'>{footerText}</div>}
```
manzoorwanijk: what if `footerText` is an empty array?

assets/src/application/ui/calendars/dateDisplay/BiggieCalendarDate/index.tsx
manzoorwanijk: Same question about `headerText` if it's an empty array.

tn3rb:
In order to convert to flexbox I had to muck around with how things were implemented in the DateCard in terms of layout and style,
which lead to having to muck around with a bunch of other elements, so there's a bunch of extra commits on this PR that change things outside of the Calendar components.

Here's a screenie of the DateCard now:

EDTRv2 Date Card

PR #2376, opened 2020-02-11
Refactor entity action menu | Closes #2352

### LINKED ISSUE #2352: Migrate Entity Action Menu to EDTR v2
The EntityActionMenu is a container that accepts menu items and displays them as series of icon buttons, either in the sidebar of the Entity Cards:

Actions Menu - Grid View

 or in the last column for Entity List Tables:

Actions Menu - List View

Here's the tricky part... we **NEED** to be able to inject new menu items into the list from **ANYWHERE**, including from other plugins. The specific use case we are facing is with the REM addon (Recurring Events Manager) which needs to be able to inject a menu item for creating/editing recurrence patterns for event dates.

 Initially this had been done using WP core JS filters (`add_filter()` and `apply_filters()`) but they had a ton of problems with those due to how JS can operate asynchronously with no guarantee that a filter will run when you want/expect it to. Another developer that was also working on the EDTR at the time but is no longer with us, insisted that the only way to solve this issue was using Portals. I looked into that but could not figure out how to make that work with this use case. My understanding was that portals had to be created **before** you could put something into them, and there was no way for them to communicate with the logic that needed to insert something into the portal. The portal just existed and you could load something in there, but you had to know ahead of time what portal you needed to use. In our use case, the menu could be created at any time due to pagination, new date created, etc, and would be specific to particular entity (like the datetime with ID 1234) and the logic inserting something into the menu needs info about the entity in order to insert the correct data.

In `assets\ZZZ\components\ui\entity-actions-menu` this was achieved simply by using a function that allowed any code anywhere to register a menu item callback that would then be executed when the menu needed to be rendered. The callback would then receive the entity object for the datetime that was currently being displayed. 

The logic in the REM addon simply looked like this:

```js
import { registerEntityActionsMenuItem } from '@eventespresso/components';
import RecurringEventsManagerApp from './recurring-events-manager-app';

const eventId = parseInt( window.eeEditorEventId, 10 );

registerEntityActionsMenuItem(
	'datetime',
	'rem',
	( dateEntity, index ) => {
		return (
			<RecurringEventsManagerApp
				key={ index }
				index={ index }
				eventId={ eventId }
				dateEntity={ dateEntity }
			/>
		);
	}
);
```

and executed as soon as possible after code was loaded.

So when a datetime card was rendered, that callback would fire and an icon button for running the REM logic would get added to the datetime's action menu. Seemed to work fine but the other dev was NOT happy about it and insisted that portals were the way to go. Sorry for the long novel, but just wanted to make sure you have all of the background info on this.

Anyways, I don't care whether my solution above is not used. I am only interested in the most robust and reliable solution being implemented. If the other dev was correct that portals are the most stable and effective way to do what this needs to do, then I want THAT to be the solution. Don't worry about offending me if that is the case, trust me, I can handle being wrong about something. IF there is some completely different solution that can be utilized that will be more bullet proof than the ones I have discussed here, then that's what we want.

Within the logic for the existing EntityActionMenu there is a lot of garbage that is only there to restrict what elements could be injected into a menu (ie: `EntityActionMenuItem`). This can easily be replaced with a Typescript interface, but any logic testing for that may need to test for a base type, because there are still different kinds of menu items that get added (icon buttons, dropdown menus, etc), although maybe a single interface will work fine.

Any new logic added for EDTR v2:

- can go into `assets\src\application\ui\layout\entityActionMenu`

- must utilize TS

- must have tests written for any non-display logic

- result in menu items that can be navigated sequentially using the keyboard (a11y)

manzoorwanijk:
This PR:
- Creates a subscription based entity action menu
- Allows addons to hooks into (subscribe to) actons menu to add/remove entries
- Closes #2352


assets/src/application/ui/layout/entityActionMenu/types.ts
```diff
@@ -0,0 +1,59 @@
...
+export type EntityMenuItems<MenuKey extends string> = {
+	// menuKey: render()
+	[K in MenuKey]?: React.ReactNode;
+};
+
+export type MenuRegistry<EntityType extends string = '', MenuKey extends string = ''> = {
+	[K in EntityType]?: {
+		// entityId
+		[key: string]: EntityMenuItems<MenuKey>;
+	};
```
nicolad: a comment might be helpful here :smiley:

assets/src/application/ui/layout/entityActionMenu/useEntityActions.ts
```diff
@@ -0,0 +1,77 @@
...
+		}
+
+		const subscriptionId = uuidv4();
+
+		updateSubscription({ id: subscriptionId, callback, options, action: 'add' });
+
+		// to unsubscribe
+		return () => {
+			updateSubscription({ id: subscriptionId, action: 'remove' });
+		};
```
nicolad: Nice, the same pattern as with useEffect cleanup.

manzoorwanijk: Yes

assets/src/application/ui/layout/entityActionMenu/types.ts
manzoorwanijk: Added a comment above there.

assets/src/application/ui/layout/entityActionMenu/useEntityActions.ts
```diff
@@ -0,0 +1,77 @@
...
+
+	const setSubscriptions = (subscriptions: Subscriptions<EntityType, MenuKey>): void => {
+		updateEntityActions('subscriptions', subscriptions);
+	};
+
+	const updateSubscription = ({ id, callback, options, action }: UpdateSubscriptionProps<EntityType, MenuKey>) => {
+		const subscriptions = getSubscriptions();
+
+		const newSubscriptions =
+			action === 'add' ? { ...subscriptions, [id]: { callback, options } } : omit([id], subscriptions);
```
nicolad: Optional suggestion:
```
const newSubscription = { [id]: { callback, options } };
const newSubscriptions = action === 'add' ? { ...subscriptions, ...newSubscription } : omit([id], subscriptions);
```

nicolad: the idea is that that ternary is too long, so I tried to split somehow :smiley:

manzoorwanijk: `callback` and `options` are only passed when adding a subscription

assets/src/application/ui/layout/entityActionMenu/useEntityActions.ts
```diff
@@ -0,0 +1,77 @@
...
+
+		setSubscriptions(newSubscriptions);
+	};
+
+	const updateEntityActions = (key: keyof EntityActionsData<EntityType, MenuKey>, value: any) => {
+		window[NAMESPACE] = assocPath([domain, 'entityActions', key], value, window[NAMESPACE]);
```
nicolad: immutable 
image

assets/src/application/ui/layout/entityActionMenu/useEntityActionsManager.ts
```diff
@@ -0,0 +1,37 @@
...
+): EntityActionsManager<MenuKey> => {
+	const [menuItems, setMenuItems] = useState([]);
+
+	const getMapKey = (itemKey: MenuKey) => `${entityType}:${entityId}:${itemKey}`;
+
+	const registerMenuItem = (key: MenuKey, render: () => React.ReactNode): void => {
+		const mapKey = getMapKey(key);
+		if (!menuItems.includes(mapKey)) {
+			setMenuItems((items) => [...items, mapKey]);
+			menuRegistry = assocPath([entityType, entityId, key], render(), menuRegistry);
```
nicolad: this `render()` is a bit confusing at first glance, a comment or an extraction would be helpful

assets/src/domain/eventEditor/ui/datetimes/dateCard/DateActionsMenu.tsx
```diff
@@ -0,0 +1,32 @@
...
+}) => {
+	const menuItems = useDatesActionMenuItems(entity);
+
+	const style: CSSProperties = {
+		position: 'absolute',
+		right: '.5rem',
+		textAlign: 'right',
+		flexDirection: layout === 'vertical' ? 'column' : 'row',
+		zIndex: 1,
+		...(position === 'top' && {
```
nicolad: :+1:

assets/src/application/ui/layout/entityActionMenu/useEntityActionsManager.ts
manzoorwanijk: done

assets/src/domain/eventEditor/ui/datetimes/hooks/useDatesActionMenuHandler.tsx
```diff
@@ -1,29 +1,34 @@
...
-					key={date.id + 'assignTickets'}
-					{...menuItemProps}
-				/>
+				<EspressoButton icon='form' onClick={() => console.log('You clicked TAM')} {...menuItemProps} />
 			));
+
+			registerMenuItem('deleteTicket', () => {
+				const { isLoaded } = useStatus();
+				/* Delete button should be hidden to avoid relational inconsistencies */
+				return isLoaded(TypeName.tickets) && <DeleteDateButton id={date.id} {...menuItemProps} />;
```
nicolad: :+1:

assets/src/domain/eventEditor/ui/entityActionMenu/types.ts
```diff
@@ -1,24 +1,28 @@
...
 
-export type SubscriptionCallback<E extends Entity, MenuKey extends string> = (
-	data: SubscriptionData<E>,
-	entityActionsManager: EntityActionsManager<MenuKey>
-) => void;
+export type EntitySubscriptionCallback<
+	E extends Entity,
+	MenuKey extends string,
+	AO = AdditionalSubscriptionCbOptions
+> = SubscriptionCallback<MenuKey, SubscriptionData<E>, AO>;
```
nicolad: :nerd_face:

assets/src/domain/eventEditor/ui/entityActionMenu/useEntityActionMenuItems.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+	const subscriptions = getSubscriptions({ entityType: filterByEntityType ? entityType : null });
+
+	Object.values(subscriptions).forEach(({ callback }) => {
+		callback({ entityType, entity }, entityActionsManager, additionalOptions);
+	});
+
+	const menuItems = getMenuItems();
+
+	return Object.entries<React.ReactType>(menuItems).map(([menuKey, Component], i) => <Component key={menuKey + i} />);
```
nicolad: :grin:

PR #2382, opened 2020-02-12
Avoid fatal errors when permanently deleting registrations

joshfeck:

This doesn't fully solve the problem (the problem being there's a registration without a transaction) but it does avoid a fatal error and displays an error message instead. I'm not sure how some sites are ending up with registrations without transactions, but we've seen more than a few reports lately. Including:
https://eventespresso.com/topic/error-message-when-trying-to-delete-something-in-the-trash/
https://eventespresso.com/topic/permanently-delete-from-trash-errors/
https://eventespresso.com/topic/error-from-word-press/
https://eventespresso.com/topic/500-errors-critical-error-on-wordpress/?view=all#post-306784

## How has this been tested

- [x] Trash a registration 
- [x] directly manipulate the database and remove the related transaction's row in the db (or give it a new TXN ID if you want to change it back later)
- [x] Go to Event Espresso > Registrations > Trash and try to permanently delete the registration

## Checklist

* [x] My code is tested.


sethshoultes:
Looks like this fixes the fatal error and shows an error notice instead:
image

PR #2393, opened 2020-02-13
Check for event object before generating iCal file

joshfeck:

In the current version (and past versions) of EE4 if you delete an event, its related datetimes are not deleted. One problem that can happen is if there's an email with an ical link, and you click on it, and the event has been deleted, the server will return a fatal error. This PR adds a check to avoid the fatal error. 

## How has this been tested

- [x] Create an event
- [x] View the event in another browser window/tab
- [x] Trash, then perm. delete the event
- [x] switch back to that other browser window/tab (don't refresh!) and click the iCal link that's displayed next to the datetime
While you will not get an iCal file, you will not get a fatal error 
- [x] Also test an iCal link on a regular un-deleted event 

## Checklist

* [x] My code is tested.


sethshoultes:
The fatal error is gone and shows a blank page if the iCal link is clicked.

PR #2403, opened 2020-02-14
Replace @blueprintjs Toaster with AntD Notifications; closes #2396.

### LINKED ISSUE #2396: Replace @blueprintjs Toaster with AntD Notifications
- see https://ant.design/components/notification/

- would be nice if existing toaster hooks API could be maintained to reduce refactoring required in other functions and components, but feel free to look at usage within existing locations like Apollo, and if it will be easier and faster to just strip it all out and start from scratch, then just do that and/or feel free to change accordingly if you see room for any improvement (ie: better functionality, etc)

- convert existing toasts as follows:
    - useSuccessToast() => notification.success()
    - useErrorToast() => notification.error()
    - useInfoToast() => notification.info()
    - useLoadingToast() => notification.open() - with loading-3-quarters icon
    - useDismissToast() => notification.close() - OR... look at the "Update Message Content" section and consider using that for changing notices in some scenarios, such as when entity loading is completed or a CRUD operation finishes, instead of closing the previous notice and opening a second one

- also add new
    - useWarningToast() => notification.warning()

- keep notifications in the bottom right corner by default, but allow that to be controlled via API

- you can likely keep the ToasterHash function for generating message IDs but feel free to change if you have an easier/better idea

nicolad:
- Closes: https://github.com/eventespresso/event-espresso-core/issues/2396
- For more info check issue's [description](https://github.com/eventespresso/event-espresso-core/issues/2396#issue-565130501)
- Loading notification should be discussed separately, there are some possible alternative solutions:
  - https://ant.design/components/spin/
  - or suspense: https://itnext.io/what-the-heck-is-this-in-react-suspense-c5e641e487a


assets/src/application/services/notification/index.ts
```diff
@@ -0,0 +1,53 @@
...
+const getHash = (message): string => {
+	const msg = JSON.stringify(message);
+	return msg
+		.split('')
+		.reduce((a, b) => {
+			a = (a << 5) - a + b.charCodeAt(0);
+			return a & a;
+		}, 0)
+		.toString();
+};
```
tn3rb: this is likely going to be problematic because the AntD typing has the `message` prop set as a `ReactNode` whereas this is expecting a `string` (and will likely blow up if a `ReactNode` is run through `.split('')`). So we might have to use something else like a timestamp for the message keys.

assets/src/application/services/notification/index.ts
```diff
@@ -0,0 +1,53 @@
...
+
+const notification = (type: 'success' | 'info' | 'warning' | 'error' | 'loading') => ({
+	description,
+	duration,
+	message,
+	placement,
+}: ArgsProps): void => {
```
tn3rb: might want to have the inner callback return the message key (vitally important if we end up having to use a timestamp) so that consuming code can use it to edit or dismiss notices.

ie:

```
const loadingKey = loadingNotification(loadingNotice);

if (!loading) {
	dismissNotification(loadingKey );
}
```

assets/src/domain/eventEditor/services/apollo/queries/tickets/useFetchTickets.ts
```diff
@@ -40,7 +41,7 @@ const useFetchTickets = (skipFetch: boolean = null): FetchEntitiesResult<Tickets
 
 	if (!initialized) {
 		toaster.loading(loading, toasterMessage);
-		toaster.error(error);
+		successNotification({ message: `datetimes initialized` });
```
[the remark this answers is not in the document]
nicolad: oops,
thx

assets/src/domain/eventEditor/services/apollo/queries/tickets/useFetchTickets.ts
```diff
@@ -25,12 +26,12 @@ const useFetchTickets = (skipFetch: boolean = null): FetchEntitiesResult<Tickets
...
 			setIsLoaded(TypeName.tickets, true);
 			toaster.dismiss(toasterMessage);
-			toaster.success(`tickets initialized`);
+			successNotification({ message: `tickets initialized` });
 		},
 		onError: (error): void => {
 			setIsError(TypeName.tickets, true);
 			toaster.dismiss(toasterMessage);
-			toaster.error(error);
+			successNotification({ message: `datetimes initialized` });
```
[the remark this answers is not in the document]
nicolad: updated

assets/src/application/services/toaster/useInitToaster.tsx
```diff
@@ -12,7 +13,7 @@ const useInitToaster: ToasterInit = ({ loadingMessage, successMessage }): Toaste
 
 		if (!initialized) {
 			toaster.loading(loadingFlag, loadingMessage);
-			toaster.error(loadingError);
+			successNotification({ message: loadingError });
```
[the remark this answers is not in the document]
nicolad: it's embarrassing, thx

assets/src/application/services/notification/index.ts
nicolad: Used `uuid` instead: https://github.com/eventespresso/event-espresso-core/pull/2403/commits/1829b9e4f691e65fee7b6d2a7c3ad5a7822a8e1e

assets/src/application/services/notification/index.ts
nicolad: Thx for the suggestion.

nicolad: @tn3rb I have mixed feelings about this `useLoadingToast`, let's pull this outside this PR.

nicolad: The `dismissNotification` is confusing to me because for example if an operation did not finished I shouldn't be able to dismiss that and go ahead with another operation that might be based on the result of first operation.

nicolad: Another option might be `Spin` component, but still not :100: sure about that:
https://ant.design/components/spin/

PR #2405, opened 2020-02-17
Refactor inline edit input | Closes #2391

### LINKED ISSUE #2391: Convert InlineEditInput to AntD Typography Adapter
Currently the `assets\src\application\ui\input\InlineEditInput\index.tsx` component is too generic and looks like it is designed to work with an input that fires update commands on blur.

The AntD Typography components with the `editable` prop only fire their update callback when editing is complete and the user presses `Enter` so I don't think the state management is required. We do however want to wrap the component in our own component so that our application code is decoupled from it. I'm thinking we may want to have an adapter for each of the AntD Typography components:

- `InlineEditTitle` (or `InlineEditHeading`) using `antd.Typography.Title`

- `InlineEditText` using `antd.Typography.Text`

- `InlineEditParagraph` (or `InlineEditTextArea`) using `antd.Typography.Paragraph`

Which could utilize shared resources/sub-components if need be.

I would also like the following CSS applied in order to hide those edit icons from sighted users until they hover over the text:

```css
.ant-typography .ant-typography-edit {
	margin-left: -1rem;
	opacity: 0.001;
	transform: scale(0.001);
	transition: all 0.1s ease;
	transition-delay: 0.3s;
}
.ant-typography:hover .ant-typography-edit {
	margin-left: 0.5rem;
	opacity: 1;
	transform: scale(1);
}
```

I don't think there will ever be any need for this to integrate with RFF.

Would also be nice if keyboard support could be tested to see if it works, and if not, attempt to add it (see `assets\ZZZ\components\form\inputs\inline-edit-input\index.js` for how this was done previously)

manzoorwanijk:
This PR:
- Refactors the inline edit input to create  components viz. `InlineEditText`, `InlineEditHeading` and `InlineEditTextArea`
- Updates consumer components
- Closes #2391

### DIFF
--- a/assets/src/application/ui/input/CurrencyInput/index.tsx
+++ b/assets/src/application/ui/input/CurrencyInput/index.tsx
@@ -1,10 +1,9 @@
 // @ts-nocheck
-import React, { useState } from 'react';
+import React from 'react';
 import Currency from 'react-currency-formatter';
-import { EditableText } from '@blueprintjs/core/lib/esm';
 
-import { EspressoButton } from '@application/ui/input';
-import InlineEditInput from '../InlineEditInput';
+import { InlineEditText } from '@appInputs/InlineEditInput';
+import { parsedAmount } from '@appServices/utilities/money';
 
 const nullFunc = (args?: any) => {};
 
@@ -13,41 +12,24 @@ interface CurrencyInputProps {
 	amount: string | number;
 	placeholder?: string;
 	onConfirm?: (result?: { amount: string | number; id: string }) => void;
-	onCancel?: (id?: string) => void;
 }
 
-const CurrencyInput: React.FC<CurrencyInputProps> = ({
-	id = '',
-	amount = 0,
-	placeholder = '',
-	onConfirm = nullFunc,
-	onCancel = nullFunc,
-}) => {
-	const [editing, setEditing] = useState(false);
-	return editing ? (
-		<InlineEditInput
-			component={EditableText}
-			key={id}
-			isEditing={editing}
-			placeholder={placeholder}
-			defaultValue={amount}
-			value={amount}
-			onCancel={() => {
-				setEditing(false);
-				if (typeof onCancel === 'function') {
-					onCancel(id);
-				}
-			}}
-			onConfirm={(amount: string) => {
-				setEditing(false);
-				onConfirm({ amount, id });
-			}}
-			selectAllOnFocus
-		/>
-	) : (
+const CurrencyInput: React.FC<CurrencyInputProps> = ({ id = '', amount = 0, onConfirm = nullFunc }) => {
+	const safeAmount = parsedAmount(amount);
+	return (
 		<>
-			<Currency quantity={amount} />
-			<EspressoButton icon='edit' onClick={() => setEditing(true)} />
+			<Currency quantity={safeAmount} />
+			<InlineEditText
+				key={id}
+				onChange={(value: string) => {
+					const newAmount = parsedAmount(value);
+					if (newAmount !== amount) {
+						onConfirm({ amount: newAmount, id });
+					}
+				}}
+			>
+				{safeAmount}
+			</InlineEditText>
 		</>
 	);
 };
--- a/assets/src/application/ui/input/CurrencyInput/index.tsx
+++ b/assets/src/application/ui/input/CurrencyInput/index.tsx
@@ -10,10 +10,10 @@ interface CurrencyInputProps {
 	id: string;
 	amount: string | number;
 	placeholder?: string;
-	onConfirm?: (result?: { amount: string | number; id: string }) => void;
+	onChange?: (result?: { amount: string | number; id: string }) => void;
 }
 
-const CurrencyInput: React.FC<CurrencyInputProps> = ({ id = '', amount = 0, onConfirm = nullFunc }) => {
+const CurrencyInput: React.FC<CurrencyInputProps> = ({ id = '', amount = 0, onChange = nullFunc }) => {
 	const { formatAmount, beforeAmount, afterAmount } = useMoneyDisplay();
 	const formattedAmount = formatAmount(amount);
 	return (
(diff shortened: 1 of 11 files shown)


assets/src/application/ui/input/InlineEditInput/Editable.tsx
```diff
@@ -0,0 +1,32 @@
...
+import './style.css';
+import { EditableProps } from './types';
+
+const { Text, Title, Paragraph } = Typography;
+
+const EditableTypes = {
+	text: Text,
+	heading: Title,
+	textarea: Paragraph,
+};
```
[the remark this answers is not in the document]
manzoorwanijk: You can't use types as values.

assets/src/application/ui/input/InlineEditInput/InlineEditHeading.tsx
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
+
+import Editable from './Editable';
+import { HeadingProps } from './types';
+
+const InlineEditHeading: React.FC<HeadingProps> = ({ ...rest }) => {
+	return <Editable inputType='heading' level={4} {...rest} />;
+};
```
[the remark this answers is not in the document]
manzoorwanijk: What I have done is to set `level={4}` by default, if a value is passed, it will be changed.

assets/src/application/ui/input/CurrencyInput/index.tsx
```diff
@@ -10,10 +10,10 @@ interface CurrencyInputProps {
 	id: string;
 	amount: string | number;
 	placeholder?: string;
-	onConfirm?: (result?: { amount: string | number; id: string }) => void;
+	onChange?: (result?: { amount: string | number; id: string }) => void;
```
nicolad: tested this change and works well;
initially thought that it will be triggered on every keyup / input value change

assets/src/application/ui/input/CurrencyInput/index.tsx
```diff
@@ -1,53 +1,36 @@
...
-			<EspressoButton icon='edit' onClick={() => setEditing(true)} />
+			{beforeAmount}
+			<InlineEditText
+				key={id}
+				onChange={(value: string) => {
+					const newAmount = parsedAmount(value);
+					if (newAmount !== amount) {
+						onChange({ amount: newAmount, id });
+					}
+				}}
```
nicolad: perhaps this can be extracted into `onChangeHandler` variable ?

assets/src/application/ui/input/InlineEditInput/Editable.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+const Editable = ({ inputType, onChange, ellipsis, className, ...rest }: EditableProps) => {
+	const Component = EditableTypes[inputType];
+
+	const htmlClasses = classNames('ee-inline-edit', `ee-inline-edit-${inputType}`, className);
```
nicolad: `className` ?

assets/src/application/ui/input/InlineEditInput/InlineEditHeading.tsx
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
+
+import Editable from './Editable';
+import { HeadingProps } from './types';
+
+const InlineEditHeading: React.FC<HeadingProps> = ({ ...rest }) => {
+	return <Editable inputType='heading' level={4} {...rest} />;
```
nicolad: what about:
```
const level = rest.level || 4;
<Editable {...rest} inputType='heading' level={level}  />;
```
?
this way there is no guessing

assets/src/application/ui/input/InlineEditInput/InlineEditText.tsx
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
+import { TextProps } from 'antd/es/typography/Text';
+
+import Editable from './Editable';
+
+const InlineEditText: React.FC<TextProps> = ({ ...rest }) => {
+	return <Editable inputType='text' ellipsis {...rest} />;
```
nicolad: passing `rest` this way isn't possible to override input type?

assets/src/application/ui/input/InlineEditInput/InlineEditTextArea.tsx
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
+
+import Editable from './Editable';
+import { TextAreaProps } from './types';
+
+const InlineEditTextArea: React.FC<TextAreaProps> = ({ ...rest }) => {
+	return <Editable inputType='textarea' ellipsis={{ rows: 3, expandable: true }} {...rest} />;
```
nicolad: same here, let's pass `rest` first, this way we will be sure that input type isn't overriden
```
	return <Editable {...rest} inputType='textarea' ellipsis={{ rows: 3, expandable: true }} />;
```

assets/src/application/ui/input/InlineEditInput/Editable.tsx
manzoorwanijk: That is a prop and should not be mutated :)

assets/src/application/ui/input/InlineEditInput/InlineEditText.tsx
manzoorwanijk: That's restricted by types.

assets/src/application/ui/input/InlineEditInput/InlineEditTextArea.tsx
manzoorwanijk: Consumers should be able to make any customizations that may be necessary.

assets/src/application/ui/input/InlineEditInput/InlineEditHeading.tsx
manzoorwanijk: I don't think that makes it any different.

assets/src/application/ui/input/InlineEditInput/Editable.tsx
nicolad: understood, in this case I would go with:
```
const Editable = ({ inputType, onChange, ellipsis, ...rest }: EditableProps) => {
```
and
```
const className = classNames('ee-inline-edit', `ee-inline-edit-${inputType}`, rest.className);

```

PR #2408, opened 2020-02-18
Update useSuccessToast to make it use AntD notifications; closes #2396.

### LINKED ISSUE #2396: Replace @blueprintjs Toaster with AntD Notifications

nicolad:

- Updates `useSuccessToast` item  from https://github.com/eventespresso/event-espresso-core/issues/2396

### DIFF
--- a/assets/src/application/services/toaster/useSuccessToast.tsx
+++ b/assets/src/application/services/toaster/useSuccessToast.tsx
@@ -4,8 +4,10 @@ import { ArgsProps } from 'antd/lib/notification';
 
 import { SuccessToast } from './types';
 
+type SuccessToastCallback = (...ArgsProps) => void;
+
 const useSuccessToast: SuccessToast = (toaster, hash) =>
-	useCallback<any>(
+	useCallback<SuccessToastCallback>(
 		(message = '', description, duration, placement) => {
 			const args: ArgsProps = {
 				message,
--- a/assets/src/application/services/toaster/useSuccessToast.tsx
+++ b/assets/src/application/services/toaster/useSuccessToast.tsx
@@ -1,24 +1,22 @@
-import { Intent } from '@blueprintjs/core';
 import { useCallback } from 'react';
+import { notification } from 'antd';
+import { ArgsProps } from 'antd/lib/notification';
 
-import { SuccessToast, SuccessToastCallback } from './types';
+import { SuccessToast } from './types';
 
 const useSuccessToast: SuccessToast = (toaster, hash) =>
-	useCallback<SuccessToastCallback>(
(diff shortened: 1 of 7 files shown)


nicolad:
Without `setTimeout` it was showing only the last notification:
Peek 2020-02-18 15-39

assets/src/application/services/toaster/useSuccessToast.tsx
```diff
@@ -1,24 +1,22 @@
...
-						intent: Intent.SUCCESS,
-						icon: 'tick-circle',
-						message,
-						onDismiss,
-						timeout,
-					},
-					hash(message)
-				);
-			}
+	useCallback<any>(
```
manzoorwanijk: why `<any>`?

assets/src/application/services/toaster/useToaster.tsx
```diff
@@ -1,26 +1,17 @@
...
-		.reduce((a, b) => {
-			a = (a << 5) - a + b.charCodeAt(0);
-			return a & a;
-		}, 0)
-		.toString();
-};
-
 const useToaster = (): ToasterHook => {
 	const toaster = useContext(ToastContext);
+	const hash = uuidv4;
```
manzoorwanijk: `uuidv4` must be called to get the id - `uuidv4()`

nicolad: Updated https://github.com/eventespresso/event-espresso-core/pull/2408/commits/b08b78a3f0d9efa84e19ac3ee516a4db8432ee41

assets/src/application/services/toaster/useSuccessToast.tsx
nicolad: updated

PR #2412, opened 2020-02-18
Date details form with custom Date/Time inputs | Closes #2394

### LINKED ISSUE #2394: Build Edit Date Details Form
- opens in modal when edit button is clicked

- has datetime ID and event name in header

- subsection headers have related antd icons (might have to experiment with position)

- has following form sections and fields:
 (`  |  ` means inputs are on same line)

### Basics (profile icon)
	- Title - text 100% width
	- Description - textarea 100% width x 4 lines

### Date & Time (calendar icon)
	- Start Date - date picker 50% width  | Start Time - time picker 50% width
	- End Date - date picker 50% width  | End Time - time picker 50% width

### Details (control icon)
	- Capacity - number 33% width
	- Archived - toggle switch

- other details like help text, validation settings, etc, can be obtained by looking at `assets\ZZZ\editor\events\dates-and-times\editor-date\edit-form\use-date-entity-input-config.js`

- wire up to Apollo mutations and provide all feedback accordingly, ie: "updating", "success", or "error"

manzoorwanijk:
This PR:
- Adds AntD 4 to deps
- Adds custom `/components` to `EspressoForm` viz. `DatePicker`, `TimePicker`
- Creates Date edit form
- Creates a `CustomDate`  object to act as middleware between AntD and `date-fns`
- Creates TS `Event` type and `useEvent` hook.
- Closes #2394


assets/src/application/ui/forms/espressoForm/components/index.ts
```diff
@@ -0,0 +1,3 @@
+export { default as DatePicker } from './DatePicker';
+
+export { default as TimePicker } from './TimePicker';
```
tn3rb: how come these are not part of  `application/ui/forms/espressoForm/adapters/index.ts ` ?

assets/src/application/ui/forms/espressoForm/constants.ts
```diff
@@ -0,0 +1,3 @@
+export const dateFormat = 'yyyy-MM-dd';
+
+export const timeFormat = 'HH:mm:ss';
```
tn3rb: might want to move these to: `assets\src\application\constants\date-fns-formats.ts`

OR utilize the `dateTimeFormats` property from `useConfigData()`

assets/src/domain/shared/ui/dateRangeInput/dateDisplay.tsx
```diff
@@ -1,7 +1,8 @@
 import React from 'react';
-import { Icon, Tag } from 'antd';
+import { Tag } from 'antd';
 import classNames from 'classnames';
 import { format, isValid } from 'date-fns';
+import { ArrowRightOutlined } from '@ant-design/icons';
 
 const FORMAT = 'EE MMM eo, yyyy';
 const FORMAT_TIME = 'EE MMM eo, yyyy p';
```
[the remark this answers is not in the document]
manzoorwanijk: Remember we decided not to couple `EspressoForm` with EDTR code? 😄

assets/src/application/ui/forms/espressoForm/components/index.ts
manzoorwanijk: Adapters are for RFF `<->` AntD integration. These are simple Date/Time picker components that don't use `momentjs` like those of AntD pickers. They use vanila JS `Date`. These components can be used anywhere either in `EspressoForm` or even in EDTR.

assets/src/application/ui/forms/espressoForm/constants.ts
manzoorwanijk: Making `EspressoForm` dependent upon EDTR config is not a good idea.

assets/src/domain/shared/ui/dateRangeInput/dateDisplay.tsx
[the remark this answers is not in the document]
tn3rb: i was suggesting the constants in the application folder which would be shared by any domain and not exclusive to the EDTR (ie: `/domain/eventEditor/`)...
unless you mean something else?

assets/src/application/ui/forms/espressoForm/components/index.ts
tn3rb: this might seem kinda convoluted, but we should use adapters for both RFF and AntD, ie:

- AntD adapters to decouple us from that lib
- RFF adapters that wrap the above components with the Field component

the basic AntD adapters could go in `assets/src/application/ui/input/`
and the RFF Field adapters could go in `assets/src/application/ui/forms/espressoForm/adapters/`

reason being is that it is possible we will need inputs from time to time that are not going to be integrated with RFF

manzoorwanijk: It's not possible to create an adapter for RFF alone because RFF suplies `FieldRenderProps` which you have to pass down to the actual input component, in our case AntD components.

assets/src/application/ui/forms/espressoForm/constants.ts
tn3rb: the constants I suggested and the config are both part of the application, making them globally available to all domains.
but `EspressoForm` is also part of the application layer anyways and therefore would not be dependent on `/domain/eventEditor/` because of dependencies within its same layer.

Later on down the road when we are building out the UI for another use case that requires forms (maybe the registration form on the frontend of the site) we will still have anything in the application and infrastructure layers available to us. 

What we need to avoid doing is making logic outside of a domain use case dependent on logic within a particular domain (like the EDTR).
ie, we can NOT have the registration form on the frontend of the site dependent on the Event Editor!

assets/src/domain/shared/ui/dateRangeInput/dateDisplay.tsx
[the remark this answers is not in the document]
manzoorwanijk: Those formats are specific to AntD datepicker because they use momentJS. It can be consumed directly from `EspressoForm` when you use it.

assets/src/application/ui/forms/espressoForm/renderers/FormRenderer.tsx
```diff
@@ -29,21 +29,23 @@ const FormRenderer: React.FC<FormRendererProps> = (props) => {
 	} = props;
 
 	const formOutput = (
-		<div className='form-wrapper'>
-			<Form layout={layout}>
-				{sections.length ? <RenderSections sections={sections} /> : null}
+		<div className='ee-form'>
```
nicolad: nice and simple;
with my commit I was thinking that the container div should be in the `EspressoForm`

assets/src/application/ui/forms/espressoForm/adapters/DatePicker.tsx
```diff
@@ -0,0 +1,17 @@
...
+import { parse } from 'date-fns';
+import { DatePicker as CustomDatePicker } from '../components';
+import { dateFormat } from '../constants';
+
+const DatePicker = ({ input, ...rest }) => {
+	return (
+		<CustomDatePicker
+			{...input}
+			{...rest}
+			value={input.value ? parse(input.value, dateFormat, new Date()) : input.value}
```
nicolad: it's easier to read this when it is extracted as a separate variable

assets/src/application/ui/forms/espressoForm/adapters/types.ts
```diff
@@ -1,19 +1,24 @@
...
+
 export interface SelectOptionProps extends OptionProps {
 	optgroup?: string;
 	options?: Array<OptionProps>;
 }
 
 export type SelectOptions = Array<SelectOptionProps>;
 
-export type FormItemProps = AntdFormItemProps;
+export type FormItemProps = Omit<AntdFormItemProps, 'children'>;
```
nicolad: nice, it's better to omit, than to make it optional

assets/src/application/ui/forms/espressoForm/components/TimePicker.tsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+import DatePicker from './DatePicker';
+import { PickerTimeProps } from 'antd/es/date-picker/generatePicker';
+import { Omit } from 'antd/es/_util/type';
```
nicolad: very interesting

assets/src/domain/eventEditor/ui/datetimes/dateForm/useDateFormConfig.ts
```diff
@@ -0,0 +1,126 @@
...
+	name?: string;
+	description?: string;
+	capacity?: number;
+	dateTime?: {
+		startDate: string;
+		startTime: string;
+		endDate: string;
+		endTime: string;
+	};
+	isTrashed?: boolean;
```
nicolad: should this be derived from Date type?

manzoorwanijk: You mean a custom type?

assets/src/domain/eventEditor/ui/datetimes/dateForm/useDateFormConfig.ts
```diff
@@ -0,0 +1,126 @@
...
+	const startDate = start ? parseISO(start) : PLUS_ONE_MONTH;
+	const endDate = end ? parseISO(end) : PLUS_TWO_MONTHS;
+
+	const { onSubmit } = config;
+
+	const onSubmitFrom: DateFormConfig['onSubmit'] = ({ dateTime, ...rest }, form, ...restParams) => {
+		// convert "dateTime" object to proper "startDate" and "endDate"
+		const startDateStr = `${dateTime.startDate} ${dateTime.startTime}`;
+		const endDateStr = `${dateTime.endDate} ${dateTime.endTime}`;
+		const formatStr = `${dateFormat} ${timeFormat}`;
```
nicolad: perhaps, in the future, this logic might be extracted into some kind of utility functions and added some tests

assets/src/domain/eventEditor/services/apollo/types.ts
```diff
@@ -19,6 +19,27 @@ export interface EntityEdge<E = Entity, ConnectionTypeName = string> {
...
+	isPrimary: boolean;
+	isSoldOut: boolean;
+	isUpcoming: boolean;
+	length: number;
+	order: number;
+	reserved: number;
+	sold: number;
+	startDate: string;
+	status: DatetimeStatus;
+}
```
tn3rb: these are datetime properties !?!?!?

assets/src/domain/eventEditor/ui/datetimes/dateForm/useDateFormConfig.ts
```diff
@@ -1,14 +1,51 @@
...
+	isTrashed?: boolean;
+}
+
+type DateFormConfig = EspressoFormProps<DateFormShape>;
+
+const useDateFormConfig = (id: EntityId, config?: EspressoFormProps): DateFormConfig => {
+	const { name, description, capacity, startDate: start, endDate: end } = useDatetimeItem({ id }) || {};
+
+	const startDate = start ? parseISO(start) : PLUS_ONE_MONTH;
+	const endDate = end ? parseISO(end) : PLUS_TWO_MONTHS;
```
tn3rb: 👍

assets/src/domain/eventEditor/services/apollo/types.ts
manzoorwanijk: No, not at all! You can have a look at the GQL schema

assets/src/application/ui/forms/espressoForm/components/index.ts
tn3rb:
```js
const AntdAdapter = ({ name, value, etc, ...rest }) => {
    return <AntdComponent name={name} value={value} etc={etc} {...rest} />;
}

const rffAdapter = ({ ...rest }) => {
    return <Field component={ AntdAdapter } {...rest} />;
}
```

how would something similar to the above not work?


just looked again at what you have going on in:

- `\espressoForm\adapters` and 
- `\espressoForm\adapters\MappedField.tsx`
- `\espressoForm\fields\Field.tsx`
- `\espressoForm\renderers\FieldRenderer.tsx`

and it's basically the same thing as my example above... basically using spread operators to pass thru any additional props.
You're doing everything dynamically though, which i think is great, but the only thing that would need to change would be the import path in `MappedField` for all of the base components. Those could simply come from `assets/src/application/ui/input/`

manzoorwanijk: I will try that approach to see how it works. Thanks 👍

nicolad: agree, we might need to pass raw fields at some point in some specific cases or edge cases, so it would be nice to have a little bit of decoupling between rff and antd

assets/src/domain/eventEditor/ui/datetimes/dateForm/useDateFormConfig.ts
nicolad: I meant `Datetime` type, sorry for the confusion

assets/src/domain/eventEditor/services/apollo/types.ts
```diff
@@ -19,21 +19,17 @@ export interface EntityEdge<E = Entity, ConnectionTypeName = string> {
...
 	isSoldOut: boolean;
 	isUpcoming: boolean;
-	length: number;
 	order: number;
-	reserved: number;
-	sold: number;
-	startDate: string;
-	status: DatetimeStatus;
+	shortDesc: string;
 }
```
tn3rb: better, but should this not have all of the fields present in the PHP type definition: `core\domain\services\graphql\types\Event.php` ???

assets/src/application/ui/forms/espressoForm/adapters/Select.tsx
```diff
@@ -7,16 +7,16 @@ const SelectField: React.FC<SelectProps> = ({ input, mode, options, ...selectPro
...
 					))}
 				</Select.OptGroup>
 			);
 		}
 		return (
-			<Select.Option {...optionProps} key={`${optionProps.value}${index}`}>
+			<Select.Option {...optionProps} value={optionProps.value} key={`${optionProps.value}${index}`}>
 				{label}
 			</Select.Option>
 		);
```
nicolad: The exported default should reflext the filename.

nicolad: One solution is to update this:
`import { Select as DefaultSelect } from 'antd';
`
or smth similar

nicolad: or to rename the file

assets/src/application/constants/momentFormats.ts
```diff
@@ -0,0 +1,3 @@
+export const defaultDateFormat = 'yyyy-MM-dd';
+
+export const defaultTimeFormat = 'HH:mm:ss';
```
manzoorwanijk: They are not momentJS formats, they are `date-fns` equivalent formats for what AntD time picker passes to `onChange`

tn3rb: ahh ok... i got confused by this comment you made previously:

> Those formats are specific to AntD datepicker because they use momentJS. It can be consumed directly from EspressoForm when you use it.

from: https://github.com/eventespresso/event-espresso-core/pull/2412/files#r381373500

so these formats are used by us with date-fns to convert things to the same format use by moment.js when using the AntD datepicker.

I'll adjust this accordingly

PR #2416, opened 2020-02-20
Ticket Edit Form using EspressoForm config | Closes #2395

### LINKED ISSUE #2395: Create Edit Form for Ticket  Entity
- opens in modal when edit button is clicked

- has ticket ID and event name in header

- subsection headers have related antd icons (might have to experiment with position)

- has following form sections and fields:
 (`  |  ` means inputs are on same line)

### Basics (profile icon)
	- Title - text 100% width
	- Description - textarea 100% width x 4 lines
    - Price - text 50% width (calculator icon button)
    - Taxable - toggle switch  |  Reverse Calculate - toggle switch

### Ticket Sales (calendar icon)
	- Start Date - date picker 50% width  |  Start Time - time picker 50% width
	- End Date - date picker 50% width  |  End Time - time picker 50% width

### Details (control icon)
	- Quantity For Sale - number 33% width  |  Number of Uses - number 33% width
	- Minimum Quantity - number 33% width  |  Maximum Quantity - number 33% width
	- Required Ticket - toggle switch  |  Default Ticket - toggle switch
	- Archived - toggle switch

- other details like help text, validation settings, etc, can be obtained by looking at `assets\ZZZ\editor\events\tickets\editor-ticket\edit-form\use-ticket-form-input-config.js`

- wire up to Apollo mutations and provide all feedback accordingly, ie: "updating", "success", or "error"

manzoorwanijk:
This PR:
- Extracts some logic related to dates/times from date edit form config for reusability.
- Locks down `EspressoForm` types to type check field names in form configs
- Improves date edit form config
- Creates Ticket edit form config
- Fixes issues with Optimistic responses caused by the new way of form submission.
- Closes #2395

here is the form demo
image


assets/src/application/ui/forms/espressoForm/renderers/FieldRenderer.tsx
```diff
@@ -1,4 +1,5 @@
 import React from 'react';
+import classnames from 'classnames';
```
nicolad: Seems that it was used as `classNames`
image

assets/src/domain/eventEditor/services/apollo/mutations/datetimes/useOptimisticResponse.ts
```diff
@@ -1,5 +1,5 @@
 import { useCallback } from 'react';
-import { pathOr } from 'ramda';
+import { pathOr, filter, compose, not, isNil } from 'ramda';
```
nicolad: :+1:

assets/src/domain/eventEditor/services/apollo/mutations/datetimes/useOptimisticResponse.ts
```diff
@@ -40,19 +40,21 @@ const useOptimisticResponse = (): OptimisticResCb => {
 		let espressoDatetime = {
 			__typename: 'EspressoDatetime',
 		};
+		// Get rid of null or undefined values
+		const filtereInput = filter(compose(not, isNil), input);
```
nicolad: In CRA I've notice `.filter(Boolean)` being used for the same purpose;
your solution is also nice

manzoorwanijk: `.filter(Boolean)` would filter out the values like `0`, `""`, `false`

manzoorwanijk: Here is the difference
image

assets/src/application/ui/forms/espressoForm/renderers/FieldRenderer.tsx
manzoorwanijk: Updated

assets/src/domain/eventEditor/ui/tickets/ticketForm/useTicketFormConfig.tsx
```diff
@@ -0,0 +1,177 @@
...
+		return onSubmit(values, form, ...restParams);
+	};
+
+	const adjacentFormItemProps = {
+		labelCol: { span: 10 },
+		wrapperCol: { span: 14 },
+		labelAlign: 'left' as 'left',
+		style: {
+			width: '49%',
+			display: 'inline-flex',
```
nicolad: we'll need to improve this later or set these as default and allow to be overriden

assets/src/domain/eventEditor/ui/tickets/useAddTicketModal.ts
```diff
@@ -0,0 +1,36 @@
...
+import { useTicketMutator, CreateTicketInput } from '@edtrServices/apollo/mutations';
+import useEvent from '@edtrServices/apollo/queries/events/useEvent';
+
+const useAddTicketModal: EditorModal = () => {
+	const { createEntity } = useTicketMutator();
+	const { closeEditor } = useEditorModal();
+
+	const { name: eventName = '' } = useEvent() || {};
+
+	const onClose = useCallback<ModalClose>((): void => {
```
nicolad: If this callback isn't passing any props maybe we can simplify this and avoid callback?

assets/src/domain/eventEditor/ui/tickets/useEditTicketModal.ts
```diff
@@ -0,0 +1,36 @@
...
+import { useTicketMutator, UpdateTicketInput } from '@edtrServices/apollo/mutations';
+import useEvent from '@edtrServices/apollo/queries/events/useEvent';
+
+const useEditTicketModal: EditorModal = (entityId) => {
+	const { updateEntity } = useTicketMutator(entityId);
+	const { closeEditor } = useEditorModal();
+
+	const { name: eventName = '' } = useEvent() || {};
+
+	const onClose = useCallback<ModalClose>((): void => {
```
nicolad: Same here, can we avoid using `useCallback` here?

assets/src/domain/shared/services/utils/processDateAndTime.ts
```diff
@@ -0,0 +1,32 @@
...
+		const startDateStr = `${dateTime.startDate} ${dateTime.startTime}`;
+		startDate = parse(startDateStr, formatStr, backupDate);
+	}
+
+	if (dateTime.endDate && dateTime.endTime) {
+		const endDateStr = `${dateTime.endDate} ${dateTime.endTime}`;
+		endDate = parse(endDateStr, formatStr, backupDate);
+	}
+	return { startDate, endDate };
+};
```
nicolad: :+1:

assets/src/domain/eventEditor/ui/tickets/ticketForm/useTicketFormConfig.tsx
manzoorwanijk: Agreed, that's why I added that `className`

assets/src/domain/eventEditor/ui/tickets/useEditTicketModal.ts
manzoorwanijk: It has a dependency `closeEditor`

tn3rb: the dependency doesn't matter at all, he's suggesting we simply return a callback that is not memoized, ie:

```js
const onClose = (): void => {
    closeEditor('editTicket');
};
```

but the problem with that is a new function will be created every time this hook is called, which could easily cause unnecessary re-renders in components that consume this callback. Since this is ultimately using a provider to get the required logic, I think it's fair to use memoization here.

nicolad: Great explanation, @tn3rb, thx.
My understanding is that this `useCallback` is preventing re-renders from a different cause/source.
To me, it looks weird to apply memoization to a close handler or any other similar handler which is just dispatching an action.

nicolad: We need to pinpoint the exact source of re-renders.
Applying `useCallback` in cases where it is not required makes it hard to reason about the code.

tn3rb: > My understanding is that this useCallback is preventing re-renders from a different cause/source.

Not sure I totally grok what you are saying...

It's not preventing from the source, ie: where `closeEditor()` is coming from, it's preventing re-renders "downstream (ie: consumers of this hook). In other words, any components that receive the `onClose()` callback will unnecessarily re-render every time this hook is called, unless  `onClose()` is memoized. Sorry if I'm repeating something you already totally get.

> We need to pinpoint the exact source of re-renders.

I agree, and normally a component will only render when its props change. The problem with passing functions as props is that they will ALWAYS be considered changed because of the way that props are compared, even if the function itself has not undergone any changes. 

For example, in the following totally contrived example, a function is passed as a prop to another component:

```js
const ComponentA = () => {
    const getBobsName = () => 'Bob';
    return (
        <ComponentB callback={ getBobsName } />
    );
}
```

`ComponentB` will be re-rendered everytime `ComponentA` renders because it will see the incoming `getBobsName` function as being different than the one it was passed previously, even though there is no possible way for that function to be anything different. Memoizing that callback however would prevent `ComponentB` from re-rendering. In the case of a callback as simple as this, it's probably more expensive to memoize the result than it is to simply rerun the function, so memoization here would be unnecessary. The callback that Manzoor memoized is pulling data from a Provider, so one would think that is more expensive of an operation and therefore memoization is likely useful.

So anytime a callback of any kind of higher complexity is being passed as a prop, it's likely wise to just memoize it.

Now that said, in this case, since we are dealing with a hook, we could check if the callback is actually being passed as a prop to another component. If it is NOT, and the callback is being utilized immediately within the consuming component, then the memoization is unnecessary (assuming the hook is not used elsewhere). But without even looking, we know that this is not the case here, because the  `onClose()` and `onSubmit()` callbacks are most certainly being passed along to another component to be used as actions that are triggered by form events.

So bottom line:  any callback that is being passed as a prop to another component should be memoized, **UNLESS** that callback is so painfully simple that memoization would likely decrease performance.

But I totally agree with your comments with regards to other forms of memoization,  I also appreciate your vigilance with regards to maintaining and the readability of code 👍 So plz do not hesitate to comment on things like this again in the future, as it is always good to be reminded about these things.

assets/src/domain/eventEditor/services/apollo/mutations/datetimes/useOptimisticResponse.ts
```diff
@@ -40,19 +40,21 @@ const useOptimisticResponse = (): OptimisticResCb => {
 		let espressoDatetime = {
 			__typename: 'EspressoDatetime',
 		};
+		// Get rid of null or undefined values
+		const filteredInput = filter(compose(not, isNil), input);
```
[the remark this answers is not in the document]
manzoorwanijk: We definitely can but `Empty` is not the proper description for that because it only filters nil values - `null`,  `undefined`

assets/src/application/constants/dateFnsFormats.ts
```diff
@@ -23,3 +23,6 @@ export const LOCALIZED_DATE_AND_TIME_FORMAT = 'Pp'; // 12/31/2020 11:59:59 PM
 export const LOCALIZED_DATE_AND_TIME_SHORT_FORMAT = 'PPpp'; // December 31, 2020 11:59:59 PM
 export const LOCALIZED_DATE_AND_TIME_LONG_FORMAT = 'PPPppp'; // December 31st, 2020 at 11:59:59 PM GMT+08:00
 export const LOCALIZED_DATE_AND_TIME_FULL_FORMAT = 'PPPPpppp'; // Thursday, December 31st, 2020 at 11:59:59 PM GMT+08:00
+
+export const CONVERT_TO_MOMENT_DATE_FORMAT = 'yyyy-MM-dd';
+export const CONVERT_TO_MOMENT_TIME_FORMAT = 'HH:mm:ss';
```
manzoorwanijk: Nice 👍

assets/src/application/services/utilities/predicates/removeNullAndUndefined/index.ts
```diff
@@ -0,0 +1,7 @@
+import { filter, compose, not, isNil } from 'ramda';
+
+type Obj = { [key: string]: any };
+
+const removeNullAndUndefined = <T>(filterable: T[] | Obj): T[] | Obj => filter(compose(not, isNil), filterable);
```
manzoorwanijk: Types here are not good, how can two types be same when some of its props have been removed? It's only possible if return type is `Partial<T>` .

assets/src/application/ui/layout/EntityCard/styles.scss
```diff
@@ -41,9 +41,49 @@
...
+	@media only screen and (max-width: 480px) {
+		flex-flow: column nowrap;
+		&__menu,
+		&__sidebar,
+		&__details-wrapper {
+			width: 100%;
+		}
+		&__menu {
+			align-items: stretch;
+			box-sizing: border-box;
```
manzoorwanijk: Cool 👍

assets/src/domain/eventEditor/services/apollo/mutations/datetimes/useOptimisticResponse.ts
```diff
@@ -40,19 +41,21 @@ const useOptimisticResponse = (): OptimisticResCb => {
 		let espressoDatetime = {
 			__typename: 'EspressoDatetime',
 		};
+		// Get rid of null or undefined values
+		const filteredInput = removeNullAndUndefined(input);
```
manzoorwanijk: better

PR #2429, opened 2020-02-24
Create hidden input | Closes #2426

### LINKED ISSUE #2426: Create Hidden Input Component
there are some cases where we will need a hidden input that integrates with RFF
can be added to `assets\src\application\ui\input`

manzoorwanijk:
This PR:
- Creates a hidden input for EspressoForm
- Closes #2426

### DIFF
--- a/assets/src/application/ui/forms/espressoForm/renderers/FieldRenderer.tsx
+++ b/assets/src/application/ui/forms/espressoForm/renderers/FieldRenderer.tsx
@@ -11,6 +11,11 @@ const FieldRenderer: React.FC<FieldRendererProps> = (props) => {
 
 	const formItemLayout = useFormItemLayout();
 
+	// no layout stuff needed for hidden field
+	if (props.fieldType === 'hidden') {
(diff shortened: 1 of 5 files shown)


assets/src/application/ui/forms/espressoForm/renderers/FieldRenderer.tsx
```diff
@@ -11,6 +11,11 @@ const FieldRenderer: React.FC<FieldRendererProps> = (props) => {
 
 	const formItemLayout = useFormItemLayout();
 
+	// no layout stuff needed for hidden field
+	if (props.fieldType === 'hidden') {
+		<MappedField {...rest} />;
```
nicolad: is it missing a return here or I don't fully understand what is the intention, because below we have another:
`<MappedField {...rest} />` inside `FormItem`

manzoorwanijk: Already pushed a fix

Issue #1447, opened 2019-08-01
Message template previews initiated from event editor should use the loaded event as the base for the preview


joshfeck:
Currently the preview just uses whatever event gets loaded by the preview data handler and Darren suggested the preview should consider the loaded event.

When the the "edit" link for a message template (or create link) is clicked from the event editor, it should use the event being edited.  Here's some things that will have to happen to make this possible

* [ ] Include EVT_ID in the request from the event editor.
* [ ] Modify the messages preview generator so that it passes along the EVT_ID to the preview data handler.
* [ ] Make sure that if EVT_ID is passed along to the preview data handler that it allows using that event even if it is NOT published.

joshfeck:
https://eventespresso.com/topic/email-previews-and-test-emails/

sethshoultes:
We have two customers this week that are confused by this issue:
https://secure.helpscout.net/conversation/1075997485/742037/
https://secure.helpscout.net/conversation/1055626170/734585/

PR #2447, opened 2020-02-25
Fet/live event msg preview

Pebblo:
See #1447 

The message system already checks for an event id set on the request for it to use for data so this just adds the that to the evt_id query var it uses.

## How has this been tested
(Make sure the event you are editing is not the 'first' event you created, meaning you can't test this on a 'fresh' install as you won't know if you're getting the correct event data)

Edit an event (make a note of the event id) and click on the edit link for a message type, then preview that message type.

Check that the preview now relates to the event in question.
Confirm the shortcodes do actually relate to the current event.

You'll know pretty quick if this isn't working, it'll error or you'll get the preview data and not your current event.

Click the link to go back to the message editor and confirm the evt_id query var is still set and this it is the correct ID. 

Switch recipients and confirm the above is still set and its the correct ID.

Repeat the above a draft event.

## Checklist

* [x] My code is tested.

### DIFF
--- a/admin_pages/messages/Messages_Admin_Page.core.php
+++ b/admin_pages/messages/Messages_Admin_Page.core.php
@@ -2139,6 +2139,11 @@ class Messages_Admin_Page extends EE_Admin_Page
         }
 
         EE_Registry::instance()->REQ->set('GRP_ID', $this->_req_data['GRP_ID']);
+        
+        // if we have an evt_id set on the request, use it.
+        $EVT_ID = isset($this->_req_data['evt_id']) && ! empty($this->_req_data['evt_id'])
+        ? absint($this->_req_data['evt_id'])
+        : false;
 
 
         // get the preview!
@@ -2156,7 +2161,7 @@ class Messages_Admin_Page extends EE_Admin_Page
         // let's add a button to go back to the edit view
         $query_args = array(
             'id'      => $this->_req_data['GRP_ID'],
-            'evt_id'  => $this->_req_data['evt_id'],
+            'evt_id'  => $EVT_ID,
             'context' => $this->_req_data['context'],
             'action'  => 'edit_message_template',
         );
--- a/admin_pages/messages/Messages_Admin_Page.core.php
+++ b/admin_pages/messages/Messages_Admin_Page.core.php
@@ -1308,6 +1308,10 @@ class Messages_Admin_Page extends EE_Admin_Page
             ? absint($this->_req_data['id'])
             : false;
 
+        $EVT_ID = isset($this->_req_data['evt_id']) && ! empty($this->_req_data['evt_id'])
+        ? absint($this->_req_data['evt_id'])
+        : false;
+
         $this->_set_shortcodes(); // this also sets the _message_template property.
         $message_template_group = $this->_message_template_group;
         $c_label = $message_template_group->context_label();
--- a/admin_pages/messages/Messages_Admin_Page.core.php
+++ b/admin_pages/messages/Messages_Admin_Page.core.php
@@ -1738,6 +1738,7 @@ class Messages_Admin_Page extends EE_Admin_Page
                 'messenger'    => $message_template_group->messenger(),
                 'context'      => $context,
                 'GRP_ID'       => $GRP_ID,
+                'evt_id'       => $EVT_ID,
                 'action'       => 'preview_message',
             ),
             $this->_admin_base_url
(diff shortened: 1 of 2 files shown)


Pebblo:
Josh, I'm not sure if you want to review this or Brent but its basically just adding a query var to various links used by the messages system.

The code to pull a specific event was already in the system and then the everything uses the event for the related objects (except registration data) so that's all that was needed.

admin_pages/messages/Messages_Admin_Page.core.php
```diff
@@ -1308,6 +1308,10 @@ protected function _edit_message_template()
             ? absint($this->_req_data['id'])
             : false;
 
+        $EVT_ID = isset($this->_req_data['evt_id']) && ! empty($this->_req_data['evt_id'])
+        ? absint($this->_req_data['evt_id'])
+        : false;
```
tn3rb: although PHP allows it, we want to avoid mixing data types in vars as much as possible because it helps to avoid an entire category of bugs that result from mismatched types.
So instead of mixing integers and booleans here, it's better to just stick with integers and use `0` instead of `false` which should still work the same in most conditionals since it evaluates as "falsey"

ie:

```php
        $EVT_ID = isset($this->_req_data['evt_id']) && ! empty($this->_req_data['evt_id'])
        ? absint($this->_req_data['evt_id'])
        : 0;
```

that said, you should verify (if you haven't already) that the models or any other logic downstream that will be consuming the results of this code, are NOT performing a strict comparison and expecting to see `false`

ie:

```php
if ( isset($this->_req_data['evt_id'] && $this->_req_data['evt_id'] === false ) { // do something }
```

which is easy enough to fix simply by replacing the strict comparison with a loose comparison:

```php
if ( isset($this->_req_data['evt_id'] && $this->_req_data['evt_id'] ) { // do something }
```

or keeping things strict but incorporating type conversion

```php
if ( isset($this->_req_data['evt_id'] && (int)$this->_req_data['evt_id'] > 0 ) { // do something }
```

or 

```php
if ( isset($this->_req_data['evt_id'] && absint($this->_req_data['evt_id']) > 0 ) { // do something }
```

or ... 

```php
// insert your preferred approach here
```

core/libraries/messages/EE_messenger.lib.php
```diff
@@ -556,7 +556,7 @@ protected function _get_admin_content_events_edit($message_types, $extra)
                 }
                 // if there is no $default_value then we set it as the global
                 $default_value = empty($default_value) ? $mtpgID : $default_value;
-                $edit_url = EEH_URL::add_query_args_and_nonce(array( 'page' => 'espresso_messages', 'action' => 'edit_message_template', 'id' => $default_value ), admin_url('admin.php'));
+                $edit_url = EEH_URL::add_query_args_and_nonce(array( 'page' => 'espresso_messages', 'action' => 'edit_message_template', 'id' => $default_value, 'evt_id' => $event_id ), admin_url('admin.php'));
```
tn3rb: might also improve readability here to extract those query vars into a variable
(i know this has not been done so much in a lot of older code, but moving forwards, it is a good practice)

ex:

```php
$query_args = [
	'page' => 'espresso_messages',
	'action' => 'edit_message_template',
	'id' => $default_value,
	'evt_id' => $event_id
];
$edit_url = EEH_URL::add_query_args_and_nonce($query_args, admin_url('admin.php'));
```

admin_pages/messages/Messages_Admin_Page.core.php
Pebblo: Whilst I don't have any issues with changing the above to always be an int, I went with using `false` because that's how it's set up for `GRP_ID` just above. I did dig into it a little previously and I think the reason it currently uses `false` is because we use [add_query_arg()](https://developer.wordpress.org/reference/functions/add_query_arg/) to build out the query string.

Setting a query var's value to false, removes the pair from the string and so avoids things like:

`admin.php?page=espresso_messages&message_type=registration&messenger=email&context=attendee&GRP_ID=3&evt_id=0&action=preview_message&preview_message_nonce=0608dc82cf`

For the default preview links when editing them through Event Espresso -> Messages.

The rest of the system handles `evt_id=0` just fine but I could add some conditionals to remove it if preferred?

tn3rb: ahh ok... it didn't occur to me to consider `add_query_arg()` 

that's perfectly fine then as is, thnx for explaining.

core/libraries/messages/EE_messenger.lib.php
Pebblo: Kk, done.

I did the create_url code the same way seeing how it was right next to it anyway.

admin_pages/messages/Messages_Admin_Page.core.php
Pebblo: Thanks to Brents eagle eye I found a notice thrown when previewing messages from the default message templates editor, no evt_id is set on that request but I was trying to use it.

joshfeck:
This works well.

PR #2470, opened 2020-02-28
Fixed issue of wait list button when Google recaptcha is enabled

hoseinrafiei:
This fix is related to https://github.com/eventespresso/eea-wait-lists/issues/21 issue. I checked the JS code of Google Recaptcha and found that it is disabling the submit button after validating the token.

I didn't removed the code completely but I added a new data parameter to be able to handle when the button should be disabled and when shouldn't.  Then I added this feature for ticket selection form and turned it off for the wait-list form. Now the submit button on wait-list form won't disable after Google recaptcha validation.

### DIFF
--- a/caffeinated/modules/recaptcha_invisible/assets/espresso_invisible_recaptcha.js
+++ b/caffeinated/modules/recaptcha_invisible/assets/espresso_invisible_recaptcha.js
@@ -99,7 +99,8 @@ function espressoLoadRecaptcha() {
                             setTimeout(
                                 function() {
                                     $submit.click();
-                                    if (eeRecaptcha.disable_submit !== true) {
+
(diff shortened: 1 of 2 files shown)


tn3rb:
@joshfeck 

we'll need to verify that this does not adversely affect the "View Details" button, although I don't think it will

sethshoultes:
Testing notes: https://github.com/eventespresso/eea-wait-lists/issues/21#issuecomment-589812035

sethshoultes:
This is working for me. 

Testing notes:
- Created a sold-out event with a wait list scenario
- Clicked the sign up for wait list button
- Clicked the join wait list button without form details and triggered form validation
- Entered details into form and completed the form successfully
- I also verified the view details, register now, finalize registration, and view order confirmation buttons are working as expected

joshfeck:
I also tested this, works as expected with no regressions found.

Issue #2451, opened 2020-02-25
Move Entity Modal Hooks into `/hooks/` Folder


tn3rb:
- the following modal hooks:
     - `/domain/eventEditor/ui/datetimes/useAddDatetimeModal.ts`
     - `/domain/eventEditor/ui/datetimes/useEditDatetimeModal.ts`
     - `/domain/eventEditor/ui/tickets/useAddTicketModal.ts`
     - `/domain/eventEditor/ui/tickets/useEditTicketModal.ts`

- should be moved (accordingly) into 
     - `/domain/eventEditor/ui/datetimes/hooks/`
     - `/domain/eventEditor/ui/tickets/hooks/`

PR #2482, opened 2020-03-03
Modal changes | Closes #2480, #2451

### LINKED ISSUE #2480: Modal Changes
- see `\application\ui\layout\editorModal`

- in the recent TAM PR, a new modal component was added that is not integrated with RFF (React Final Form) and will likely find additional usages as time passes. To clarify that the existing modal **IS** integrated with RFF, we should rename it from `EditorModal` to `FormModal` and update all usages accordingly.

- when opening `EditorModal` to make changes to an entity, the modal initially appears to way down in the bottom right corner of the screen with over half of the modal completely off screen. After not even a second, the modal centers itself correctly in the middle of the screen. In contrast, then non-RFF modal does not do this and appears to originate from the center of the screen and remain centered. We need to find out **why** the `EditorModal` behaves like this and fix it

- the `EditorModal` header text should include the DBID when editing an entity, and maybe change "Update" to "Edit", example: "Edit datetime #123"

- the debug data displayed at the bottom of the forms should be moved into some kind of collapsible container such as https://ant.design/components/collapse/

manzoorwanijk:
This PR:
- Changes`/application/ui/layout/editorModal` to `/application/ui/layout/formModal` and renames the underling hooks/types/paths/components
- Changes`/shared/ui/editorModal` to `/shared/ui/formModal` and renames the underling hooks/types/paths/components
- Changes `/application/ui/forms/formModal` to `/application/ui/forms/modalForm`, which IMHO is a better representation of the coponent.
- Updates all the consumer components to use the renamed components/hooks
- Fixes the modal position when it opens
- Adds `dbId` to form modal title
- Extracts and makes form DebugInfo collapsable
- Moves Entity Modal Hooks into `/hooks/` Folder
- Closes #2480, #2451 

Here is the demo of the modal and `DebugInfo`
demo

With commit [a2750a0](https://github.com/eventespresso/event-espresso-core/pull/2482/commits/a2750a078e9736e806bead4163326c759db878be), I have the form/modal header and footer always visible and made only the body scrollable (without scrollbars).
Here is the demo for that. You may see the difference from the above demo
demo


assets/src/application/ui/display/DebugInfo.tsx
```diff
@@ -0,0 +1,38 @@
+import React, { CSSProperties } from 'react';
+import { Collapse } from 'antd';
+
+const { Panel } = Collapse;
+
+const dataStyle: CSSProperties = {
```
nicolad: can this be moved to a `.scss` file?

manzoorwanijk: Yes, but I thought it's a dev only component, so no need to extract to avoid bundling that with production CSS

assets/src/application/services/context/FormModalProvider.tsx
```diff
@@ -0,0 +1,16 @@
+import React, { createContext } from 'react';
```
nicolad: Good idea to have a context for a form modal :+1:

assets/src/application/ui/forms/modalForm/RenderModalForm.tsx
```diff
@@ -58,17 +44,18 @@ const FormModalForm: React.FC<FormModalFormProps> = ({
...
 			onOk={onClose}
-			style={modalStyle}
 			okText={__('Submit')}
 			okButtonProps={submitButton}
 			cancelText={__('Reset')}
 			cancelButtonProps={resetButton}
 			onCancel={onClose}
 			width={'80%'}
+			wrapClassName='ee-modal-form'
+			bodyStyle={{ padding: 0 }}
```
nicolad: hmm, is this a temp?

assets/src/application/ui/forms/modalForm/styles.scss
```diff
@@ -0,0 +1,25 @@
+.ee-modal-form {
+	& > div {
+		box-sizing: border-box;
+		max-width: 1200px;
+		min-height: 50%;
+		min-width: 320px;
+		height: calc(100vh - 100px);
```
nicolad: :taco:

assets/src/domain/shared/ui/formModal/useEditors.ts
```diff
@@ -0,0 +1,28 @@
+import { useMemo } from 'react';
+
+import { FormModalsHook } from './types';
+import useAddDatetimeModal from '@edtrUI/datetimes/hooks/useAddDatetimeModal';
+import useEditDatetimeModal from '@edtrUI/datetimes/hooks/useEditDatetimeModal';
+import useAddTicketModal from '@edtrUI/tickets/hooks/useAddTicketModal';
+import useEditTicketModal from '@edtrUI/tickets/hooks/useEditTicketModal';
+import useTicketPriceCalculatorModal from '@edtrUI/tickets/ticketPriceCalculator/hooks/useTicketPriceCalculatorModal';
+
+const useEditors: FormModalsHook = ({ entityId, entityDbId }) => {
```
nicolad: handy :1st_place_medal:

assets/src/application/ui/display/DebugInfo.tsx
nicolad: understood
`emotion` would be a great alternative here (just thinking out loud)

assets/src/application/ui/forms/modalForm/RenderModalForm.tsx
manzoorwanijk: No, it's to avoid hardcoded padding set by AntD, I have added that padding in `.scss`

PR #2505, opened 2020-03-05
Reset pagination when entity count changes | Closes #2492

### LINKED ISSUE #2492: Reset Pagination After Entity List Filters Change
#### Steps to Reproduce

- refresh page and wait till everything loads but don't touch any filters or pagination yet

- if you do not have more than 6 dates in the list already, then change the pagination "per page" value to 2

 - click to view page 2 of the dates

 - now click to view filters, and from the "show" dropdown, select "all active and upcoming" (assuming that your dates list has at least one active or upcoming date in it)

 - the "no results found" notice should appear

 - if you now click on the "page 1" button in the pagination, a result should appear

Looks like we need to reset the pagination if the number of results changes as a result of other filters.

manzoorwanijk:
This PR:
- Makes changes to `useEntityPagination` hook to make sure the pagination resets when entity count changes.
- Closes #2492

### DIFF
--- a/assets/src/application/ui/layout/entityList/pagination/useEntityPagination.ts
+++ b/assets/src/application/ui/layout/entityList/pagination/useEntityPagination.ts
@@ -17,6 +17,7 @@ const useEntityPagination = <E extends Entity>({ entities }: EntityListComponent
 		if (previousCount.current !== total) {
 			// reset page number to 1
 			setPageNumber(1);
+			previousCount.current = total;
 		}
 	}, [total]);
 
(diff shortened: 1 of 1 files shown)


assets/src/application/ui/layout/entityList/pagination/useEntityPagination.ts
```diff
@@ -10,6 +10,13 @@ const useEntityPagination = <E extends Entity>({ entities }: EntityListComponent
 	const [perPage, setPerPage] = useState(6);
 	const paginatedEntities = slice<E>(perPage * (pageNumber - 1), perPage * pageNumber, entities);
 	const total = entities.length;
+	const previousCount = useRef(total);
```
nicolad: clever :+1:

PR #2511, opened 2020-03-06
MOAR Filter Bar Tweaks and Other Various Style Fixes

tn3rb:
- more fixes, primarily for a11y involving focus states for inputs and trying to make everything look consistent

- did a little code cleanup and extraction too

- still having a problem where using the keyboard to navigate the filter bar inputs and pressing ENTER to see dropdown options will submit the WP editor form... and AntD does not provide a click event for the select input (not them directly but react-select or whatever lib they are using) might create an issue for this

### DIFF
--- a/assets/src/application/services/utilities/cancelClickEvent.ts
+++ b/assets/src/application/services/utilities/cancelClickEvent.ts
@@ -1,0 +1,23 @@
+import React from 'react';
+
+const { console } = window;
+
+export type ClickEvent = React.KeyboardEvent | React.MouseEvent;
+
+interface CancelClickInterface {
+	(click: ClickEvent, source?: string): void;
+}
+
+/**
+ * utility for blocking click events
+ * and displaying debug data in dev environments
+ */
+export const cancelClickEvent: CancelClickInterface = (click, source = ''): void => {
+	if (click && typeof click.preventDefault === 'function') {
+		click.preventDefault();
+		click.stopPropagation();
+		if (source !== '') {
+			console.log('%c >> CLICK <<', 'font-size: 13px; color: yellow;', source, click);
+		}
+	}
+};
--- a/assets/src/application/ui/input/BaseInput/style.scss
+++ b/assets/src/application/ui/input/BaseInput/style.scss
@@ -6,4 +6,71 @@
 		margin: var(--ee-margin-micro) var(--ee-margin-micro) 0;
 		white-space: nowrap;
 	}
+
+	.ant-input,
+	.ant-select .ant-select-selector,
+	.ant-select:not(.ant-select-disabled) .ant-select-selector,
+	.ant-select-single:not(.ant-select-customize-input) .ant-select-selector {
+		background-color: var(--ee-background-color);
+		background-clip: padding-box;
+		border: 2px solid var(--ee-color-grey-10);
+		border-radius: var(--ee-border-radius-small);
+		box-shadow: none;
+		box-sizing: border-box !important;
+		color: var(--ee-default-text-color);
+		display: inline-block;
+		font-size: var(--ee-font-size-default);
+		line-height: var(--ee-font-size-bigger);
+		height: var(--ee-icon-button-size) !important;
+		transition: border-color 0.15s ease-in-out, box-shadow 0.15s ease-in-out;
+		vertical-align: middle;
+		width: 100%;
+	}
+
+	.ant-input {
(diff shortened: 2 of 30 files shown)


assets/src/application/services/utilities/cancelClickEvent.ts
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+
+const { console } = window;
```
nicolad: `console` will work without this, since window is the global scope

assets/src/application/ui/input/BaseInput/style.scss
```diff
@@ -6,4 +6,71 @@
...
+		box-shadow: none;
+		box-sizing: border-box !important;
+		color: var(--ee-default-text-color);
+		display: inline-block;
+		font-size: var(--ee-font-size-default);
+		line-height: var(--ee-font-size-bigger);
+		height: var(--ee-icon-button-size) !important;
+		transition: border-color 0.15s ease-in-out, box-shadow 0.15s ease-in-out;
+		vertical-align: middle;
+		width: 100%;
```
nicolad: pure gold: the order and the reset

assets/src/application/ui/input/BaseInput/style.scss
```diff
@@ -6,4 +6,71 @@
...
+
+		&:hover,
+		&:focus,
+		&.ant-select-focused {
+			.ant-select-selector,
+			.ant-select-selector:focus,
+			.ant-select-selector:hover,
+			&:hover .ant-select-selector,
+			&:focus .ant-select-selector {
+				border-color: unset;
```
nicolad: quite readable nesting

assets/src/application/ui/input/InlineEditInput/InlineEditHeading.tsx
```diff
@@ -4,8 +4,8 @@ import Editable from './Editable';
 import { HeadingProps } from './types';
 
 const InlineEditHeading: React.FC<HeadingProps> = ({ ellipsis, level = 4, ...rest }) => {
-	const enableEllipsis = ellipsis ? ellipsis : { rows: 1, expandable: false };
-	return <Editable inputType='heading' ellipsis={enableEllipsis} level={level} {...rest} />;
+	const ellipsisOptions = ellipsis ? ellipsis : { rows: 2, expandable: false };
+	return <Editable ellipsis={ellipsisOptions} level={level} {...rest} inputType='heading' />;
```
nicolad: why `inputType='heading'` is at the end and `{...rest}` in the middle?

assets/src/application/ui/layout/entityList/filterBar/Collapsible.tsx
```diff
@@ -1,19 +1,10 @@
...
-type voidFn = () => void;
-
-interface CollapsibleProps {
-	entityFilters: JSX.Element;
-	listId: string;
-	searchText: string;
-	setSearchText: voidFn;
-	showEntityFilters: boolean;
-}
-
```
nicolad: :taco:

assets/src/application/ui/layout/entityList/filterBar/EntityListFilterBar.tsx
```diff
@@ -31,69 +26,16 @@ const EntityListFilterBar: React.FC<EntityListFilterBarProps> = ({ entityFilters
 		view,
 	}: EntityListFilterState = filterState;
 
-	const listViewButton = React.useMemo(
```
nicolad: noicee !!!111!

assets/src/application/ui/layout/entityList/filterBar/types.ts
```diff
@@ -0,0 +1,35 @@
+import { EntityListFilterState } from './useEntityListFilterState';
```
nicolad: yeah, it's better to have these grouped

assets/src/application/ui/input/InlineEditInput/InlineEditHeading.tsx
tn3rb: meant to write a comment to explain that but forgot... :\

the inputType prop has strict values and this order prevents it from being overwritten, not that that would be likely to happen, but basically just means that `InlineEditHeading` will **always** use `inputType='heading'`

assets/src/application/ui/input/BaseInput/style.scss
tn3rb: despite it being a little crazy because of how that component works

PR #2513, opened 2020-03-06
Refactor and centralize enity filter state | Closes #2507

### LINKED ISSUE #2507: Entity list doesn't update when adding or deleting an entity
When adding or deleting an entity, the list doesn't seem to update. The most probable cause is the filterbar paginated entities.

manzoorwanijk:
This PR:
- Refactors entity filter state management
- Improves EntityList and pagination
- Creates shared filter state types for EDTR entities
- Creates EntityList context
- Creates datetime filter state manager
- Updates components to use the new filter state
- Partially closes #2507


assets/src/domain/eventEditor/services/filterState/datetimes/useFilteredDatetimes.ts
```diff
@@ -0,0 +1,38 @@
...
+
+	// Avoid synchronous state update
+	useEffect(() => {
+		if (total !== savedTotal) {
+			setTotal(savedTotal);
+		}
+	}, [total, savedTotal]);
+
+	return paginatedDatetimes;
+};
```
tn3rb: At the top of the `DatesList` component we have:

```js
	const filteredDates = useFilteredDatetimes();
	const filterState = useDatesListFilterState();
```

and then here we are also retrieving the filter state.

why not do it once, either:

- retrieve filter state in `DatesList` and then pass to this hook

ie: 

```js
	// in DatesList
	const filterState = useDatesListFilterState();
	const filteredDates = useFilteredDatetimes( filterState );
```

or

- only retrieve filter state here and then return it AND the filtered entities list to the consuming component

ie: 

```js
	// in DatesList
	const { filteredDates, filterState } = useFilteredDatetimes();
```


That said... 

can we discuss trying to find a way to run ALL of this filtering within the `EntityList` component in the `\application\ui\layout` layer instead of separately within each specific entity list in the domain layer?

ie: define universal interfaces for:

- entity filters ( entities, filter predicate(s), value(s) to filter by ) : used by `@sharedEntities/datetimes/predicates/filters` and other filters

- search filters (although these could likely just run as part of the above filters, but could also be centralized within the app layer)

- entity sorters ( entities, sorting predicate(s), value(s) to sort by ) : used by `@sharedEntities/datetimes/predicates/sorters` and other sorters

then pass the filter and sorter callbacks ( or arrays of callbacks which would allow for multiple filters and/or sorters to be added dynamically )
along with the values (or arrays of values) for the filters and sorters to the Entity List state manager, which would run the filters, followed by the sorters, and THEN run the pagination filters.

Why go to that trouble? 

Because we're definitely going to be doing this kind of thing again and again, and it just seems to me that it would make more sense to have that logic performed once in one central spot than to have to duplicate it all over the place. So basically it would be really great if we could figure out how to simply pass the filter UI along with a set of filter state setters and getters to one central reducer which would process everything.

You will see this is how I did things in EDTR v1 if you look at `assets\ZZZ\data\eventespresso\filter-state\index.js` and how it was implemented in `assets\ZZZ\editor\events\dates-and-times\editor-date\filter-bar\use-dates-list-filter-state-setters.js` and `assets\ZZZ\editor\events\dates-and-times\editor-date\filter-bar\use-dates-list-filter-state.js` but plz note that this was using redux, which of course is not too different than the `useReducer()` hook. The thing I didn't do however was figure out how to abstract that final step and have the actual filtering, searching, sorting, and pagination ALSO handled by a single centralized core system. In other words, I still had to do what you see in `assets\ZZZ\editor\events\dates-and-times\editor-date\filter-bar\use-filtered-dates-list.js` within the dateslist itself, but we were not using TypeScript at the time and so being able to define an interface for an abstracted system to handle that would have been far too difficult.

But you are a better JS dev than I am, and we are also using TypeScript now, so I think you could take this to the next level where I couldn't before.

So let's take the Dates list and break it down...

it currently has a search field, a field for setting the sorting, and 2 unique filters for dates although we want to split one of those into two, so let's just say three filters:

- search filter
- status filter (expired, active, upcoming, trashed, etc)
- sales volume filter (100% sold, 90+% sold, 75+% sold, etc)
- start date end date display modifier (not really a filter)
- sorting field

plus pagination, which  should be identical for any list.

But what if we wanted to add more filters in the future?

- venue filter
- specific month/year/date range filter
- etc

and what if an add-on needed to inject its own filters?

- Recurring Events Manager specific filter
- Wait List add-on related filter

SO ideally, any system we build will need to be able to dynamically add filters as needed.

It initially looks to me like there are a combination of the following things:

- filters - reduce the number of entities in a list based on a predicate and a supplied value 
- display modifiers - don't change the number of entities in a list, but modify how the entities are displayed (start or end date, grid view vs list view, etc)
- sorting modifiers - reorder the list based on a predicate and value to sort by
- pagination - total entities, entities per page, page number, should always work the same for any list

Each filter could consist of the following:

- a setter for updating the current value
- a getter for retrieving the current value
- a predicate for reducing the list based on the current value

We should be able to define a bunch of universal interfaces for managing state for any number of filters, using an ID for the list and a name for each filter, plus whatever else:

```
filter setter {
	list ID,
	filter ID,
	current filter value
}

filter getter {
	list ID,
	filter ID,
}

filter predicate {
	list ID,
	filter ID,
	entity list to filter
}

add filter {
	list ID,
	filter ID,
	filter predicate
}

remove filter (optional) {
	list ID,
	filter ID
}

get filters {
	list ID,
}
```

So anytime the Entity List is rendered, the Filter State manager would retrieve all of the filters that have been set for the given list (based on the List ID), then run all of the filter predicates through the reducer with the list of entities.

Sorting would work nearly identical to the filtering, just replace the word "filter" in the pseudo code above with "sorting" (or equivalent), so that you can set and get values as well as sorting predicates for a specific list, and then run those upon rendering.

Because the search field will likely be pretty universal, all we would really need to know is a set of entity fields to search for and everything else can be the same for any entity (i might be wrong about that though). So that would just need an interface like:

```
search term setter {
	list ID,
	search term,
}

search term getter {
	list ID,
}

add entity search field {
	list ID,
	field name (or fields),
}

remove entity search field (optional) {
	list ID,
	field name,
}

get entity search fields {
	list ID,
}
```

Then anytime the search field receives input for an entity list, the Filter State manager would simply pull the search fields from the state and then run all of the entities through some generic `EntityFieldSearchReducer` with the current search term prior to running any other filters (or maybe after since text search is likely more computationally expensive). 

Display Modifiers would likely only require getters and setters, as the Entity List and Filter State Manager would not have to do anything upon rendering because the specific entity list view components could retrieve those values themselves (ie: for toggling display of start or end date). The exception to this would be the "view" filter which the Entity List component itself would use for deciding whether to render the Card Views or Table View for the entities (although we could have other views in the future for different UIs so maybe that could be rethought and improved too... but not now).

For pagination, the same logic used for setting and getting display modifiers could be used for storing the values needed by the EntityPagination component:

- previous total (to decide whether to reset)
- current page number
- entities per page 

so the same getters and setters could control those values, which would then be used to slice the final filtered, searched, and sorted array of entities.

Finally, my most sincere apologies for bringing this all up after you have already put all of this great work into this. I wasn't really aware that issue #2507 would result in such a massive refactoring, and that got me thinking again about how I originally wanted this to all work as well as what challenges we may face in the future. I'll try harder to be more on the ball so that your efforts are not wasted like this again, not that this was a waste because you may still be able to salvage some of what you have, and it's always good to try solving problems from different angles.

I must say that I am very impressed with your TypeScript skills every time I review one of these PRs where you are tackling something more complex.

If you see this on Monday morning and are unsure about anything and/or want to discuss this further before taking any action, then feel free to just shelve it and work on something else until I get online and you have the opportunity to curse at me! 😆

assets/src/domain/eventEditor/ui/datetimes/datesList/filterBar/controls/DatetimesToShowControl.tsx
```diff
@@ -0,0 +1,85 @@
...
+						value: DatetimesToShow.expiredOnly,
+						label: __('all expired dates'),
+					},
+					{
+						value: DatetimesToShow.trashedOnly,
+						label: __('trashed dates only'),
+					},
+				]}
+				onChange={setDatetimesToShow}
+				size='large'
```
tn3rb: this actually conflicts with the styles the way I have them set up, so you don't need to add this for select fields

assets/src/domain/eventEditor/ui/datetimes/datesList/index.tsx
```diff
@@ -31,4 +30,12 @@ const DatesList: React.FC = () => {
...
-export default DatesList;
+const WrappedDatesList: React.FC = () => {
+	return (
+		<DatetimesListProvider>
+			<DatesList />
+		</DatetimesListProvider>
+	);
+};
+
+export default WrappedDatesList;
```
tn3rb: 👍 

I still think HoCs are useful in some cases

although technically i guess this really isn't an HoC 🤷‍♂

assets/src/domain/eventEditor/ui/tickets/ticketsList/filterBar/controls/TicketsSortedByControl.tsx
```diff
@@ -26,25 +26,25 @@ const TicketsSortedByControl: React.FC<TicketsSortedByControlProps> = React.memo
...
 			<SelectInput
-				label={__('sort')}
+				label={__('sort by')}
 				className='ee-ticket-list-filter-bar-order-select'
 				value={sortTicketsBy}
 				options={[
 					{
 						value: SortTicketsBy.date,
-						label: __('by ticket sale date'),
+						label: __('ticket sale date'),
```
tn3rb: nice change 👍

assets/src/domain/shared/services/filterState/types.ts
```diff
@@ -0,0 +1,28 @@
...
+}
+
+export interface EntityFilterStateManager extends EntityFilterState {
+	setDisplayStartOrEndDate: (displayStartOrEndDate: DisplayStartOrEndDate) => void;
+}
+
+export type EntityFilterStateReducer<EFS extends EntityFilterState, EFA extends EntityFilterAction<any>> = (
+	state: EFS,
+	action: EFA
+) => EFS;
```
tn3rb: these sound like governmental organizations!

assets/src/application/ui/layout/entityList/EntityList.tsx
```diff
@@ -3,9 +3,8 @@ import { __ } from '@wordpress/i18n';
 import { Divider, Typography } from 'antd';
 
 import { EmptyState, ErrorIndicator, LoadingIndicator } from '@appDisplay/index';
+import { EntityListFilterState } from './filterBar';
 import EntityListFilterBar from '@appLayout/entityList/filterBar/EntityListFilterBar';
```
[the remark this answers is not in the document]
manzoorwanijk: Agreed

assets/src/application/ui/layout/entityList/filterBar/types.ts
```diff
@@ -0,0 +1,40 @@
...
+
+export interface EntityListFilterStateManager<SortBy = BasicSortBy> extends EntityListFilterState<SortBy> {
+	setSortBy: (sortBy: SortBy) => void;
+	setPerPage: (newPageNumber: number, newPerPage: number) => void;
+	setPageNumber: (page: number) => void;
+	setTotal: (total: number) => void;
+	setSearchText: (text: string) => void;
+	setGridView: VoidFunction;
+	setListView: VoidFunction;
+	toggleEntityFilters: VoidFunction;
```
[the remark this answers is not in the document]
manzoorwanijk: They exist in TS 😄

assets/src/application/ui/layout/entityList/filterBar/useEntityListFilterStateManager.ts
```diff
@@ -0,0 +1,134 @@
...
+	BasicSortBy,
+	EntityListFilterState,
+	EntityListFilterStateManager,
+	EntityListFilterStateReducer,
+} from './types';
+
+type ELFSM<SortBy = BasicSortBy> = EntityListFilterStateManager<SortBy>;
+
+const useEntityListFilterStateManager = <SortBy = BasicSortBy>(defaultSortBy: SortBy): ELFSM<SortBy> => {
+	type FSM = ELFSM<SortBy>;
```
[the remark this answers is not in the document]
manzoorwanijk: Yeah, you can

assets/src/application/ui/layout/entityList/filterBar/useEntityListFilterStateManager.ts
```diff
@@ -0,0 +1,134 @@
...
+				return { ...state, sortBy };
+
+			case 'SET_VIEW':
+				return { ...state, view };
+
+			case 'TOGGLE_ENTITY_FILTERS':
+				return { ...state, showEntityFilters: !state.showEntityFilters };
+
+			default:
+				throw new Error('Unexpected action');
```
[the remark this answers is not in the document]
manzoorwanijk: `__DEV__` takes care of that, which already enabled during webpack `development` mode.

assets/src/domain/eventEditor/services/context/EntityListContext/index.tsx
```diff
@@ -0,0 +1,17 @@
...
+export const DatetimesListProvider: React.FC<ContextProviderProps> = ({ children }) => {
+	const filters = useDatesListFilterStateManager();
+	return <DatetimesListContext.Provider value={{ filters }}>{children}</DatetimesListContext.Provider>;
+};
+
+/* export const TicketListContext = createContext<EntityListContextProps<TicketsFilterStateManager>>(null);
+export const TicketListProvider: React.FC<ContextProviderProps> = ({ children }) => {
+	const filters = useDatesListFilterState();
+	return <TicketListContext.Provider value={{ filters }}>{children}</TicketListContext.Provider>;
+}; */
```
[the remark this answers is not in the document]
manzoorwanijk: Agreed

assets/src/domain/eventEditor/services/filterState/datetimes/useFilteredDatetimes.ts
manzoorwanijk: I didn't pass `filterState` to `useFilteredDatetimes()` because we may need filtered dates at some other places like TAM.

I am not with this
```js
const { filteredDates, filterState } = useFilteredDatetimes();
```
because filtered dates should not return `filterState` IMHO - Single Responsibility Principle

I agree with rest of the suggestions to let `EntityList` to the filtering and sorting, but that can be done in another PR to make it fully extensible and abstracted.

assets/src/domain/eventEditor/ui/datetimes/datesList/index.tsx
manzoorwanijk: I don't think bad about HOCs. I used similar approach in TAM form content context

image

assets/src/application/ui/layout/entityList/EntityList.tsx
```diff
@@ -3,9 +3,8 @@ import { __ } from '@wordpress/i18n';
 import { Divider, Typography } from 'antd';
 
 import { EmptyState, ErrorIndicator, LoadingIndicator } from '@appDisplay/index';
-import EntityListFilterBar from '@appLayout/entityList/filterBar/EntityListFilterBar';
-import useEntityListFilterState from '@appLayout/entityList/filterBar/useEntityListFilterState';
-import { useEntityPagination } from '@appLayout/entityList/pagination';
+import { EntityListFilterStateManager } from './filterBar';
+import EntityListFilterBar from './withValidFilterState';
```
nicolad: should this reflect the actual file name?
`import withValidFilterState from './withValidFilterState';`

manzoorwanijk: May be, but it has to be PascalCase to be able to be used as JSX element.

assets/src/application/ui/layout/entityList/filterBar/filterState/useEntityListFilterStateManager.ts
```diff
@@ -0,0 +1,101 @@
...
+
+import {
+	BasicSortBy,
+	EntityListFilterState,
+	EntityListFilterStateManager,
+	EntityListFilterStateReducer,
+} from './types';
+import getReducer from './reducer';
+
+type ELFSM<SortBy = BasicSortBy> = EntityListFilterStateManager<SortBy>;
```
nicolad: you lost me here :)

nicolad: small comment might be helpful here

assets/src/application/ui/layout/entityList/filterBar/filterState/useEntityListFilterStateManager.ts
```diff
@@ -0,0 +1,101 @@
...
+
+const useEntityListFilterStateManager = <SortBy = BasicSortBy>(defaultSortBy: SortBy): ELFSM<SortBy> => {
+	type FSM = ELFSM<SortBy>;
```
nicolad: we might need a dictionary soon :smiley:

assets/src/application/ui/layout/entityList/withValidFilterState.tsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+
+import { EntityListFilterBar, EntityListFilterBarProps } from './filterBar';
+
+const withValidFilterState = (FilterBarComponent: React.ReactType): React.FC<EntityListFilterBarProps> => ({
+	filterState,
+	...props
+}) => {
+	return filterState ? <FilterBarComponent filterState={filterState} {...props} /> : null;
```
nicolad: maybe filterstate might have a safe default values and we might not need this check?

assets/src/domain/eventEditor/services/context/EntityListContext/index.tsx
```diff
@@ -0,0 +1,11 @@
+import React, { createContext } from 'react';
+
+import { EntityListContextProps, ContextProviderProps } from '../types';
+import { DatetimesFilterStateManager, useDatesListFilterStateManager } from '../../filterState';
+
+export type DatetimesListContextProps = EntityListContextProps<DatetimesFilterStateManager>;
+export const DatetimesListContext = createContext<DatetimesListContextProps>(null);
+export const DatetimesListProvider: React.FC<ContextProviderProps> = ({ children }) => {
+	const filters = useDatesListFilterStateManager();
+	return <DatetimesListContext.Provider value={{ filters }}>{children}</DatetimesListContext.Provider>;
```
nicolad: `yup`, no more prop drilling here :taco:

assets/src/application/ui/layout/entityList/withValidFilterState.tsx
manzoorwanijk: Better be safe

assets/src/domain/eventEditor/services/filterState/datetimes/types.ts
```diff
@@ -0,0 +1,44 @@
...
+}
+
+export type DatetimesFilterActionType = 'SET_DATETIMES_TO_SHOW' | EntityFilterActionType;
+
+export interface DatetimesFilterAction
+	extends Partial<DatetimesFilterState>,
+		EntityFilterAction<DatetimesFilterActionType> {}
+
+export interface DatetimesFilterStateManager
+	extends EntityListFilterStateManager<SortBy>,
```
nicolad: not sure what this `<SortBy>` does?

assets/src/domain/eventEditor/services/filterState/datetimes/useFilteredDatetimes.ts
```diff
@@ -0,0 +1,39 @@
...
+	const paginatedDatetimes = useMemo<Array<Datetime>>(() => {
+		return sortedDatetimes.slice(perPage * (pageNumber - 1), perPage * pageNumber);
+	}, [sortedDatetimes, perPage, pageNumber]);
+
+	// Avoid synchronous state update
+	useEffect(() => {
+		if (total !== filteredDatetimes.length) {
+			setTotal(filteredDatetimes.length);
+		}
+	}, [total, filteredDatetimes]);
```
nicolad: interesting

assets/src/domain/eventEditor/services/filterState/datetimes/types.ts
manzoorwanijk: It sets the proper union types that you pass to `sortBy()`. Application `SortBy` is only `'name' | 'id'`. If a domain wants more `SortBy` options, it should be passed here. I could leave it as string but you know that makes it prone to mistakes.

PR #2515, opened 2020-03-09
IconMenuItem Refactor; closes #2481.

### LINKED ISSUE #2481: IconMenuItem Changes
- see `\application\ui\display\IconMenuItem`

- originally a wrapper for Icon buttons added to an `EntityActionMenu` that would also show a count for related entities, but will likely have other uses.

- rename component/folder to `ItemCount` and update any references accordingly

- remove div wrapper and **remove clock !?!?!** (dunno why that was even there)

- add `showZero` and `offset={[-5, 5]}` props to Badge

- apply following classes to Badge component:

     - incoming className prop if one exists
     - generic class like `ee-item-count`
     - `ee-item-count--has-items` if incoming `count` prop is greater than zero
     - `ee-item-count--no-items` if incoming `count` prop is zero

- edit scss file and:

     - remove non-applicable styles
     - update class names to match above (following BEM)
     - remove pretty much everything except background color since AntD takes care of the rest
     - use background color vars currently in use in `ee-has-items` and `ee-no-items` classes

- import into `AssignDatesButton` and `AssignTicketsButton` and wrap those buttons with new ItemCount component

- figure out how to get count of related entities from Apollo cache and use that for the value of the `count` prop (talk to Manzoor about how to do this, unless you happen to be Manzoor, in which case simply mutter to yourself)

nicolad:
Closes: https://github.com/eventespresso/event-espresso-core/issues/2481


nicolad:
here badge is crossing the card edge: 

image
***
one solution might be to display badge on the left side, but not sure if that is the best idea:
image
***

another option would be to remove paddings from the middle block:
image

assets/src/application/ui/display/ItemCount/index.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+import './style.scss';
+
+interface ItemCountProps extends BadgeProps {
+	children?: React.ReactNode;
+	id?: string;
+	showClock?: boolean;
+}
+
+const ItemCount = ({ children, count, ...props }: ItemCountProps) => {
```
manzoorwanijk:
```suggestion
const ItemCount: React.FC<ItemCountProps> = ({ children, count, ...props }) => {
```

assets/src/application/ui/display/ItemCount/index.tsx
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+import { Badge } from 'antd';
+import { BadgeProps } from 'antd/lib/badge';
+import classNames from 'classnames';
+
+import './style.scss';
+
+interface ItemCountProps extends BadgeProps {
+	children?: React.ReactNode;
```
manzoorwanijk: You don't need to add `children` here when the type is `React.FC`

assets/src/application/ui/display/ItemCount/index.tsx
nicolad: thx :bowing_man:

assets/src/application/ui/display/ItemCount/index.tsx
nicolad: updated, thx

PR #2528, opened 2020-03-10
Refactor Loading Indicator and Migrate Loading Notice; closes #2360.

### LINKED ISSUE #2360: Refactor Loading Indicator and Migrate Loading Notice
I'm thinking that we will require more than one type of component for providing feedback on loading states:

- a "Loading Notice" for showing that a large collection, such as an entire list of entities or new admin screen, is in the process of being loaded. This would be used as placeholder for a list or screen or form when it is loading.

- a smaller "Loading Indicator" for providing user feedback when an action such as saving or deleting is in progress. This could be embedded in other components such as buttons.

### Loading Notice

- see `assets\ZZZ\components\ui\loading-notice`

- move to `assets\src\application\ui\display\loadingNotice`

- can use `<Spin size="large" />` from https://ant.design/components/spin/

- should be rendered inside a large container with an accompanying notice

- can also be wrapped around a container (ie: container is child) which will fade container

### Loading Indicator

- move to `assets\src\application\ui\display\loadingIndicator\`

- delete `assets\src\application\ui\display\LoadingIndicator.js`

- can use https://ant.design/components/spin/ with loading icon:

```js
const antIcon = <Icon type="loading-3-quarters" spin />;
ReactDOM.render(<Spin indicator={antIcon} />, mountNode);
```

nicolad:
- Closes: https://github.com/eventespresso/event-espresso-core/issues/2360


assets/src/application/ui/display/loadingNotice/types.ts
```diff
@@ -0,0 +1,8 @@
+import { SpinProps, SpinSize } from 'antd/lib/spin';
+
+export interface LoadingNoticeProps {
+	className?: string;
+	loading?: boolean | SpinProps;
```
nicolad: the same approach as here: https://github.com/ant-design/ant-design/blob/master/components/list/index.tsx#L44

assets/src/application/ui/display/loadingNotice/LoadingNotice.tsx
```diff
@@ -0,0 +1,28 @@
...
+import classNames from 'classnames';
+import { Spin } from 'antd';
+import { _x, sprintf } from '@wordpress/i18n';
+
+import { LoadingNoticeProps } from './types';
+
+const LoadingNotice: React.FC<LoadingNoticeProps> = ({ children, loading, size, ...props }) => {
+	const className = classNames(props.className, 'ee-loading-notice', `ee-loading-notice-${size}`);
+	const tip = props.tip || sprintf(_x('loading%s', 'loading...', 'event_espresso'));
+	let loadingProp = loading;
```
nicolad: the same as here: https://github.com/ant-design/ant-design/blob/master/components/list/index.tsx#L184

assets/src/application/ui/display/loadingNotice/LoadingNotice.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+	if (typeof loadingProp === 'boolean') {
+		loadingProp = {
+			spinning: loadingProp,
+		};
+	}
+
+	return (
+		<div className={className}>
+			<Spin size={size} tip={tip} {...loadingProp}>
```
nicolad: wrapped as here: https://github.com/ant-design/ant-design/blob/master/components/list/index.tsx#L278

assets/src/application/ui/display/loadingIndicator/LoadingIndicator.tsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+import { LoadingOutlined } from '@ant-design/icons';
+import { Spin } from 'antd';
+import { SpinProps } from 'antd/lib/spin';
+
+const LoadingIndicator: React.FC<SpinProps> = (props) => {
+	const antIcon = <LoadingOutlined style={{ fontSize: 24 }} spin />;
+
+	return <Spin indicator={antIcon} {...props} />;
```
nicolad: result: https://codesandbox.io/s/8oy2s

nicolad:
The output of `LoadingNotice`:
Peek 2020-03-10 11-32

assets/src/application/ui/display/loadingIndicator/LoadingIndicator.tsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+import { LoadingOutlined } from '@ant-design/icons';
+import { Spin } from 'antd';
+import { SpinProps } from 'antd/lib/spin';
+
+const LoadingIndicator: React.FC<SpinProps> = (props) => {
+	const antIcon = <LoadingOutlined style={{ fontSize: 24 }} spin />;
```
manzoorwanijk: We should avoid inline styling or at least avoid using object literals in JSX attributes. It creates unnecessary rerenders.

nicolad: it was based on antd example:
image

nicolad: but I do agree with your comment, thx, suggestion applied

PR #2529, opened 2020-03-10
Move related entities to badge | Closes #2526

### LINKED ISSUE #2526: Move Related Tickets and Related Dates Lists into Ticket Assignment Manager Button Badge Tooltips
- on the DateCard before the `EntityDetailsPanel` the DBIDs for the related tickets are retrieved and displayed as a series of tags

- similarly on the TicketCard, the DBIDs for the related dates are displayed as a series of tags

- open the respective Ticket Assignments Buttons for each of the above components. ie: the `AssignTicketsButton` for the `DateActionsMenu`, and the `AssignDatesButton` for the `TicketActionsMenu`

- you will see similar logic that retrieves the related entities for each of the above

- generate a string of IDs similar to what is done in the Entity Card components above, then combine that with the additional text from those components, and pass that as the "title" prop to the `ItemCount` component

- when finished, hovering over the relations count badge for the assignments buttons should display a tooltip like: "Related Tickets: 1 2 3" or "Related Dates: 4 5 6"

- make any adjustments to the `ItemCount` component as necessary

- make sure that the `ItemCount` title prop does NOT display numbers if that is all it is passed. This can be done by setting the title to `' '` 

- now go back to DateCard and TicketCard and remove all of that relations logic

manzoorwanijk:
This PR:
- Improves types and extracts reducer for relations manager
- Updates `<ItemCount />` to add support for `title`
- Improves and simplifies query hooks
- Adds and improves query hooks for relations. Newly created hooks include `useRelatedDatetimes()`, `useRelatedTickets` and `useRelatedPrices`.
- Moves related Dates and Tickets to related badge tooltip
- Removes related entities from cards
- Closes #2526

Here is the demo of how it looks
image

### DIFF
--- a/assets/src/application/services/apollo/relations/reducer.ts
+++ b/assets/src/application/services/apollo/relations/reducer.ts
@@ -1,0 +1,78 @@
+import { pathOr, assocPath, dissocPath, clone } from 'ramda';
+import { RelationsReducer, RelationalData } from './types';
+
+const reducer: RelationsReducer = (state, action) => {
+	const { entity, entityId, relation, relationId, relationIds } = action;
+	let newState: RelationalData, relations: string[];
+	switch (action.type) {
+		case 'INITIALIZE':
+			return action.data;
+
+		case 'ADD_RELATION':
+			relations = pathOr([], [entity, entityId, relation], state);
+			// If the relation already exists
+			if (relations.includes(relationId)) {
+				return state;
+			}
+			newState = assocPath([entity, entityId, relation], [...relations, relationId], state);
+			return newState;
+
+		case 'REMOVE_RELATION':
+			newState = clone(state);
+			// existing relation list.
+			relations = pathOr([], [entity, entityId, relation], newState);
+			// if relationId is given remove it from the list.
+			if (relationId) {
+				return assocPath(
+					[entity, entityId, relation],
+					relations.filter((id) => id !== relationId),
+					newState
+				);
+			}
+
+			if (relations.length) {
+				/**
+				 * If we are here, it means that we have values for `entityId` in `relation`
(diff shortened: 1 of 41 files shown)


assets/src/application/services/apollo/relations/reducer.ts
```diff
@@ -0,0 +1,78 @@
...
+			if (relations.includes(relationId)) {
+				return state;
+			}
+			newState = assocPath([entity, entityId, relation], [...relations, relationId], state);
+			return newState;
+
+		case 'REMOVE_RELATION':
+			newState = clone(state);
+			// existing relation list.
+			relations = pathOr([], [entity, entityId, relation], newState);
```
nicolad: good candidate for `nullish coalescing operator` :D :D

assets/src/application/ui/display/ItemCount/index.tsx
```diff
@@ -10,15 +10,18 @@ interface ItemCountProps extends BadgeProps {
...
 		'ee-item-count--no-items': count === 0,
 	});
 	const offset = props.offset || [-5, 5];
 
+	const countBadge = <span className='ant-badge-count'>{count}</span>;
+	const countNode = title ? <Tooltip title={title}>{countBadge}</Tooltip> : countBadge;
+
 	return (
-		<Badge {...props} className={className} count={count} offset={offset} showZero>
+		<Badge {...props} className={className} count={countNode} offset={offset} showZero>
```
nicolad: :+1:

assets/src/application/services/apollo/relations/reducer.ts
manzoorwanijk: That will be like this
```suggestion
			relations = newState?.[entity]?.[entityId]?.[relation] ?? [];
```

PR #2532, opened 2020-03-10
Create Inline Infinity Edit Fields for Capacity and Quantity | Closes #2525

### LINKED ISSUE #2525: Use InlineEditInput for DateDetailsPanel Capacity & TicketDetailsPanel Quantity Fields
- previously `assets\ZZZ\components\ui\entity-details-panel\entity-details.js` would pass entity details marked "editable" to the `InlineEditInput` component, but now that has to be done manually, which is better because it allows any JSX element to be used as a value.

- in `\domain\eventEditor\ui\datetimes\datesList\cardView\DateDetails.tsx`, the capacity field needs to be editable

- likewise in `\domain\eventEditor\ui\tickets\ticketsList\cardView\TicketDetails.tsx`, the quantity field needs to be editable

- create new components for both of the required inputs then import into the above files for passing to `EntityDetailsPanel` (similar to how `DateRegistrationsLink` and `TicketRegistrationsLink` are handled)

- in the `onChange` handler for the `DateDetails` capacity input, after updating the datetime capacity field, we need to also do the following:

     - retrieve all related tickets

     - loop thru the tickets and check for ticket quantities greater than then new datetime capacity

     - set those ticket quantities to be the same as the new datetime capacity

     - ticket quantities less than then new datetime capacity should NOT be changed

manzoorwanijk:
This PR:
- Improves `parseInfinity`
- Creates `InlineEditInfinity` component in `/application/ui/input/InlineEditInput`
- Creates `useUpdateRelatedTickets` and `useTicketQuantityForCapacity` hooks
- Creates `DateCapacity` component to use in `DateDetails` panel
- Creates `TicketQuantity` component to use in `TicketDetails` panel
- Closes #2525


assets/src/application/ui/input/InlineEditInput/InlineEditInfinity.tsx
```diff
@@ -0,0 +1,47 @@
...
+			break;
+		default:
+			output = `${value}`;
+			break;
+	}
+
+	const htmlClasseName = classNames('ee-infinity-sign', className);
+
+	return (
+		<Editable {...rest} editable={editable} inputType='text' className={htmlClasseName}>
```
nicolad: :taco: for placing spread operator in front :+1:

PR #2539, opened 2020-03-11
Improve and wire up entity search | Closes #2538

### LINKED ISSUE #2538: Filter Bar Search Input No Searchie
type stuff in filter bar search input but it no searchie!!!

why it no searchie?

manzoorwanijk:
This PR:
- Converts `entityFieldSearch` to `entityListSearch` and improves it
- Adds strict type checking for `entityListSearch` and multi field support locked down by types
- Wires up the entity search for Dates and Tickets
- Closes #2538


assets/src/application/services/utilities/text/entityListSearch/index.test.ts
```diff
@@ -0,0 +1,78 @@
...
+	},
+	{
+		id: 'RGF0ZXRpbWU6ODU=',
+		dbId: 86,
+		name: 'another title',
+		description: null,
+	},
+];
+
+const searhArgs: EntityListSearchProps<TestEntity> = {
```
nicolad: `searchArgs`

assets/src/application/services/utilities/text/entityListSearch/index.test.ts
```diff
@@ -0,0 +1,78 @@
...
+	entities,
+	searchFields: ['name', 'description'],
+	searchText: '',
+};
+
+describe('entityListSearch', () => {
+	it('should return two results by name and description', () => {
```
nicolad: `description` is being mentioned, but is not checked in the assertion

manzoorwanijk: It it used because description contains that keyword

assets/src/domain/eventEditor/services/filterState/datetimes/useFilteredDatetimes.ts
```diff
@@ -9,17 +10,39 @@ import sorters from '@sharedEntities/datetimes/predicates/sorters';
...
 	// Flter the list
 	const filteredEntities = useMemo<Array<Datetime>>(() => {
 		return filters({ dates, show: datetimesToShow });
 	}, [dates, datetimesToShow]);
 
+	// search entities
+	const searchResults = useMemo<Array<Datetime>>(() => {
+		if (!searchText) {
+			return filteredEntities;
+		}
```
nicolad: maybe we can move this conditional inside of `entityListSearch`, so that we don't duplicate this logic outside, because the same conditional is applied in `useFilteredTickets.ts `

manzoorwanijk: That's already there, it's just to bail early to avoid running the search filter. This will be improved when Filter Bar is extracted.

nicolad: yes, just saw that conditional inside :+1:

PR #2546, opened 2020-03-12
Added main menu item to date action menu.

nicolad:
- Added main menu item to dates action menu

## How has this been tested
image
***
image


nicolad:
I am not sure about copy date action, should this be in v2, @tn3rb ?

nicolad:
Also noticed that current style is different from the old one:
image
***
since it's reusing some of the GB's styles:
image
***
open to any suggestions

assets/src/domain/eventEditor/ui/datetimes/datesList/actionsMenu/DateEntityMainMenuItem.tsx
```diff
@@ -0,0 +1,50 @@
...
+					title: __('edit date'),
+					icon: Icon.EDIT,
+					onClick: onEditClick,
+					// onClick: useOpenEditor(useEventDateEditorId(dateEntity)),
+				},
+				// {
+				// 	title: __('copy date'),
+				// 	icon: 'admin-page',
+				// 	onClick: useCopyDateEntity(dateEntity),
+				// },
```
nicolad: this is commented out, because there is a pending feedback asked from @tn3rb

assets/src/domain/eventEditor/ui/datetimes/datesList/actionsMenu/DateEntityMainMenuItem.tsx
```diff
@@ -0,0 +1,50 @@
...
+				{
+					title: __('trash date'),
+					icon: 'trash',
+					onClick: () => deleteEntity({ id }),
+				},
+			]}
```
tn3rb: i'm wondering if it might be better to change this and instead just pass a JSX element for each menu item instead of some configuration details

pros and cons to both approaches...
___
### Defining Menu Items in a Config Object (like above)
#### pros
- relatively simple to add new menu items
- all menu items should render the same because they would all use the same logic
- tabbing and a11y features can likely be better achieved for same reason as above
- tooltips would apply to the entire menu item in a consistent and reliable manner
- guaranteed safety and consistency with respects to what can be added to menu since items are defined solely by config object
#### cons
- not reusable in any way and potentially less DRY if a similar button is required elsewhere to perform the same thing (like delete a date) 
- violation of SRP - logic is not self contained, meaning the menu component also contains logic for handling events (like `deleteEntity()`)
- less versatile. the same thing that make this approach more consistent also makes it less able to handle a completely different use case
___
### Supplying Menu Items as JSX Elements
#### pros
- also relatively simple to add new menu items since any renderable node could be used
- far more reusable and DRY because the JSX elements passed to the menu could potentially be reused anywhere
- better adherence to SRP since the `DateEntityMainMenu` would simply aggregate other items for display
- far more versatile because nearly anything could be added to a menu (technically there would be no barriers)
- menu items could potentially still be controlled via TS and or using a wrapper to maintain consistency
#### cons
- menu items may not render the same because they could be anything
- tabbing and a11y features may be more difficult to achieve because of lack of control over menu items
- tooltips could also be troublesome if responsibility is left solely to the incoming component, and/or duplicated by the dropdown 

will need to think about this because I keep flip flopping back n forth thinking about which is better

what do you think?

assets/src/application/ui/display/espressoIcon/svgPaths.ts
```diff
@@ -79,6 +79,7 @@ export const svgPath: SvgPath = {
 		'0v2H6V7h11zM4 11c.55 0 1 .45 1 1s-.45 1-1 1-1-.45-1-1 .45-1 1-1zm13 ' +
 		'0v2H6v-2h11zM4 15c.55 0 1 .45 1 1s-.45 1-1 1-1-.45-1-1 .45-1 1-1zm13 ' +
 		'0v2H6v-2h11z',
+	more: 'M17 7V5H3v2h14zm0 4V9H3v2h14zm0 4v-2H3v2h14z',
```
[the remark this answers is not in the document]
nicolad: yeah, baby-SVG :D

assets/src/domain/eventEditor/ui/datetimes/datesList/actionsMenu/DateEntityMainMenuItem.tsx
nicolad: Yes, agree that this can be made more reusable.
Another approach we can take inspiration from is the antd select: https://ant.design/components/select/

manzoorwanijk: Yes, it is better like this:
```tsx
<Menu>
    <MenuItem>Label 1</MenuItem>
    <MenuItem>Label 2</MenuItem>
    <MenuItem>Label 3</MenuItem>
</Menu>
```
The API could expose both `Menu` and `MenuItem`

assets/src/application/ui/layout/entityActionsMenu/mainMenu/Copy.tsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+import { __ } from '@wordpress/i18n';
+
+import { EspressoDropdownMenuItem } from '@application/ui/layout/espressoDropdownMenu';
+import { Icon } from '@application/ui/input';
+import { MenuItemProps } from './types';
+
+const Copy: React.FC<MenuItemProps> = ({ entityType, onClick }) => {
```
nicolad: Naming and location might/will be improved.

assets/src/application/ui/layout/entityActionsMenu/mainMenu/Edit.tsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+import { __ } from '@wordpress/i18n';
+
+import { EspressoDropdownMenuItem } from '@application/ui/layout/espressoDropdownMenu';
+import { Icon } from '@application/ui/input';
+import { MenuItemProps } from './types';
+
+const Edit: React.FC<MenuItemProps> = ({ entityType, onClick }) => {
+	return <EspressoDropdownMenuItem icon={Icon.EDIT} onClick={onClick} title={__('edit ' + entityType)} />;
```
nicolad: oops, edited to `trash`

nicolad:
Current result compared to old version:

image

assets/src/application/ui/layout/entityActionsMenu/entityMenuItems/Trash.tsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import { __, sprintf } from '@wordpress/i18n';
+
+import { EspressoDropdownMenuItem } from '@application/ui/layout/espressoDropdownMenu';
+import { Icon } from '@application/ui/input';
+import { MenuItemProps } from './types';
+
+const Trash: React.FC<MenuItemProps> = ({ entityType, onClick, ...props }) => {
+	const title = sprintf(__('trash %s'), entityType);
```
manzoorwanijk: `entityType` is [not being translated](https://github.com/eventespresso/event-espresso-core/pull/2546/files#diff-bd9d38bb96c35744dcd2864ab0732210R31) from the source. It's better to explicitly get the title from consumer and set default to `__('copy')`. The consumer can then pass `__('copy datetime')`

assets/src/application/ui/layout/entityActionsMenu/entityMenuItems/types.ts
```diff
@@ -0,0 +1,4 @@
+export type MenuItemProps = {
+	entityType: 'datetime' | 'ticket';
```
manzoorwanijk: This will be untranslatable. Better not to use it for output strings.

assets/src/application/ui/layout/espressoDropdownMenu/MenuItem.tsx
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
+
+import { EspressoButton } from '@application/ui/input';
+import { EspressoDropdownMenuItemProps as Props } from './types';
+
+const EspressoDropdownMenuItem: React.FC<Props> = ({ icon, onClick, title, ...props }) => {
+	// console.log('EspressoDropdownMenuItem props', props);
```
manzoorwanijk: oops?

assets/src/application/ui/layout/espressoDropdownMenu/types.ts
```diff
@@ -0,0 +1,17 @@
+import { AnyObject } from '@appServices/utilities/types';
+
+export interface EspressoDropdownMenuProps extends AnyObject {
+	children: React.ReactNode;
+	className?: string;
+	icon?: string | null;
```
manzoorwanijk: no need to use `null` as union type when we use TS `strict: false`

assets/src/domain/eventEditor/ui/datetimes/datesList/actionsMenu/dateEntityMainMenuItem/CopyDate.tsx
```diff
@@ -0,0 +1,34 @@
...
+			'isUpcoming',
+			'length',
+			'status',
+			'__typename',
+		],
+		datetime
+	);
+	const onClick = () => createEntity(newDatetime);
+
+	return <Copy {...props} entityType='datetime' onClick={onClick} />;
```
manzoorwanijk: Better to pass the title here `__('copy date')`

assets/src/domain/eventEditor/ui/datetimes/datesList/actionsMenu/dateEntityMainMenuItem/EditDate.tsx
```diff
@@ -0,0 +1,23 @@
...
+	const datetime = useDatetimeItem({ id });
+	const { openEditor } = useFormModal();
+	const onClick = () =>
+		openEditor({
+			editorId: 'editDatetime',
+			entityId: id,
+			entityDbId: datetime.dbId,
+		});
+
+	return <Edit {...props} entityType='datetime' onClick={onClick} />;
```
manzoorwanijk: Same here about title

assets/src/application/ui/layout/entityActionsMenu/entityMenuItems/Trash.tsx
nicolad: updated, thx:
https://github.com/eventespresso/event-espresso-core/pull/2546/commits/62b59b24f731f99dcf120e00d95aeea6ce616af0

assets/src/application/ui/layout/entityActionsMenu/entityMenuItems/types.ts
nicolad: remove this: https://github.com/eventespresso/event-espresso-core/pull/2546/commits/62b59b24f731f99dcf120e00d95aeea6ce616af0

assets/src/application/ui/layout/espressoDropdownMenu/MenuItem.tsx
nicolad: removed, thx

assets/src/application/ui/layout/espressoDropdownMenu/types.ts
nicolad: I've added that from here: https://github.com/WordPress/gutenberg/tree/master/packages/components/src/dropdown-menu#icon
***
but removed now, and left only:
`icon?: string;`

assets/src/domain/eventEditor/ui/datetimes/datesList/actionsMenu/dateEntityMainMenuItem/CopyDate.tsx
nicolad: done, thx

assets/src/domain/eventEditor/ui/datetimes/datesList/actionsMenu/dateEntityMainMenuItem/EditDate.tsx
nicolad: updated, thx

assets/src/application/ui/layout/entityActionsMenu/entityMenuItems/Copy.tsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import { __ } from '@wordpress/i18n';
+
+import { EspressoDropdownMenuItem } from '@application/ui/layout/espressoDropdownMenu';
+import { Icon } from '@application/ui/input';
+import { MenuItemProps } from './types';
+
+const Copy: React.FC<MenuItemProps> = ({ onClick, ...props }) => {
+	const title = props.title || __('copy');
```
manzoorwanijk: 👍

assets/src/domain/eventEditor/ui/datetimes/datesList/actionsMenu/dateEntityMainMenuItem/EditDate.tsx
```diff
@@ -0,0 +1,24 @@
...
+	const { id } = useDatetimeContext();
+	const datetime = useDatetimeItem({ id });
+	const { openEditor } = useFormModal();
+	const onClick = () =>
+		openEditor({
+			editorId: 'editDatetime',
+			entityId: id,
+			entityDbId: datetime.dbId,
+		});
+	const title = __('edit datetime');
```
manzoorwanijk: nice

assets/src/domain/eventEditor/ui/datetimes/datesList/actionsMenu/dateEntityMainMenuItem/CopyDate.tsx
```diff
@@ -0,0 +1,35 @@
...
+			'isExpired',
+			'isSoldOut',
+			'isTrashed',
+			'isUpcoming',
+			'length',
+			'status',
+			'__typename',
+		],
+		datetime
+	);
```
manzoorwanijk: IMHO, it's better and safe o use `pick` instead of `omit`, just in case any other field is added in future.

assets/src/domain/eventEditor/ui/datetimes/datesList/actionsMenu/dateEntityMainMenuItem/CopyDate.tsx
```diff
@@ -0,0 +1,24 @@
...
+import { useDatetimeMutator } from '@edtrServices/apollo/mutations';
+
+const CopyDate = (props) => {
+	const { id } = useDatetimeContext();
+	const datetime = useDatetimeItem({ id });
+	const { createEntity } = useDatetimeMutator();
+	const newDatetime = pick(
+		['capacity', 'description', 'endDate', 'isPrimary', 'name', 'order', 'reserved', 'sold', 'startDate'],
+		datetime
+	);
```
manzoorwanijk: Looks great now 👍

assets/src/domain/eventEditor/ui/datetimes/datesList/actionsMenu/dateEntityMainMenuItem/CopyDate.tsx
tn3rb: that's a double edged sword because you might want to also copy those properties as well...
impossible to know, but yours is likely the safer option because you have to explicitly add fields you want

assets/src/application/ui/layout/espressoDropdownMenu/Menu.tsx
```diff
@@ -0,0 +1,113 @@
...
+			className: 'components-dropdown-menu__popover',
+		},
+		popoverProps
+	);
+
+	return (
+		<Dropdown
+			className={classNames(' components-dropdown-menu', className)}
+			popoverProps={mergedPopoverProps}
+			renderToggle={({ isOpen, onToggle }) => {
```
tn3rb: the callback passed here could likely be extracted to its own component which would better adhere to SRP (same with renderContent)

assets/src/application/ui/layout/espressoDropdownMenu/style.scss
```diff
@@ -20,21 +20,31 @@
...
+		font-weight: inherit;
+		margin: 0 0 5px !important; // important was needed because it is conflicting with EspressoButton styles
 		outline: none;
-		padding: 6px;
+		padding: 0 !important;
 		width: 100% !important; // important was needed because it is conflicting with EspressoButton styles
+
+		&:focus:not(:disabled):not([aria-disabled='true']):not(.is-secondary) {
+			color: #191e23;
+			background-color: var(--ee-color-grey-14);
```
tn3rb:
```suggestion
			color: var(--ee-text-on-grey-14);
			background-color: var(--ee-color-grey-14);
```

You can get a [WCAG](https://www.w3.org/WAI/standards-guidelines/wcag/) compliant text color for any background color var, simply by replacing the word `color` with `text-on`

examples: 

background color | text color
--- | ---
`--ee-` `color` `-grey-14` | `--ee-` `text-on` `-grey-14`
`--ee-` `color` `-primary` | `--ee-` `text-on` `-primary`
`--ee-` `color` `-purple` | `--ee-` `text-on` `-purple`
`--ee-` `color` `-yellow-super-low-contrast` | `--ee-` `text-on` `-yellow-super-low-contrast`

assets/src/application/ui/layout/espressoDropdownMenu/Menu.tsx
nicolad: done

nicolad: for simplicity and since, those 2 components aren't exposed, I've named those:
`Content` and `Toggle`, based on the prop that is accepting those

assets/src/domain/eventEditor/ui/datetimes/datesList/actionsMenu/dateEntityMainMenuItem/CopyDate.tsx
```diff
@@ -0,0 +1,24 @@
...
+	const onClick = () => createEntity(newDatetime);
```
manzoorwanijk: Here you will need to do this:
```ts
import useRelations from '@appServices/apollo/relations/useRelations';
const { getRelations } = useRelations();
const tickets = getRelations({
	entity: 'datetimes',
	entityId: id,
	relation: 'tickets',
});
```

Then 
```ts
const onClick = () => createEntity({...newDatetime, tickets});
```

nicolad: thank you, sir
done: https://github.com/eventespresso/event-espresso-core/pull/2546/commits/ce349879323a4c6d16149d0b6dfcde21eeaf8732

tn3rb: Can we plz move that entire process into the onclick callback so that we don't need to do anything (copy date, retrieve relations, etc) until the copy button actually gets clicked.

assets/src/application/ui/layout/espressoDropdownMenu/style.scss
nicolad: very handy and smart

assets/src/domain/eventEditor/ui/datetimes/datesList/actionsMenu/dateEntityMainMenuItem/CopyDate.tsx
nicolad: :eye: of the :tiger: 

yet another good observation :+1:

nicolad: done: https://github.com/eventespresso/event-espresso-core/pull/2546/commits/8ce917bd7722cb2ad765fc6812ca85f138c492ff

assets/src/application/ui/layout/espressoDropdownMenu/Content.tsx
```diff
@@ -0,0 +1,39 @@
+import React, { Children, cloneElement } from 'react';
+import classNames from 'classnames';
+import { NavigableMenu } from '@wordpress/components';
+
+import { AnyObject } from '@appServices/utilities/types';
+import { mergeProps } from './utils';
+
+interface Props extends AnyObject {
```
nicolad: I kept this locally to avoid polluting the related `types.ts`

assets/src/domain/eventEditor/ui/datetimes/datesList/actionsMenu/dateEntityMainMenuItem/CopyDate.tsx
```diff
@@ -0,0 +1,33 @@
...
+import { __ } from '@wordpress/i18n';
+
+import { Copy } from '@application/ui/layout/entityActionsMenu/entityMenuItems';
+import { useDatetimeContext } from '@edtrHooks/index';
+import { useDatetimeItem } from '@edtrServices/apollo/queries';
+import { useDatetimeMutator } from '@edtrServices/apollo/mutations';
+import { useRelations } from '@appServices/apollo/relations';
+
+const CopyDate = (props) => {
+	const onClick = () => {
```
nicolad: @tn3rb I've got this error, had to revert:

image

tn3rb: oh dang it... right... i always forget about that with hooks


... found a possible solution here: https://blog.logrocket.com/frustrations-with-react-hooks/

scroll down to the "**Return an executable function from the Hook**" heading (no anchors to link to)

nicolad: I've tried the solution proposed there:

image

***
but still got that error:

image

nicolad: What we can do instead is to render an actual component with that hook logic upon `onClick` event, such that component will return null, but still allows hook execution in function body.
Anyway we need a global pattern/ solution not just for this specific action, but for other buttons coupled to hooks as well. Let's pull this issue out of this PR.

tn3rb: ok just add this functionality in with the hooks all being called at the top of the button component
and we'll create an issue for researching solutions and update later.

assets/src/application/ui/layout/espressoDropdownMenu/Menu.tsx
```diff
@@ -0,0 +1,43 @@
...
+			className={classNames(' components-dropdown-menu', className)}
+			popoverProps={mergedPopoverProps}
+			renderToggle={({ isOpen, onToggle }) => (
+				<Toggle icon={icon} isOpen={isOpen} label={label} onToggle={onToggle} toggleProps={toggleProps} />
+			)}
+			renderContent={({ onClose }) => (
+				<Content children={children} label={label} menuProps={menuProps} onClose={onClose} />
+			)}
+		/>
+	);
```
tn3rb: ya this will be much easier to focus on exactly what you want

PR #2559, opened 2020-03-17
Create EntityTable Component | Closes #2553

### LINKED ISSUE #2553: Create EntityTable Component
In #2551 `/application/ui/layout/entityList/EntityCardList.tsx` was added for abstracting and controlling how the card views were handled for the `EntityList`. We should do something similar for the table view portion of the entity list and add a `EntityTable` component in the same folder. This has the following benefits:

- acts as an adapter and abstracts away the direct use of `ResponsiveTable` which makes it easier to change in the future if need be

- standardizes how props are passed to the table component (kinda same point as above)

- allows all entity list table views to be modified from one location if need be (also kinda sorta same point as above)

manzoorwanijk:
This PR:
- Creates `EntityTable` component inside `/application/ui/layout/entityList`
- Updates dates `TableView` and the related components
- Updates tickets `TableView` and the related components
- Closes #2553


assets/src/application/ui/layout/entityList/EntityTable.tsx
```diff
@@ -0,0 +1,35 @@
...
+	bodyRowGenerator,
+	className,
+	entities,
+	filterState,
+	headerRowGenerator,
+	tableCaption,
+	tableId,
+}: EntityTableProps<E, FS>) => {
+	const bodyRows = entities.map((entity) => bodyRowGenerator({ entity, filterState }));
+	const headerRow = headerRowGenerator(filterState);
```
nicolad: nice :+1:

assets/src/application/ui/layout/entityList/types.ts
```diff
@@ -2,27 +2,44 @@ import React from 'react';
...
+export type HeaderRowGeneratorFn<FS extends ELFSM> = (filerState: FS) => HeaderRow;
+
+export interface EntityTableProps<E extends Entity, FS extends ELFSM> extends EntityListViewProps<E, FS> {
+	bodyRowGenerator: BodyRowGeneratorFn<E, FS>;
+	headerRowGenerator: HeaderRowGeneratorFn<FS>;
+	tableCaption?: string;
+	tableId?: string;
+}
+
 export type EntityListComponent<E extends Entity, FS extends ELFSM> = React.ComponentType<EntityListViewProps<E, FS>>;
```
nicolad: cleeean :+1:

assets/src/domain/eventEditor/ui/datetimes/datesList/tableView/TableView.tsx
```diff
@@ -2,39 +2,28 @@ import React from 'react';
...
-			headerRows={[headerRows]}
-			metaData={{
-				tableId: 'date-entities-list-view',
-				tableCaption: __('Event Dates'),
-			}}
+		<EntityTable
+			entities={entities}
+			filterState={filterState}
+			bodyRowGenerator={datesListTableRow}
+			headerRowGenerator={datesListTableHeader}
```
nicolad: maybe we should name these 2 as the props that are receiving the values?
```
bodyRowGenerator={bodyRowGenerator}
headerRowGenerator={headerRowGenerator}
```

nicolad: and the related files

PR #2560, opened 2020-03-17
Entity edit forms: Empty box for no limit | Closes #1699

### LINKED ISSUE #1699: Display empty text box for "no limit" or "no maximum" instead of a -1
> For datetime and ticket limits, I prefer having “” be the value for “no limit” rather than “-1”. It’s fine to store it in the DB as “-1” but IMO an empty text box is more immediately understood to mean “nothing entered here/not in use” whereas “-1” requires reading some documentation somewhere to know what that means.

per Mike's comment in:

http://p2.eventespresso.com/2019/10/10/edtr-test-server/#comment-20394

Usage in ticket editor:
Screen Shot 2019-10-14 at 1 37 53 PM
(Quantity, Maximum Quantity, & Number of Uses fields)

Usage in datetime editor:
Screen Shot 2019-10-14 at 1 40 45 PM
(Capacity field)

manzoorwanijk:
This PR adds support for empty/blank field for no limit for the fields that can be infinity.
- Creates `formatInfinity` utility in `/application/services/utilities/number`
- Adds support for `parseAsInfinity` prop in `EspressoForm`
- Updates date and ticket form configs to use `parseAsInfinity`
- Closes #1699

### DIFF
--- a/assets/src/application/services/utilities/number/formatInfinity.ts
+++ b/assets/src/application/services/utilities/number/formatInfinity.ts
@@ -7,7 +7,7 @@ const formatInfinity: FormatInfinity = (value, defaultValue = '') => {
 	if (isInfinite(value)) {
 		return defaultValue;
 	}
-	return `${value}`;
+	return value.toString();
 };
 
 export default formatInfinity;
--- a/assets/src/application/services/utilities/number/formatInfinity.ts
+++ b/assets/src/application/services/utilities/number/formatInfinity.ts
@@ -1,0 +1,13 @@
+import isInfinite from './isInfinite';
+
+type FormatInfinity = (value: number, defaultValue?: string) => string;
+
+const formatInfinity: FormatInfinity = (value, defaultValue = '') => {
+	// If it is infinity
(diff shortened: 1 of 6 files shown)


assets/src/application/services/utilities/number/formatInfinity.ts
```diff
@@ -0,0 +1,13 @@
+import isInfinite from './isInfinite';
+
+type FormatInfinity = (value: number, defaultValue?: string) => string;
+
+const formatInfinity: FormatInfinity = (value, defaultValue = '') => {
+	// If it is infinity
+	if (isInfinite(value)) {
+		return defaultValue;
+	}
+	return `${value}`;
```
nicolad: `toString`?

nicolad: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toString

manzoorwanijk: Done

PR #2561, opened 2020-03-17
Add Entity List Legend Components; closes #2547.

### LINKED ISSUE #2547: Add Entity List Legend Components
In order for users to fully understand the EDTR UI we need to add a legend to the bottom of each entity list in order to describe the various icons and colour codings used throughout.

Here's an example of the legend currently in use below the Events Admin List Table

This is something that we will definitely utilize in the UIs for many different use cases, so ideally whatever we build should be highly reusable.

As well, just like the Entity Action Menus and Filter Bars, we will also likely require the ability to inject additional legend items into specific legends from addons, third party code, etc. So we will also want to wire this up using the fancy new registry logic that Manzoor just developed (or is still in the process of developing. That is not critical atm though so it is something we can add later. For now we can just pass an array of legend items.

Initially I see this being composed of three main parts:

1. abstract `EntityListLegend` component located in the `/application/ui/display` folder, which would be responsible for managing items and controlling their layout and display. This will likely also require a `LegendItem` component that will represent a single item. It's possible we may want to have two kinds of legend item components with one for icon based legend items and another for the colour swatches used to represent entity statuses. Because the legend is rather large, we might want to also have them hidden initially like we do with the filter bar filters and then only display them when an icon button is clicked. Let's use the AntD `<CompassFilled />` icon which could also rotate when clicked. I think we'll also want to have some text like "show legend" and "hide legend" beside the legend. The legends should be integrated directly with the EntityList and based on the ListId used for each.

2. a `DatesListLegend` for configuring the legend for the Dates list. This will require legend items for every icon used in the list as well as color swatches for each datetime status.

3. a `TicketsListLegend` for configuring the legend for the Tickets list. This will require legend items for every icon used in the list as well as color swatches for each ticket status.

nicolad:
- Implements `EntityListLegend` which gets configuration from dates and tickets and displays two columns of legend items.
- Closes: https://github.com/eventespresso/event-espresso-core/issues/2547


assets/src/application/ui/display/EntityListLegend/index.tsx
```diff
@@ -0,0 +1,50 @@
...
+const EntityListLegend: React.FC<Props> = ({ legendConfig }) => {
+	const { icons, swatches } = legendConfig;
+	return (
+		<Row className='ee-row'>
+			<Col span={12}>
+				<dl className='ee-list-table-legend'>
+					{Object.entries(icons).map(([icon, text]) => {
+						return (
+							<dt>
+								<EspressoIcon icon={icon} svgSize={18} />
```
nicolad: @manzoorwanijk I do get this error here, any hints or TS voodoo fixes?

image

nicolad:
@tn3rb furthermore when this will be hooked to registry I was thinking to make a small layout improvement:
is there are more swatches than icons, then I will slice the swatch array into 2 columns

assets/src/domain/eventEditor/ui/datetimes/datesList/config.ts
```diff
@@ -0,0 +1,16 @@
...
+		[Icon.GROUPS]: 'View Registrations for Event',
+		[Icon.TICKET]: 'View Assigned Tickets',
+	},
+	swatches: {
+		DTT: 'Trashed',
+		DTE: 'Expired',
+		DTS: 'Sold Out',
+		DTA: 'Active',
+	},
+};
```
nicolad: @tn3rb what are your thoughts on this?

assets/src/application/ui/layout/entityList/filterBar/EntityListFilterBar.tsx
```diff
@@ -17,8 +20,10 @@ import './style.scss';
 const EntityListFilterBar = <FS extends ELFSM>({
 	domain,
 	filterState,
+	legendConfig,
 	listId,
 }: EntityListFilterBarProps<FS>): JSX.Element => {
+	const [showLegend, setShowLegend] = useState(false);
```
nicolad: @tn3rb pulled down the state here, as discussed in chat.

assets/src/application/ui/display/EntityListLegend/index.tsx
```diff
@@ -0,0 +1,51 @@
...
+	const { icons, swatches } = legendConfig;
+
+	return (
+		<Row className='ee-row'>
+			<Col span={12}>
+				<dl className='ee-list-table-legend'>
+					{Object.entries(icons).map(([icon, description]) => {
+						return (
+							<dt>
+								<EspressoIcon icon={icon} svgSize={18} />
```
manzoorwanijk:
```suggestion
								<EspressoIcon icon={icon as Icon} svgSize={18} />
```

assets/src/application/ui/display/EntityListLegend/types.ts
```diff
@@ -0,0 +1,12 @@
+import { Icon } from '@appDisplay/espressoIcon';
+
+type Icons = {
+	[key in Icon]?: string;
+};
```
manzoorwanijk: My suggestion here would be to create an array of objects with descriptive keys like this:
```ts
type IconDescriptions = Array<{
    icon: Icon;
    description: string;
}>
```

assets/src/application/ui/display/EntityListLegend/types.ts
```diff
@@ -0,0 +1,12 @@
...
+
+export interface LegendConfig {
+	icons: Icons;
+	swatches: {
+		[key: string]: string;
+	};
```
manzoorwanijk:
```suggestion
	swatches: AnyObject<string>;
```

assets/src/application/ui/display/EntityListLegend/index.tsx
nicolad: thx @manzoorwanijk

assets/src/application/ui/display/EntityListLegend/types.ts
nicolad: agree with more explicit approach

assets/src/application/ui/display/EntityListLegend/index.tsx
```diff
@@ -0,0 +1,51 @@
...
+	legendConfig: LegendConfig;
+}
+
+const EntityListLegend: React.FC<Props> = ({ legendConfig }) => {
+	const { icons, swatches } = legendConfig;
+
+	return (
+		<Row className='ee-row'>
+			<Col span={12}>
+				<dl className='ee-list-table-legend'>
```
tn3rb: Although I really like the fact that you used a description list here, which is the most semantic markup for this use case, this isn't responsive the way it is setup.

I think we have two options here:

1. convert this completely to use AntD `Row` and `Col` components but use their responsive sizing props

ie:

```js
		<Row className='ee-row ee-list-table-legend' gutter={[16, { xs: 8, sm: 16, md: 24, lg: 32 }]}>
			{Object.entries(icons).map(([icon, description]) => {
				return (
					<Col xs={{ span: 24 }} sm={{ span: 12 }} md={{ span: 8 }} lg={{ span: 4 }}>
						<EspressoIcon icon={icon} svgSize={18} />
						<span className='ee-legend-description'>{description}</span>
					</Col>
				);
			})}
			{Object.entries(swatches).map(([swatchClassName, description]) => {
				return (
					<Col xs={{ span: 24 }} sm={{ span: 12 }} md={{ span: 8 }} lg={{ span: 4 }}>
						<span
							className={classNames('ee-status-legend', 'ee-status-background-color-' + swatchClassName)}
						></span>
						<span className='ee-legend-description'>{description}</span>
					</Col>
				);
			})}
		</Row>
```

2. drop the AntD components completely and use a description list (because I think your choice of dl, dt was great), but then use CSS grid to control the display

The former is easier but I think the latter would be better, at least from a semantics perspective which is somewhat important for a11y. We could also abstract either of those to be components in the app layer UI folder, whereas an abstract "List Table Legend" could go in with the Entity List component similar to how the filter bar and pagination are bundled with it. The specific "legend" components would of course need to go in the appropriate `/domain/` folder.

As well, I think this would be better if it simply mapped a single array of legend items to whatever kind of grid we end up using. 

How about we do the following:

- create a `LegendItem` component
- use additional components like `LegendItemIcon` and `LegendItemSwatch` for controlling those two aspects of things
- have those passed in via a single prop that is typed for those two components
- then either create:
     - `/application/ui/layout/grid` which wraps the AntD grid and sets up some common sizes for various breakpoints
     - `/application/ui/layout/descriptionList` which would use `<dl>`, `<dt>`, and `<dd>` for markup and CSS grid for controlling the display and responsiveness

This will better follow Single Responsibility Principle (SRP) imho since the individual legend item components will control their display as opposed to the list doing so. This allows for more variability of items to be added to these kinds of lists. For example, what if in the future we want to add a third kind of legend item? I also think having a list of one specific component will make it easier to handle adding items via the registry since it can be restricted to that specific type.

assets/src/application/ui/layout/entityList/EntityList.tsx
```diff
@@ -56,7 +57,12 @@ const EntityList = <E extends Entity, ELFS extends EntityListFilterStateManager<
 			<Title className='ee-entity-list__header' level={3}>
 				{headerText}
 			</Title>
-			<EntityListFilterBar domain={domain} filterState={filterState} listId={listId} />
+			<EntityListFilterBar
+				domain={domain}
+				filterState={filterState}
+				legendConfig={legendConfig}
```
tn3rb: I didn't like this at first because I thought these were going to display along with the filters but after checking out the branch and trying this I really like that this keeps the UI cleaner and has all of the buttons in the same place. 👍

assets/src/domain/eventEditor/ui/datetimes/datesList/config.ts
```diff
@@ -0,0 +1,16 @@
+import { Icon } from '@appDisplay/espressoIcon';
+import { LegendConfig } from '@application/ui/display/EntityListLegend/types';
+
+export const legendConfig: LegendConfig = {
+	icons: {
+		[Icon.EDIT]: 'Edit Event Details',
+		[Icon.GROUPS]: 'View Registrations for Event',
+		[Icon.TICKET]: 'View Assigned Tickets',
```
tn3rb:
```suggestion
		[Icon.EDIT]: __('Edit Event Date Details'),
		[Icon.GROUPS]: __('View Registrations for this Date'),
		[Icon.TICKET]: __('Manage Ticket Assignments'),
		[Icon.TRASH]: __('Move Date to Trash'),
```

assets/src/domain/eventEditor/ui/tickets/ticketsList/config.ts
```diff
@@ -0,0 +1,16 @@
+import { Icon } from '@appDisplay/espressoIcon';
+import { LegendConfig } from '@application/ui/display/EntityListLegend/types';
+
+export const legendConfig: LegendConfig = {
+	icons: {
+		[Icon.EDIT]: 'Edit Event Details',
+		[Icon.CALENDAR]: 'View Assigned Dates',
+		[Icon.CALCULATOR]: 'View Ticket Price Calculator',
```
tn3rb:
```suggestion
		[Icon.EDIT]: 'Edit Ticket Details',
		[Icon.CALENDAR]: 'Manage Date Assignments',
		[Icon.CALCULATOR]: 'Ticket Price Calculator',
		[Icon.TRASH]: __('Move Ticket to Trash'),
```

assets/src/application/ui/display/EntityListLegend/types.ts
nicolad: updated

assets/src/application/ui/display/EntityListLegend/types.ts
nicolad: updated

assets/src/domain/eventEditor/ui/datetimes/datesList/config.ts
```diff
@@ -0,0 +1,16 @@
...
+	},
+	swatches: {
+		DTT: 'Trashed',
+		DTE: 'Expired',
+		DTS: 'Sold Out',
+		DTA: 'Active',
```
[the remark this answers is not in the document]
nicolad: added `DTU` and made those translatable

assets/src/application/ui/layout/entityList/filterBar/buttons/LegendButton/index.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+	return (
+		<div className='ee-toggle-legend-btn'>
+			<EspressoButton
+				buttonType={EspressoButtonType.MINIMAL}
+				className={className}
+				icon={icon}
+				id={id}
+				onClick={toggleLegend}
+				tooltip={__('show filters')}
```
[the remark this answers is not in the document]
nicolad: done, also updated the other button:

image
***
image

assets/src/application/ui/layout/entityList/filterBar/buttons/LegendButton/index.tsx
```diff
@@ -0,0 +1,41 @@
...
+			/>
+			<label className='esprs-button-label' htmlFor={id}>
+				{__(`${showLegend ? 'hide' : 'show'} legend`)}
+			</label>
```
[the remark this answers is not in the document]
nicolad: ok, lemme know the result

assets/src/domain/eventEditor/ui/datetimes/datesList/config.ts
nicolad: updated, thx

assets/src/domain/eventEditor/ui/tickets/ticketsList/config.ts
nicolad: updated, thank you

assets/src/application/ui/display/EntityListLegend/index.tsx
nicolad: @tn3rb please elaborate this:
`have those passed in via a single prop that is typed for those two components`

assets/src/domain/eventEditor/ui/tickets/ticketsList/config.ts
```diff
@@ -0,0 +1,16 @@
...
+	},
+	swatches: {
+		DTT: 'Trashed',
+		DTE: 'Expired',
+		DTS: 'Sold Out',
+		DTA: 'Active',
```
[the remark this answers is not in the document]
nicolad: updated and added `aria-label`s

assets/src/application/ui/display/EntityListLegend/index.tsx
tn3rb: in its simplest form, imagine the `LegendItem` component having two props:

- symbol: for displaying an icon or status swatch

- description: the text that explains what the legend item symbol represents

The typing for the props might then be something like :

```ts
interface LegendItemProps {
	symbol: Icon | Swatch; 
	description: string;
}
```

where `Icon | Swatch` would correspond to the `LegendItemIcon` and `LegendItemSwatch` components.

Now that said, there's multiple ways this could be done.

If you go the route of creating a more abstracted "DescriptionList" component, then instead of having  `LegendItem`, `LegendItemIcon`, and `LegendItemSwatch` components, it might be better to have generic `ListItem` and `ColorSwatch` components that we use in conjunction with `EspressoIcon` and then simply combine those together when creating an `AdminLegend` component.

The primary reason I'm suggesting this level of abstraction is because it's very likely that we will require these kinds of components for other use cases. As well, when it comes to extending these things via Manzoor's registry, we want to be able to easily restrict what gets added to things. So in this case, we don't want add-ons or third party devs co-opting an admin legend to add in some completely different and unexpected content. So only having one specific component that can be registered will clamp down on what can be done with those elements.

tn3rb:
re: labels and tooltips

Josh suggested the following:

button labels

as well as the following comment:

> the tooltips can indicate a bit more like “show this” or “hide that”

so let's do that, but also increase the spacing between buttons to help differentiate the labels from each other

assets/src/application/ui/layout/descriptionList/DescriptionList.tsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+
+import { DescriptionListProps } from './types';
+import './styles.scss';
+
+const DescriptionList: React.FC<DescriptionListProps> = ({ dataSource }) => {
```
nicolad: I've used the same prop as in antd `List`: https://github.com/ant-design/ant-design/blob/master/components/list/index.tsx#L39

assets/src/application/ui/layout/descriptionList/DescriptionList.tsx
```diff
@@ -0,0 +1,21 @@
...
+	return (
+		<dl className='ee-description-list-grid'>
+			{dataSource.map(({ description, term }) => {
+				return (
+					<div>
+						<dt>{term}</dt>
+						<dd>{description}</dd>
+					</div>
```
nicolad: This can be wrapped in `div`, based on MDN:

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dl

image

nicolad:
This is how it looks now:

image

assets/src/application/ui/display/ColorSwatch/index.tsx
```diff
@@ -0,0 +1,15 @@
...
+import classNames from 'classnames';
+
+import './style.scss';
+
+interface Props {
+	className?: string;
+	label?: string;
+}
+
+const ColorSwatch: React.FC<Props> = ({ className, label }) => {
```
nicolad: I thought that `application/ui/display` is a good place for this component.

assets/src/application/ui/display/EntityListLegend/types.ts
```diff
@@ -0,0 +1,12 @@
+import { AnyObject } from '@appServices/utilities/types';
+import { Icon } from '@appDisplay/espressoIcon';
+
+type IconType = {
```
nicolad: I named this `IconType` because I already imported `Icon` type.

assets/src/application/ui/layout/descriptionList/styles.scss
```diff
@@ -0,0 +1,15 @@
+.ee {
+	&-description-list-grid {
```
nicolad: It's a bit tricky to make this grid dynamic.
What I would like to implement in a separate PR is to define `grid-template-rows` based on some passed prop or to compute somehow it's value.

assets/src/application/ui/display/ColorSwatch/index.tsx
tn3rb: ya that's perfect

assets/src/application/ui/layout/descriptionList/styles.scss
tn3rb: try this for the parent container (dl?)

```css
		display: grid;
		gap: var(--ee-margin-big);
		grid-template-columns: repeat(auto-fit, minmax(SET_PX_WIDTH_YOU_WANT, 1fr));
		grid-auto-rows: SET_PX_HEIGHT_YOU_WANT;
```

gotta say:

```csss
.ee {
	&-
```

this is very clever namespacing 👍

assets/src/application/ui/layout/descriptionList/DescriptionList.tsx
tn3rb: I still think this, combined with some of what you added in `EntityListLegend`, would be good to extract into separate components

ie:

```js
interface DescriptionListItemProps {
	className: string;
	description: string;
	term: string | EspressoIcon | ColorSwatch; // or React.Node or whatever
}

const DescriptionListItem: React.FC<DescriptionListItemProps > = ({ className, description, term }) => (
	<div className={`${className} ee-description-list-item`}>
		<dt>{term}</dt>
		<dd>{description}</dd>
	</div>
);
```

then the interface for `DescriptionList` could type for an array of `DescriptionListItem[]`

and `EntityListLegend` would do something like:

```js
	const iconsSource = icons.map(({ icon, description }) => {
		return (
			<DescriptionListItem
				className='ee-legend-description'
				description={description}
				term={
					<EspressoIcon aria-label={description} icon={icon as Icon} svgSize={18} />
				}
			/>
		);
	});
```

because then everything is components

assets/src/application/ui/layout/descriptionList/styles.scss
nicolad: updated:

image

***
will have to tweak more when registry will be coupled

assets/src/application/ui/display/ColorSwatch/style.scss
```diff
@@ -0,0 +1,12 @@
+/** -------------------------------------------------------------------------
+ * Event Espresso ColorSwatch Stylesheet
+ * © 2020 Event Espresso
+ * ------------------------------------------------------------------------- */
+
+.ee {
+	&-color-swatch {
+		display: inline-block;
+		height: 18px;
+		width: 18px;
```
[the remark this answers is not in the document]
nicolad: `var(--ee-font-size-big)` seems a little too big, at least compared to icons

image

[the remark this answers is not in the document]
nicolad: applied default size

image

assets/src/application/ui/layout/descriptionList/DescriptionList.tsx
nicolad: done here: https://github.com/eventespresso/event-espresso-core/pull/2561/commits/4e25febdd9d3cb85ab05fcd7f2dfe7665449afa0

nicolad: it is a bit tricky to setup:
`dataSource: DescriptionListItemProps[]`, have set for now:
`dataSource: JSX.Element[]`
***
more info in this thread: https://eventespresso.slack.com/archives/C080DDVLM/p1584627543023800

assets/src/application/ui/layout/descriptionList/types.ts
```diff
@@ -5,5 +5,5 @@ export interface DescriptionListItemProps {
 }
 
 export interface DescriptionListProps {
-	dataSource: JSX.Element[];
+	dataSource: DescriptionListItemProps[];
```
nicolad: @tn3rb please note that in order to preserve this type @manzoorwanijk gave his :brain: and some changes has been made

assets/src/application/ui/display/ColorSwatch/style.scss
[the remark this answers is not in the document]
tn3rb: ya that looks POIFECT 👌

assets/src/application/ui/layout/entityList/filterBar/Collapsible.tsx
```diff
@@ -2,42 +2,25 @@ import React, { useRef } from 'react';
...
-	showEntityFilters = false,
-}) => {
+const Collapsible: React.FC<CollapsibleProps> = ({ children, show = false }) => {
 	const ref = useRef<HTMLDivElement>();
 	const { height } = useRect(ref);
 	const props = useSpring({
-		height: showEntityFilters ? height : 0,
-		opacity: showEntityFilters ? 1 : 0,
+		height: show ? height : 0,
+		opacity: show ? 1 : 0,
```
tn3rb: nice adherence to DRY

beyond the scope of this PR, but we could likely extract this into its own component for use outside of the filter bar

PR #2571, opened 2020-03-18
Add a copyPromotionLineItem method to CopyRegistrationService.php.

Pebblo:
This PR adds a method that does the heavy lifting for https://github.com/eventespresso/eea-attendee-mover/pull/15

This new method takes the 'old' and 'new' registration objects.

Pulls the total line item from the transaction and then uses that + the tickets from the registration objects to pull the event line items for each (this would need refactoring for additional promotion scopes) and then applies any promotions that applied to the old event line item, to the new one.

What I don't like about this is the reliance on the promotions add-on but not sure which way to prevent that.

---

## Checklist

* [ ] My code is tested.


core/domain/services/registration/CopyRegistrationService.php
```diff
@@ -174,4 +176,45 @@ public function copyPaymentDetails(
...
+     * @throws RuntimeException
+     * @throws UnexpectedEntityException
+     * @throws EE_Error
+     */
+    public function copyPromotionLineItems(
+        EE_Registration $target_registration,
+        EE_Registration $registration_to_copy
+    ) {
+        $transaction = $registration_to_copy->transaction();
+        if ($transaction instanceof \EE_Transaction) {
```
tn3rb: plz import all classes so that all dependencies are listed at the top of the file
so in other words, this will change to 

```suggestion
        if ($transaction instanceof EE_Transaction) {
```

and then you'll add:

```php
use EE_Transaction;
```

at the top with the other imports.

Then do this for all other classes still using the root namespace `\`

core/domain/services/registration/CopyRegistrationService.php
```diff
@@ -174,4 +176,45 @@ public function copyPaymentDetails(
...
+            $new_event_line_item = EEH_Line_Item::get_event_line_item_for_ticket($total_line_item, $target_registration->ticket());
+            $promo_line_items = EEH_Line_Item::get_line_items_of_object_type($old_event_line_item, \EEM_Line_Item::OBJ_TYPE_PROMOTION);
+
+            // Bail if we don't have any promotion line items.
+            if (empty($promo_line_items)) {
+                return;
+            }
+
+            // Don't increment 'usage' for copied promotions
+            add_filter('FHEE__EED_Promotions__add_promotion_line_item__bypass_increment_promotion_scope_uses', '__return_true');
```
tn3rb: 👍

core/domain/services/registration/CopyRegistrationService.php
```diff
@@ -174,4 +176,45 @@ public function copyPaymentDetails(
...
+    public function copyPromotionLineItems(
+        EE_Registration $target_registration,
+        EE_Registration $registration_to_copy
+    ) {
+        $transaction = $registration_to_copy->transaction();
+        if ($transaction instanceof \EE_Transaction) {
+            $total_line_item = $transaction->total_line_item();
+            $old_event_line_item = EEH_Line_Item::get_event_line_item_for_ticket($total_line_item, $registration_to_copy->ticket());
+            $new_event_line_item = EEH_Line_Item::get_event_line_item_for_ticket($total_line_item, $target_registration->ticket());
+            $promo_line_items = EEH_Line_Item::get_line_items_of_object_type($old_event_line_item, \EEM_Line_Item::OBJ_TYPE_PROMOTION);
```
tn3rb: do we need to cancel these too? or was that already done (i don't remember and don't have phpStorm open atm to look)

garthkoyle:
@Pebblo how should this be tested?

joaovictorangeline:
@Pebblo Just a reminder on how to test this one.

PR #2573, opened 2020-03-19
Prevent Form Submission when ENTER Key Pressed on AntD Select Inputs; closes #2568.

### LINKED ISSUE #2568: Son of Prevent Form Submission when ENTER Key Pressed on AntD Select Inputs
plz see #2527

Same fix applied in that PR needs to be applied to basically any and all AntD Select inputs throughout the EDTR to prevent them from submitting the WP form when ENTER is pressed.

nicolad:
- Adds `onPressEnter` `preventDefault` to prevent default form enter behavior.
- Optionally added types from antd and made related adjustments.
- Closes: https://github.com/eventespresso/event-espresso-core/issues/2568

### DIFF
--- a/assets/src/application/ui/forms/espressoForm/config.tsx
+++ b/assets/src/application/ui/forms/espressoForm/config.tsx
@@ -49,7 +49,7 @@ export const formConfig: EspressoFormProps = {
 					name: 'email',
 					label: 'Email',
 					fieldType: 'text',
-					htmlType: 'email',
+					type: 'email',
 				},
 				{
 					name: 'phone',
@@ -241,13 +241,13 @@ export const formConfig: EspressoFormProps = {
 			label: 'Price',
 			fieldType: 'text',
 			addonAfter: '$',
-			htmlType: 'number',
+			type: 'number',
 		},
 		{
 			name: 'password',
 			label: 'Password',
 			fieldType: 'text',
-			htmlType: 'password',
+			type: 'password',
 			desc: 'Please enter a strong password',
 		},
 		{
--- a/assets/src/application/ui/forms/espressoForm/config.tsx
+++ b/assets/src/application/ui/forms/espressoForm/config.tsx
@@ -241,13 +241,13 @@ export const formConfig: EspressoFormProps = {
 			label: 'Price',
 			fieldType: 'text',
 			addonAfter: '$',
-            htmlType: 'number',
+			htmlType: 'number',
 		},
 		{
 			name: 'password',
 			label: 'Password',
 			fieldType: 'text',
-            htmlType: 'password',
+			htmlType: 'password',
 			desc: 'Please enter a strong password',
 		},
 		{
--- a/assets/src/application/ui/forms/espressoForm/config.tsx
+++ b/assets/src/application/ui/forms/espressoForm/config.tsx
@@ -49,7 +49,7 @@ export const formConfig: EspressoFormProps = {
 					name: 'email',
 					label: 'Email',
 					fieldType: 'text',
-					type: 'email',
+					htmlType: 'email',
 				},
 				{
 					name: 'phone',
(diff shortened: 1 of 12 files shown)


assets/src/application/ui/input/Text.tsx
```diff
@@ -1,8 +1,9 @@
 import React from 'react';
 import { Input } from 'antd';
+import { InputProps } from 'antd/lib/input';
 
-const Text = ({ input, htmlType = 'text', ...rest }) => {
-	return <Input {...input} {...rest} type={htmlType} />;
+const Text: React.FC<InputProps> = ({ type = 'text', ...props }) => {
```
nicolad: Because `InputProps` doesn't have `htmlType`, I've tracked down why is that and found next:
- `InputProps` extends `Omit<React.InputHTMLAttributes<HTMLInputElement>, 'size' | 'prefix'>`: https://github.com/ant-design/ant-design/blob/master/components/input/Input.tsx#L15
which has this type: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts#L2057
- In `Button` component antd uses `htmlType` although which was a bit confusing: https://github.com/ant-design/ant-design/blob/master/components/button/button.tsx#L100

assets/src/application/ui/input/Number.tsx
```diff
@@ -1,8 +1,9 @@
 import React from 'react';
 import { InputNumber } from 'antd';
+import { InputNumberProps } from 'antd/lib/input-number';
 
-const NumberField = ({ input, ...rest }) => {
-	return <InputNumber {...input} {...rest} />;
```
nicolad: in this case, it didn't make sense to pass these separately

nicolad:
In few related components I saw missing types and where it was straightforward, I've added missing types.

assets/src/application/ui/forms/espressoForm/config.tsx
```diff
@@ -49,7 +49,7 @@ export const formConfig: EspressoFormProps = {
 					name: 'email',
 					label: 'Email',
 					fieldType: 'text',
-					htmlType: 'email',
+					type: 'email',
```
tn3rb: this may have been this way to avoid a conflict with RFF... 
or is just to differentiate between the other "type" used in the form config (ie: fieldType)

nicolad: ok, adjusted back, thx

nicolad:
related to `Select` and `SelectInput`, I've searched now in 4 places to make sure I don't miss something and I didn't found `onPressEnter` callback:
- docs: https://ant.design/components/select/
- source code: https://github.com/ant-design/ant-design/blob/master/components/select/index.tsx
- local definition files
- have searched through antd repo - found that `onPressEnter` is used also internally also by `Tag`, `Table` and `ResizableTextArea`, but currently we don't use any of those as inputs : https://github.com/ant-design/ant-design/search?p=1&q=onPressEnter&unscoped_q=onPressEnter

***
local definition file:

image

assets/src/application/ui/input/Number.tsx
manzoorwanijk: It does make sense passing them separately. `input` object props have to be passed as props to `<InputNumber />`. Not just `input`, there is `meta` object as well but we haven't used it here.
For details about the shape of `input` and `meta`, see RFF [FieldRenderProps](https://final-form.org/docs/react-final-form/types/FieldRenderProps)

assets/src/application/ui/input/Text.tsx
```diff
@@ -1,8 +1,13 @@
 import React from 'react';
 import { Input } from 'antd';
+import { InputProps } from 'antd/lib/input';
 
-const Text = ({ input, htmlType = 'text', ...rest }) => {
-	return <Input {...input} {...rest} type={htmlType} />;
```
manzoorwanijk: Here as well, `input` object must be passed to the input.

nicolad: I don't understand what you mean, `props` includes `input` prop, why this needs to be passed separately and the spreaded each individually.

nicolad: this was clarified in the chat, thx

nicolad:
found a workaround for the missing `onPressEnter` on select by using `onInputKeyDown`

Peek 2020-03-20 17-04

PR #2577, opened 2020-03-20
Refactor Ticket Price Calculator | Closes #2567

### LINKED ISSUE #2567: Remove RFF from Ticket Price Calculator
The Ticket Price Calculator (TPC) in both EDTR v1 as well as the current implementation have been using React Final Form (RFF) for managing form state. In retrospect, this unfortunately turned out to be a bad decision because there were a few subtle issues that I have been completely unable to resolve, primarily:

 - RFF requires the use of form decorators for responding to changes in form data but does not allow those callbacks to change from render to render. In order to recalculate the base price or total however, the values need to take into account additional price modifiers added to the form, which changes the content of the decorator callbacks because they now include additional modifiers in their calculations. This appears to be an incompatibility between `final-form-calculate` and `react-final-form-arrays` (or could still be because of my implementation).

Anyways, at this point it should simplify things greatly by just not using RFF at all for the Ticket Price Calculator. This will however require the existing logic to be refactored.

Here's a rough breakdown of the logic in the order that it needs to occur:

- TPC button is clicked for a specific ticket

- current ticket details need to be retrieved from the Apollo cache

- the ticket's related price entities need to be retrieved 

- the price's related price type entities need to be retrieved 

- to avoid updates to the above entities while the user is editing things within the TPC, new objects and TS types should be generated for use within the calculator. A single `PriceModifier` object/type should be sufficient as the only details needed from the ticket are the current total (which can be calculated from the individual prices) and the value for the new `reverseCalculate` field.

example TPC data object:

```js
{
	ticket: {
		id: abc123, // for tracking ticket in apollo cache
		total: 19.95,
		reverseCalculate: false,
	},
	priceModifiers: [
		{
			id: def456, // for tracking existing price in apollo cache
			type: BASE_PRICE,
			label: 'name',
			desc: 'blah blah',
			amount: 15,
		},
		{
			id: 'ghi789', // for tracking existing price in apollo cache
			type: FIXED_SURCHARGE,
			label: 'extra fee',
			desc: 'just cuz',
			amount: 4.95,
		},
		{
			id: 'NEW', // indicates new price to create upon submission
			type: FED_TAX,
			label: 'tax',
			desc: 'theft',
			amount: 10,
		},
	],
}
```

- the array of `PriceModifier` objects, plus values for the `ticketTotal` and `reverseCalculate` need to be initialized and saved to local state.

- the TPC renders with a single table row for each `PriceModifier` plus a footer row showing the `ticketTotal`. Here's a screenshot from a previous build

:

- the first row should always be for the tickets base price and should NOT have a delete button available because every ticket should have a base price even if the ticket is free, in which case the base price simply has an amount of `0`. The price type selector for the base price should also be disabled since we don't want users changing that to something else.

- each row after that should have a price type selector with "base price" removed as an option (there can only be one!) and a delete button in the "Actions" column.

- the last row of the table body should be for adding a new price modifier which is the same as the previous rows except instead of a delete button it has a button for committing the addition of the new modifier to the ticket price. I'm considering removing this and requiring the user to click the add button first prior to having the fields available because as is, I'm worried that people will enter data into the fields and forget to click the "Add" button and then potentially lose their data. If the add button is clicked we should display an error if no amount has been entered (label and desc can be empty).

- the final row in the table has only two fields. A number input for setting the ticket total price, and a button for toggling the calculation direction, ie: the `reverseCalculate` value. If set to `false` (the default value), then the ticket total field is disabled and its value is calculated starting with the base price and applying each modifier in order. If `reverseCalculate` is set to `true`, then the ticket total field isenabled and its value is used as the starting point to calculate the base price by applying each modifier in reverse order.

- so depending on the `reverseCalculate` value, we either calculate the ticket total or the ticket base price when any of the following events occurs:

     - base price amount is changed (and `reverseCalculate` is `false`)

     - regular price modifier "Price Type" is changed (ie: surcharge, discount, dollar, %)

     - regular price modifier amount is changed

     - new price modifier is added

     - ticket total amount is changed (and `reverseCalculate` is `true`)

- **here's the tricky part** recalculating the total or base price can NOT trigger the calculations to run again (this was another issue with RFF because it would fire its decorators for a field even if the input was disabled causing the updating to recurse). As well, there needs to be a delay between when an input value changes and when the calculator recomputes values because we don't want that happening while someone is still in the process of entering a number. 500ms is sufficient.

- upon final submission, the existing ticket in cache that the data was taken from needs to be updated if either the ticket total or value for `reverseCalculate` has changed. Similarly for prices, any values for `PriceModifier` objects that have an ID corresponding to an existing price in the apollo cache need to be updated. All other `PriceModifier` objects should result in the creation of new prices. We also need to somehow track price modifiers that are removed (maybe via a `trashed` flag in the TPC data?) so that those price objects can be trashed.

manzoorwanijk:
This PR removes the RFF dependency for Ticket price calculator by creating a separate state management for it.
- Closes #2567


assets/src/domain/eventEditor/ui/tickets/ticketPriceCalculator/data/types.ts
```diff
@@ -0,0 +1,48 @@
...
+	deletedPrices: Array<EntityId>;
+}
+
+export type DataActionType =
+	| 'UPDATE_TICKET_PRICE'
+	| 'TOGGLE_REVERSE_CALCULATE'
+	| 'ADD_PRICE'
+	| 'UPDATE_PRICE'
+	| 'DELETE_PRICE'
+	| 'ADD_PRICE_TO_DELETED'
```
tn3rb: I'm assuming:

- DELETE_PRICE is for simply removing newly added modifiers

- ADD_PRICE_TO_DELETED is for scheduling existing prices for deletion

manzoorwanijk: - `DELETE_PRICE` removes a price modifier
- `ADD_PRICE_TO_DELETED` adds the deleted price ID to the bucket which we can use in `onSubmit`

assets/src/domain/eventEditor/ui/tickets/ticketPriceCalculator/data/useDataReducer.ts
```diff
@@ -0,0 +1,88 @@
...
+				const newPrices =
+					typeof index !== 'undefined' ? insert(index, price, state.prices) : append(price, state.prices);
+				return {
+					...state,
+					prices: newPrices,
+				};
+
+			case 'UPDATE_PRICE':
+				// find the index of the price to update
+				const priceIndex = findIndex(propEq('id', id))(state.prices);
```
tn3rb: might also be able to use `entityHasGuid()` from `/domain/shared/services/predicates/selectionById` 

or better yet... 

move the above into that file (modified accordingly of course) and then import here.

Question:

is there any benefit to using the curried version of this:

```js
findIndex(propEq('id', id))(state.prices)
```

vs:

```js
findIndex(propEq('id', id), state.prices)
```

???

assets/src/domain/eventEditor/ui/tickets/ticketPriceCalculator/data/useDataStateManager.ts
```diff
@@ -0,0 +1,90 @@
...
+			if (!isNew) {
+				dispatch({
+					type: 'ADD_PRICE_TO_DELETED',
+					id,
+				});
+			}
+			dispatch({
+				type: 'DELETE_PRICE',
+				id,
+			});
```
tn3rb: yup... that's what I figured :)

assets/src/domain/eventEditor/ui/tickets/ticketPriceCalculator/data/useDataReducer.ts
manzoorwanijk: No extra benefit.

manzoorwanijk: Removed the curried version

manzoorwanijk: used `entityHasGuid` 👍

assets/src/domain/eventEditor/ui/tickets/ticketPriceCalculator/utils/calculateTicketTotal.ts
```diff
@@ -0,0 +1,31 @@
...
+	}
+
+	/**
+	 * @todo Make sure the base price (or non percent price) is at the beginning of the list,
+	 * otherwise if a percent price ends up at the beginning, it will have
+	 * no effect on total becaue of multiply by intial total which is 0.
+	 * May be same applies to the sorting in `calculateBasePrice`
+	 */
+	const sortedModifiers = sortByPriceOrderIdAsc(allPrices);
+	const newTicketTotal = reduce(ticketTotalCalculator, 0, sortedModifiers);
```
manzoorwanijk: @tn3rb I need your attention/help here. This problem wasted my time until I figured out why the percent prices were being ignored 🙄

assets/src/domain/eventEditor/ui/tickets/ticketPriceCalculator/utils/calculateTicketTotal.ts
```diff
@@ -0,0 +1,31 @@
...
+	}
+
+	/**
+	 * @todo Make sure the base price (or non percent price) is at the beginning of the list,
+	 * otherwise if a percent price ends up at the beginning, it will have
+	 * no effect on total because of multiply by intial total which is 0.
+	 * May be same applies to the sorting in `calculateBasePrice`
+	 */
+	const sortedModifiers = sortByPriceOrderIdAsc(allPrices);
+	const newTicketTotal = reduce(ticketTotalCalculator, 0, sortedModifiers);
```
manzoorwanijk: @tn3rb I need your attention/help here. This problem wasted my time until I figured out why the percent prices were being ignored 🙄

tn3rb: yes that's why the base price always needs to be first (or ticket total when reverse calculating) because the base price is it's own type and can not be a percent

assets/src/domain/eventEditor/ui/tickets/ticketPriceCalculator/stateListeners/usePriceChangeListener.ts
```diff
@@ -0,0 +1,32 @@
...
+	const { prices } = useDataState();
+
+	const [priceAmountsStr, setPriceAmountsStr] = useState('');
+	const [booleanFlagsStr, setBooleanFlags] = useState('');
+	useEffect(() => {
+		// To avoid triggering the change on every render
+		// collect all the prices (excluding empty/zero)
+		// convert to JSON to only trigger when the value changes
+		const priceAmounts = JSON.stringify(prices.map((price) => price.amount).filter(Boolean));
+		const booleanFlags = JSON.stringify(prices.map((price) => pick(pirceBooleanFlags, price)));
```
manzoorwanijk: Inspiration from https://twitter.com/dan_abramov/status/1104414272753487872

assets/src/domain/eventEditor/ui/tickets/ticketPriceCalculator/stateListeners/useInitStateListeners.ts
```diff
@@ -0,0 +1,38 @@
...
+			updateBasePrice();
+		} else {
+			updateTicketTotal();
+		}
+	}, [reverseCalculate, updateBasePrice, updateTicketTotal]);
+
+	// Subscribe to price related changes.
+	usePriceChangeListener(calculatePrice);
+	// Subscribe to reverseCalculate changes
+	useReverseCalcChangeListener(calculatePrice);
```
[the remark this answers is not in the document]
manzoorwanijk: I observed some decimal precision issue. I agree that it shouldn't need to recalculate the price

assets/src/domain/eventEditor/ui/tickets/ticketPriceCalculator/stateListeners/usePriceTypeChangeListener.ts
```diff
@@ -0,0 +1,62 @@
...
+		// To avoid triggering the change on every render
+		// convert to JSON to only trigger when the value changes
+		const newPriceTypesStr = JSON.stringify(prices.map((price) => price.priceType));
+		setPriceTypesStr(newPriceTypesStr);
+	}, [prices]);
+	// Because of the deps, it will run only when priceTypes changes
+	useEffect(() => {
+		const newPriceTypeMapping = prices.reduce((acc, price) => {
+			return assocPath([price.id], price.priceType, acc);
+		}, {});
```
tn3rb: can move contents of second `useEffect` into the first and wrap in a conditional that compares `priceTypesStr` against `newPriceTypesStr` before running the calculator

assets/src/domain/eventEditor/ui/tickets/ticketPriceCalculator/stateListeners/usePriceChangeListener.ts
```diff
@@ -11,22 +11,23 @@ const usePriceChangeListener: StateChangeListenerHook = (calculatePrice) => {
...
 	useEffect(() => {
 		// To avoid triggering the change on every render
 		// collect all the prices (excluding empty/zero)
 		// convert to JSON to only trigger when the value changes
 		const newPriceAmountsStr = JSON.stringify(prices.map((price) => price.amount).filter(Boolean));
 		const newBooleanFlagsStr = JSON.stringify(prices.map((price) => pick(priceBooleanFlags, price)));
+
+		if (newPriceAmountsStr === priceAmountsStr && newBooleanFlagsStr === booleanFlagsStr) {
+			return;
+		}
```
tn3rb: immediate return is even better than what I suggested 👍

assets/src/domain/eventEditor/ui/tickets/ticketPriceCalculator/buttons/PriceModifierActions.tsx
```diff
@@ -1,24 +1,27 @@
...
-import { TpcModifierFormRowProps } from '../types';
+import { PriceModifierProps } from '../types';
 
-const PriceModifierActions: React.FC<TpcModifierFormRowProps> = ({
-	index,
-	name,
-	price,
-	fields: { push, remove, reset, sort },
-}) => {
+// just temporary
```
nicolad: some day maybe we will be using `emotion` for this kind of scope styles, not today :)

assets/src/domain/eventEditor/ui/tickets/ticketPriceCalculator/buttons/PriceModifierActions.tsx
```diff
@@ -1,24 +1,27 @@
...
-			<AddPriceModifierButton key={key} name={name} price={price} push={push} reset={reset} sort={sort} />
-		);
-	} else if (!price.isBasePrice) {
-		actions.push(<DeletePriceModifierButton key={key} index={index} remove={remove} />);
+
+	actions.push(<AddPriceModifierButton key={'add'} index={index} />);
+
+	if (!price.isBasePrice) {
+		actions.push(<DeletePriceModifierButton key={'delete'} price={price} />);
 	}
```
nicolad: maybe it's a personal preference, but I'd rather use short circuit evaluation for this kind of logic - it's a bit cleaner and I don't need to know anything about that additional array

assets/src/domain/eventEditor/ui/tickets/ticketPriceCalculator/buttons/TicketPriceCalculatorButton.tsx
```diff
@@ -1,33 +1,32 @@
...
-		<EspressoButton
-			icon='calculator'
-			onClick={onClick}
-			tooltip={__('ticket price calculator')}
-			tooltipProps={{ placement: 'left' }}
-			{...buttonProps}
-		/>
+		<>
+			<EspressoButton
+				icon='calculator'
```
nicolad: why not `icon={Icon.CALENDAR}` :) ?

PR #2578, opened 2020-03-20
Filter Bar Changes: part one; partially closes #2476.

### LINKED ISSUE #2476: Filter Bar Changes
## For Entity List Filter Bar

- see: `\application\ui\layout\entityList\filterBar\EntityListFilterBar.tsx`

- change all references to "List" view (and its variations) to "Table" view and update all related code

- change all references to "Grid" view (and its variations) to "Card" view and update all related code

- update styles so that:

     - all references to WordPress GB classes like `components-icon-button` are removed and replaced accordingly

     - the `ee-active-filters` applies the same color affects as the hover state

     - styles can be converted to sass if that helps with the refactor

## For Dates List Filter Bar

- see `\domain\eventEditor\ui\datetimes\datesList\filterBar\controls\ShowDatesControl.tsx`

- split the "ShowDates" filter and state into two separate components:

     - one for "Status" that includes the following:
          - any (all Dates)
          - activeUpcoming <= default
          - activeOnly
          - upcomingOnly
          - nextActiveUpcomingOnly
          - soldOutOnly
          - expiredOnly
          - recentlyExpiredOnly
          - trashedOnly

     - one for "Sales" that includes the following:
          - any (all Dates) <= default
          - above90Capacity
          - above75Capacity
          - above50Capacity
          - below50Capacity

- update all related logic accordingly

## For Tickets List Filter Bar

- see `\domain\eventEditor\ui\tickets\ticketsList\filterBar\controls\ShowTicketsControl.tsx`

- split the "ShowTickets" filter and state into two separate components:

     - one for "Status" that includes the following:
          - any (all Tickets)
          - onSaleAndPending <= default
          - onSaleOnly 
          - pendingOnly 
          - nextOnSaleOrPendingOnly 
          - soldOutOnly 
          - expiredOnly
          - recentlyExpiredOnly
          - trashedOnly

     - one for "Sales" that includes the following:
          - any (all Tickets) <= default
          - above90Sold 
          - above75Sold 
          - above50Sold 
          - below50Sold 

- update all related logic accordingly

nicolad:
### For Entity List Filter Bar

- see: `\application\ui\layout\entityList\filterBar\EntityListFilterBar.tsx`

- change all references to "List" view (and its variations) to "Table" view and update all related code

- change all references to "Grid" view (and its variations) to "Card" view and update all related code

- update styles so that:

     - all references to WordPress GB classes like `components-icon-button` are removed and replaced accordingly

     - the `ee-active-filters` applies the same color affects as the hover state

     - styles can be converted to sass if that helps with the refactor

## How has this been tested
Peek 2020-03-20 20-23


assets/src/application/ui/layout/entityList/filterBar/style.scss
```diff
@@ -3,6 +3,8 @@
  * © 2018 Event Espresso
  * ------------------------------------------------------------------------- */
 
+$filter-btn: '.esprs-button.ant-btn.ee-entity-list-filter-bar__btn';
```
nicolad: more info: https://css-tricks.com/snippets/sass/use-sass-variable-selector/

assets/src/application/ui/layout/entityList/filterBar/style.scss
```diff
@@ -68,6 +40,40 @@
 		}
 	}
 
+	&__btn {
+		&.esprs-button.ant-btn {
+			&#{$filter-btn} {
```
nicolad: this was used to increase specificity, more info: https://www.tobymackenzie.com/blog/2016/05/10/duplicate-selectors-increase-specificity/

assets/src/application/ui/layout/entityList/filterBar/style.scss
```diff
@@ -68,6 +40,40 @@
...
+		}
+		&__wrap {
+			align-items: center;
+			display: flex;
+			flex-flow: column nowrap;
+			padding: var(--ee-padding-tiny);
+			width: 90px;
+			#{$filter-btn} {
+				&#{$filter-btn} {
+					&.ee-active-filters {
```
nicolad: same here, this was used to increase specificity, more info: https://www.tobymackenzie.com/blog/2016/05/10/duplicate-selectors-increase-specificity/

***
there might be a cleaner way to do this, dunno ATM

assets/src/application/ui/layout/entityList/filterBar/buttons/FilterButtonWrap.tsx
```diff
@@ -5,7 +5,7 @@ import { FilterButtonWrapProps } from '../types';
 
 const FilterButtonWrap: React.FC<FilterButtonWrapProps> = ({ children, id, label }) => {
 	return (
-		<div className='ee-filter-bar-filter-btn-wrap'>
+		<div className='ee-entity-list-filter-bar__btn__wrap'>
```
nicolad: I am not a BEM master, hope this is ok :)

assets/src/application/ui/layout/entityList/filterBar/style.scss
tn3rb: that looks like a button within another button ?!?!?! is that right??

would this not be equivalent to:

```css
		&.esprs-button.ant-btn {
			.esprs-button.ant-btn.ee-entity-list-filter-bar__btn {
```

assets/src/application/ui/layout/entityList/filterBar/Collapsible.tsx
```diff
@@ -19,7 +19,7 @@ const Collapsible: React.FC<CollapsibleProps> = ({ children, show = false }) =>
 
 	return (
 		<animated.div style={props}>
-			<div className='ee-filter-bar-filter-collapsible' ref={ref}>
+			<div className='ee-entity-list-filter-bar__collapsible' ref={ref}>
```
tn3rb: BEM FTW 👍

assets/src/application/ui/layout/entityList/filterBar/style.scss
nicolad: based on my understanding, this would be the same:
```
&.esprs-button.ant-btn {
			&.esprs-button.ant-btn.ee-entity-list-filter-bar__btn {
```

nicolad: we need to find a better solution to fight antd specificity, because even if this duplication works I am not :100: happy with this approach since it's a bit convoluted

assets/src/application/ui/layout/entityList/filterBar/buttons/FilterButtonWrap.tsx
tn3rb: ya I'm certainly not a master either considering I just discovered this recently but I don't think I've ever seen nested elements like that. I think that would go against at least part of what BEM is trying to achieve.

it appears the above violation happens because you are merging two block names together

ie: entity list is one block and filter bar is another block vs the filter bar being an element of the entity list. 

yes you _could_ look at the filter bar being an element of the entity list, but it's probably far more accurate to view it as its own thing and therefore as a separate block.

So that said, if you remove `entity-list` from that selector, then it's easier to fit into the BEM schema:

- BLOCK: `ee-filter-bar`
- ELEMENT: `btn-wrap`
- MODIFIER: ???

with the resulting selector being: `ee-filter-bar__btn-wrap`

Then for the Entity List block the elements might be things like:

- `ee-entity-list__header` <== insert filter bar here
- `ee-entity-list__body` <== the actual entity list goes here (or maybe `ee-entity-list__list`)
- `ee-entity-list__footer` <== pagination and other buttons go here

A possible good rule of thumb for differentiating something as belonging to its own block might be to look if it is in its own folder or not. So in this case, despite the filter bar being within the `entityList` folder, it is also within its own `filterBar` folder and therefore should be considered its own block. I'm just making up that rule on the spot, it's not anything I have read anywhere so I am (as always) totally open for suggestions, etc.

assets/src/application/ui/layout/entityList/filterBar/buttons/TableView.tsx
```diff
@@ -0,0 +1,28 @@
...
+import classNames from 'classnames';
+import { __ } from '@wordpress/i18n';
+
+import { EspressoButton, EspressoButtonType, Icon } from '@application/ui/input';
+import FilterButtonWrap from './FilterButtonWrap';
+import { TableViewFilterButtonProps } from '../types';
+
+const TableViewFilterButton: React.FC<TableViewFilterButtonProps> = ({ listId, setTableView, view }) =>
+	React.useMemo(() => {
+		const className = classNames('ee-entity-list-filter-bar__btn', { 'ee-active-filters': view === 'table' });
```
tn3rb: ay some of these existing selector names are not very BEM friendly eh?

I think any refactoring pertaining to that is beyond the scope of this PR however and should therefore be done in another one.

assets/src/application/ui/layout/entityList/filterBar/style.scss
tn3rb: oh right you had another `&` in there... but even so, that would essentially evaluate to the equivalent of:

```css
.esprs-button.ant-btn.esprs-button.ant-btn.ee-entity-list-filter-bar__btn
```

which is just a bit redundant....
since `$filter-btn` equals '.esprs-button.ant-btn.ee-entity-list-filter-bar__btn';

then can't this just be:

```css
	&__btn {
		&#{$filter-btn} {
```

???

nicolad: yes, that is simpler, thx:
https://github.com/eventespresso/event-espresso-core/pull/2578/commits/2e6e34df811e691238954098c128353a7a2c71cd

assets/src/application/ui/layout/entityList/filterBar/buttons/FilterButtonWrap.tsx
nicolad: Improved that button wrap: https://github.com/eventespresso/event-espresso-core/pull/2578/commits/ca40c5f1c1823b1dabfbcb818385d48b128ffb4c

assets/src/application/ui/layout/entityList/filterBar/buttons/TableView.tsx
nicolad: I've renamed `ee-entity-list-filter-bar` to `ee-filter-bar` and added modifiers

assets/src/application/ui/layout/entityList/filterBar/buttons/CardView.tsx
```diff
@@ -9,7 +9,7 @@ import { CardViewFilterButtonProps } from '../types';
 
 const CardViewFilterButton: React.FC<CardViewFilterButtonProps> = ({ listId, setCardView, view }) =>
 	React.useMemo(() => {
-		const className = classNames('ee-entity-list-filter-bar__btn', { 'ee-active-filters': view === 'card' });
+		const className = classNames('ee-filter-bar__btn', { 'ee-filter-bar__btn--active': view === 'card' });
```
tn3rb: aaaa yeah 👍

Issue #1022, opened 2019-03-14
Remove link to event's checkins when context is already viewing event's checkins


joshfeck:
Originally reported by @Pebblo in https://events.codebasehq.com/projects/event-espresso/tickets/9368

Note in the screenshot the outlined text:
https://monosnap.com/file/3iJIimoy5UKzdihZbKbY01riRsgwjw#

The outlined text is a hyperlink that goes to a list of checkins for that event. It's a helpful link if you navigate to the "all events" Event Check-In tab by clicking on the tab from the "All registrations from all events" page.

Where it's not helpful is when you're already viewing a page of checkins for the event. 

Could we either a) remove the column entirely when on that view or b) remove the link and leave the text?

tn3rb:
should just remove the column from that view since it has no additional information to add and therefore no purpose at that point

joshfeck:
resolved in #2579

PR #2579, opened 2020-03-20
Remove link to event's checkins when context is already viewing event's checkins

Pebblo:
See: https://github.com/eventespresso/event-espresso-core/issues/1022

Rather than jumping in checking for an evt_id in various places, I moved the current check for adding the checkbox column, below the default columns array and then removed the 'Event' column from within that conditional.

See #1022 

## How has this been tested
Go to Event Espresso -> Registrations -> Event Check-in

You should see an 'event' column shown on that page.

Click on the event name to view check-ins for that event, on that page, you should not see the event link/column

## Checklist

* [x] My code is tested.


joshfeck:
Very nice indeed.

Issue #1021, opened 2019-03-14
Make the *Event*->Registrations->'Event Check-in' tab contextual


joshfeck:
Originally reported by @Pebblo in https://events.codebasehq.com/projects/event-espresso/tickets/8520

> Go to Event Espresso -> Events -> Event -> Registrations.
> 
> So that lists the registrations specific to THAT event. Now click on the 'Event Check-in' tab.
> 
> Your taken to a page that lists all registrations for all events. 
> 
> You need to select the event from the dropdown and filter those registrations.
> 
> Can we make that tab go directly to check-ins for the event we are viewing registrations for?

joshfeck:
This would go to the same URL that the "View Checkins" link that's displayed when you hover over a row in the Event Espresso > Registration admin list table.

PR #2582, opened 2020-03-20
Make the Event Check-in tab contextual

Pebblo:
See: #1021

This makes the 'Event Check-in' tab contextual, meaning if you are viewing registrations for a specific event and you click on that tab, it opens checks for the event in question.

The nav URL value has a hardcoded 'action' because I don't think we can pull it any other way as its set right in that section, if no EVT_ID it uses null, which then defaults to the 'normal' link for all check-ins.

Would it be worth altering how the page config [nav url's are generated](https://github.com/eventespresso/event-espresso-core/blob/master/core/admin/EE_Admin_Page.core.php#L1478-L1483) here so we can pass values to them? I've seen `'extra_request'` passed around in a few places, something similar here that can be passed into the query_args?

## How has this been tested
Go to Event Espresso -> Registrations.

Confirm that the 'Event Check-in' tab works as expected and show all check-ins.

Back to the registration list, open up a specific events registrations list.

Now confirm that clicking on the Event Check-in link lists the checkings for the specific event you were viewing.

Now, on the check-in list, click on the 'Event Check-in' tab again, this time it should load all check-ins for all events again.

## Checklist

* [x] My code is tested.

### DIFF
--- a/caffeinated/admin/extend/registrations/Extend_Registrations_Admin_Page.core.php
+++ b/caffeinated/admin/extend/registrations/Extend_Registrations_Admin_Page.core.php
@@ -49,6 +49,9 @@ class Extend_Registrations_Admin_Page extends Registrations_Admin_Page
         $reg_id = ! empty($this->_req_data['_REG_ID']) && ! is_array($this->_req_data['_REG_ID'])
             ? $this->_req_data['_REG_ID']
             : 0;
+        $EVT_ID = ! empty($this->_req_data['event_id']) && ! is_array($this->_req_data['event_id'])
+            ? $this->_req_data['event_id']
+            : 0;
         $new_page_routes = array(
             'reports'                      => array(
                 'func'       => '_registration_reports',
--- a/caffeinated/admin/extend/registrations/Extend_Registrations_Admin_Page.core.php
+++ b/caffeinated/admin/extend/registrations/Extend_Registrations_Admin_Page.core.php
@@ -119,7 +119,7 @@ class Extend_Registrations_Admin_Page extends Registrations_Admin_Page
                     'label'      => esc_html__('Event Check-In', 'event_espresso'),
                     'order'      => 10,
                     'persistent' => true,
-                    'url'        => !empty($EVT_ID)
+                    'url'        => !empty($EVT_ID) && $this->get_current_view() !== 'event_registrations'
                         ? self::add_query_args_and_nonce(array('action' => 'event_registrations', 'event_id' => $EVT_ID), $this->_admin_base_url)
                         : null,
                 ),
--- a/caffeinated/admin/extend/registrations/Extend_Registrations_Admin_Page.core.php
+++ b/caffeinated/admin/extend/registrations/Extend_Registrations_Admin_Page.core.php
@@ -119,6 +119,9 @@ class Extend_Registrations_Admin_Page extends Registrations_Admin_Page
                     'label'      => esc_html__('Event Check-In', 'event_espresso'),
                     'order'      => 10,
                     'persistent' => true,
+                    'url'        => !empty($EVT_ID)
+                        ? self::add_query_args_and_nonce(array('action' => 'event_registrations', 'event_id' => $EVT_ID), $this->_admin_base_url)
+                        : null,
                 ),
                 'help_tabs'     => array(
                     'registrations_event_checkin_help_tab'                       => array(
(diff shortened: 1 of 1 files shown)


tn3rb:
re:

> Would it be worth altering how the page config nav url's are generated  here so we can pass values to them? I've seen 'extra_request' passed around in a few places, something similar here that can be passed into the query_args?

ya you could check for a "query_args" key or something instead of "url" in the `$config['nav']` array.
that code you linked to ould have to call a new function that ran a series of if-else clauses because you wouldn't be able to do it properly inline.

ex:

```php
	$this->_nav_tabs[ $slug ] = array(
		'url'       => $this->parseNavTabURL(),
		'link_text' => isset($config['nav']['label'])
		// other code

	// somewhere else in the class
    protected function parseNavTabURL()
    {
		if ( isset($config['nav']['url']) ) {
			return $config['nav']['url'];
		}

		$query_args = isset($config['nav']['query_args']) )
			? $config['nav']['query_args'] // may be better to use the args on the next line as a default and then merge these into it
			: array('action' => $slug);

		return self::add_query_args_and_nonce(
			$query_args,
			$this->_admin_base_url
		);
	}
```

Pebblo:
Thanks, Brent.

I went through with this and ended up with:

```php
protected function parseNavURL()
    {
        if (isset($config['nav']['url'])) {
            return isset($config['nav']['url']);
        }

        $query_args = array('action' => $slug);
        
        if (! empty($config['nav']['query_args'])) {
            $query_args = array_merge($query_args, $config['nav']['query_args']);
        }

        return self::add_query_args_and_nonce(
            $query_args,
            $this->_admin_base_url
        );
    }
```

Problem is, I don't think we can assume we always want every tab to ignore query_args if we are already on that view so then we still need to check for both an `EVT_ID` and `get_current_view()` and end up with something like this:

```php
'query_args' => !empty($EVT_ID) && $this->get_current_view() !== 'event_registrations'
                        ? array($EVT_ID)
                        : null`
```

I know I could just pass EVT_ID and loop through query_args using only non-empty values etc but its starting to feel like a solution to a problem that doesn't really need solving, or I'm looking at this wrong?

tn3rb:
sry for the totally late response... this PR slipped under my radar...

re:
> I don't think we can assume we always want every tab to ignore query_args if we are already on that view

right, so that's something that you either have to:

- decide upon at the source before calling `parseNavURL()` from `Extend_Registrations_Admin_Page`, similar to what you are doing now in this PR, and then only pass the query args you want, and then also pass `$query_args` to `parseNavURL()` and fallback to a default if nothing is recieved, ex:


```php
    /**
     * @param array      $config
     * @param array      $query_args 
     * @param string     $slug
     * @return bool|string
     * @since $VID:$
     */
    protected function parseNavURL(array $config, array $query_args , $slug)
    {
        if (isset($config['nav']['url'])) {
            return isset($config['nav']['url']);
        }

        $query_args = ! empty($query_args) ? $query_args : ['action' => $slug];

        return EE_Admin_Page::add_query_args_and_nonce(
            $query_args,
            $this->_admin_base_url
        );
    }
```

AND/OR

- somehow incorporate into the logic of `parseNavURL()` so that the decision can be made there. For example, you could supply an array of `Callables` (ie: callback functions) that `parseNavURL()` could loop through and pass the  query args to for modification, so maybe something like:


```php
    /**
     * @param array      $config
     * @param array      $query_args 
     * @param string     $slug
     * @param callable[] $filters
     * @return bool|string
     * @since $VID:$
     */
    protected function parseNavURL(array $config, array $query_args , $slug, array $filters = [])
    {
        if (isset($config['nav']['url'])) {
            return isset($config['nav']['url']);
        }

        $query_args = ! empty($query_args) ? $query_args : ['action' => $slug];

        foreach ($filters as $filter) {
            $query_args = is_callable($filter) ? $filter($query_args) : $query_args; 
        }

        return EE_Admin_Page::add_query_args_and_nonce(
            $query_args,
            $this->_admin_base_url
        );
    }
```

of course, if you wanted to make the query arg modification open to the global scope, you could use a WP filter, whereas the above would restrict it to classes that can call `parseNavURL()`, ie:


```php
    /**
     * @param array      $config
     * @param array      $query_args 
     * @param string     $slug
     * @return bool|string
     * @since $VID:$
     */
    protected function parseNavURL(array $config, array $query_args , $slug)
    {
        if (isset($config['nav']['url'])) {
            return isset($config['nav']['url']);
        }

        $query_args = ! empty($query_args) ? $query_args : ['action' => $slug];

        $query_args = apply_filters(
			'FHEE__EE_Admin_Page_parseNavURL__query_args',
			$query_args,
			$config,
			$slug
		);

        return EE_Admin_Page::add_query_args_and_nonce(
            $query_args,
            $this->_admin_base_url
        );
    }
```

PR #2587, opened 2020-03-23
Add eventStatus schema

joshfeck:

This adds one new property to the event structured data: eventStatus. Event Espresso doesn't have an equivalent event status for every provided event status option, so only "postponed", "cancelled", and "scheduled" will be supported for now. 

Note: Existing filter hooks could be used to alter ld+json output to show eventStatus as "EventMovedOnline" or "EventRescheduled".

## How has this been tested

- [x] Check the page source of a published event page, it should now include '"eventStatus": "https://schema.org/EventScheduled",' within the ld+json structured data
- [x] Set an event to Postponed, and check its page source, it should now include "eventStatus": "https://schema.org/EventPostponed",'
- [x] Set an event to Cancelled status, and check its page source, it should now include "eventStatus": "https://schema.org/EventCancelled",'
- [x] Test the ld+json with the Structured data testing tool available at https://search.google.com/structured-data/testing-tool/u/0/

## Checklist

* [x] My code is tested.

### DIFF
--- a/core/helpers/EEH_Schema.helper.php
+++ b/core/helpers/EEH_Schema.helper.php
@@ -50,15 +50,14 @@ class EEH_Schema
         $template_args['event_end'] = $primary_datetime->end_date(DateTime::ATOM);
         unset($primary_datetime);
         switch ($event->status()) {
-            case 'cancelled':
+            case EEM_Event::cancelled:
                 $event_status = 'EventCancelled';
                 break;
-            case 'postponed':
+            case EEM_Event::postponed:
                 $event_status = 'EventPostponed';
                 break;
             default:
                 $event_status = 'EventScheduled';
-                break;
         }
         $template_args['event_status'] = $event_status;
         $template_args['currency'] = EE_Registry::instance()->CFG->currency->code;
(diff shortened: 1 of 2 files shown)


core/helpers/EEH_Schema.helper.php
```diff
@@ -48,6 +49,18 @@ public static function add_json_linked_data_for_event(EE_Event $event)
...
+            case 'cancelled':
+                $event_status = 'EventCancelled';
+                break;
+            case 'postponed':
+                $event_status = 'EventPostponed';
+                break;
+            default:
+                $event_status = 'EventScheduled';
+                break;
+        }
```
tn3rb: It's best to use constants for the values returned for status because it ensures they are valid
and you don't _need_ a break for `default` if it is the last case in the switch
(doesn't cause any problems if left in however, but can be removed)

```suggestion
        switch ($event->status()) {
            case EEM_Event::cancelled:
                $event_status = 'EventCancelled';
                break;
            case EEM_Event::postponed:
                $event_status = 'EventPostponed';
                break;
            default:
                $event_status = 'EventScheduled';
        }
```

sethshoultes:
This looks good to me:
image

image

Issue #2595, opened 2020-03-24
Fix Tickets List Filter Bar isChained Filter


tn3rb:
### steps to reproduce issue

- open filters for dates list

- try different **status** filter options until you find an option that gives either 1 single datetime as a result or zero datetimes

- open filters for ticket list

- click on new **status** filter and select "all tickets for all dates"

- take note of what tickets are showing and total number of tickets (we don't actually display that number but number of pages will suffice)

- click on the "**isChained**" filter button beside the **status** filter

- the **status** filter options _should_ change to "all tickets for above dates only" (or something like that) and the number of tickets in the entity list should change accordingly to reflect only those related to the dates in the dates list (zero tickets if zero dates are displayed)

### other issues

- the "**isChained**" filter button should default to being chained to the dates list (ie: default value === true)

- let's add a label above the button similar to how all of the other filters have a label (ex: status, sales, etc). This label can simply be "link"

- the hover state for the "**isChained**" filter button also needs tweaking when the button is active. It needs the fill color for the icon to be set to `var(--ee-color-white)`

**isChained** filter button hover state:

![isChained button hover state](https://user-images.githubusercontent.com/1751030/77450663-d6954580-6db0-11ea-9fe3-8898a21f6861.png)

### mobile view

- on small screens the filters all look really good, but could be expanded to 100% width

- the isChained button however loses its tooltip because there is no hover state for touch devices

- let's duplicate the tooltip text for ALL Espresso Buttons and add it to the DOM after the button in a div that only gets displayed on screens less than 480px. In fact, it may be best to create an adapter for AntD's `Tooltip` component and add this logic to that

here's an image showing what I'm thinking:

![isChained button on mobile](https://user-images.githubusercontent.com/1751030/77452746-88ce0c80-6db3-11ea-9ba1-7c7c3347f62d.png)

plz feel free to split up into more than one PR

nicolad:
will try to reproduce, but atm chain tooltip is showing:

![image](https://user-images.githubusercontent.com/1477453/77473433-24946280-6e1e-11ea-9de7-261f2c54fefa.png)

PR #2602, opened 2020-03-24
Fix Tickets List Filter Bar isChained Filter; closess #2595.

nicolad:
- Closes: https://github.com/eventespresso/event-espresso-core/issues/2595


assets/src/application/ui/layout/entityList/filterBar/style.scss
```diff
@@ -85,6 +106,10 @@ $filter-btn: '.esprs-button.ant-btn.ee-filter-bar__btn';
 		white-space: pre;
 		width: auto;
 
+		@include max782px {
+			width: 100%;
+		}
```
nicolad: tooltip needs to be excluded here

assets/src/application/ui/input/EspressoButton/Tooltip.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+import { isEmpty } from '@appServices/utilities/string';
+
+interface TicketsChainedButtonProps extends TooltipPropsWithTitle {
+	button: JSX.Element;
+	buttonText?: string;
+	tooltip?: string;
+}
+
+const Tooltip: React.FC<TicketsChainedButtonProps> = ({ button, buttonText, tooltip, ...props }) => {
```
nicolad: still WIP

assets/src/application/ui/input/EspressoButton/Tooltip.tsx
```diff
@@ -0,0 +1,35 @@
+import React from 'react';
+import classNames from 'classnames';
+import { isMobile } from 'react-device-detect';
```
tn3rb: hmmmm... if we have to go to this extent to do this then maybe we shouldn't bother...

I assumed things could be hidden solely using CSS but hadn't considered nesting and such...

That said, we should also consider what Gutenberg has since that will most likely already be available:
https://github.com/WordPress/gutenberg/tree/master/packages/viewport

and as a final thought, WordPress (at least in the admin) adds a `mobile` class to the `body` tag for small screens (although i don't know what pixel value that starts at) and it's super easy to detect that without having to load an entire library to do something that has essentially already been done:

```js
const body = document.getElementsByTagName("BODY")[0];
const isMobile = body  ? body.classList.contains('mobile') : false;
```

that would make a nice little utility, or could even be dumped into our config data

assets/src/application/ui/layout/entityList/filterBar/style.scss
tn3rb: maybe not... it won't do anything on mobile since there is no hover state right?

assets/src/domain/eventEditor/ui/tickets/ticketsList/filterBar/controls/TicketsChainedButton.tsx
```diff
@@ -15,18 +15,19 @@ interface TicketsChainedButtonProps {
...
 	const icon = isChained ? Icon.LINK : Icon.UNLINK;
+	const label = __('link');
 	const tooltip = isChained
 		? __('tickets list is linked to dates list and is showing tickets for above dates only')
 		: __('tickets list is unlinked and is showing tickets for all event dates');
 
 	return (
-		<>
-			<label className='esprs-button-label screen-reader-text'>{__('Only Show Tickets for Dates Above')}</label>
+		<BaseInput label={label}>
```
tn3rb: i thought of the perfect name for this... convert it to an HOC and call it `withLabel` and then wrap the input exports at the source (or the imports in the implementation)

assets/src/application/ui/input/EspressoButton/Tooltip.tsx
```diff
@@ -0,0 +1,35 @@
...
+	buttonText?: string;
+	tooltip?: string;
+}
+
+const Tooltip: React.FC<TicketsChainedButtonProps> = ({ button, buttonText, tooltip, ...props }) => {
+	const mobileTooltip = isEmpty(buttonText) ? (
+		<div>
+			{button}
+			<div className='esprs-button__tooltip'>{tooltip}</div>
+		</div>
```
tn3rb: if you give this wrapper div a classname then you could use flexbox to position things inside it

assets/src/application/ui/input/EspressoButton/Tooltip.tsx
```diff
@@ -0,0 +1,35 @@
...
+		button
+	);
+
+	return isMobile ? (
+		mobileTooltip
+	) : (
+		<DefaultTooltip title={tooltip} {...props}>
+			{button}
+		</DefaultTooltip>
+	);
```
tn3rb: sorry this ended up being a bit more convoluted than expected...
couldn't the mobile "tooltip" (might be better to call it `mobileHelpText`) just be appended onto the button with `display:none` set for any screen above 480px?

assets/src/domain/eventEditor/ui/tickets/ticketsList/filterBar/controls/TicketsChainedButton.tsx
nicolad: I've added `withLabel` HOC : https://github.com/eventespresso/event-espresso-core/pull/2602/commits/a6fa54c21355a004cb90cf4ceb3f6f17b9797f02

nicolad: since `withLabel` needs `label` prop which is not passed to `EspressoButton`, we can't just `wrap the exports at the source`

nicolad: TBH this approach with HOC is a bit convoluted.
A simpler solution I'd see is to just pass `label` to `EspressoButton` and deal at that level or even simpler, create a component `EspressoButtonWithLabel` which is not a HOC but a wrapper around `EspressoButton` which accepts `label` along those `EspressoButtonProps`

assets/src/application/ui/layout/entityList/filterBar/style.scss
```diff
@@ -48,7 +48,7 @@ $filter-btn: '.esprs-button.ant-btn.ee-filter-bar__btn';
 			margin: 0;
 			&:focus,
 			&:hover {
-				border: 1px solid var(--ee-color-primary-low-contrast);
+				border: 2px solid var(--ee-color-primary-low-contrast);
```
nicolad: This change was needed because of this issue:

Peek 2020-03-25 14-32

assets/src/application/ui/layout/entityList/filterBar/style.scss
nicolad: yes, you're right, nothing has been changed, this is how it looks:

image

nicolad: without `width: 100%`

image

assets/src/application/ui/input/EspressoButton/Tooltip.tsx
nicolad: finished

assets/src/application/ui/input/EspressoButton/Tooltip.tsx
nicolad: thx, have used `@wordpress/viewport`

nicolad: I don't want to rely on a class on body element, the reason being that:
1. WP can update that name and break our functionality
2. it's quite easy to alter those classnames, with jQuery or native JS

nicolad: `@wordpress/viewport` gives us more control and we are not affected by outside changes of classnames and such

assets/src/application/ui/input/EspressoButton/Tooltip.tsx
nicolad: done, thx

assets/src/application/ui/input/EspressoButton/Tooltip.tsx
nicolad: updated, but I can't apply css since we need a different HTML structure to display differently on mobile

image

assets/src/application/ui/display/viewport/index.js
```diff
@@ -0,0 +1,2 @@
+export { default as ifViewportMatches } from './if-viewport-matches';
+export { default as withViewportMatch } from './with-viewport-match';
```
nicolad: the reason why this package was added is because of this error:
image
***

I removed the code that wasn't needed and was causing the problems

assets/src/application/ui/display/viewport/with-viewport-match.js
```diff
@@ -0,0 +1,60 @@
+/**
+ * External dependencies
+ */
+import { mapValues } from 'lodash';
```
manzoorwanijk: Why not kill `lodash`?

assets/src/application/ui/input/EspressoButton/Tooltip.tsx
```diff
@@ -0,0 +1,36 @@
...
+import { Tooltip as DefaultTooltip } from 'antd';
+import { TooltipPropsWithTitle } from 'antd/lib/tooltip';
+import { withViewportMatch } from '@appDisplay/viewport';
+import { __ } from '@wordpress/i18n';
+
+import { isEmpty } from '@appServices/utilities/string';
+import './style.scss';
+
+interface TicketsChainedButtonProps extends TooltipPropsWithTitle {
+	button: JSX.Element;
```
manzoorwanijk:
```suggestion
	button: React.ReactNode;
```

assets/src/application/ui/input/withLabel/index.tsx
```diff
@@ -0,0 +1,21 @@
...
+		<div className={'ee-base-input'}>
+			{label && (
+				<label className='ee-base-input-label' htmlFor={id}>
+					{label}
+				</label>
+			)}
+			{Component}
+		</div>
+	);
+};
```
manzoorwanijk: That `Component` is a bit misleading there, its rendered as `{Component}` which means it's a node not a component.

```suggestion
const withLabel = (output: React.ReactNode): React.FC<Props> => {
	const Component: React.FC<Props> = ({ id, label } ) => {
		return (
			<div className={'ee-base-input'}>
				{label && (
					<label className='ee-base-input-label' htmlFor={id}>
						{label}
					</label>
				)}
				{output}
			</div>
		);
	}

	return Component;
};
```

assets/src/application/ui/display/viewport/with-viewport-match.js
nicolad: agree,
I have just copied over this package and wanted first to know if you both are ok with pulling this

assets/src/application/ui/input/withLabel/index.tsx
nicolad: as I wrote, I am not :100: happy with using HOC in this specific case

assets/src/application/ui/input/EspressoButton/Tooltip.tsx
nicolad: got failing tests and had to revert back, sorry

assets/src/application/ui/input/EspressoButton/Tooltip.tsx
tn3rb: great points, i agree

assets/src/application/ui/layout/entityList/filterBar/style.scss
tn3rb: that looks much better... think I prefer the 100% width fields for mobile.

assets/src/application/ui/display/viewport/with-viewport-match.js
tn3rb: ya let's not use lodash

assets/src/application/ui/layout/entityList/filterBar/buttons/FilterButtonWrap.tsx
```diff
@@ -1,17 +1,18 @@
...
-	return (
-		<div className='ee-filter-bar__btn-wrap'>
-			{children}
+const FilterButtonWrap: React.FC<FilterButtonWrapProps> = ({ children, id, isMobile, label }) => (
+	<div className='ee-filter-bar__btn-wrap'>
+		{children}
+		{!isMobile && (
 			<label className='ee-filter-btn-label' htmlFor={id}>
 				{label}
 			</label>
```
tn3rb: why are you removing this label on mobile?
isn't this for adding the labels like "table view", "card view", etc ?

assets/src/application/ui/layout/entityList/filterBar/style.scss
tn3rb: ok... might have to apply that change in a bunch of other places too

assets/src/application/ui/input/EspressoButton/Tooltip.tsx
nicolad: ended up cloning `@wordpress/viewport` HOC, because of an error related to jest

assets/src/application/ui/layout/entityList/filterBar/style.scss
```diff
@@ -73,6 +79,21 @@ $filter-btn: '.esprs-button.ant-btn.ee-filter-bar__btn';
...
+	&__chain {
+		&--active {
+			&#{$filter-btn} {
+				&:focus,
+				&:hover {
+					svg {
+						path {
+							fill: var(--ee-color-white);
+						}
+					}
```
tn3rb: the icons still have a white background for their focus state (but change to a blue border), so using this style for `:focus` would essentially make the icon disappear

```suggestion
				&:hover {
					svg {
						path {
							fill: var(--ee-color-white);
						}
					}
```

assets/src/application/ui/display/viewport/with-viewport-match.js
nicolad: replaced with `ramda` alternative

assets/src/application/ui/display/viewport/with-viewport-match.js
```diff
@@ -0,0 +1,60 @@
...
+			// a hook that calls other hooks always in the same order (because the query never changes).
+			// eslint-disable-next-line react-hooks/rules-of-hooks
+			return useViewportMatch( breakpointName, operator );
+		} );
+	return createHigherOrderComponent( ( WrappedComponent ) => {
+		return pure( ( props ) => {
+			const queriesResult = useViewPortQueriesResult();
+			return <WrappedComponent { ...props } { ...queriesResult } />;
+		} );
+	}, 'withViewportMatch' );
```
tn3rb: this could probably simplified by removing `createHigherOrderComponent` (which only appends the HOC name to the display name for the wrapped component) and `pure` which is just a form of caching by performing a shallow equals check on the porps before rendering. We _may_ want to add the latter to this component directly, but we could do so just using memoization.

assets/src/application/ui/input/withLabel/index.tsx
```diff
@@ -0,0 +1,25 @@
...
+	id?: string;
+	label?: string;
+}
+
+const withLabel = (output: React.ReactNode): React.FC<Props> => {
+	const Component: React.FC<Props> = ({ id, label } ) => {
+		return (
+			<div className={'ee-base-input'}>
+				{label && (
+					<label className='ee-base-input-label' htmlFor={id}>
```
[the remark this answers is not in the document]
nicolad: done: https://github.com/eventespresso/event-espresso-core/pull/2602/commits/524539501f42548bea54efdfa77e72ca0bc3d9ac

assets/src/application/ui/input/BaseInput/style.scss
```diff
@@ -1,5 +1,7 @@
-.ee-base-input {
```
nicolad: will remove `.ee-base-input` reference when `BaseInput` will be removed

assets/src/application/ui/layout/entityList/filterBar/buttons/FilterButtonWrap.tsx
nicolad: without that we'll have this doubling label situation;

image

image

assets/src/application/ui/display/viewport/with-viewport-match.js
nicolad: have simplified, thx

assets/src/application/ui/input/BaseInput/style.scss
nicolad: `BaseInput` have been removed and replaced with `withLabel`

assets/src/application/ui/layout/entityList/filterBar/buttons/FilterButtonWrap.tsx
tn3rb: ahhh ok... ya I get it now, didn't realize the extra label was being added to those too

assets/src/application/ui/input/EspressoButton/Tooltip.tsx
```diff
@@ -0,0 +1,36 @@
...
+	buttonText?: string;
+	isMobile?: boolean;
+	tooltip?: string;
+}
+
+const Tooltip: React.FC<TicketsChainedButtonProps> = ({ button, buttonText, isMobile, tooltip, ...props }) => {
```
nicolad: this wasn't named `EspressoTooltip` because it's used only internally to `EspressoButton`

PR #2608, opened 2020-03-25
EDTR React.lazy/Suspense

manzoorwanijk:
This PR prepares EDTR to make use of `React.lazy` and `React.Suspense`


tn3rb:
I experimented with this a bit and was able to push the lazy loading down into the dateslist so that the filter bar and other aspects of the entity list load immediately and you only have to wait for the actual entities. It's not anywhere close to perfect as the filters are not yet able to provide the list with the filtered entities.  I also had to bypass some of the TS because it was giving me grief.

For the filters, I'm wondering if it's possible for a component that is nested within a context provider to update a value on that provider for others to utilize. Maybe have the component feeding the provider save the filtered entities in state, and pass that list to the provider as well as a setter for updating the state.

So checkout the `EDTR/lazy-suspense-2-edtr-boogaloo` branch and see if you can move that along even further.

Let's not spend too much time on this now though because it would probably be more advantageous to put time into other optimizations first.

Here's a screenie of just the Dates List lazy loading (warning: this event has a LOT Of dates and takes a while to load)

Dates List Lazy Loading

tn3rb:
@manzoorwanijk 

Pushed all of my changes to branch `EDTR/lazy-suspense-2-edtr-boogaloo`.

The last commit was an attempt to get the dates list to show a loading notice when you change the filters (for example changing the status filter from "all upcoming" to "expired only") but it just doesn't work because the query to get the dates and filter them happens to fast and any perceived delay is occurring solely during the actual rendering of the new date cards.

So plz feel free to reset that branch to the second to last commit and/or just take whatever you want from that branch.

manzoorwanijk:
Here is the demo of the latest changes.
demo

assets/src/domain/eventEditor/ui/datetimes/datesList/DatesList.tsx
```diff
@@ -0,0 +1,49 @@
...
+		<Suspense fallback={<LoadingNotice />}>
+			<CardView />
+		</Suspense>
+	);
+
+	const RenderTableView = () => (
+		<Suspense fallback={<LoadingNotice />}>
+			<TableView />
+		</Suspense>
+	);
```
manzoorwanijk, line 29: This is here temporarily. It can be extracted to separate components.

nicolad:
I don't remember seeing this error before:

image
***
will check later on master branch

assets/src/application/services/utilities/memo/getPropsAreEqual.ts
```diff
@@ -0,0 +1,22 @@
...
+
+type PropsAreEqual<P extends {}> = (
+	prevProps: Readonly<React.PropsWithChildren<P>>,
+	nextProps: Readonly<React.PropsWithChildren<P>>
+) => boolean;
+
+/**
+ * Generates the comparison function that can be used as second argument to React.memo()
+ */
+const getPropsAreEqual = <P extends {}>(pathToValue: Parameters<typeof path>[0]): PropsAreEqual<P> => {
```
nicolad, line 12: is there a comparison measurement with and without this function?

assets/src/application/ui/layout/entityList/filterBar/filterState/useEntityListFilterStateManager.ts
```diff
@@ -20,66 +20,81 @@ const useEntityListFilterStateManager = <SortBy = BasicSortBy>(defaultSortBy: So
...
-		});
-	};
+	const setSortBy: FSM['setSortBy'] = useCallback(
+		(sortBy) => {
+			dispatch({
+				type: 'SET_SORT_BY',
+				sortBy,
+			});
+		},
+		[dispatch]
```
nicolad, line 30: why `dispatch` is set as dep, but not `sortBy`?

assets/src/domain/eventEditor/services/context/EntityListContext/DatetimesListContext.tsx
```diff
@@ -1,14 +1,32 @@
...
+
+	const filtersStr = JSON.stringify(filters);
+	// memoize filter state
+	const filterState = useMemo(() => filters, [filtersStr]);
+
+	const filteredEntitiesList = useFilteredEntities(domain, datesList, datetimes, filterState);
+
+	const cacheIds = JSON.stringify(getCacheIds(filteredEntitiesList));
+	// memoize filtered entities
+	const filteredEntities = useMemo(() => filteredEntitiesList, [cacheIds]);
```
nicolad, line 27: :+1:

assets/src/domain/eventEditor/services/context/EntityListContext/TicketsListContext.tsx
```diff
@@ -1,14 +1,32 @@
...
+	const filterState = useMemo(() => filters, [filtersStr]);
+
+	const filteredEntitiesList = useFilteredEntities(domain, ticketsList, datetimes, filterState);
+
+	const cacheIds = getCacheIds(filteredEntitiesList).join(':');
+	// memoize filtered entities
+	const filteredEntities = useMemo(() => filteredEntitiesList, [cacheIds]);
+
+	const value: TicketsListContextProps = { filterState, filteredEntities };
+
```
nicolad, line 30: this piece looks like that one from the dates list, is it worth of extracting this?

assets/src/application/ui/layout/entityList/filterBar/filterState/useEntityListFilterStateManager.ts
manzoorwanijk: `sortBy` is not in outer scope 😄

assets/src/domain/eventEditor/ui/datetimes/datesList/cardView/DateCard.tsx
```diff
@@ -68,6 +65,6 @@ const DateCard: React.FC<EntityListItemProps> = React.memo(({ id }) => {
 			/>
 		</DatetimeProvider>
 	) : null;
-});
+}, getPropsAreEqual(['entity', 'cacheId']));
```
nicolad, line 68: might be a bit more readable:
`export default React.memo(DateCard, getPropsAreEqual(['entity', 'cacheId']))`

assets/src/domain/eventEditor/services/context/EntityListContext/TicketsListContext.tsx
manzoorwanijk: It does look like that, may be we can extract it later.

assets/src/domain/eventEditor/ui/datetimes/datesList/cardView/DateCard.tsx
manzoorwanijk: agreed 👍

assets/src/application/services/utilities/memo/getPropsAreEqual.ts
manzoorwanijk: `React.memo` does not necessarily prevent re-renders, it's just a performance optimization. I haven't measured the difference.

assets/src/domain/eventEditor/ui/datetimes/datesList/cardView/DateCard.tsx
manzoorwanijk: Done

assets/src/application/ui/layout/entityList/filterBar/EntityListFilterBar.tsx
```diff
@@ -25,15 +25,11 @@ const EntityListFilterBar = <FS extends ELFSM>({
...
-		setCardView,
-		setTableView,
-		setSearchText,
-		showEntityFilters,
-		toggleEntityFilters,
-		view,
-	} = filterState;
+
+	const [showEntityFilters, setShowEntityFilters] = useState(false);
+	const toggleEntityFilters = () => setShowEntityFilters((v) => !v);
```
manzoorwanijk, line 30: Local state for toggle to improve performance.

assets/src/domain/eventEditor/services/context/EntityListContext/DatetimesListContext.tsx
tn3rb, line 27: would be better to move some of this memoization into `useFilteredEntities()` because then it will apply to both the dates list, tickets list, and any other entity list we build in the future.

So for example this code:

```js
	const cacheIds = JSON.stringify(getCacheIds(filteredEntitiesList));
	// memoize filtered entities
	const filteredEntities = useMemo(() => filteredEntitiesList, [cacheIds]);
```

could go into into `useFilteredEntities()` and automatically return a memoized entity list, so the code here in `DatetimesListContext` would just be

```js
const filteredEntities = useFilteredEntities(domain, datesList, datetimes, filterState);
```

but it would still be better to memoize the filterstate here since it's not really used elsewhere and will likely always be accessed via this context provider.

Might also be handy to return the stringified entity list cacheIDs for use in other components so that they don't have to recreate that logic like:

```js
const { entityListCacheKey, filteredEntities } = useFilteredEntities(domain, datesList, datetimes, filterState);
```

assets/src/domain/eventEditor/ui/datetimes/datesList/cardView/DateCard.tsx
```diff
@@ -6,23 +6,20 @@ import DateDetails from './DateDetails';
...
-import useDatetimeItem from '@edtrServices/apollo/queries/datetimes/useDatetimeItem';
 import statusBgColorClassName from '@sharedEntities/datetimes/helpers/statusBgColorClassName';
 
 import EntityCard from '@appLayout/EntityCard';
-import { EntityListItemProps } from '@appLayout/entityList';
+import type { EntityListItemProps } from '@appLayout/entityList';
 import { InlineEditHeading, InlineEditTextArea } from '@appInputs/InlineEditInput';
 import { useDatesListFilterState } from '@edtrServices/filterState';
 import { useDatetimeMutator } from '@edtrServices/apollo/mutations';
+import type { Datetime } from '@edtrServices/apollo/types';
```
tn3rb, line 16: what's the benefit/reason for doing this?

assets/src/domain/eventEditor/ui/datetimes/datesList/DatesList.tsx
tn3rb: let's just do that now then plz since it seems to be relevant to the scope of this PR

core/domain/services/admin/events/editor/AdvancedEditorData.php
```diff
@@ -169,7 +169,9 @@ protected function getEditorData($eventId)
 
         $i18n = self::getJedLocaleData('event_espresso');
 
-        return compact('event', 'graphqlEndpoint', 'currentUser', 'generalSettings', 'i18n');
+        $assets_url = EE_PLUGIN_DIR_URL . 'assets/dist/';
+
+        return compact('event', 'graphqlEndpoint', 'currentUser', 'generalSettings', 'i18n', 'assets_url');
```
tn3rb, line 174:
```suggestion
        $assetsUrl = EE_PLUGIN_DIR_URL . 'assets/dist/';

        return compact('event', 'graphqlEndpoint', 'currentUser', 'generalSettings', 'i18n', 'assetsUrl');
```

assets/src/domain/eventEditor/interfaces/types.ts
```diff
@@ -13,6 +13,7 @@ export interface EventData {
 }
 
 export interface EEEditorData {
+	assets_url?: string;
```
tn3rb, line 16:
```suggestion
	assetsUrl?: string;
```

assets/src/domain/eventEditor/services/publicPath.ts
```diff
@@ -0,0 +1,14 @@
...
+@typescript-eslint/camelcase,
+@typescript-eslint/no-unused-vars,
+no-var
+*/
+/* global __webpack_public_path__ */
+declare var __webpack_public_path__: string;
+__webpack_public_path__ = window?.eeEditorData?.assets_url;
+
+// not needed
+export const assets_url = __webpack_public_path__;
```
tn3rb, line 14:
```suggestion
__webpack_public_path__ = window?.eeEditorData?.assetsUrl;

// not needed
export const assetsUrl = __webpack_public_path__;
```

assets/src/domain/eventEditor/ui/datetimes/datesList/cardView/DateCard.tsx
manzoorwanijk: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html#type-only-imports-exports

assets/src/domain/eventEditor/services/context/EntityListContext/DatetimesListContext.tsx
manzoorwanijk: I agree with moving that to `useFilteredEntities`, but not with the last suggestion - SRP.

assets/src/application/ui/layout/entityList/EntityCardList.tsx
```diff
@@ -14,9 +14,4 @@ const EntityCardList = <E extends Entity>({ EntityCard, entities }: EntityCardLi
 	);
 };
 
-export default React.memo(EntityCardList, (prevProps, nextProps) => {
-	const prevValue = JSON.stringify(getCacheIds(prevProps.entities));
-	const nextValue = JSON.stringify(getCacheIds(nextProps.entities));
-
-	return prevValue === nextValue;
-});
```
nicolad: :taco: :hamburger: :curry:

assets/src/application/services/utilities/memo/entityListCacheIdString.ts
```diff
@@ -0,0 +1,6 @@
+import { Entity } from '@appServices/apollo/types';
+import { getCacheIds } from '@appServices/predicates';
+
+const entityListCacheIdString = <E extends Entity>(entities: E[]): string => JSON.stringify(getCacheIds(entities));
```
nicolad, line 4: nice, atomic extraction :+1:

PR #2609, opened 2020-03-26
EDTR v1 Asset Purge

tn3rb:
Started to look into why css was not loading for production builds, and wanted to cut down on the amount of assets and bundles being built and loaded. All references to EDTR v1 assets have simply been commented out for the time being, as this is just an initial step to see what issues may arise.

You'll notice that some of the config data parsing was modified. This was because initially I forgot to keep the inline `eejs` data object as a dependency for the EDTR assets so the config data was undefined. This tool a while to figure out however, because the resulting errors didn't point to the config data at all and looked like a problem with asset paths.


.npmrc
```diff
@@ -1 +1,2 @@
-save-exact = false
\ No newline at end of file
+save-exact = false
+scripts-prepend-node-path = true
```
nicolad: this is the fix for this warning:

image

package.json
```diff
@@ -136,6 +136,7 @@
 	},
 	"devDependencies": {
 		"@apollo/react-testing": "^3.1.3",
+		"@babel/compat-data": "^7.9.0",
```
nicolad: this is an attempt to fix this error:

image

nicolad: based on this article: https://github.com/angular/angular-cli/issues/17262
but this can be wrong

package.json
```diff
@@ -245,5 +244,8 @@
 		"webpack-bundle-analyzer": "^3.6.1",
 		"webpack-cli": "^3.3.11",
 		"webpack-merge": "^4.2.2"
+	},
+	"resolutions": {
+		"@babel/preset-env": "7.9.0"
```
nicolad: based on this: https://stackoverflow.com/questions/60780664/could-not-find-plugin-proposal-numeric-separator

package.json
```diff
@@ -47,19 +46,12 @@
 			"last 1 safari version"
 		]
 	},
-	"babel": {
-		"presets": [
-			"react-app"
-		]
-	},
 	"dependencies": {
```
nicolad: some of the deps were moved in dev deps, mainly because of:
```
The difference between these two, is that devDependencies are modules which are only required during development, while dependencies are modules which are also required at runtime.
```
more info: https://medium.com/@dylanavery720/npmmmm-1-dev-dependencies-dependencies-8931c2583b0c

nicolad:
Still having this issue:

image
***
which got me to this:
- https://stackoverflow.com/questions/58086396/typescript-runtime-error-cannot-read-property-length-of-undefined
- https://github.com/microsoft/TypeScript/issues/21801

nicolad:
> your changes look great
> what do you want to do next with this?
> just clean up all of the commented out code and try running with it?
> 
> we should maybe cut a branch from master just prior to merging this and call it something like `EDTR/old-webpack` just for reference


there is still failing build, need to fix that first and also check also that weird `Cannot read property 'end' of undefined` which is also happening now in https://github.com/eventespresso/event-espresso-core/pull/2623

assets/src/application/services/config/useConfigData.ts
```diff
@@ -18,29 +18,25 @@ import { JsDataProps, ConfigDataProps } from './types';
...
+			admin: data?.paths?.admin_url || '',
+			home: data?.paths?.site_url || '',
 		} as SiteUrlProps),
 		timezone: Timezone({
-			city: data.default_timezone && data.default_timezone.pretty ? data.default_timezone.pretty : '',
-			name: data.default_timezone && data.default_timezone.string ? data.default_timezone.string : '',
-			offset: data.default_timezone && data.default_timezone.offset ? data.default_timezone.offset : 0,
+			city: data?.default_timezone?.pretty || '',
+			name: data?.default_timezone?.string || '',
+			offset: data?.default_timezone?.offset || 0,
```
manzoorwanijk: Optional chaining makes the code cleaner

nicolad:
`Apparently in the latest versions of babel there are problems.`
https://stackoverflow.com/a/60792728/4094575

PR #2612, opened 2020-03-26
Fix EDTR timezone issues | Closes #2610

### LINKED ISSUE #2610: Timezone Issues
Here's a couple of screenshots showing:

### - the same dates in EDTR v2 (order is not the same)

If you look at the times for the "Tax Rate Seminar" date, you'll see that the start time in the legacy editor is 2pm but in the new EDTR it is 10am. 

Why?

Because the legacy editor is displaying times in the timezone set for the site but the EDTR is displaying times in my local timezone. My site's timezone is currently set to Eastern Daylight Time UTC+3 and my local timezone is currently Pacific Daylight Time UTC+7. This is why there is a 4 hour difference between the two.

So the requirements for this issue are simple:

- display **ALL** dates in the timezone set for the site

- save all dates to the database in UTC

... much, MUCH, **MUCH** easier said than done (well not really _that_ difficult, but kind of a constant game of whack-a-mole). 

I've recently added a new function `formatDateForSite()` to **useTimeZoneTime** (`assets\src\application\services\hooks\useTimeZoneTime.ts`) which may be useful for fixing this issue.

manzoorwanijk:
This PR fixes the timezone related issues for EDTR:
- Creates TZ converter and formatter functions in `useTimeZoneTime`
- Updates components inside `/application/ui/calendars` to use TZ formatting
- Changes the way dates are prepared before being sent to the server.
- Updates form inputs to use site timezone
- Closes #2610


assets/src/domain/shared/services/utils/dateAndTime.ts
```diff
@@ -3,32 +3,34 @@ import { formatISO, isValid, parse, parseISO } from 'date-fns';
...
+	endDate: string;
 }
 
-interface StartAndEndDate {
-	startDate: Date;
-	endDate: Date;
+export interface DateAndTime extends Partial<StartAndEndDate> {
+	startTime?: string;
+	endTime?: string;
 }
```
tn3rb: 👍

assets/src/application/services/hooks/useTimeZoneTime.ts
```diff
@@ -2,18 +2,22 @@ import { path } from 'ramda';
 import { parseISO, isValid } from 'date-fns';
 
 import useConfig from '../config/useConfig';
-import { localToUtc, utcToLocal } from '../utilities/date';
```
tn3rb: might as well delete these functions from here since they are not used anywhere else and are kinda useless without the timezone, which basically means you would have to duplicate the logic in `useTimeZoneTime.ts` in order to use the functions, which would be pointless.

PR #2614, opened 2020-03-26
Add Component for Switching Date Display in Entity Cards; closes #2598.

### LINKED ISSUE #2598: Add Component for Switching Date Display in Entity Cards
Currently the `DateCard` and `TicketCard` components are using the `CalendarDateRange` component for their sidebar content but this does not change views when the "**display**" filter option is changed.

- create a new component in `assets\src\application\ui\calendars\dateDisplay` called `CalendarDateSwitcher` (open to better ideas for name)

- implement logic similar to that in `assets\ZZZ\editor\events\dates-and-times\editor-date\grid-view\event-date-calendar-date.js` but abstract things so that it will work with any entity simply by providing the necessary props such as:

     - `showDate` (or `displayDate`)

     - `startDate`

     - `endDate`

     - `headerText` & `footerText`

     - `sidebarColorClass`

- utilize this new component in both `DateCard` and `TicketCard`

- wire up to the filter bar "**display**" filter and ensure that dates switch and display correctly as the filter changes

nicolad:
- The previous PR was closed after reading with `EDTR/master` and had to open this one.
- Closes https://github.com/eventespresso/event-espresso-core/issues/2598


assets/src/application/ui/calendars/dateDisplay/CalendarDateSwitcher/index.tsx
```diff
@@ -0,0 +1,79 @@
...
+import { DisplayStartOrEndDate } from '@sharedServices/filterState';
+import { PLUS_ONE_MONTH, PLUS_TWO_MONTHS } from '@sharedConstants/defaultDates';
+import { BiggieCalendarDate, CalendarDateRange } from '../';
+
+interface Props {
+	className?: string;
+	displayDate: DisplayStartOrEndDate;
+	endDate: string;
+	footerText?: string;
+	headerPrefix?: string;
```
tn3rb: let's remove this prop because we'll handle the text differently at the source

assets/src/application/ui/calendars/dateDisplay/CalendarDateSwitcher/index.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+const CalendarDateSwitcher: React.FC<Props> = React.memo(
+	({
+		className,
+		displayDate = DisplayStartOrEndDate.start,
+		footerText,
+		headerPrefix,
+		headerText,
+
+		...props
```
tn3rb:
```suggestion
		headerText,
		...props
```

assets/src/application/ui/calendars/dateDisplay/CalendarDateSwitcher/index.tsx
```diff
@@ -0,0 +1,79 @@
...
+import { PLUS_ONE_MONTH, PLUS_TWO_MONTHS } from '@sharedConstants/defaultDates';
+import { BiggieCalendarDate, CalendarDateRange } from '../';
+
+interface Props {
+	className?: string;
+	displayDate: DisplayStartOrEndDate;
+	endDate: string;
+	footerText?: string;
+	headerPrefix?: string;
+	headerText?: string;
```
tn3rb: what might work better for these text labels is to have a separate object with its own interface like:

```js
CalendarDateLabels {
	headerPast: string; // sale starts
	headerFuture: string; // sale started
	footerPast: string; // sale ends
	footerFuture: string; // sale ended
}
```

then we can create a utility for switching tense based on a supplied date like

```js
const headerText = switchTenseForDate( date, labels.headerPast, labels.headerFuture );
```

and then pass that along to the appropriate calendar component.

this way the `CalendarDateSwitcher` will not only handle the responsibility of selecting what component to use for the date display, but also what labels to use.

what do you think? (i'm open for suggestions)

as well, we can simplify this PR and then figure out the above in https://github.com/eventespresso/event-espresso-core/projects/6#card-35163011 since that's what that potential issue is supposed to cover.

assets/src/application/ui/calendars/dateDisplay/CalendarDateSwitcher/index.tsx
```diff
@@ -0,0 +1,79 @@
...
+				);
+			default:
+				return (
+					<BiggieCalendarDate
+						className={className}
+						date={startDate}
+						footerText={footerText}
+						headerText={`${headerPrefix || ''}${headerText || __('starts')}`}
+					/>
+				);
```
tn3rb: maybe just get rid of the `start` case since it is the exact same thing as the `default` case

but make sure that the `default` case gets the `showTime` prop set

assets/src/application/ui/calendars/dateDisplay/CalendarDateSwitcher/index.tsx
nicolad: it removed, not sure why it's showing

nicolad: seems to be outdated

image

assets/src/application/ui/calendars/dateDisplay/CalendarDateSwitcher/index.tsx
nicolad: this is outdated

assets/src/application/ui/calendars/dateDisplay/CalendarDateSwitcher/index.tsx
nicolad: I think it might be confusing to have `start` option in `DisplayStartOrEndDate`, but not in the switch case.
I have extracted that into a variable and passed for both `default` and `start`, IMHO it's a bit cleaner and safer.

assets/src/application/ui/calendars/dateDisplay/CalendarDateSwitcher/index.tsx
nicolad: let's handle this in that issue, because currently I am trying to fix/ understand failing tests :disappointed:

package.json
```diff
@@ -138,7 +138,7 @@
 		"@apollo/react-testing": "^3.1.3",
 		"@babel/plugin-proposal-class-properties": "^7.8.3",
 		"@babel/plugin-transform-runtime": "^7.8.3",
-		"@babel/preset-env": "^7.8.6",
+		"@babel/preset-env": "^7.9.0",
```
nicolad: The error was saying:
```
 Could not find plugin "proposal-numeric-separator". Ensure there is an entry in ./available-plugins.js for it. (While processing: "/home/travis/build/eventespresso/event-espresso-core/node_modules/babel-preset-react-app/index.js$0")
```
and I've updated this package to latest master where that plugin exists: 
https://github.com/babel/babel/blob/master/packages/babel-preset-env/src/available-plugins.js#L17

package.json
```diff
@@ -243,5 +243,8 @@
 		"webpack-assets-manifest": "^3.1.1",
 		"webpack-cli": "^3.3.11",
 		"webpack-merge": "^4.2.2"
+	},
+	"resolutions": {
```
nicolad: More info: https://classic.yarnpkg.com/en/docs/selective-version-resolutions/

package.json
```diff
@@ -137,11 +137,11 @@
...
 		"@babel/plugin-proposal-class-properties": "^7.8.3",
-		"@babel/plugin-transform-runtime": "^7.8.3",
+		"@babel/plugin-transform-runtime": "7.9.0",
 		"@babel/preset-env": "^7.9.0",
-		"@babel/preset-react": "^7.8.3",
-		"@babel/preset-typescript": "^7.8.3",
-		"@babel/runtime-corejs3": "^7.8.4",
+		"@babel/preset-react": "7.9.1",
+		"@babel/preset-typescript": "7.9.0",
+		"@babel/runtime": "7.9.0",
```
nicolad: these updates were made based on this:
- https://github.com/facebook/create-react-app/issues/8680
and
- https://github.com/facebook/create-react-app/pull/8681/files

nicolad: also I noticed they use `@babel/runtime`, not `@babel/runtime-corejs3`, updated that as well, will double check on ejected CRA

PR #2628, opened 2020-03-28
Refactor Button Tooltip and Label as HOCs

tn3rb:
Because of the improper use of hooks within our copy of the WP `withViewportMatch` component, we needed to find a different way to hide and display button labels and tooltips on mobile.

To start, I wanted to refactor the extract the Tooltip logic from EspressoButton cause SRP!!! As well, we have multiple implementations of input labels, so wanted to create a single definitive label component as well. I decided (maybe incorrectly) that these would work well as HOCs and could be used to decorate EspressoButton allowing it to be used on its own or in combination if need be.

Now here's the problem:

EspressoButton needs to be able to handle refs, and therefore any decorator HOCs need to use `React.forwardRef()`. I've tried my best but don't quite have enough Typescript Mojo-Fu to get things working correctly. TS gives me one error in each HOC (withLable && withTooltip) but compiles fine. Clicking on one of the buttons that is wrapped with the HOCs now crashes the EDTR (yeehaw) and provides a completely useless stack trace with zero references to any of our code, so I can't quite determine what's going wrong.

Anyways, was wondering if you could take a look at the two HOCs and see if you can resolve the TS issue with the refs.

Here's an article that guided my attempt to integrate the HOCs with TS: https://medium.com/@martin_hotell/react-refs-with-typescript-a32d56c4d315
which was linked to from here:
https://github.com/typescript-cheatsheets/react-typescript-cheatsheet/blob/master/README.md#forwardrefcreateref


assets/src/application/ui/display/withLabel/withLabel.tsx
```diff
@@ -0,0 +1,43 @@
...
+				</label>
+				<WrappedComponent {...(props as P)} ref={forwardedRef} />
+			</div>
+		) : (
+			<WrappedComponent {...(props as P)} ref={forwardedRef} />
+		);
+	};
+
+	const ForwardedComponentWithLabel = (props: P, ref: T) => {
+		return <WithLabel {...props} forwardedRef={ref} />;
```
tn3rb: this is the line where the TS error appears

assets/src/application/ui/input/EspressoButton/types.ts
```diff
@@ -21,16 +21,16 @@ export enum EspressoButtonType {
 	SECONDARY = 'SECONDARY',
 }
 
-export interface EspressoButtonProps extends Partial<NativeButtonProps> {
-	buttonProps?: object;
+export interface EspressoButtonProps
+	extends Partial<NativeButtonProps>,
+		Partial<withLabelProps>,
+		Partial<withTooltipProps> {
```
tn3rb: because label and tooltip props can be passed to the button but will get intercepted by the HOCs

assets/src/application/ui/display/withLabel/withLabel.tsx
manzoorwanijk: The mistake is that you have defined `forwardedRef` to be `React.RefObject<T>` in `WithLabel` props but here you are passing as just `T`.
```suggestion
	const ForwardedComponentWithLabel = (props: P, ref: React.RefObject<T>) => {
		return <WithLabel {...props} forwardedRef={ref} />;
```

assets/src/application/ui/display/withTooltip/withTooltip.tsx
```diff
@@ -0,0 +1,44 @@
...
+			<WrappedComponent {...(props as P)} aria-label={ariaLabel} ref={forwardedRef} tooltip={tooltip} />
+		);
+		return (
+			<DefaultTooltip {...tooltipProps} title={tooltip}>
+				{toolTipped}
+			</DefaultTooltip>
+		);
+	};
+
+	const ForwardedComponentWithTooltip = (props: P, ref: T) => {
```
manzoorwanijk:
```suggestion
	const ForwardedComponentWithTooltip = (props: P, ref: React.RefObject<T>) => {
```

assets/src/application/ui/input/index.ts
```diff
@@ -11,6 +11,6 @@ export { default as Text } from './Text';
 export { default as Hidden } from './Hidden';
 export { default as TextArea } from './TextArea';
 export { default as TimePicker } from './DateTimePicker/TimePicker';
-export { default as withLabel } from './withLabel';
```
manzoorwanijk: Since you have removed it from here, there are some consumer components like `SelectInput`, `SearchInput` etc. that will be affected.

assets/src/application/ui/input/EspressoButton/EspressoButton.tsx
```diff
@@ -1,20 +1,20 @@
...
+type ButtonType = React.ComponentType;
 
 /**
  * Button wrapper for adding styles
  *
  * forwardRef to be able to accept
  * onMouseEnter, onMouseLeave, onFocus, onClick events from parent
  */
-const EspressoButton = React.forwardRef<Button, EspressoButtonProps>(
+const EspressoButton = React.forwardRef<ButtonType, EspressoButtonProps>(
```
manzoorwanijk: This should be `Button` not `ButtonType`
```suggestion
const EspressoButton = React.forwardRef<Button, EspressoButtonProps>(
```

assets/src/application/ui/display/withLabel/withLabel.tsx
tn3rb: right... that makes sense, thnx

assets/src/application/ui/input/index.ts
tn3rb: those have all been updated

tn3rb:
Thanks for the TS fixes Manzoor.

Think I have fixed all of the responsive layouts and styles now and the new HOCs appear to be working well.

assets/src/application/ui/layout/descriptionList/styles.scss
```diff
@@ -1,71 +1,70 @@
...
-			display: grid;
-			grid-template-columns: max-content;
-			grid-template-rows: repeat(2, var(--ee-font-size-huge));
-			grid-auto-flow: column;
-			justify-content: start;
-			line-height: var(--ee-font-size-default);
-			place-items: center start;
-			margin-top: var(--ee-margin-small);
-			width: 100%;
+.ee-description-list {
```
tn3rb: changed this after doing a search for `ee-description-list` and getting nothing :(

assets/src/application/ui/layout/espressoDropdownMenu/Toggle.tsx
```diff
@@ -36,21 +36,21 @@ const Toggle: React.FC<Props> = ({ icon, isOpen, label, onToggle, toggleProps })
...
 			icon={icon}
-			onClick={(event: React.SyntheticEvent) => {
+			onClick={(event: React.MouseEvent) => {
 				onToggle(event);
 				if (mergedToggleProps.onClick) {
 					mergedToggleProps.onClick(event);
 				}
 			}}
-			onKeyDown={(event) => {
+			onKeyDown={(event: React.KeyboardEvent) => {
```
manzoorwanijk: Type safety for events is nice 👍

assets/src/application/ui/layout/descriptionList/styles.scss
nicolad: the:
```
.ee {
	&-description-list {
		&-grid {
```
was resulting in [`ee-description-list-grid`](https://github.com/eventespresso/event-espresso-core/blob/EDTR/master/assets/src/application/ui/layout/descriptionList/DescriptionList.tsx#L9), no styles has been applied to `ee-description-list`;
the change you've made is fine, since it's avoiding one nesting level :+1: 
***
I think I started `.ee {` namespacing when there was multiple blocks in the same stylesheet.

assets/src/application/ui/display/withTooltip/withTooltip.tsx
```diff
@@ -0,0 +1,67 @@
...
+	}) => {
+		const ariaLabel = isEmpty(buttonText) && !isEmpty(tooltip) ? tooltip : null;
+		let toolTipped: React.ReactNode;
+		if (showTooltipOnMobile) {
+			const className = classNames({
+				'ee-mobile-help-text': true,
+				'ee-mobile-help-text--short': tooltip.length < 25,
+				'ee-mobile-help-text--long': tooltip.length > 50,
+			});
+			tooltipProps = { ...tooltipProps, overlayClassName: 'ee-mobile-help-text__tooltip' };
```
nicolad, line 29: @tn3rb this is mutating the `tooltipProps` prop

PR #2630, opened 2020-03-30
EDTR Cache ID for memoization | Closes #2621

### LINKED ISSUE #2621: Add cacheID property to Entity Interface
- add `cacheID` property to Entity Interface

- do the following only when an entity is updated:

     -  remove existing `cacheID` property

     -  run entity through `JSON.stringify()`

     -  save that value as new `cacheID` property

- this value can now be used throughout the codebase for memoization purposes as a way to easily determine if an entity has been updated

ex:

```js
return useMemo( () => (
    <SomeComponent entity={entity} etc={etc} {...props} />
), [entity.cacheID] );
```

manzoorwanijk:
This PR adds readonly `cacheId` field to GQL schema to use for memoization.
- Adds `cacheId` to EDTR entities GQL queries
- Adds `getCacheIds` and `getGuids` predicates.
- Updates entity list and list item queries to use `cacheId` for memoization with new predicates
- Closes #2621


assets/src/domain/eventEditor/services/apollo/queries/priceTypes/usePriceTypes.ts
```diff
@@ -1,21 +1,24 @@
...
 	const options = usePriceTypeQueryOptions();
 	const { data } = useCacheQuery<PriceTypesList>(options);
 
-	const priceTypes = data?.espressoPriceTypes?.nodes || [];
+	const nodes = data?.espressoPriceTypes?.nodes || [];
+
+	const cacheIds = getCacheIds(nodes).join(':');
 
-	return include.length ? entitiesWithGuIdInArray(priceTypes, include) : priceTypes;
+	return useMemo(() => nodes, [cacheIds]);
```
manzoorwanijk: I removed the support for `include` param - SRP

assets/src/domain/eventEditor/services/apollo/queries/datetimes/useDatetimes.ts
```diff
@@ -1,12 +1,19 @@
...
 const useDatetimes = (): Array<Datetime> => {
 	const options = useDatetimeQueryOptions();
 	const { data } = useCacheQuery<DatetimesList>(options);
 
-	return data?.espressoDatetimes?.nodes || [];
+	const nodes = data?.espressoDatetimes?.nodes || [];
+
+	const cacheIds = getCacheIds(nodes).join(':');
+
+	return useMemo(() => nodes, [cacheIds]);
```
nicolad: is is possible to have `nodes` being changed while `cacheIds` remains the same?

assets/src/domain/eventEditor/services/apollo/queries/priceTypes/usePriceTypeForPrice.ts
```diff
@@ -21,8 +22,11 @@ const usePriceTypeForPrice = (priceId: EntityId): PriceType => {
 
 	// get the default price type object
 	const defaultPriceType = useDefaultPriceType();
-	let relatedPriceTypes = usePriceTypes(relatedPriceTypeIds);
-	relatedPriceTypes = relatedPriceTypeIds.length ? relatedPriceTypes : [];
+	const allPriceTypes = usePriceTypes();
+
+	const relatedPriceTypes = relatedPriceTypeIds.length
+		? entitiesWithGuIdInArray(allPriceTypes, relatedPriceTypeIds)
+		: [];
```
nicolad: maybe we can move this check and leave this clean?
```
 const entitiesWithGuIdInArray = <T extends Entity>(entities: T[], guidArray: EntityId[]): T[] => {
    if(guidArray.length === 0) return [];
	return filter((entity: T) => includes(entityGuId(entity), guidArray), entities);
};

const relatedPriceTypes = entitiesWithGuIdInArray(allPriceTypes, relatedPriceTypeIds);
```
?

assets/src/domain/shared/services/predicates/listPredicates.ts
```diff
@@ -0,0 +1,7 @@
+import { pluck } from 'ramda';
+
+import { Cacheable, Entity } from '@appServices/apollo/types';
+
+export const getGuids = pluck<keyof Pick<Entity, 'id'>>('id');
+
+export const getCacheIds = pluck<keyof Pick<Cacheable, 'cacheId'>>('cacheId');
```
nicolad: this is amazing, so much boilerplate code has been avoided here :taco:

nicolad: ok, not that much: `entity.map({id} => id)`
but anyway `pluck` is more cleaner and it has some interesting use cases we might use in the future:
```
R.pluck(0, [[1, 2], [3, 4]]);               //=> [1, 3]
R.pluck('val', {a: {val: 3}, b: {val: 5}}); //=> {a: 3, b: 5}
```

assets/src/domain/eventEditor/services/apollo/queries/datetimes/useDatetimes.ts
manzoorwanijk: No, `cacheId` is `uuid`.

core/domain/services/graphql/resolvers/FieldResolver.php
```diff
@@ -136,6 +138,25 @@ public function resolveId($source)
...
+     *
+     * @param mixed     $source
+     * @return string
+     * @since $VID:$
+     */
+    public function resolveCacheId($source)
+    {
+        $model_name = $this->model->item_name();
+        $ID         = $source->ID();
+        $uniqid     = uniqid();
```
nicolad: was about to ask why not `uniqueID` ?
but checked docs and it turns out this is PHP function name
https://www.php.net/manual/en/function.uniqid.php
:smiley:

assets/src/domain/eventEditor/services/apollo/queries/priceTypes/usePriceTypeForPrice.ts
manzoorwanijk: Done

assets/src/domain/shared/services/predicates/listPredicates.ts
tn3rb: I'm guessing that the way Manzoor has done this preserves the Types better than simply mapping a prop over to the returned array, so consuming code will know that it is specifically receiving an array of GUIDs or an array of CacheIDs as opposed to an array with some entity prop in it.

core/domain/services/graphql/resolvers/FieldResolver.php
```diff
@@ -136,6 +138,25 @@ public function resolveId($source)
...
+        // Since cacheId does not need to be globally unique
+        // $uniqid is sufficient, still adding the model name and ID
+        // may be we need/use them in future.
+        return "{$model_name}:{$ID}:{$uniqid}";
+    }
```
tn3rb: you can also pass a prefix to `uniqid()` for the first parameter and it's always advisable to set the second parameter to true, which increases the amount of randomness

```suggestion
    public function resolveCacheId($source)
    {
        $model_name = $this->model->item_name();
        $ID         = $source->ID();
        // Since cacheId does not need to be globally unique
        // $uniqid is sufficient, still adding the model name and ID
        // may be we need/use them in future.
        return uniqid("{$model_name}:{$ID}:", true);
    }
```

assets/src/domain/eventEditor/services/apollo/queries/datetimes/useDatetimes.ts
```diff
@@ -1,12 +1,19 @@
...
+import { getCacheIds } from '@sharedServices/predicates';
 
 const useDatetimes = (): Array<Datetime> => {
 	const options = useDatetimeQueryOptions();
 	const { data } = useCacheQuery<DatetimesList>(options);
 
-	return data?.espressoDatetimes?.nodes || [];
+	const nodes = data?.espressoDatetimes?.nodes || [];
+
+	const cacheIds = getCacheIds(nodes).join(':');
```
tn3rb: I noticed here: https://github.com/eventespresso/event-espresso-core/pull/2630/files#diff-8732518b6f67c185fb6198f60a1c3647R156

that you were using `:` in the cache IDs generated server side. 

Will that conflict with using the same character here to separate IDs?

maybe we should use `-` or `,` here ?

manzoorwanijk: It doesn't matter at all. Here is an array of `cacheId`s of the whole list that I convert to string. `.join(':')` is simpler than `JSON.stringify()`

PR #2633, opened 2020-03-30
Fix no prices after creating a ticket | Closes #2632 

### LINKED ISSUE #2632: Prices disappear when creating a ticket
Steps:
- Open TPC for a ticket to make sure it has prices
- Create a new ticket
- Wait for mutation to complete
- Open TPC for the same ticket again

Prices are gone

manzoorwanijk:
This PR fixes the issue of prices disappearing when creating a ticket which was happening because of Ticket ids not sorted when writing prices to the cache:
- Adds sorting to `ticketIn` and `datetimeIn` when writing prices and tickets cache respectively
- Updates `.map()` wherever needed to use `getGuid`.
- Closes #2632

### DIFF
--- a/assets/src/domain/eventEditor/hooks/entityListCacheUpdates/test/useUpdatePriceTypeList.test.ts
+++ b/assets/src/domain/eventEditor/hooks/entityListCacheUpdates/test/useUpdatePriceTypeList.test.ts
@@ -5,6 +5,7 @@ import { useCacheRehydration } from '@edtrServices/apollo/initialization';
 import useUpdatePriceTypeList from '../useUpdatePriceTypeList';
 import { usePriceTypeQueryOptions, usePriceTypes } from '@edtrServices/apollo/queries';
 import { ApolloMockedProvider } from '@edtrServices/context/TestContext';
+import { getGuids } from '@sharedServices/predicates';
 
 const timeout = 5000; // milliseconds
 describe('useUpdatePriceTypeList', () => {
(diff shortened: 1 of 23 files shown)


assets/src/domain/eventEditor/services/apollo/mutations/tickets/useOnCreateTicket.ts
```diff
@@ -14,7 +15,7 @@ const useOnCreateTicket = (): TicketMutationCallbackFn => {
 	const onCreateTicket = ({ proxy, datetimeIds, ticket, tickets, prices }: TicketMutationCallbackFnArgs): void => {
 		if (ticket.id) {
 			const { nodes = [] } = tickets;
-			const ticketIn = nodes.map(({ id }: Ticket) => id);
+			const ticketIn = getGuids(nodes).sort();
```
manzoorwanijk: This is the actual fix for prices

assets/src/domain/eventEditor/hooks/entityListCacheUpdates/test/useUpdatePriceTypeList.test.ts
```diff
@@ -59,7 +60,7 @@ describe('useUpdatePriceTypeList', () => {
 		);
 		await waitForUpdate({ timeout });
 
-		const cachedPriceTypeIds = cacheResult.current.map(({ id }) => id);
+		const cachedPriceTypeIds = getGuids(cacheResult.current);
```
tn3rb: 👍 great, we want to reuse logic as much as possible, even for tiny things like this (also think it reads better)

assets/src/domain/eventEditor/services/apollo/mutations/tickets/useOnCreateTicket.ts
tn3rb: wow... crazy that not sorting caused an issue

PR #2641, opened 2020-04-01
Fix confirm_email being automatically added to address info.

Pebblo:
Reported here: https://eventespresso.com/topic/confirm-email-address/

Confirmed on my site.

The new confirm_email system question is automatically added to the address question group (as it's a system question it is automatically added to either Personal info or Address Info).

This branch skips that on activation but also allows the email_confirm question to show up in the address info question group so users can fix this themselves.

We should probably remove that again and add some code to automatically remove that new system question from the group but this gets a quick fix out the door.

## How has this been tested
On a site that is NOT running 4.10.5 already, update to this branch.

Enable the address question on an event and confirm that the confirm email question doesn't show in that group when registering.

---

On a site already running 4.10.5, before switching to this branch if you don't have an event with address info enabled, enable it and open the registration form.

You'll see the 'confirm email' question in that group on the front end.

Switch to this branch and retest, the confirm email question should no longer show.

## Checklist

* [ ] My code is tested.

### DIFF
--- a/core/helpers/EEH_Activation.helper.php
+++ b/core/helpers/EEH_Activation.helper.php
@@ -1096,7 +1096,7 @@ class EEH_Activation implements ResettableInterface
                     $QST_ID = $wpdb->insert_id;
 
                     // skip adding email_confirm to a question group
-                    if ( $QST_system === 'email_confirm' ) {
+                    if ($QST_system === 'email_confirm') {
                         continue;
                     }
 
--- a/core/helpers/EEH_Activation.helper.php
+++ b/core/helpers/EEH_Activation.helper.php
@@ -1094,6 +1094,12 @@ class EEH_Activation implements ResettableInterface
                         array('%s', '%s', '%s', '%s', '%d', '%s', '%d', '%d', '%d', '%d')
                     );
                     $QST_ID = $wpdb->insert_id;
+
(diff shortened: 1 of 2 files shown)


core/helpers/EEH_Activation.helper.php
```diff
@@ -138,6 +138,7 @@ public static function initialize_db_content()
         EEH_Activation::insert_default_status_codes();
         EEH_Activation::generate_default_message_templates();
         EEH_Activation::create_no_ticket_prices_array();
+        EEH_Activation::removeEmailConfirmFromAddressGroup();
```
tn3rb, line 141: separating into own function is smart 👍

core/helpers/EEH_Activation.helper.php
```diff
@@ -898,19 +899,15 @@ public static function initialize_system_questions()
...
-            'address2',
-            'city',
-            'country',
-            'state',
-            'zip',
-            'phone',
+        // all system questions
+        $personal_system_group_questions = ['fname', 'lname', 'email'];
+        $address_system_group_questions = ['address', 'address2', 'city', 'country', 'state', 'zip', 'phone'];
+        $system_questions_not_in_group = ['email_confirm'];
```
tn3rb, line 905: good catch 👍

Issue #2639, opened 2020-03-31
EDTR crashes when clicking dropdown menu


manzoorwanijk:
Click on dates dropdown menu for example.

![image (1)](https://user-images.githubusercontent.com/18226415/78063898-8587e180-73ae-11ea-9efd-ffd41f64d91c.png)

manzoorwanijk:
Fixed via #2681

PR #2649, opened 2020-04-02
Refactor Dropdown Menu to Not Use WP GB 

tn3rb:
plz see: #2639 

Because the WordPress GB Dropdown menu crashes due to improper hook use, this PR attempts to create a keyboard & a11y friendly dropdown menu using existing AntD components

Current issue is the inability to set focus to first menu item when dropdown opens


nicolad:
for simplicity maybe we can use this dep: https://reacttraining.com/reach-ui/menu-button
and decide later what we want

manzoorwanijk:
Completed via #2681

PR #2654, opened 2020-04-06
Fix filter bar buttons tooltip | Closes #2652

### LINKED ISSUE #2652: Filter Bar Table View and Card View Button Tooltips Do Not Close
If you click on either of the Table View or Card View buttons in either of the filter bars, the tooltips that appear when you first click on them do not disappear after the view changes and you are left with something like this:

I'm guessing it is a memoization issue

manzoorwanijk:
This PR fixes the issue of filter bar buttons tooltip not closing. The issue was because of the current view button getting disabled thus preventing `focus` and `blur` events.

### DIFF
--- a/assets/src/application/ui/layout/entityList/filterBar/buttons/CardViewFilterButton.tsx
+++ b/assets/src/application/ui/layout/entityList/filterBar/buttons/CardViewFilterButton.tsx
@@ -6,8 +6,9 @@ import { __ } from '@wordpress/i18n';
 import { CardViewFilterButtonProps } from '../types';
 import { EspressoButton, EspressoButtonType } from '@application/ui/input';
 import { LabelPosition } from '@application/ui/display';
+import { getPropsAreEqual } from '@appServices/utilities';
 
-const CardViewFilterButton: React.FC<CardViewFilterButtonProps> = React.memo(({ listId, setCardView, view }) => {
+const CardViewFilterButton: React.FC<CardViewFilterButtonProps> = ({ listId, setCardView, view }) => {
 	const className = classNames('ee-filter-bar__btn', { 'ee-filter-bar__btn--active': view === 'card' });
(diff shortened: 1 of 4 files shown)


assets/src/application/ui/layout/entityList/filterBar/buttons/ToggleLegendButton.tsx
```diff
@@ -19,7 +20,7 @@ const ToggleLegendButton: React.FC<ToggleLegendButtonProps> = ({ listId, showLeg
 		</animated.div>
 	);
 	const filterId = `ee-toggle-legend-btn-${listId}`;
-	const tooltip = __(`${showLegend ? 'hide' : 'show'} legend`);
+	const tooltip = showLegend ? __('hide legend') : __('show legend');
```
manzoorwanijk, line 23: @nicolad please note that translation functions cannot handle variables. If you want use variables, then use `sprintf`.

assets/src/application/ui/layout/entityList/filterBar/buttons/CardViewFilterButton.tsx
```diff
@@ -6,25 +6,25 @@ import { __ } from '@wordpress/i18n';
...
 	return (
 		<EspressoButton
 			buttonType={EspressoButtonType.MINIMAL}
 			className={className}
-			disabled={view === 'card'}
 			icon={<AppstoreFilled />}
 			id={filterId}
 			label={__('card view')}
-			onClick={setCardView}
+			onClick={view !== 'card' ? setCardView : null}
```
nicolad, line 22: :+1:

assets/src/application/ui/layout/entityList/filterBar/buttons/ToggleLegendButton.tsx
nicolad: ok, thx for explanation

PR #2663, opened 2020-04-07
Add registrationCount to TicketRegistrationsLink | Closes #2658

### LINKED ISSUE #2658: Add RegCount Number Badge to RegLink Icons for Tickets and use in Table View
- in PR #2657 Manzoor added a `registrationCount` property to tickets 

- instead of displaying this as a separate number in the Entity Details Panel or in the Table View, let's simply add this as as a "badge" to the registrations link similar to how the ticket relations are shown for this date in the image below

- then add the reg link to the ticket table view instead of just showing the reg count on its own

manzoorwanijk:
This PR makes the following changes:
- Enhances `ItemCount` component to add `emphasizeZero` prop
- Adds `registrationCount` to `TicketRegistrationsLink`
- Adds `TicketRegistrationsLink` to `TableView`

`emphasizeZero={false}` vs `emphasizeZero={true}`
image

### DIFF
--- a/assets/src/application/ui/display/ItemCount/index.tsx
+++ b/assets/src/application/ui/display/ItemCount/index.tsx
@@ -15,9 +15,9 @@ interface ItemCountProps extends BadgeProps {
 const ItemCount: React.FC<ItemCountProps> = ({
 	children,
 	count,
-	emphasizeZero = true,
+	emphasizeZero,
 	title = ' ',
-	zeroCountChar = '!',
+	zeroCountChar,
 	...props
 }) => {
 	const className = classNames(props.className, 'ee-item-count', {
@@ -25,7 +25,7 @@ const ItemCount: React.FC<ItemCountProps> = ({
 		'ee-item-count--no-items': count === 0 && emphasizeZero,
 	});
 	const offset = props.offset || [7, 7];
-	const value = count > 0 ? count : zeroCountChar;
+	const value = count === 0 && typeof zeroCountChar !== 'undefined' ? zeroCountChar : count;
 
 	const countNode = (
 		<Tooltip title={title}>
--- a/assets/src/application/ui/display/ItemCount/index.tsx
+++ b/assets/src/application/ui/display/ItemCount/index.tsx
@@ -7,14 +7,22 @@ import { getPropsAreEqual } from '@appServices/utilities';
 import './style.scss';
 
 interface ItemCountProps extends BadgeProps {
-	zeroCountChar?: string | JSX.Element;
 	children: React.ReactNode;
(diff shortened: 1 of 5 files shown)


assets/src/application/ui/display/ItemCount/index.tsx
```diff
@@ -7,14 +7,22 @@ import { getPropsAreEqual } from '@appServices/utilities';
 import './style.scss';
 
 interface ItemCountProps extends BadgeProps {
-	zeroCountChar?: string | JSX.Element;
 	children: React.ReactNode;
+	emphasizeZero?: boolean;
+	zeroCountChar?: string | JSX.Element;
```
nicolad, line 12: :tada: 
alpha beta gamma **iota**

assets/src/application/ui/display/ItemCount/index.tsx
```diff
@@ -7,14 +7,22 @@ import { getPropsAreEqual } from '@appServices/utilities';
...
+	count,
+	emphasizeZero = true,
+	title = ' ',
+	zeroCountChar = '!',
+	...props
+}) => {
 	const className = classNames(props.className, 'ee-item-count', {
 		'ee-item-count--has-items': count > 0,
-		'ee-item-count--no-items': count === 0,
+		'ee-item-count--no-items': count === 0 && emphasizeZero,
```
nicolad, line 25: just to help me understand, what Is the purpose behind emphasizing zero?

assets/src/domain/eventEditor/ui/tickets/TicketRegistrationsLink.tsx
```diff
@@ -26,19 +27,22 @@ const TicketRegistrationsLink: React.FC<Props> = ({ ticket }) => {
...
-			<a
-				href={regListUrl}
-				className={'ee-editor-details-reg-url-link'}
-				target={'_blank'}
-				rel={'noopener norefferer'}
-			>
-				<EspressoIcon icon={Icon.GROUPS} svgSize={24} />
-			</a>
-		</Tooltip>
+		<ItemCount count={ticket.registrationCount} title={countTitle} zeroCountChar={'0'} emphasizeZero={false}>
```
nicolad, line 34: let's switch default to `false`, so when we'll have to enable it we will pass just like this, this way we will make props more succint:
```
<ItemCount count={ticket.registrationCount} title={countTitle} zeroCountChar={'0'} emphasizeZero>
```

assets/src/application/ui/display/ItemCount/index.tsx
manzoorwanijk: See the screenshot I have attached

Issue #2673, opened 2020-04-08
Replace Other AntD UI Display Components with Chakra UI Equivalents 


tn3rb:
#### Alert
- `\domain\eventEditor\ui\ticketAssignmentsManager\components\ErrorMessage.tsx`
- replace with Chakra UI Alert

#### Avatar
- `\application\ui\display\Avatar.tsx`
- replace with Chakra UI Avatar

#### Badge
- `\application\ui\display\Avatar.tsx`
- `\application\ui\display\ItemCount\index.tsx`
- replace with custom component created using some of the logic and css from `\components\ui\icon-menu-item` in the `EDTR/refactor-master` branch

#### Collapse/Panel
- `\application\ui\display\DebugInfo.tsx`
- replace with Chakra UI Collapse

#### Empty
- `\application\ui\display\EmptyState.tsx`
- replace with custom composition similar to what's done here: [Chakra UI Alert](https://chakra-ui.com/alert#composition) but using the subcomponents listed in the Props section directly (ie: Box, Flex, Icon) with the icon being something like:
```css
<Icon name="question-outline" size='96px' color='gray.100'/>
```

#### Popconfirm
- `\application\ui\layout\confirmDelete\ConfirmDelete.tsx`
- replace with Chakra UI AlertDialog

#### Popover
- `\application\ui\calendars\dateDisplay\TimezoneTimeInfo\index.tsx`
- replace with Chakra UI Popover

#### Result
- `\application\ui\display\ErrorIndicator.tsx`
- replace with same custom component used to replace `Empty`, but use the warning icon with `red.500` (until we get a theme set up)

#### Spin
- `\application\services\toaster\useToastManager.tsx`
- `\application\ui\display\LoadingIndicator.tsx`
- `\application\ui\display\loadingNotice\LoadingNotice.tsx`
- replace with Chakra UI Spinner

#### Typography.Paragraph, Typography.Text, & Typography.Title
- `\application\ui\display\EmptyState.tsx`
- `\application\ui\input\InlineEditInput\Editable.tsx`
- `\application\ui\layout\entityList\EntityList.tsx`
- replace with Chakra UI Textarea, Text, & Heading respectively

### steps:

- add adapters in `\infrastructure\ui\display`
- update uses listed above
- submit as separate PRs for each component

nicolad:
`Editable` component was refactored via: https://github.com/eventespresso/event-espresso-core/pull/2697

Issue #2672, opened 2020-04-07
Replace AntD Tooltip with Chakra UI Tooltip


tn3rb:
#### known uses:

- `\application\ui\calendars\dateDisplay\CalendarPageDate`
- `\application\ui\calendars\dateDisplay\TimezoneTimeInfo`
- `\application\ui\display\ItemCount`
- `\application\ui\display\withTooltip`
- `\application\ui\forms\espressoForm\renderers\FieldRenderer.tsx`
- `\domain\eventEditor\ui\datetimes\DateRegistrationsLink.tsx`
- `\domain\eventEditor\ui\tickets\TicketRegistrationsLink.tsx`

#### steps:

- add adapter in `\infrastructure\ui\display`
- update `withTooltip` to use adapter 
- update other uses listed above to use `withTooltip` (where possible and not overly complex)

PR #2636, opened 2020-03-31
Add Confirmation Modals to Event Handlers for Destructive Processes; closes #2613.

nicolad:
## Problem this Pull Request solves
- Closes: https://github.com/eventespresso/event-espresso-core/issues/2613

PR #2682, opened 2020-04-08
Replace Other AntD UI Display Components with Chakra UI Equivalents #2673

nicolad:
- Closes https://github.com/eventespresso/event-espresso-core/issues/2673


assets/src/application/ui/display/EmptyState.tsx
```diff
@@ -12,16 +10,24 @@ interface EmptyStateProps {
...
+			status='warning'
+			variant='subtle'
+			textAlign='center'
+		>
+			<AlertIcon name='question-outline' size='96px' color='gray.100' />
+
+			<AlertTitle mt={4} mb={1} fontSize='lg'>
+				{title}
+			</AlertTitle>
+			{description && <AlertDescription maxWidth='sm'>{description}</AlertDescription>}
```
nicolad, line 27: image

assets/src/application/ui/layout/confirmDelete/ConfirmDelete.tsx
```diff
@@ -1,19 +1,56 @@
...
 
-interface ConfirmDeleteProps extends Omit<PopconfirmProps, 'title'> {
+interface ConfirmDeleteProps {
+	buttonProps: EspressoButtonProps;
+	onConfirm: VoidFunction;
 	title?: string;
 }
 
-const ConfirmDelete: React.FC<ConfirmDeleteProps> = ({ children, ...props }) => {
+const ConfirmDelete: React.FC<ConfirmDeleteProps> = ({ buttonProps, onConfirm, ...props }) => {
```
nicolad, line 20: Peek 2020-04-08 23-06

nicolad: The other confirm component is in this PR: https://github.com/eventespresso/event-espresso-core/pull/2636/files#diff-c9c1763fc52e041df1c8a1448b30f578R7
***
since that depends on new models being implemented, the tightening up will happen in that PR

assets/src/application/ui/calendars/dateDisplay/TimezoneTimeInfo/index.tsx
```diff
@@ -33,14 +42,19 @@ const TimezoneTimeInfo: React.FC<OffsetInfoProps> = ({ className, date, inline =
...
 	return (
-		<Popover content={content} title={__('This Date Converted To:')} trigger={'click'} align={{ offset: [0, -60] }}>
-			<div className={htmlClassName}>
-				<Tooltip title={__('click for timezone\ninformation')} mouseEnterDelay={1}>
-					<GlobalOutlined />
-				</Tooltip>
-			</div>
+		<Popover>
+			<PopoverTrigger>
+				<IconButton aria-label={label} icon={GlobalOutlined} variant='unstyled' />
```
nicolad, line 50: tried wrapping this with a tooltip, but got an error:
image
***
seems like `PopoverTrigger` children need to be either `Button` or `IconButton` or smth that can use `forwardRef`, tried `EspressoButton` but it didn't worked

assets/src/application/ui/calendars/dateDisplay/TimezoneTimeInfo/index.tsx
```diff
@@ -13,7 +22,7 @@ export interface OffsetInfoProps {
 	inline?: boolean;
 }
 
-const TimezoneTimeInfo: React.FC<OffsetInfoProps> = ({ className, date, inline = false }) => {
+const TimezoneTimeInfo: React.FC<OffsetInfoProps> = ({ date, inline = false }) => {
```
nicolad, line 25: Peek 2020-04-09 00-40

nicolad: @tn3rb notice `IconButton` focus state, I am not :100: happy with that, but we need to have a global styling applied for that and for other similar buttons

assets/src/application/ui/calendars/dateDisplay/TimezoneTimeInfo/index.tsx
```diff
@@ -1,8 +1,17 @@
...
+	Popover,
+	PopoverTrigger,
+	PopoverContent,
+	PopoverHeader,
+	PopoverBody,
+	PopoverArrow,
+	PopoverCloseButton,
+} from '@chakra-ui/core';
 import { __ } from '@wordpress/i18n';
 import { GlobalOutlined } from '@ant-design/icons';
```
nicolad, line 13: I didn't found similar icon in chakra's lib, so left the andt until will find another

assets/src/application/ui/display/ErrorIndicator.tsx
```diff
@@ -1,7 +1,32 @@
...
-import { ResultProps } from 'antd/lib/result';
+import { Alert, AlertIcon, AlertTitle, AlertDescription, IAlert } from '@chakra-ui/core';
 
-const ErrorIndicator: React.FC<ResultProps> = (props) => <Result status='warning' {...props} />;
+interface Props extends IAlert {
+	description?: string;
+	title?: string;
+}
+
+const ErrorIndicator: React.FC<Props> = ({ description, title, ...props }) => {
```
nicolad, line 9: image

assets/src/application/ui/display/loadingNotice/LoadingNotice.tsx
```diff
@@ -1,33 +1,26 @@
...
 import classNames from 'classnames';
-import { Spin } from 'antd';
+import { Spinner } from '@appDisplay/index';
 import { _x, sprintf } from '@wordpress/i18n';
 
 import { LoadingNoticeProps } from './types';
 import './style.scss';
 
-const LoadingNotice: React.FC<LoadingNoticeProps> = ({ children, loading, size = 'large', ...props }) => {
+const LoadingNotice: React.FC<LoadingNoticeProps> = ({ children, size = 'lg', ...props }) => {
```
nicolad, line 9: Here is the label being exposed to screen readers:
image

nicolad:
`Badge` and `Editable` is not refactored in this PR. will be done in follow-up PR.

assets/src/application/ui/calendars/dateDisplay/BiggieCalendarDate/index.tsx
```diff
@@ -37,7 +38,7 @@ export const BiggieCalendarDate: React.FC<BiggieCalendarDateProps> = ({
 	if (!isValid(dateObject)) {
 		return null;
 	}
-	className += ' ee-biggie-calendar-date__wrapper';
```
nicolad, line 40: we shall not mutate props

assets/src/application/ui/calendars/dateDisplay/TimezoneTimeInfo/index.tsx
```diff
@@ -1,19 +1,25 @@
...
 
-export interface OffsetInfoProps {
-	className?: string;
+export interface Props {
 	date: Date;
-	inline?: boolean;
 }
 
-const TimezoneTimeInfo: React.FC<OffsetInfoProps> = ({ className, date, inline = false }) => {
+const TimezoneTimeInfo: React.FC<Props> = ({ date }) => {
```
nicolad, line 22: image

assets/src/application/ui/display/DebugInfo.tsx
```diff
@@ -1,15 +1,14 @@
...
+import { Button, Collapse } from '@chakra-ui/core';
+import styled from '@emotion/styled';
+
+const Pre = styled.pre`
+	border-radius: 5px;
+	box-sizing: border-box;
+	padding: 1em 2em;
+	color: #a9ce47;
+	background-color: #26203d;
+`;
```
nicolad, line 11: these styles were applied taking into account that this component is rendered only in dev env

assets/src/application/ui/display/EmptyState.tsx
```diff
@@ -12,16 +10,26 @@ interface EmptyStateProps {
 
 const EmptyState = ({ className, children, description, title }: EmptyStateProps) => {
```
nicolad: component className has been applied
image

nicolad:
got some weird failing tests, will need to investigate closer:

image

assets/src/infrastructure/ui/display/Alert/Error.tsx
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+import { Alert, AlertIcon, AlertTitle, AlertDescription } from '@chakra-ui/core';
+
+const Error = ({ description, title }) => {
```
manzoorwanijk, line 4: No props defined

assets/src/infrastructure/ui/display/Alert/index.tsx
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+import { IAlert } from '@chakra-ui/core';
+
+import Error from './Error';
+
+interface Props extends IAlert {
+	description: string;
+	title: string;
+}
```
manzoorwanijk, line 9: Lets use a specific name for props for all the components.

```suggestion
interface AlertProps extends IAlert {
	description: string;
	title: string;
}
```

assets/src/infrastructure/ui/display/Avatar.tsx
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+import { Avatar as ChakraAvatar, AvatarBadge, IAvatar, BoxProps } from '@chakra-ui/core';
+
+interface Props extends IAvatar {
```
manzoorwanijk, line 4:
```suggestion
interface AvatarProps extends IAvatar {
```

assets/src/infrastructure/ui/display/Avatar.tsx
```diff
@@ -0,0 +1,26 @@
...
+	badgeProps?: BoxProps;
+	userName?: string;
+}
+
+const Avatar = ({ badgeProps, userName, ...avatarProps }: Props) => {
```
manzoorwanijk, line 9:
```suggestion
const Avatar: React.FC<AvatarProps> = ({ badgeProps, userName, ...avatarProps }) => {
```

assets/src/infrastructure/ui/display/DebugInfo.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+	const dataToRender = asJson ? JSON.stringify(data, null, 2) : data;
+
+	const output = <Pre>{dataToRender}</Pre>;
+
+	if (!asCollapse) {
+		return output;
+	}
+
+	const handleToggle = () => setShow(!show);
```
manzoorwanijk, line 35: `useCallback`

assets/src/infrastructure/ui/display/EmptyState.tsx
```diff
@@ -0,0 +1,38 @@
...
+import { Alert, AlertIcon, AlertTitle, AlertDescription } from '@chakra-ui/core';
+
+interface EmptyStateProps {
+	children?: React.ReactNode;
+	className?: string;
+	description: string;
+	title?: string;
+}
+
+const EmptyState = ({ children, description, title, ...props }: EmptyStateProps) => {
```
manzoorwanijk, line 12:
```suggestion
const EmptyState: React.FC<EmptyStateProps> = ({ children, description, title, ...props }) => {
```

assets/src/infrastructure/ui/display/confirmDelete/ConfirmDelete.tsx
```diff
@@ -0,0 +1,57 @@
...
+							{__('Yes')}
+						</Button>
+					</AlertDialogFooter>
+				</AlertDialogContent>
+			</AlertDialog>
+		</>
+	);
+};
+
+export default ConfirmDelete;
```
manzoorwanijk, line 57: IMHO, `ConfirmDelete` does not belong to infrastructure. It belongs to `application` but will use components from infrastructure.

nicolad: that will create another level of abstraction, thus making reasoning a bit more convoluted

nicolad: a good example of abstraction is BaseInput and then other inputs are composed based on that base components

nicolad: help me understand the composition between `infrastructure` and `application` layers

assets/src/infrastructure/ui/display/EmptyState.tsx
nicolad: thx @manzoorwanijk

assets/src/infrastructure/ui/display/Alert/index.tsx
nicolad: agree, I was `Props` on purpose, since it wasn't exported;
but it's better to have consistency; updated

assets/src/infrastructure/ui/display/Alert/Error.tsx
nicolad: added, thx

assets/src/infrastructure/ui/display/Avatar.tsx
nicolad: updated, thx

assets/src/infrastructure/ui/display/Avatar.tsx
nicolad: done

assets/src/infrastructure/ui/display/DebugInfo.tsx
nicolad: done

assets/src/infrastructure/ui/display/TimezoneTimeInfo/index.tsx
```diff
@@ -0,0 +1,61 @@
...
+				<PopoverTrigger>
+					<IconButton aria-label={label} icon={GlobalOutlined} variant='unstyled' />
+				</PopoverTrigger>
+				<PopoverContent zIndex={4}>
+					<PopoverArrow />
+					<PopoverCloseButton className='ee-timezone-info__close-button' color='var(--ee-color-black)' />
+					<PopoverHeader color='var(--ee-color-black)'>{__('This Date Converted To:')}</PopoverHeader>
+					<PopoverBody>{content}</PopoverBody>
+				</PopoverContent>
+			</Popover>
```
tn3rb, line 56: **THIS** composition of things from Chakra would be appropriate as some kind of "popover" adapter within the infrastructure layer as long as it does NOT set classes like "ee-timezone-info" or apply labels like "This Date Converted To" because those are specific to a single use case.

Remember the little description I posted yesterday about how layers are setup:

```
third party code -> infrastructure -> application -> domain
```

which can also be worded as:

```
external services & objects -> adapters -> framework -> use cases
```

There's two main ideas that are really important to understand about how they interact:

1. the flow of dependencies (which in the simplest form can be thought of as imports) goes from left to right (also described as flowing from the outer layers to the inner layers, because the domain layer is essentially wrapped bv the other layers: like in this [image](https://camo.githubusercontent.com/081561c762f6fcb27bccd8c647e7ef0709c7d21f/68747470733a2f2f63646e2e7261776769742e636f6d2f6a6b70686c2f636c6561722d6172636869746563747572652f62316439643366322f696d616765732f636c6561722d6172636869746563747572652d646f6d61696e2d6170706c69636174696f6e2d636c69656e742d74696572732e737667) or this more [complex image](https://herbertograca.files.wordpress.com/2018/11/100-explicit-architecture-svg.png)):

    - third party code obviously can not depend on any of our code 

    - adapters should not depend on anything other than third party code as they are simply a way of preparing external logic for our use within our application (as well as keeping things decoupled)

    - the application layer should consist primarily of its own logic but can utilize adapters to compose services and elements that can then be used within all of our specific use cases

    - the domain layer is all of our specific use cases and will primarily depend on the application layer but can also directly import adapters from the infrastructure layer (although it's better to utilize logic from the app layer). Because this layer represents the specific use cases, none of its logic can be utilized in any of the other layers. For example, the current use case we are working on is the event editor in the admin. If any of that specific logic was utilized within a component in the application (or infrastructure) layer, then that component would not be usable in any other use cases (or more accurately, it would indicate some logic that is not specific to the event editor and should therefore be relocated to another layer, or at least into the "shared" folder within the domain layer). You wouldn't want content that requires user authentication popping up in non-authed use cases.

2. logic becomes more and more specific and specialized as you move from the outer layers to the inner layers

    - considering the outermost layer is comprised of third party code / external services, it's not hard to see how that would be the most generalized type of code. For example, a database service or a UI library like Chakra could be used to build just about anything as they do not define any specific use case.

    - the adapters in the infrastructure layer are a an intermediary step that prepare those completely generic external services and objects for use within our own application. So you can see that things suddenly got quite a bit more specialized because the logic in that layer is only usable by our application.

    - the application layer takes this a step further by building up services and objects that are more and more specialized (like the Entity List component) but still not specific to one specific use case.

    - the domain layer is of course comprised of all of the specific use cases, although we only have one at the moment (the admin event editor), and consists of the most specialized logic of all of the layers. Anything within a specific use case should be unusable and/or unwanted in any other use case or layer. If that is not the case, then it indicates that the logic should be extracted into a location with greater scope.

Let me know if you have any questions about this. Maybe we can call discuss this at the next dev meeting so that everyone is on the same page (I should have done this already so my apologies for any confusion)

assets/src/infrastructure/ui/display/Alert/index.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+const Alert: React.FC<AlertProps> = ({ description, status, title }) => {
+	if (status === 'error') {
+		return <Error description={description} title={title} />;
+	}
+
+	return null;
+};
```
tn3rb, line 17: what's the point of this?!?!?

why add a totally new component that either returns another component or nothing ?!?! 

Why would anyone use this?

plz remove this

assets/src/infrastructure/ui/display/Alert/Error.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+interface ErrorProps {
+	description: string;
+	title: string;
+}
+
+const Error: React.FC<ErrorProps> = ({ description, title }) => {
+	return (
+		<Alert
+			status='error'
```
tn3rb, line 12: Had a comment written out about how the `EmptyState` and `ErrorIndicator` components should not be in the infrastructure layer but you've gone and changed things so it's not applicable anymore

### 😆 

This `Error` component you have now is better but could still be improved.

This is still a bit too specific for an adapter because it is applying a somewhat specific use case to the underlying Chakra `Alert` component (ie: an error)

So what you need to do is create a single adapter for the  `Alert` component that can be used by any component in the app layer that needs to display some sort of static notification (as opposed to a popup notice). Then reinstate the `EmptyState`, `ErrorIndicator`, and any other components in the app layer that need this, and have those use the new adapter instead of importing from Chakra directly.

The new adapter for the Chakra `Alert` component could be named something like:

- Notification

- Notice

- Banner (meh)

- Alert (although this sounds more like it's just for warnings)

 Then we can also create other more specific components in the app layer for things like:

- success notice

- error notice

- info notice

- warning notice

that would simply apply a bit of extra configuration to the new adapter component. Client code in specific use cases can then utilize these components in the app layer, but only have to add a few extra details, like the actual text to be displayed for the notice as well as set some other style related props like "notice size" etc

assets/src/infrastructure/ui/display/Avatar.tsx
```diff
@@ -0,0 +1,26 @@
...
+		);
+	}
+
+	if (userName) {
+		const size = avatarProps.size || 'lg';
+		return <ChakraAvatar {...avatarProps} name={userName} size={size} />;
+	}
+
+	return <ChakraAvatar {...avatarProps} />;
+};
```
tn3rb, line 24: this is much much better as an adapter 👍

assets/src/infrastructure/ui/display/DebugInfo.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+	return (
+		<>
+			<Button size='sm' onClick={handleToggle} mt='1rem'>
+				{show ? 'Hide' : 'Show'} Debug Info
+			</Button>
+			<Collapse mt={4} isOpen={show}>
+				{output}
+			</Collapse>
+		</>
```
tn3rb, line 45: too specific for an adapter.

turn this into something like:

- Collapsable

- ShowHideContainer

- ???

and **THEN** have the `DebugInfo` component in the app layer, import this and use it for doing its thing.

The way you have this right now, this logic can **only** be used for displaying debug data, whereas this adapter should be usable anywhere we want to have some content that can be hidden and displayed via a toggle, such as the filter bar filters. As is, you would what? Create another component in the infrastructure layer for displaying the filter bar filters that uses the same Chakra components? 

The things in this layer need to be completely generic but completely configurable so that they can be utilized in any other layer

assets/src/infrastructure/ui/display/EmptyState.tsx
```diff
@@ -0,0 +1,38 @@
...
+	className?: string;
+	description: string;
+	title?: string;
+}
+
+const EmptyState: React.FC<EmptyStateProps> = ({ children, description, title, ...props }) => {
+	const className = classNames(props.className, 'ee-empty-state');
+
+	return (
+		<Alert
```
tn3rb, line 16: plz see my comments about `/infrastructure/ui/display/Alert/Error.tsx`

assets/src/infrastructure/ui/display/ErrorIndicator.tsx
```diff
@@ -0,0 +1,32 @@
...
+import { Alert, AlertIcon, AlertTitle, AlertDescription, IAlert } from '@chakra-ui/core';
+
+interface Props extends IAlert {
+	description?: string;
+	title?: string;
+}
+
+const ErrorIndicator: React.FC<Props> = ({ description, title, ...props }) => {
+	return (
+		<Alert
```
tn3rb, line 11: plz see my comments about `/infrastructure/ui/display/Alert/Error.tsx`

this is duplicating the same logic as other components in this layer

assets/src/infrastructure/ui/display/LoadingIndicator.tsx
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
+import { ISpinnerProps } from '@chakra-ui/core';
+
+import Spinner from './Spinner';
+
+const LoadingIndicator: React.FC<ISpinnerProps> = (props) => {
+	return <Spinner {...props} size='lg' />;
+};
```
tn3rb, line 8: move this back into the app layer where it can have some sensible defaults applied, like setting a default text notice to "loading..."

The `Spinner` component you are importing here is the adapter, everything else that needs a spinner can go into the app layer and import it. Don't make duplicate adapters.

assets/src/infrastructure/ui/display/confirmDelete/ConfirmDelete.tsx
nicolad: done, also added the variation I was talking about yesterday: 
https://github.com/eventespresso/event-espresso-core/pull/2682/commits/532bca0daeb3191b8ee366122cca0532c7eee742
*** 
next will add `ConfirmClose` variation

assets/src/infrastructure/ui/display/TimezoneTimeInfo/index.tsx
```diff
@@ -0,0 +1,61 @@
...
+				<strong>{__("The Website's Time Zone")}</strong>
+			</div>
+			<div className={'ee-focus-priority-6'}>{formatDateForSite(date)}</div>
+			<br />
+			<div className={'ee-focus-priority-8'}>
+				<strong>{__('UTC (Greenwich Mean Time)')}</strong>
+			</div>
+			<div className={'ee-focus-priority-6'}>{formatUtcDateForSite(date)}</div>
+		</div>
+	);
```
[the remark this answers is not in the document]
nicolad: adapted: https://github.com/eventespresso/event-espresso-core/pull/2682/commits/7cdfc1edf5ea839195220e8783f9059f44d2aecb

assets/src/infrastructure/ui/display/TimezoneTimeInfo/index.tsx
nicolad: Yes, it clicked, thx.
I was confused/ biased by how `antd` components were implemented before this approach.
Some of `antd` components being wrapped inside ours, but other being used directly.

assets/src/infrastructure/ui/display/Alert.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+const Alert: React.FC<Props> = ({ description, title, ...props }) => (
+	<ChakraAlert
+		{...props}
+		variant='subtle'
+		flexDirection='column'
+		justifyContent='center'
+		textAlign='center'
+		height='200px'
+	>
```
nicolad, line 17: status is passed from the consumer:

image

assets/src/infrastructure/ui/display/Alert/index.tsx
nicolad: removed

assets/src/infrastructure/ui/display/Alert/Error.tsx
nicolad: named `Alert` for simplicity: https://github.com/eventespresso/event-espresso-core/pull/2682/commits/04660d043aca0de7fc59ac25f10c87a5e76cd9c4
***
the `ErrorMessage` which is consuming it has a fairly limited API surface:

```
...
	if (asAlert) {
		return <Alert description={errorMessage} status='error' title={__('Error')} />;
	}
...
```

assets/src/infrastructure/ui/display/DebugInfo.tsx
nicolad: updated, but let's not abstract away the logic into `ShowHideContainer`, because:
1. each component standalone would be useless
2. I'd still have to `show` prop around, this is too much of abstraction IMHO

nicolad: your pong on this?

assets/src/infrastructure/ui/display/EmptyState.tsx
nicolad: updated

assets/src/infrastructure/ui/display/ErrorIndicator.tsx
nicolad: updated `ErrorIndicator`

assets/src/infrastructure/ui/display/DebugInfo.tsx
tn3rb: oh no I didn't mean to split the button out from the `Collapse` container...

I just meant to make this generic so that any use case that requires a container that can be toggled in and out via a button could then use this component, whereas now it's defined as `DebugInfo` which limits the uses (even if just in name)

assets/src/infrastructure/ui/display/Alert.tsx
tn3rb: noice 👍

assets/src/infrastructure/ui/display/DebugInfo.tsx
nicolad: ok, these were separated, lemme know if that's fine

assets/src/infrastructure/ui/display/LoadingIndicator.tsx
nicolad: applied "loading..." as label for screen readers, since `Spinner` does not accept `children` prop as can be seen here:

image

assets/src/application/ui/display/TimezoneTimeInfo/index.tsx
```diff
@@ -33,16 +29,17 @@ const TimezoneTimeInfo: React.FC<OffsetInfoProps> = ({ className, date, inline =
 			<div className={'ee-focus-priority-6'}>{formatUtcDateForSite(date)}</div>
 		</div>
 	);
-	const htmlClassName = classNames(className, 'ee-timezone-info', { 'ee-timezone-info--inline': inline });
+	const label = __('click for timezone\ninformation');
+	const iconProps = {
+		icon: 'info-outline',
```
nicolad, line 34: have used `GlobalOutlined` from antd icons here, there was a type issue and ended up placing this temporary.
I have copied global outlined path, but all icons needs to be added in a separate PR as follows: https://chakra-ui.com/icon#adding-custom-icons

nicolad:
got this last failing test:

image

***
I am switching to https://github.com/eventespresso/event-espresso-core/issues/2672 to see if refactoring that would help

assets/src/infrastructure/ui/display/AlertDialog.tsx
```diff
@@ -0,0 +1,52 @@
...
+							{__('Yes')}
+						</Button>
+					</AlertDialogFooter>
+				</AlertDialogContent>
+			</ChakraAlertDialog>
+		</>
+	);
+};
+
+export default AlertDialog;
```
manzoorwanijk, line 52: `AlertDialog` should also handle the normal error alerts which may not necessarily have footer buttons or may be just the `OK` button. Also a prop like `alertType` could be passed as `success|error|warning|info` and an appropriate icon could be used in `AlertDialogHeader`

manzoorwanijk: An example of the above is the TAM alert dialog when there are orphan entities.

assets/src/infrastructure/ui/display/Collapse.tsx
```diff
@@ -0,0 +1,24 @@
...
+		<>
+			<Button size='sm' onClick={handleToggle} mt='1rem'>
+				{show ? 'Hide' : 'Show'} {buttonText && buttonText}
+			</Button>
+			<ChakraCollapse {...props} isOpen={show} mt={4}>
+				{children}
+			</ChakraCollapse>
+		</>
+	);
+};
```
manzoorwanijk, line 22: 👍

assets/src/infrastructure/ui/display/Popover.tsx
```diff
@@ -0,0 +1,44 @@
...
+import { Icons } from '@chakra-ui/core/dist/theme/icons';
+import { __ } from '@wordpress/i18n';
+
+type IconProps = {
+	label: string;
+	icon: Icons;
+};
+
+interface PopoverProps {
+	content: JSX.Element;
```
manzoorwanijk, line 21:
```suggestion
	content: React.ReactNode;
```

assets/src/infrastructure/ui/display/Popover.tsx
```diff
@@ -0,0 +1,44 @@
...
+	header?: string;
```
manzoorwanijk, line 22:
```suggestion
interface PopoverProps {
	content: React.ReactNode;
	header?: React.ReactNode;
```

nicolad: thx

assets/src/infrastructure/ui/display/AlertDialog.tsx
nicolad: I understand this, but based on `AlertDialog` docs:
```
AlertDialog and it's components composes the Modal component, the only exception is that it requires a leastDestructiveRef which is similar to the initialFocusRef in Modal
```
I think it's better to leave this component as it is to avoid pushing too much logic here and instead create another component for the use case you described, but which would use `Modal` instead.
What do you think?

assets/src/application/ui/display/confirm/Confirm.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+const Confirm: React.FC<ConfirmProps> = ({ type, ...props }) => {
+	if (type === 'delete') {
+		return <ConfirmDelete {...props} />;
+	}
+
+	return <AlertDialog {...props} />;
+};
+
+export default Confirm;
```
tn3rb, line 19: I would do this the other way around

Having a single "master" component like this will require a more complex API in order to configure things, which in turn makes client code more complex. It's better to have an abstracted "parent" component that hides the complexities away, and then create simple straightforward facades that take care of some additional configuration, so that client code can be as simple and clean as possible. (there are of course exceptions where having a single master component works better, like when you are dynamically generating content based on some other configuration)

So instead of importing a `Confirm` component into client code and then having to pass a prop to tell it you actually want a `ConfirmDelete` popup, it's easier, simpler, and more clear for the client code to just import the `ConfirmDelete` component directly. THEN, the `ConfirmDelete` component can internally use the `Confirm` component which internally uses the `AlertDialog` adapter.

Another reason to do things this way: currently the state management for controlling whether the `AlertDialog` is open or not is inside the adapter. But that logic does not belong there. That logic belongs in the app layer.

Here's what I want you to do:

- rename `Confirm` to `useConfirmDialog` (or maybe `useAlertDialog`?) - yes we are going to transform it into a hook

- move all of the state management logic and extra implementation details from `AlertDialog` into this new hook

- also move the actual import and configuration of the `AlertDialog` adapter into this new hook

- import the `useConfirm` hook into `ConfirmDelete` and `ConfirmClose` and whatever other confirm dialog components we need/want

- have these other components pass their configuration/implementation details to `useConfirmDialog` which will then return the configured and ready to go `AlertDialog`

- return the "configured and ready to go" `AlertDialog` for use in client code

This approach accomplishes the following:

- `AlertDialog` stays as a simple straightforward adapter that does nothing except decouple our application from third party code. If we need to change UI libraries in the future (or if there are breaking changes in the implementation of that same third party component) we are insulated from those changes, and only need to refactor the adapter to handle the external changes

- `useConfirmDialog` hides the complexities of state management and adapter configuration and can also handle things like swapping what components to use (like button vs link triggers) so that that logic does not have to be dealt with inside our other components.

- the above allows "child" components like `ConfirmDelete` and `ConfirmClose` to be much simpler because the hook can do most of the heavy lifting. For example, both of those "child" components could have a prop like `useLinkTrigger` and if set to true would pass a `LinkProps` object instead of ` `ButtonProps` object to `useConfirmDialog` which would use the presence of that object in a conditional to decide what trigger component to use. The "child" components could still have logic for exposing some of their configuration to client code for overriding a set of internal defaults, just like you have done in `ConfirmDelete` with the `title`

- client code requires far less configuration because 90% of the time, the "child" components should work just fine as is, but can be reconfigured via props for the edge cases. As well, since client code is dealing with the simplified and specialized "child" components, the APIs will be much easier to understand and manage as opposed to having to figure out how to get a single "master" Confirm component to spit out the correct thing you need.

- ideally, if we need/want to add a new type of confirm dialog component, we don't need to touch any other code. We simply need to import the `useConfirmDialog` hook into the new component and configure things the way we want. The way you are doing things now would require the `Confirm` component to be modified every time you want to add a new type of confirm dialog component (assuming that `Confirm` would allow this new component to be utilized via its `type` prop`).

so the final result should work something like this:

- the `AlertDialog` adapter is dead simple and is only used for managing differences in props between our app and the Chakra UI library. It does NOTHING else.

- `useConfirmDialog` handles all of the state management and as much shared logic as possible in order to hide that complexity and keep the final user APIs as simple as possible

- "child" components like `ConfirmDelete` and `ConfirmClose` utilize and configure `useConfirmDialog` via props, and provide a set of sensible defaults that can be exposed as needed to client code for further customization.

- client code utilizing the "child" components can hopefully in most cases not have to do anything in terms of customization because the sensible defaults in the child components work fine as is

assets/src/infrastructure/ui/display/AlertDialog.tsx
manzoorwanijk: Yeah I am OK with that.

assets/src/application/ui/display/confirm/Confirm.tsx
```diff
@@ -0,0 +1,49 @@
...
+						{__('No')}
+					</Button>
+				}
+				header={buttonProps?.tooltip}
+				isOpen={isOpen}
+				leastDestructiveRef={cancelRef}
+				okButton={
+					<Button variantColor='red' onClick={onDelete} ml={3}>
+						{__('Yes')}
+					</Button>
```
nicolad, line 40: was getting this error, this is why I didn't used `EspressoButton`:

image

assets/src/infrastructure/ui/display/AlertDialog.tsx
```diff
@@ -0,0 +1,52 @@
...
+						</AlertDialogHeader>
+					)}
+					<AlertDialogBody>{title}</AlertDialogBody>
+					<AlertDialogFooter>
+						<Button ref={cancelRef} onClick={onClose}>
+							{__('No')}
+						</Button>
+						<Button variantColor='red' onClick={onDelete} ml={3}>
+							{__('Yes')}
+						</Button>
```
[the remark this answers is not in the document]
nicolad: updated: https://github.com/eventespresso/event-espresso-core/pull/2682/commits/d4ffc6ca41e3c84e76740e0abc6015a9a0cd4b8e

assets/src/infrastructure/ui/display/Collapse.tsx
```diff
@@ -0,0 +1,24 @@
...
+}
+
+const Collapse: React.FC<Props> = ({ buttonText, children, ...props }) => {
+	const [show, setShow] = React.useState(false);
+	const handleToggle = useCallback(() => setShow(!show), [show]);
+
+	return (
+		<>
+			<Button size='sm' onClick={handleToggle} mt='1rem'>
+				{show ? 'Hide' : 'Show'} {buttonText && buttonText}
```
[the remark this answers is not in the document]
nicolad: simplified: https://github.com/eventespresso/event-espresso-core/pull/2682/commits/ad7eca3a5bcfe9419f5d74f5ba7418bd8ecb2212

assets/src/infrastructure/ui/display/AlertDialog.tsx
```diff
@@ -0,0 +1,52 @@
...
+import { AlertDialogProps } from './types';
+
+const AlertDialog: React.FC<AlertDialogProps> = ({ buttonProps, onConfirm, title }) => {
+	const [isOpen, setIsOpen] = React.useState(false);
+	const cancelRef = React.useRef();
+	const onClose = () => setIsOpen(false);
+	const onDelete = () => {
+		onClose();
+		onConfirm();
+	};
```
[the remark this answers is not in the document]
nicolad: replaced the order, but in this case since `onConfirm` is not an async operation, both will be fired at almost the same time
***
in order to execute `onClose` after `onConfirm`, the last needs to returns a promise

assets/src/infrastructure/ui/display/Popover.tsx
```diff
@@ -0,0 +1,44 @@
...
+	Popover as ChakraPopover,
+	PopoverTrigger,
+	PopoverContent,
+	PopoverHeader,
+	PopoverBody,
+	PopoverArrow,
+	PopoverCloseButton,
+} from '@chakra-ui/core';
+import { Icons } from '@chakra-ui/core/dist/theme/icons';
+import { __ } from '@wordpress/i18n';
```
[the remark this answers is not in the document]
nicolad: yes, had cleaned up

assets/src/infrastructure/ui/display/Popover.tsx
```diff
@@ -0,0 +1,44 @@
...
+interface PopoverProps {
+	content: JSX.Element;
+	header?: string;
+	iconProps: IconProps;
+}
+
+const Popover: React.FC<PopoverProps> = ({ content, header, iconProps, ...props }) => (
+	<ChakraPopover {...props}>
+		<PopoverTrigger>
+			<IconButton aria-label={iconProps.label} icon={iconProps.icon} variant='unstyled' />
```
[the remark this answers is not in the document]
nicolad: Great observation, updated: https://github.com/eventespresso/event-espresso-core/pull/2682/commits/c94915c055adec1635e18056383da7e357f5caa0

assets/src/application/ui/display/confirm/Confirm.tsx
nicolad: updated: https://github.com/eventespresso/event-espresso-core/pull/2682/commits/73646067896bc6a7278ffb59a47ef7d37c5db594
***
`ConfirmClose` will be added in this PR along with related logic
https://github.com/eventespresso/event-espresso-core/pull/2636

assets/src/domain/shared/entities/tickets/predicates/filters/nextOnSaleOrPendingOnly/index.test.ts
```diff
@@ -1,16 +0,0 @@
...
-	it('should return an empty array if tickets are not on sale', () => {
-		const updatedTickets = tickets.map((ticket) => {
-			const startDate = formatISO(new Date(2018, 8, 18, 19, 0, 52));
-			const endDate = formatISO(new Date(2018, 9, 18, 19, 0, 52));
-			return { ...ticket, startDate, endDate, isOnSale: false, isPending: false };
-		});
-		const filteredTickets = nextOnSaleOrPendingOnly(updatedTickets);
-		expect(filteredTickets).toEqual([]);
-	});
-});
```
nicolad, line 16: I can't comprehend why this test was causing that test error:

image
***
have removed that for now, since I don't know how this can be fixed;
had tried skipping it, did not worked

PR #2690, opened 2020-04-13
Date and Time pickers | Closes #2671

### LINKED ISSUE #2671: Replace AntD DatePicker and TimePicker
see #2667

#### known uses:

- `\application\ui\input\DateTimePicker\DatePickerBase.tsx`
- `\application\ui\input\DateTimePicker\TimePickerBase.tsx`

#### steps:

- discuss, debate, engage in fisticuffs regarding which external lib to use

- add adapters in `\infrastructure\ui\inputs`

- ensure that dates display in site TZ (but has option to override with others?)

- ensure that dates are passed to rest of system in UTC

- update uses listed above

- submit as separate PRs for each component

manzoorwanijk:
This PR:
- Creates `Datepicker` and `Timepicker` from Uber Base UI in `/application/ui/input/dateTime`
- Updates RFF adapters to use new pickers
- Updates Date and Ticket edit forms to use the new pickers

NOTE: `yarn` needed for this.


assets/src/application/ui/input/dateTime/index.tsx
```diff
@@ -0,0 +1,27 @@
...
+	);
+};
+
+export const Timepicker: React.FC<TimepickerProps> = (props) => {
+	return (
+		<Suspense fallback={<LoadingNotice />}>
+			<LazyTimepicker {...props} />
+		</Suspense>
+	);
+}; */
```
manzoorwanijk, line 23: I wanted to lazy load the pickers to see the bundle cost but it didn't help, may be because our production CSS build is not properly set up yet. Leaving it here, may be we can use it

assets/src/domain/eventEditor/ui/datetimes/dateForm/useDateFormConfig.ts
```diff
@@ -39,10 +38,10 @@ const useDateFormConfig = (id: EntityId, config?: EspressoFormProps): DateFormCo
...
 		dateTime: {
-			startDate: format(startDate, CONVERT_TO_MOMENT_DATE_FORMAT),
-			startTime: format(startDate, CONVERT_TO_MOMENT_TIME_FORMAT),
-			endDate: format(endDate, CONVERT_TO_MOMENT_DATE_FORMAT),
-			endTime: format(endDate, CONVERT_TO_MOMENT_TIME_FORMAT),
+			startDate: startDate,
+			startTime: startDate,
+			endDate: endDate,
+			endTime: endDate,
 		},
```
tn3rb, line 45: imho this is the kind of thing (the format/shape that dates are supplied as) that we should be trying to encapsulate within our infrastructure adapters because if we ever have to change datepickers in the future, then this code will also need to change. Ideally our adapters would always receive the exact same input (JS Date object, ISO date string, etc) and then convert to whatever the third party code expects (JS Date object, moment object, ISO date string, etc).

assets/src/application/ui/input/dateTime/BaseUIProvider.tsx
```diff
@@ -0,0 +1,16 @@
...
+
+const BaseUIProvider: React.FC = ({ children }) => {
+	return (
+		<StyletronProvider value={engine}>
+			<BaseProvider theme={LightTheme}>{children}</BaseProvider>
+		</StyletronProvider>
+	);
+};
+
+export default BaseUIProvider;
```
tn3rb, line 16: imho this could be part of the infrastructure adapter for the datepicker

assets/src/application/ui/input/dateTime/withBaseProvider.tsx
```diff
@@ -0,0 +1,15 @@
...
+	props: Props
+) => {
+	return (
+		<BaseUIProvider>
+			<Component {...props} />
+		</BaseUIProvider>
+	);
+};
+
+export default withBaseProvider;
```
tn3rb, line 15: imho this could be part of the infrastructure adapter for the datepicker and timepicker and wrap the exports for those adapters right at the source

assets/src/application/ui/forms/espressoForm/adapters/TimePicker.tsx
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+
+import { Timepicker } from '@infraUI/inputs/dateTime';
+import { FieldRendererProps } from '../types';
+
+const TimePicker: React.FC<FieldRendererProps> = ({ input: { onChange, ...input }, meta, ...rest }) => {
+	return (
+		<Timepicker
+			{...input}
```
nicolad, line 9: now I know why this is done this way
:)

assets/src/domain/eventEditor/ui/datetimes/dateForm/useDateFormConfig.ts
nicolad: good one, agree, the end components and the application layer shouldn't be worried too much about these formatting

PR #2693, opened 2020-04-14
Refactor EspressoButton and EspressoIcon | Closes #2675

### LINKED ISSUE #2675: Update EspressoButton and EspressoIcon to work with Chakra UI and Other Updated Components
#### known uses:

- `\application\ui\input\EspressoButton\EspressoButton.tsx`
- `\application\ui\display\espressoIcon\EspressoIcon.tsx`
- `\domain\eventEditor\ui\ticketAssignmentsManager\components\table\BodyCell.tsx`

#### steps:

- add adapters in `\infrastructure\ui\inputs` and `\infrastructure\ui\display` respectively
- update uses listed above
- ensure all implementations function as expected

manzoorwanijk:
This PR refactors `EspressoButton` and `EspressoIcons` to lift their burden.
- Creates `IconButton` in `/infrastructure/ui/inputs`
- Create standalone icon components in `/application/ui/display/espressoIcons/collection`
- Refactor `EspressoButton` to separate icon rendering
- Updates all the consumer components


assets/src/application/ui/display/confirm/useConfirm.tsx
```diff
@@ -1,12 +1,15 @@
-import React, { useCallback, useMemo } from 'react';
-import { Button, useDisclosure } from '@chakra-ui/core';
+import React, { useCallback } from 'react';
+import { useDisclosure } from '@chakra-ui/core';
```
nicolad, line 2: I also have used  `useDisclosure` in `Confirm` comp in the app layer, but not sure what's the best approach, if we want to keep external deps at this layer or not, just thinking outloud :thinking:

assets/src/application/ui/display/confirm/useConfirm.tsx
```diff
@@ -1,12 +1,15 @@
...
 import { __ } from '@wordpress/i18n';
 
 import { EspressoButton } from '@application/ui/input';
 import { AlertDialog } from '@infraUI/display';
+import { Button } from '@infraUI/inputs';
 
 import { ConfirmProps } from './types';
 
+// TODO: fix file name and add suport for EspressoIconButton e.g. for TPC delete price button
```
nicolad, line 11: it was initially `Confirm`, but after this short comment: https://github.com/eventespresso/event-espresso-core/pull/2682#discussion_r406828007
it was renamed/ repurpose to `useConfirm`

assets/src/application/ui/display/confirm/useConfirm.tsx
```diff
@@ -19,19 +22,11 @@ const Confirm: React.FC<ConfirmProps> = ({ buttonProps, onConfirm, title }) => {
 		<>
 			<EspressoButton {...buttonProps} onClick={onOpen} />
 			<AlertDialog
-				cancelButton={
-					<Button ref={cancelRef} onClick={onClose}>
-						{__('No')}
-					</Button>
-				}
+				cancelButton={<Button buttonText={__('No')} ref={cancelRef} onClick={onClose} />}
```
nicolad, line 25: taking a fresh look now at the `Button` API, the `buttonText` doesn't help very much, because it's clear that the text is pertaining to button not to :crocodile: or anything else
I know, this is how it was before, just noticed that this is a bit verbose, we can use just `text` or pass it as `children`, dunno

assets/src/application/ui/display/espressoIcons/EspressoIcon.tsx
```diff
@@ -0,0 +1,39 @@
...
+ */
+const EspressoIcon = React.forwardRef<SVGSVGElement, EspressoIconProps>(
+	({ name, svgSize = 20, className, ...otherProps }, ref) => {
+		const path = svgPath[name];
+
+		if (!path) {
+			return null;
+		}
+
+		const htmlClass = classNames(className, 'dashicon', `dashicons-${name}`, 'espresso-icon');
```
nicolad, line 18: `className`

nicolad: oh, I see why this was used this way, what about:

```
		const className = classNames(otherProps.className, 'dashicon', `dashicons-${name}`, 'espresso-icon');
```
and `className` can be removed from the destructurization from the above

assets/src/application/ui/display/espressoIcons/EspressoIcon.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+		return (
+			<svg
+				role={'img'}
```
nicolad, line 22: why not `role='img'`?
want to understand the difference 
between this :point_up: and :point_right: `role={'img'}`

assets/src/application/ui/display/espressoIcons/EspressoIcon.tsx
```diff
@@ -0,0 +1,39 @@
...
+				aria-hidden={true}
+				focusable={false}
+				xmlns='http://www.w3.org/2000/svg'
+				width={svgSize}
+				height={svgSize}
+				viewBox='0 0 20 20'
```
nicolad, line 28: this is standard for all icons?
just took a look at GB's, they don't have this set as default: https://github.com/WordPress/gutenberg/blob/master/packages/primitives/src/svg/index.js#L89

assets/src/application/ui/display/espressoIcons/collection/Copy.tsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+
+import { EspressoIconProps, Icon } from '../types';
+import EspressoIcon from '../EspressoIcon';
+
+const Copy = React.forwardRef<SVGSVGElement, Partial<EspressoIconProps>>((props, ref) => (
+	<EspressoIcon {...props} name={Icon.COPY} ref={ref} />
+));
+
+Copy.displayName = 'Copy';
```
nicolad, line 10: yes, this will be helpful in dev tool, to avoid those anonymouse functions 
:+1:

assets/src/application/ui/display/espressoIcons/collection/GlobalOutlined.tsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+
+import { EspressoIconProps, Icon } from '../types';
+import EspressoIcon from '../EspressoIcon';
+
+const GlobalOutlined = React.forwardRef<SVGSVGElement, Partial<EspressoIconProps>>((props, ref) => (
```
nicolad, line 6: :+1:

assets/src/application/ui/input/EspressoButton/EspressoButton.tsx
```diff
@@ -14,15 +13,15 @@ type ButtonType = React.ComponentType<EspressoButtonProps>;
  * forwardRef to be able to accept
  * onMouseEnter, onMouseLeave, onFocus, onClick events from parent
  */
-const EspressoButton = React.forwardRef<Button, EspressoButtonProps>(
+const EspressoButton = React.forwardRef<typeof Button, EspressoButtonProps>(
```
nicolad, line 16: I noticed this pattern in few other cases how is this different than the previous version w/o `typeof`?

assets/src/application/ui/input/EspressoButton/EspressoIconButton.tsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import classNames from 'classnames';
+
+import { IconButton, IconButtonProps } from '@infraUI/inputs';
+import { EspressoIconButtonProps } from './types';
+import { withLabel, withTooltip } from '../../display';
+
+type ButtonType = React.ComponentType<EspressoIconButtonProps>;
+
+const EspressoIconButton = React.forwardRef<typeof IconButton, IconButtonProps>(
```
nicolad, line 10: let's name it just `IconButton`
we had a convo about `Espresso` prefix in slack last week and @tn3rb agreed that we can go with a shorter version w/o the prefix

assets/src/application/ui/input/EspressoButton/EspressoButton.tsx
```diff
@@ -14,15 +13,15 @@ type ButtonType = React.ComponentType<EspressoButtonProps>;
...
 	(
 		{
-			icon,
-			onClick,
-			buttonText,
 			buttonSize = EspressoButtonSize.DEFAULT,
+			buttonText,
 			buttonType = EspressoButtonType.DEFAULT,
 			className: htmlClass,
+			icon,
```
nicolad, line 23: do we still need to pass this prop here, since we have that separated `EspressoIconButton`?
and also you proposed yesterday to have the separation of concerns in this component with which I do agree
also chakra have the same approach: https://chakra-ui.com/iconbutton

assets/src/application/ui/layout/entityList/filterBar/buttons/CardViewFilterButton.tsx
```diff
@@ -13,16 +13,16 @@ const CardViewFilterButton: React.FC<CardViewFilterButtonProps> = ({ listId, set
 	const filterId = `ee-card-view-btn-${listId}`;
 
 	return (
-		<EspressoButton
-			buttonType={EspressoButtonType.MINIMAL}
+		<EspressoIconButton
 			className={className}
-			icon={<AppstoreFilled />}
+			icon={() => <AppstoreFilled />}
```
nicolad, line 18: maybe a separate PR to copy the svg paths from `@ant-design/icons`

assets/src/domain/eventEditor/ui/datetimes/datesList/actionsMenu/AssignTicketsButton.tsx
```diff
@@ -24,14 +25,17 @@ const AssignTicketsButton: React.FC<EntityListItemProps> = React.memo(({ entity
 		? `${__('Related Tickets:')} ${relatedTicketDbIds.join(', ')}`
 		: __('There are no tickets assigned to this datetime. Please click the ticket icon to update the assignments.');
 
+	const tooltipProps = { placement: 'right' as 'right' };
```
nicolad, line 28: image

assets/src/application/ui/display/confirm/useConfirm.tsx
tn3rb: If it was a separate package that could be installed on its own then that would be fine, but as is, if we want/need to change UI library in the future, then this will need refactoring. It's ok to directly import low-level utility packages like Ramda and date-fns but for components we should use adapters. This kinda falls in between but because it is part of a UI library, should be decoupled imho.

potential options:

- create an adapter for `useDisclosure`

- build our own custom hook(s) for managing common state patterns

- use another library for non-UI utilities like: https://hooks.umijs.org/hooks/state/use-toggle

assets/src/application/ui/display/confirm/useConfirm.tsx
tn3rb: ya the previous API was because the button was handling multiple things, so I agree that this could be changed now.
might be better to just use children which allows for anything whereas a "text" prop implies a `string` only, but I'm ok with either.

assets/src/application/ui/display/espressoIcons/EspressoIcon.tsx
tn3rb: there's no difference technically

I usually use brackets for string props because then it is easier to convert the value to a variable (not the case for the "role" attribute of course) and i can just do the same thing every time out of habit without having to think about it.

just my two cents worth, but if both of you really prefer using the `prop='string'` pattern over  `prop={'string'}` then i am happy to change.

assets/src/application/ui/display/espressoIcons/EspressoIcon.tsx
tn3rb: doesn't this depend on the scale that the SVG path was written in?

assets/src/application/ui/display/confirm/useConfirm.tsx
nicolad: +1 for adapter, the easiest and with less friction option ATM

assets/src/application/ui/display/espressoIcons/EspressoIcon.tsx
nicolad: I am ok with both but would prefer a consistent approach, such as there is no need to think twice when seeing smth basic like this

assets/src/application/ui/input/EspressoButton/EspressoButton.tsx
manzoorwanijk: That's just an extra step

assets/src/application/ui/input/EspressoButton/EspressoIconButton.tsx
manzoorwanijk: We already have `IconButton` in infrastructure folder.

assets/src/application/ui/input/EspressoButton/EspressoButton.tsx
manzoorwanijk: We can pass the button that is shown on the left side of the text.

assets/src/application/ui/layout/entityList/filterBar/buttons/CardViewFilterButton.tsx
manzoorwanijk: Yes

assets/src/domain/eventEditor/ui/datetimes/datesList/actionsMenu/AssignTicketsButton.tsx
manzoorwanijk: That's TS string to literal casting 😄

assets/src/application/ui/input/EspressoButton/EspressoButton.tsx
nicolad: ok, but what is the purpose?

assets/src/application/ui/display/espressoIcons/EspressoIcon.tsx
manzoorwanijk: Done

assets/src/application/ui/display/confirm/useConfirm.tsx
manzoorwanijk: I agree with creating an adapter

assets/src/application/ui/input/EspressoButton/EspressoButton.tsx
manzoorwanijk: `IconButton` cannot have text. So when we need both Icon and text, we will pass that icon and text to the button.

assets/src/application/ui/forms/espressoForm/config.tsx
```diff
@@ -23,10 +23,10 @@ export const formConfig: EspressoFormProps = {
 		return errors;
 	},
 	submitButton: {
-		label: 'Save Changes',
+		buttonText: 'Save Changes',
```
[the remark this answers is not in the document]
manzoorwanijk: That's just a test config file, an example usage of espressoForm

PR #2694, opened 2020-04-14
Fix logic for loading custom espresso_default.css from /uploads/

joshfeck:
If someone were to copy the espresso_default.css stylesheet to /wp-content/uploads/espresso/css, their copy would not load. The logic was kind of broken, and the URL to stylesheet needs to be provided, not the path.

## How has this been tested

- [x] Copy the espresso_default.css into /wp-content/uploads/espresso/css
- [x] Make a noticeable change to some styles and clear browser cache
- [x] Check an event page to verify the change takes effect
- [x] Also verify if no espresso_default.css in /wp-content/uploads/espresso/css then the stylesheet provided by the plugin loads as expected

## Checklist

* [x] My code is tested.


Pebblo:
I've also tested this and it works as expected, passing back to get a once over on my last commit before merging.

PR #2704, opened 2020-04-17
Show alert upon closing modal via escape keypress; partially close #2613.

### LINKED ISSUE #2613: Add Confirmation Modals to Event Handlers for Destructive Processes
Currently a user can click on buttons (or perform keypresses) for destructive events, like deleting an entity or closing a form, and there is nothing to prevent that from happening if the click or keypress was accidental.

Therefore we need to add a confirmation step prior to performing any kind of destructive logic

- delete buttons
- modal cancel button
- modal close button
- ESC key (onKeyPress handler)

The [AntD Popconfirm component](https://ant.design/components/popconfirm/) is designed specifically for this use case

nicolad:
- Creates `ModalWithAlert` in app layer to add alert in error cases
- Partially closes: https://github.com/eventespresso/event-espresso-core/issues/2613

## How has this been tested
Peek 2020-04-17 16-40
***
Peek 2020-04-17 16-42

### DIFF
--- a/assets/src/application/ui/layout/modal/ModalWithAlert.tsx
+++ b/assets/src/application/ui/layout/modal/ModalWithAlert.tsx
@@ -1,0 +1,50 @@
+import React, { useCallback, useEffect, useMemo } from 'react';
+import { useDisclosure } from '@chakra-ui/core';
+import { __ } from '@wordpress/i18n';
+import { canUseDOM } from 'exenv';
+import { ESCAPE } from '@wordpress/keycodes';
+
+import { ButtonProps } from '@infraUI/inputs';
+import { AlertDialog } from '@infraUI/display';
+import { Button as ButtonAdapter } from '@infraUI/inputs';
+import { Modal, ModalProps } from '@infraUI/layout/modal';
+
+interface Props extends ModalProps {
+	showAlertOnEscape: boolean;
+}
+
+const ModalWithAlert: React.FC<Props> = ({ children, showAlertOnEscape, ...props }) => {
+	const { isOpen, onOpen, onClose } = useDisclosure();
+	const cancelRef = React.useRef();
+	const onEscape = ({ keyCode }) => keyCode === ESCAPE && onOpen();
+
+	useEffect(() => {
+		canUseDOM && document.addEventListener('keydown', onEscape);
+
+		return () => {
+			canUseDOM && document.removeEventListener('keydown', onEscape);
+		};
+	}, []);
+
+	return (
+		<>
+			<Modal {...props} closeOnEsc={!showAlertOnEscape}>
+				{children}
+			</Modal>
+			{showAlertOnEscape && (
+				<AlertDialog
+					cancelButton={<ButtonAdapter buttonText={__('No')} ref={cancelRef} onClick={onClose} />}
+					header={__('Are you sure you want to close this modal?')}
+					isOpen={isOpen}
+					leastDestructiveRef={cancelRef}
+					okButton={
+						<ButtonAdapter buttonText={__('Yes')} variantColor='red' onClick={props.onClose} ml={3} />
+					}
+					onClose={onClose}
+				/>
+			)}
+		</>
+	);
+};
+
+export default ModalWithAlert;
--- a/assets/src/application/ui/layout/modal/ModalWithAlert.tsx
+++ b/assets/src/application/ui/layout/modal/ModalWithAlert.tsx
@@ -10,12 +10,18 @@ import { Button as ButtonAdapter } from '@infraUI/inputs';
 import { Modal, ModalProps } from '@infraUI/layout/modal';
 
 interface Props extends ModalProps {
+	cancelBtnText?: string;
+	header?: string;
+	okBtnText?: string;
 	showAlertOnEscape: boolean;
 }
 
 const ModalWithAlert: React.FC<Props> = ({ children, showAlertOnEscape, ...props }) => {
 	const { isOpen, onOpen, onClose } = useDisclosure();
 	const cancelRef = React.useRef();
+	const cancelBtnText = props.cancelBtnText || __('No');
+	const header = props.header || __('Are you sure you want to close this?');
+	const okBtnText = props.okBtnText || __('Yes');
 	const onEscape = ({ keyCode }) => keyCode === ESCAPE && onOpen();
 
 	useEffect(() => {
@@ -33,12 +39,12 @@ const ModalWithAlert: React.FC<Props> = ({ children, showAlertOnEscape, ...props
 			</Modal>
 			{showAlertOnEscape && (
 				<AlertDialog
-					cancelButton={<ButtonAdapter buttonText={__('No')} ref={cancelRef} onClick={onClose} />}
-					header={__('Are you sure you want to close this modal?')}
+					cancelButton={<ButtonAdapter buttonText={cancelBtnText} ref={cancelRef} onClick={onClose} />}
+					header={header}
 					isOpen={isOpen}
 					leastDestructiveRef={cancelRef}
 					okButton={
-						<ButtonAdapter buttonText={__('Yes')} variantColor='red' onClick={props.onClose} ml={3} />
+						<ButtonAdapter buttonText={okBtnText} variantColor='red' onClick={props.onClose} ml={3} />
 					}
 					onClose={onClose}
 				/>
--- a/assets/src/application/ui/layout/modal/ModalWithAlert.tsx
+++ b/assets/src/application/ui/layout/modal/ModalWithAlert.tsx
@@ -1,10 +1,9 @@
-import React, { useCallback, useEffect, useMemo } from 'react';
+import React, { useEffect } from 'react';
 import { useDisclosure } from '@chakra-ui/core';
 import { __ } from '@wordpress/i18n';
-import { canUseDOM } from 'exenv';
+import { canUseDOM } from '@appServices/utilities/dom';
 import { ESCAPE } from '@wordpress/keycodes';
 
-import { ButtonProps } from '@infraUI/inputs';
 import { AlertDialog } from '@infraUI/display';
 import { Button as ButtonAdapter } from '@infraUI/inputs';
 import { Modal, ModalProps } from '@infraUI/layout/modal';
--- a/assets/src/application/ui/layout/modal/ModalWithAlert.tsx
+++ b/assets/src/application/ui/layout/modal/ModalWithAlert.tsx
@@ -28,10 +28,14 @@ const ModalWithAlert: React.FC<Props> = ({ children, showAlertOnEscape, ...props
 	};
 
 	useEffect(() => {
-		canUseDOM && document.addEventListener('keydown', onEscape);
+		if (canUseDOM) {
+			document.addEventListener('keydown', onEscape);
+		}
 
 		return () => {
-			canUseDOM && document.removeEventListener('keydown', onEscape);
+			if (canUseDOM) {
+				document.removeEventListener('keydown', onEscape);
+			}
 		};
 	}, []);
 
--- a/assets/src/application/ui/layout/modal/ModalWithAlert.tsx
+++ b/assets/src/application/ui/layout/modal/ModalWithAlert.tsx
@@ -21,7 +21,11 @@ const ModalWithAlert: React.FC<Props> = ({ children, showAlertOnEscape, ...props
 	const cancelBtnText = props.cancelBtnText || __('No');
 	const header = props.header || __('Are you sure you want to close this?');
 	const okBtnText = props.okBtnText || __('Yes');
-	const onEscape = ({ keyCode }) => keyCode === ESCAPE && onOpen();
+	const onEscape = ({ keyCode }): void => {
+		if (keyCode === ESCAPE) {
+			onOpen();
+		}
+	};
 
 	useEffect(() => {
 		canUseDOM && document.addEventListener('keydown', onEscape);
--- a/assets/src/infrastructure/ui/layout/modal/modalCloseButtonProps/index.tsx
+++ b/assets/src/infrastructure/ui/layout/modal/modalCloseButtonProps/index.tsx
@@ -1,4 +1,3 @@
-import React from 'react';
 import { __ } from '@wordpress/i18n';
 
 import { ButtonProps } from '@infraUI/inputs';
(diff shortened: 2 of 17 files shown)


package.json
```diff
@@ -74,6 +74,7 @@
 		"date-fns": "^2.11.1",
 		"date-fns-tz": "^1.0.10",
 		"emotion-theming": "^10.0.27",
+		"exenv": "^1.2.2",
```
nicolad, line 77: Added this since we are already using that indirectly via Chakra: https://github.com/chakra-ui/chakra-ui/blob/master/packages/chakra-ui/package.json#L31

assets/src/infrastructure/ui/layout/modal/modalCloseButtonProps/styles.scss
```diff
@@ -0,0 +1,17 @@
...
+		top: 8px;
+		right: 12px;
+		min-width: auto;
+		width: var(--ee-padding-big);
+		height: var(--ee-padding-big);
+		svg {
+			height: var(--ee-border-radius-big);
+			margin: 0;
+			width: var(--ee-border-radius-big);
+		}
```
nicolad, line 15: I like this better: 

image

nicolad: I needed `--ee-border-radius-big` to match `12px` which is the same dimension as in chakra.

nicolad: Open to better variable names than `--ee-border-radius-big`

assets/src/application/ui/forms/espressoForm/utils.ts
```diff
@@ -126,7 +126,7 @@ export const yupToFinalFormErrors = async <T>(validationSchema: ObjectSchema, va
 	} catch (err) {
 		// type annotate the error
 		const error: ValidationError = err;
-		return error.inner.reduce((formError, innerError) => {
+		return error?.inner?.reduce((formError, innerError) => {
```
nicolad, line 129: this change was discussed with @manzoorwanijk

assets/src/domain/eventEditor/ui/ticketAssignmentsManager/components/ModalContainer.tsx
```diff
@@ -10,9 +10,9 @@ const ModalContainer: React.FC<ModalContainerProps> = ({ isOpen, onClose, ...pro
 
 	let title = '';
 	if (assignmentType === 'forDate') {
-		title = sprintf(__('Ticket Assignment Manager for Datetime: %s - %s'), entity.dbId, entity.name);
+		title = sprintf(__('Ticket Assignment Manager for Datetime: %s - %s'), String(entity.dbId), entity.name);
 	} else if (assignmentType === 'forTicket') {
-		title = sprintf(__('Ticket Assignment Manager for Ticket: %s - %s'), entity.dbId, entity.name);
+		title = sprintf(__('Ticket Assignment Manager for Ticket: %s - %s'), String(entity.dbId), entity.name);
```
nicolad, line 15: image

nicolad:
Got some tests which are failing only on Travis:

image
***
locally:

image

***

assets/src/application/ui/layout/modal/ModalWithAlert.tsx
```diff
@@ -0,0 +1,50 @@
...
+				{children}
+			</Modal>
+			{showAlertOnEscape && (
+				<AlertDialog
+					cancelButton={<ButtonAdapter buttonText={__('No')} ref={cancelRef} onClick={onClose} />}
+					header={__('Are you sure you want to close this modal?')}
+					isOpen={isOpen}
+					leastDestructiveRef={cancelRef}
+					okButton={
+						<ButtonAdapter buttonText={__('Yes')} variantColor='red' onClick={props.onClose} ml={3} />
```
tn3rb, line 41: let's change all of these i18n strings into variables and then also add optional props for them with these as the defaults, ie:

```js
const cancelBtnText = cancelButtonText ? cancelButtonText : __('No');
// etc
```

As well, let's change the header text to just be: 

```
__('Are you sure you want to close this?')
```

as most non-devs won't understand what a modal is

assets/src/domain/eventEditor/ui/ticketAssignmentsManager/components/ModalContainer.tsx
tn3rb: crazy... never had that before

assets/src/infrastructure/ui/layout/modal/modalCloseButtonProps/index.tsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import { __ } from '@wordpress/i18n';
+
+import { ButtonProps } from '@infraUI/inputs';
+import { Close } from '@appDisplay/icons';
+
+import './styles.scss';
+
+const modalCloseButtonProps: ButtonProps = {
+	className: 'confirm-close',
```
tn3rb, line 10: plz namespace all class names, ie: `ee-confirm-close`

assets/src/infrastructure/ui/layout/modal/modalCloseButtonProps/styles.scss
tn3rb: ya I'll most likely tweak things once everything is converted, but that doesn't look bad as is

package.json
tn3rb: LOLZ I added a function with the exact same name for the toaster
it's likely the exact same logic too, since I think that package is a duplication of internal logic from React

assets/src/application/ui/layout/modal/ModalWithAlert.tsx
nicolad: > as most non-devs won't understand what a modal is

good one, thx

nicolad: updated this, with a slightly changed approach, but with the same result;
thx for suggestion

assets/src/infrastructure/ui/layout/modal/modalCloseButtonProps/index.tsx
nicolad: thx, updated

package.json
nicolad: got tricked by chakra: https://github.com/chakra-ui/chakra-ui/blob/master/packages/chakra-ui/src/Modal/index.js#L22
***
but followed your approach, which is less dependent on external stuff

assets/src/infrastructure/ui/layout/modal/modalCloseButtonProps/styles.scss
nicolad: thanks

assets/src/application/ui/layout/modal/ModalWithAlert.tsx
```diff
@@ -0,0 +1,55 @@
...
+	const header = props.header || __('Are you sure you want to close this?');
+	const okBtnText = props.okBtnText || __('Yes');
+	const onEscape = ({ keyCode }) => keyCode === ESCAPE && onOpen();
+
+	useEffect(() => {
+		canUseDOM && document.addEventListener('keydown', onEscape);
+
+		return () => {
+			canUseDOM && document.removeEventListener('keydown', onEscape);
+		};
```
manzoorwanijk, line 31: For clarity, better use `if (canUseDOM) {}`

assets/src/domain/eventEditor/ui/ticketAssignmentsManager/components/ModalContainer.tsx
```diff
@@ -10,9 +10,9 @@ const ModalContainer: React.FC<ModalContainerProps> = ({ isOpen, onClose, ...pro
 
 	let title = '';
 	if (assignmentType === 'forDate') {
-		title = sprintf(__('Ticket Assignment Manager for Datetime: %s - %s'), entity.dbId, entity.name);
+		title = sprintf(__('Ticket Assignment Manager for Datetime: %s - %s'), String(entity.dbId), entity.name);
```
manzoorwanijk, line 13: Always better to use language constructs (which are faster) instead of calling functions. Although, it won't make a big difference here but still it's better.
```suggestion
		title = sprintf(__('Ticket Assignment Manager for Datetime: %s - %s'), `${entity.dbId}`, entity.name);
```

assets/src/application/ui/layout/modal/ModalWithAlert.tsx
```diff
@@ -0,0 +1,55 @@
...
+	showAlertOnEscape: boolean;
+}
+
+const ModalWithAlert: React.FC<Props> = ({ children, showAlertOnEscape, ...props }) => {
+	const { isOpen, onOpen, onClose } = useDisclosure();
+	const cancelRef = React.useRef();
+	const cancelBtnText = props.cancelBtnText || __('No');
+	const header = props.header || __('Are you sure you want to close this?');
+	const okBtnText = props.okBtnText || __('Yes');
+	const onEscape = ({ keyCode }) => keyCode === ESCAPE && onOpen();
```
manzoorwanijk, line 24: Not good for TS, try to set the return type of the function to see the problem with that magical shortcut 😄

assets/src/domain/eventEditor/ui/ticketAssignmentsManager/components/ModalContainer.tsx
manzoorwanijk: I had that many times 😄

assets/src/application/ui/layout/modal/ModalWithAlert.tsx
nicolad: agree, made it TS-friendly:
```
	const onEscape = ({ keyCode }): void => {
		if (keyCode === ESCAPE) {
			onOpen();
		}
	};
```
***
was in one-liner mood before :smiley:

assets/src/domain/eventEditor/ui/ticketAssignmentsManager/components/ModalContainer.tsx
nicolad: I would prefer clear and clean code over minor performance optimization.
It's not the case here, but in order to know exactly what is the performance gain or loss, I'd have to measure both cases and see exact number. And, if I'd do that probably there wouldn't be very noticeable difference.

nicolad: However, let's take aside performance discussion.
I think there is a bit of the problem with `sprintf`.
As I understand `sprintf` is returning a string, then why I should care about these type coercions?
I think `sprintf` should be smart enough to this and maybe other check and I should just pass whatever I want. If I pass `undefined` or `null`, ignore it, otherwise do your smart job and please don't bother me with type coercion.

nicolad: I love the DX `classnames` have, I can pass whatever I want, it can be `null`, object or string and then it's smart enough to figure out by itself what needs to be the result.

nicolad: At some point, maybe, we should create our own wrapper for `sprintf` and make it a bit smarter.

nicolad: `classnames` is awesome:

```
classNames('foo', 'bar'); // => 'foo bar'
classNames('foo', { bar: true }); // => 'foo bar'
classNames({ 'foo-bar': true }); // => 'foo-bar'
classNames({ 'foo-bar': false }); // => ''
classNames({ foo: true }, { bar: true }); // => 'foo bar'
classNames({ foo: true, bar: true }); // => 'foo bar'

// lots of arguments of various types
classNames('foo', { bar: true, duck: false }, 'baz', { quux: true }); // => 'foo bar baz quux'

// other falsy values are just ignored
classNames(null, false, 'bar', undefined, 0, 1, { baz: null }, ''); // => 'bar 1'
```

assets/src/application/ui/layout/modal/ModalWithAlert.tsx
nicolad: I followed the same example as in chakra: https://github.com/chakra-ui/chakra-ui/blob/master/packages/chakra-ui/src/Modal/index.js#L131
***
but if the clarity is an issue, then it's better to make it more explicit, thx for letting me know this :+1:

assets/ZZZ/components/ui/styles/themes/default/index.js
```diff
@@ -1,8 +0,0 @@
-import './color-variables.css';
```
nicolad, line 1: was getting this error:

image

nicolad: we have this in `EDTR/refactor-master` in case we might need this, no need to keep this here

config/webpack.common.js
```diff
@@ -211,15 +211,6 @@ const config = [
...
 	},
-	{
-		entry: {
-			'eventespresso-core-css-default': pathToEDTRv1 + 'components/ui/styles/themes/default/index.js',
-		},
-		module: moduleConfigWithJsAndCssRules,
-		watchOptions: {
-			poll: 1000,
-		},
-	},
```
nicolad, line 222: attempt #2 for:

image

assets/src/domain/eventEditor/services/apollo/queries/prices/test/useFetchPrices.test.ts
```diff
@@ -38,7 +38,6 @@ describe('useFetchPrices', () => {
 
 		await waitForNextUpdate({ timeout }); // wait for response
 
-		expect(result.current.error).toBeDefined();
```
nicolad, line 41: locally I was getting error because of this line, weird

config/webpack.common.js
nicolad, line 222: this file will completely/ safely removed in https://github.com/eventespresso/event-espresso-core/pull/2609

package.json
```diff
@@ -95,6 +95,7 @@
 	},
 	"devDependencies": {
 		"@apollo/react-testing": "^3.1.4",
+		"@babel/compat-data": "^7.9.0",
```
nicolad, line 98: Added this temporary, to make this failing pass:

image

nicolad: As I wrote in another comment, this non-sense will be removed/ adjusted here: https://github.com/eventespresso/event-espresso-core/pull/2609

assets/src/domain/eventEditor/services/apollo/queries/prices/test/useFetchPrices.test.ts
tn3rb: @manzoorwanijk any ideas about this?

Issue #2710, opened 2020-04-20
Fix Unmet Peer Dependencies


tn3rb:
Have been noticing a slew of warnings when running Yarn updates:

```
warning " > @apollo/react-hooks@3.1.4" has incorrect peer dependency "graphql@^14.3.1".
warning "@apollo/react-hooks > @apollo/react-common@3.1.4" has unmet peer dependency "apollo-utilities@^1.3.2".
warning "@apollo/react-hooks > @apollo/react-common@3.1.4" has incorrect peer dependency "graphql@^14.3.1".
warning "@chakra-ui/core > toasted-notes > @reach/alert > @reach/component-component@0.1.3" has unmet peer dependency "prop-types@^15.6.2".
warning "@wordpress/components > @emotion/native@10.0.27" has unmet peer dependency "react-native@>=0.14.0 <1".
warning "@wordpress/components > react-dates > react-with-styles@3.2.3" has unmet peer dependency "react-with-direction@^1.1.0".
warning "@wordpress/url > react-native-url-polyfill@1.1.2" has unmet peer dependency "react-native@*".
warning "antd > rc-picker@1.4.2" has unmet peer dependency "dayjs@^1.8.18".
warning " > apollo-cache-inmemory@1.6.5" has incorrect peer dependency "graphql@0.11.7 || ^0.12.0 || ^0.13.0 || ^14.0.0".
warning "apollo-cache-inmemory > apollo-cache@1.3.4" has incorrect peer dependency "graphql@^0.11.0 || ^0.12.0 || ^0.13.0 || ^14.0.0".
warning "apollo-cache-inmemory > apollo-utilities@1.3.3" has incorrect peer dependency "graphql@^0.11.0 || ^0.12.0 || ^0.13.0 || ^14.0.0".
warning " > apollo-client@2.6.8" has incorrect peer dependency "graphql@^0.11.0 || ^0.12.0 || ^0.13.0 || ^14.0.0".
warning "apollo-client > apollo-link@1.2.13" has incorrect peer dependency "graphql@^0.11.3 || ^0.12.3 || ^0.13.0 || ^14.0.0".
warning " > apollo-link-batch-http@1.2.13" has incorrect peer dependency "graphql@^0.11.0 || ^0.12.3 || ^0.13.0 || ^14.0.0".
[...]

PR #2717, opened 2020-04-21
Fix Unmet Peer Dependencies #2710

nicolad:
- The title says it all


nicolad:
Known issue: https://github.com/chakra-ui/chakra-ui/issues/469

config/webpack.config.js
```diff
@@ -52,7 +51,7 @@ const sassModuleRegex = /\.module\.(scss|sass)$/;
 
 // This is the production and development configuration.
 // It is focused on developer experience, fast rebuilds, and a minimal bundle.
-module.exports = function(webpackEnv) {
+module.exports = function (webpackEnv) {
```
nicolad, line 54: prettier was here

package.json
```diff
@@ -20,7 +20,6 @@
 		"ci:build": "npm run prod",
 		"ci:test": "npm run test-js",
 		"dev": "node scripts/watch.js",
-		"generate": "graphql-codegen",
```
nicolad, line 23: this script isn't needed, it was used only once to generated type definition from server-side schema, but then we discussed that we want to do that manually

PR #2722, opened 2020-04-21
Fix zIndex for dropdown menu; closes #2687.

### LINKED ISSUE #2687: Fix zIndex for Chakra components
Tooltip is not displayed in Modal and dropdown hides behind wp admin menu.

image

Chakra provides some zIndex values out of the box to control the stacking order of components.
https://chakra-ui.com/theme#z-index

nicolad:
- Fixes `z-index` for `DropdownMenu`
- Closes https://github.com/eventespresso/event-espresso-core/issues/2687

## How has this been tested
image


assets/src/application/ui/layout/dropdownMenu/DropdownMenu.tsx
```diff
@@ -1,8 +1,10 @@
 import React from 'react';
 
+import { Menu } from '@infraUI/layout/menu';
+
 import DropdownMenuList from './DropdownMenuList';
 import DropdownToggle from './DropdownToggle';
-import { Menu } from '@infraUI/layout/menu';
+
```
nicolad, line 7: a bit of grouping

assets/src/application/ui/layout/dropdownMenu/DropdownMenuItem.tsx
```diff
@@ -4,10 +4,11 @@ import classNames from 'classnames';
 import { MenuItem } from '@infraUI/layout/menu';
 import { DropdownMenuItemProps } from './types';
 
-const DropdownMenuItem: React.FC<DropdownMenuItemProps> = ({ icon: Icon, title, className, ...rest }) => {
-	const newClassName = classNames('components-dropdown-menu__menu-item', className);
```
nicolad, line 8: `newClassName` name does not reflect what is happening here;
if this is new, then we should have old one somewhere?

assets/src/application/ui/layout/dropdownMenu/DropdownMenuList.tsx
```diff
@@ -2,11 +2,13 @@ import React, { Children, cloneElement } from 'react';
...
+import './style.scss';
+
+const DropdownMenuList: React.FC<MenuListProps> = ({ children, placement = 'top', ...props }) => {
+	const className = classNames('ee-dropdown-menu__list', props.className);
 
-const DropdownMenuList: React.FC<MenuListProps> = ({ children, className, placement = 'top', ...rest }) => {
-	const newClassName = classNames(' components-dropdown-menu', className);
 	return (
-		<MenuList placement={placement} {...rest} className={newClassName}>
+		<MenuList {...props} className={className} placement={placement}>
```
nicolad, line 11: putting `placement` prop in front doesn't make any difference in this specific case

assets/src/application/ui/layout/dropdownMenu/style.scss
```diff
@@ -0,0 +1,7 @@
+.ee-dropdown-menu {
+	&__list {
+		&[role='menu'] {
```
nicolad, line 3: this is for specificity ... and for `a11y` :)

assets/src/application/services/theme/theme.ts
```diff
@@ -2,8 +2,11 @@ import { theme as chakraTheme, DefaultTheme } from '@chakra-ui/core';
 
 type CustomTheme = DefaultTheme;
 
+const zIndices = { ...chakraTheme.zIndices, popover: 1501 };
```
nicolad, line 5: Tweaked this, but no luck: 

image

nicolad: Based on my finding, that `z-index=1` is overriding the theme value here: https://github.com/chakra-ui/chakra-ui/blob/master/packages/chakra-ui/src/Menu/index.js#L334
***
so, it won't help by adjuting values at the theme level, in this case

nicolad:
image

assets/src/application/ui/input/Button/style.scss
```diff
@@ -0,0 +1,491 @@
...
+		}
+	}
+
+	&:not(.ant-btn-icon-only),
+	&:active:not(.ant-btn-icon-only),
+	&:focus:not(.ant-btn-icon-only),
+	&:visited:not(.ant-btn-icon-only) {
+		// box-sizing: content-box !important; // Need to ask/ investigate this
+		// min-height: var(--ee-font-size-big); // Also this is questionable
+		// padding: var(--ee-padding-tiny) var(--ee-padding-default) var(--ee-padding-tiny) var(--ee-padding-small); // Also this is questionable
```
nicolad, line 83: these styles was causing this:

Peek 2020-04-22 14-49

assets/src/infrastructure/ui/inputs/dateTime/style.scss
```diff
@@ -0,0 +1,3 @@
+[data-baseweb='popover'] {
+	z-index: 2;
```
nicolad, line 2: `2` was needed because these arrows were visible:

image

nicolad: end result:

image
***
image

nicolad: this also fixed datepicker
***
before:

image
***
after:

image

nicolad: I am not :100: % happy with `[data-baseweb='popover']` selector, before going with that route I tried:
- passing a className to `baseui`'s `TimePicker`, such that I can have a custom control, but that won't accept `className` prop :disappointed: 
- tried to apply styles to `role="listbox"` , but that was a nested child node and wasn't applying the fix

assets/src/application/ui/input/Button/style.scss
```diff
@@ -46,7 +46,7 @@
 		font-size: var(--ee-font-size-default);
 		font-weight: bold;
 		line-height: var(--ee-font-size-default);
-		margin: 0 0 var(--ee-margin-smaller);
+		// margin: 0 0 var(--ee-margin-smaller); Setting margins on these states is a bit confusing, needs discussion.
```
nicolad, line 49: this margin was causing this issue:

image

nicolad: I don't quite understand the reasoning behind this, it might be trying to override other styles, but not sure which one and not sure if the is still actual.

assets/src/application/ui/display/entityDetailsPanel/style.scss
```diff
@@ -50,18 +50,20 @@
...
 			box-shadow: none;
 			-webkit-box-sizing: border-box;
 			-moz-box-sizing: border-box;
 			box-sizing: border-box;
 			color: var(--ee-default-text-color);
-			display: block;
+			display: inline-flex;
 			font-size: var(--ee-font-size-default);
 			height: var(--ee-icon-button-size);
+			justify-content: center;
```
nicolad, line 66: Before:

image
***
after:

image

assets/src/application/ui/input/Button/style.scss
```diff
@@ -0,0 +1,494 @@
...
+			}
+		}
+
+		&:not(.ant-btn-icon-only),
+		&:active:not(.ant-btn-icon-only),
+		&:focus:not(.ant-btn-icon-only),
+		&:visited:not(.ant-btn-icon-only) {
+			// box-sizing: content-box !important; // Need to ask/ investigate this
+			// min-height: var(--ee-font-size-big); // Also this is questionable
+			// padding: var(--ee-padding-tiny) var(--ee-padding-default) var(--ee-padding-tiny) var(--ee-padding-small); // Also this is questionable
```
nicolad, line 89: @tn3rb I need your :brain: here, because I am not aware of the previous context;
are these still actual in event editor? because these are causing some trouble

assets/src/application/ui/input/Button/style.scss
tn3rb: those styles were not intended to be applied to icon buttons

assets/src/application/ui/display/entityDetailsPanel/style.scss
tn3rb: beauty eh 👍

assets/src/application/ui/input/Button/IconButton/index.tsx
```diff
@@ -2,24 +2,19 @@ import React from 'react';
...
+
+import './style.scss';
 
 type BtnType = React.ComponentType<IconButtonProps>;
 
 const IconButton = React.forwardRef<typeof IconButtonAdapter, IconButtonAdapterProps>(
-	({ className, icon, onClick, ...props }, ref) => {
-		const newClassName = classNames('esprs-button', className);
+	({ icon, onClick, ...props }, ref) => {
+		const className = classNames('ee-button ee-icon-button', props.className);
```
tn3rb, line 14: might cause less conflicts (although potentially less DRY) to not apply `ee-button` class to icon buttons
i think before I was trying to do too much with the espresso button component

if there is a lot of duplication for the styles but we want to keep things DRY, then maybe we can add an `ee-btn-base` class that applies the common styles for both kinds of buttons

nicolad:
There is some differences on how border is applied in different cases, part of this is because I've borrowed some stuff from v1:

image
***
I would like to make it a little more consisten, for example I see box-shadows being used in few cases for hover and focus, I think this is a good approach since it won't affect the inner sizes of the element.
But this is a bit out of the scope of the `z-index`, so I'd like to improve further in a separate PR, if that's ok.

assets/src/application/ui/input/Button/style.scss
```diff
@@ -0,0 +1,493 @@
...
+ * Event Espresso Button Stylesheet
+ * © 2020 Event Espresso
+ * ------------------------------------------------------------------------- */
+
+:root {
+	--ee-icon-button-size: 36px;
+}
+
+.ee-event-editor {
+	.ee-button.ee-button {
```
tn3rb, line 11: oooo i just realized that this won't work as soon as we add a new domain other than the event editor...

maybe using an `ee-btn-base` class for common button styles would work though because then we can target `.ee-btn-base.ee-button` which would NOT apply to any legacy buttons

nicolad:
Follow-up issue: https://github.com/eventespresso/event-espresso-core/issues/2725

assets/src/application/ui/input/Button/style.scss
nicolad: ok, thx

assets/src/application/ui/input/Button/style.scss
nicolad: thx, good suggestion

nicolad:
Passed suggestion further: https://github.com/eventespresso/event-espresso-core/issues/2725#issuecomment-617909573

PR #2732, opened 2020-04-27
Load Bundled WP GQL on EE Activation | Closes #2711

### LINKED ISSUE #2711: Load Bundled WP GQL on EE Activation and Flush Permalinks
On a new activation of EE, we need to ensure that the bundled WP GQL plugin loads so that it's rewrite rules for the `/graphql/` can be set properly, otherwise that endpoint will be a 404

manzoorwanijk:
This PR registers plugin activation and deactivation hooks to run the corresponding WP GraphQL callbacks.

### DIFF
--- a/espresso.php
+++ b/espresso.php
@@ -129,8 +129,6 @@ if (function_exists('espresso_version')) {
          */
         function espresso_plugin_deactivation()
         {
-            update_option('ee_espresso_activation', true);
-
             // Run WP GraphQL deactivation callback
             if (PHP_VERSION_ID < 70000) {
                 return;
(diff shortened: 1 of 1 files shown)


espresso.php
```diff
@@ -111,10 +111,37 @@ function espresso_version()
...
         }
 
         register_activation_hook(EVENT_ESPRESSO_MAIN_FILE, 'espresso_plugin_activation');
 
+        /**
+         * espresso_plugin_deactivation
+         */
+        function espresso_plugin_deactivation()
+        {
+            update_option('ee_espresso_activation', true);
```
tn3rb, line 132: don't need this... so plz remove

PR #2734, opened 2020-04-27
Uniformize hover and focus state border; closes #2725.

### LINKED ISSUE #2725: Refactor styles
Currently, we have 2 approaches when applying border to a button or icon button in hover/focust state:
1. using border
2. using box-shadow

I think we need to make it uniform, such that it will use only `box-shadow`, the reason being that it won't affect the element inner sizing.

nicolad:
- Refactor based on this: https://github.com/eventespresso/event-espresso-core/issues/2725#issuecomment-618629015
- Closes: https://github.com/eventespresso/event-espresso-core/issues/2725

### DIFF
--- a/assets/src/application/ui/input/Button/IconButton/index.tsx
+++ b/assets/src/application/ui/input/Button/IconButton/index.tsx
@@ -11,7 +11,7 @@ type BtnType = React.ComponentType<IconButtonProps>;
 
 const IconButton = React.forwardRef<typeof IconButtonAdapter, IconButtonAdapterProps>(
 	({ icon, onClick, ...props }, ref) => {
-		const className = classNames('ee-btn-base ee-icon-button', props.className);
+		const className = classNames('ee-btn-base ee-icon-btn', props.className);
 
 		return (
 			<IconButtonAdapter {...props} className={className} icon={icon} onClick={onClick} tabIndex={0} ref={ref} />
--- a/assets/src/application/ui/input/Button/IconButton/index.tsx
+++ b/assets/src/application/ui/input/Button/IconButton/index.tsx
@@ -11,7 +11,7 @@ type BtnType = React.ComponentType<IconButtonProps>;
 
 const IconButton = React.forwardRef<typeof IconButtonAdapter, IconButtonAdapterProps>(
 	({ icon, onClick, ...props }, ref) => {
-		const className = classNames('ee-btn-base ee-icon-btn', props.className);
+		const className = classNames('ee-icon-btn', props.className);
 
 		return (
 			<IconButtonAdapter {...props} className={className} icon={icon} onClick={onClick} tabIndex={0} ref={ref} />
--- a/assets/src/application/ui/layout/entityList/RegistrationsLink/index.tsx
+++ b/assets/src/application/ui/layout/entityList/RegistrationsLink/index.tsx
@@ -14,7 +14,12 @@ interface Props {
 const RegistrationsLink: React.FC<Props> = ({ href, title }) => {
 	return (
 		<Tooltip title={title}>
-			<a href={href} className={'ee-editor-details-reg-url-link'} target={'_blank'} rel={'noopener norefferer'}>
+			<a
+				href={href}
+				className={'ee-btn-base ee-editor-details-reg-url-link'}
+				target={'_blank'}
+				rel={'noopener norefferer'}
+			>
 				<Groups svgSize={24} />
 			</a>
 		</Tooltip>
--- a/assets/src/infrastructure/ui/inputs/style.scss
+++ b/assets/src/infrastructure/ui/inputs/style.scss
@@ -1,0 +1,78 @@
+.input-select-textarea-hover {
+	border-color: var(--ee-color-primary-low-contrast);
+	color: var(--ee-color-primary-high-contrast);
+	box-shadow: 0 0 0 1px var(--ee-color-primary-low-contrast);
+}
+
+.input-select-textarea-focus {
+	border-color: var(--ee-color-primary);
+	box-shadow: 0 0 0 1px var(--ee-color-primary);
+}
+
+.ee-input-base {
+	&.ee-input-base {
+		border-color: var(--ee-border-color);
+		border-style: solid;
+		border-width: 2px;
+		&:focus {
+			/* Visible in Windows high-contrast themes */
+			outline-color: transparent;
+			outline-width: 2px;
+			outline-style: solid;
+		}
+	}
+
+	&.ee-date-picker.ee-date-picker {
+		input[type='text'] {
+			&:hover {
+				@extend .input-select-textarea-hover;
+			}
+			&:focus {
+				@extend .input-select-textarea-focus;
+			}
+		}
+	}
+
+	&.ee-input {
+		@at-root {
+			input[type='text']#{&},
+			input[type='password']#{&},
+			input[type='date']#{&},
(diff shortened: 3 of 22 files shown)


nicolad:
image

nicolad:
image

nicolad:
image

nicolad:
image

nicolad:
image

nicolad:
image

nicolad:
image

nicolad:
image

nicolad:
image

nicolad:
image

nicolad:
image

nicolad:
@tn3rb need your :brain: here, because didn't noticed such case of on `EDTR/refactori-master`:

image

assets/src/application/ui/input/Button/IconButton/index.tsx
```diff
@@ -11,7 +11,7 @@ type BtnType = React.ComponentType<IconButtonProps>;
 
 const IconButton = React.forwardRef<typeof IconButtonAdapter, IconButtonAdapterProps>(
 	({ icon, onClick, ...props }, ref) => {
-		const className = classNames('ee-btn-base ee-icon-button', props.className);
+		const className = classNames('ee-btn-base ee-icon-btn', props.className);
```
nicolad, line 14: made it consistent to `ee-btn-base`

tn3rb:
In some ways things are better and in some ways things are not so betterer...

your code changes look good, but I think all of the transitions from EDTR v1 and WordPress Gutenberg, to Blueprint, to Ant Design, and finally (hahaha) to Chakra have just made a mess of things (especially my lame attempts to fix the Ant-D styles - sorry!).

so I'm wondering if it might be easier in the long run to attack this from a different angle:

1. start by performing a "reset" by going through the editor and commenting out a lot of the stylesheets that are applying styles to non-layout elements like buttons, headings, links, inputs, etc. So when this is done, everything except styles for layout elements like entity cards, should be reset to browser defaults.

2. then we can create a base stylesheet for holding some really generic "global" type styles to function as a default style guide of sorts. These could be encoded using variables so that they can easily be pulled into any other stylesheets. This would include our standard text, border, and color styles. We already have some of this in the `assets/src/application/ui/styles` that I created for EDTR v1 but it could be better organized

3. then we create separate PRs for refactoring individual components so that we can strip out all of the leftover cruft from Ant-D and other libraries, starting with the low-level elements in the application layer. So this would be things like buttons, inputs, etc

4. then finally we create separate PRs for refactoring individual components from the domain layer to handle any cases where the default styles are not desirable and/or don't work for whatever reason (like that globe icon button in the entity sidebars). Hopefully we can keep this to a minimum.

**ALL** of the above could either be done :

- progressively in as the smallest chunks possible and merged into EDTR/master fast and furiously 

or

- progressively merged into another branch separate from EDTR/master where we constantly rebase onto EDTR/master every time it is updated. This would might allow us to gut things with less disruption to work being done to EDTR/master, but this might also be more difficult due to the possibility of conflicts when rebasing onto master.

Anyways... the above is just me "thinking out loud" and not necessarily what I really want to do... very open to hear how you think it would be best to clean up the mish-mash of mixed up classes, styles, and formats, we have going on.

If we don't want to bother trying to do anything like the above and instead just want to keep chipping away at things until we're totally happy with the results, then here's some issues I've noticed with this current branch (again, I think a lot of these issues are caused by all of the switching between different libraries, as well as the lack of a solid foundation to begin with)

---

### filter bar inputs

have different border width:

ScreenShot_20200427155104

i think our default border style for buttons, inputs, and selects should be `2px solid var(ee-color-grey-10)` with the color changing to `var(--ee-color-primary-super-low-contrast)` on hover and `var(--ee-color-primary)` on focus (similar to what I said in 2725 but with a slightly lighter hover color)

---

### icon buttons on the entity cards

should be borderless until hovered or focused. So the reg links should not have a border:

ScreenShot_20200427155433

and we also don't want the border width to change upon focus:

ScreenShot_20200427155648

---

### timezone info icon button

similarly the timezone info icon button in the sidebar should be borderless until hovered or focused

ScreenShot_20200427155222

and then the active and focus states can be `var(--ee-color-primary-super-low-contrast)` on hover and `var(--ee-color-white)` on focus, instead of combining border color and shadow:

ScreenShot_20200427155740

---

### other button styles

other button styles are really inconsistent, with some looking  like EDTR v1 and others looking like a cross between Ant-D and Chakra, such as on the edit form modals:

ScreenShot_20200427162511

or the close confirmation popup:

ScreenShot_20200427191246

---

### style "jumps"

as well, some of the styles switch drastically when clicking on them, such as the "Add New Date" button:

ScreenShot_20200427190823

on click
ScreenShot_20200427161641

---

### LOLZ

Weirdest issue I discovered is the close button on the edit form modal:
ScreenShot_20200427162226

which jumps waaaaaaaay out of place, NOT when you click on that button, or hover over it, or interact with it in any way... but when you **edit the form** 
ScreenShot_20200427162451

---

### text inputs

Some other minor issues include the differences between the inline edit text input and the inline edit textarea:

ScreenShot_20200427155606
ScreenShot_20200427155627

as well as the datepicker styles which you were already aware of some issues, but did you also notice the light grey outline/background around the datepickers:
ScreenShot_20200427161711

---

Anyways... sorry for the novel, hard to address all of these little issues any other way.

So I really want to hear if you have any thoughts on a better way to address all of these issues and clean up the inconsistencies and mess that is there now. And just to be 100000% clear, you are not responsible for the current mess in any way. I know that it is primarily the result of my decisions to switch between various UI libraries, which I don't regret because I think it will ultimately result in us being in a better place, but I also don't want to be prescribing an approach to fixing things that you may not be happy with. So as always, plz feel free to present any ideas you have with regards to this, because I want solutions that work for all of us

nicolad:
> solid foundation


I like your out-load thought process and I also think that it's better to have a solid foundation. 

I have a comment on item 1, instead of relying on browser's default, let's use a normalize or smth similar. In this article it's described how browser default have some mismatch between FF and Chrome: https://css-tricks.com/overriding-default-button-styles/

nicolad:
Need to get rid of this stylesheet on event editor, is this possible?

image

assets/src/application/ui/layout/entityList/RegistrationsLink/index.tsx
```diff
@@ -14,7 +14,12 @@ interface Props {
 const RegistrationsLink: React.FC<Props> = ({ href, title }) => {
 	return (
 		<Tooltip title={title}>
-			<a href={href} className={'ee-editor-details-reg-url-link'} target={'_blank'} rel={'noopener norefferer'}>
+			<a
```
nicolad, line 17: @tn3rb I think we might better have a `Clickable` component here which can be a wrapper that applies button styles to an anchor.
An example: https://reakit.io/docs/clickable/

assets/src/infrastructure/ui/inputs/style.scss
```diff
@@ -0,0 +1,78 @@
...
+			input[type='text']#{&},
+			input[type='time']#{&},
+			input[type='url']#{&},
+			input[type='week']#{&} {
+				&:hover {
+					@extend .input-select-textarea-hover;
+				}
+				&:focus {
+					@extend .input-select-textarea-focus;
+				}
```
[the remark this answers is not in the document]
nicolad: it was kinda last resort solution, I am not very happy with how hard is this to read and comprehend

Issue #2418, opened 2020-02-20
Improve server-side GQL mutation error messages


manzoorwanijk:
Currently most of the GQL mutation error messages are too generic and provide lesser info about what went wrong.
They come from `mutators` folder inside `core/domain/services/graphql`.

Although, we cannot display every error message in the output, there could be some debugging mechanism that uses `if ( defined( 'GRAPHQL_DEBUG' ) )` and logs the error messages.

manzoorwanijk:
Closed via #2735

PR #2735, opened 2020-04-27
Refactor Server-Side GQL Mutation Error Handling

tn3rb:
- see #2418

- adds abstract parent class for mutation handlers

- extracts duplicated logic into above class

- refactors mutators to utilize shared logic and provide better error messages

### DIFF
--- a/core/domain/services/graphql/mutators/EntityMutator.php
+++ b/core/domain/services/graphql/mutators/EntityMutator.php
@@ -1,0 +1,179 @@
+<?php
+
+namespace EventEspresso\core\domain\services\graphql\mutators;
+
+use EE_Base_Class;
+use EEM_Base;
+use Exception;
+use GraphQL\Error\FormattedError;
+use GraphQL\Error\UserError;
+use GraphQLRelay\Relay;
+use OutOfBoundsException;
+use RuntimeException;
+use Throwable;
+
+/**
+ * Class EntityMutator
+ * Description
+ *
+ * @package EventEspresso\core\domain\services\graphql\mutators
+ * @author  Brent Christensen
+ * @since   $VID:$
+ */
+abstract class EntityMutator
+{
+
+    /**
+     * @param EEM_Base $model
+     * @param integer $ID
+     * @param string   $capability
+     * @return EE_Base_Class
+     * @throws OutOfBoundsException
+     * @throws UserError
+     * @since $VID:$
+     */
+    protected static function getEntityFromID(EEM_Base $model, $ID, $capability = 'ee_edit_events')
+    {
+        EntityMutator::checkPermissions($model, $capability);
+        return EntityMutator::getEntity($model, $ID);
+    }
+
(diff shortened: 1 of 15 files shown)


core/domain/services/graphql/mutators/EntityMutator.php
```diff
@@ -0,0 +1,177 @@
...
+     * @return EE_Base_Class
+     * @throws OutOfBoundsException
+     * @throws UserError
+     * @since $VID:$
+     */
+    protected static function getEntityFromInputData(EEM_Base $model, array $input, $capability = 'ee_edit_events')
+    {
+        EntityMutator::checkPermissions($model, $capability);
+        $ID = EntityMutator::getEntityIDFromGlobalId($model, $input);
+        return EntityMutator::getEntity($model, $ID);
```
manzoorwanijk, line 52: Why not this?
```suggestion
        $ID = self::getEntityIDFromGlobalId($model, $input);
        return self::getEntity($model, $ID);
```

manzoorwanijk:
Permanent deletion of a trashed entity results in an error which doesn't give much info. Can you please try that?

core/domain/services/graphql/mutators/EntityMutator.php
tn3rb: I prefer using the class name over `self` for a few reasons:

- in scenarios where you are just looking at a code snippet or diff where there isn't any greater context, then seeing something like `self::someMethod()` tells you NOTHING about what class the code is coming from.

- some linters give warnings about using `self` and suggest changing to full class names

- maybe in VS Code it is a pain to have to type out the full class name but in PhpStorm it literally only have to type the first couple letters of the class name and the intellisense will pop up with the current class as the first option. So using the full class name almost always requires less key presses than typing `self`

but the biggest reason of all is:

- using explicit class names instead of `self` absolutely 100% prevents copy pasta errors. If you copy n paste a method or block of code from one class into another, and that block of code happens to utilize `self` somewhere, then it is _possible_ that you could miss that and forget to change it to reference the previous class, especially if both classes have a method with the same name (like `self::ID()`).

manzoorwanijk: TIL, thank you for the explanation 👍

tn3rb:
> Permanent deletion of a trashed entity results in an error which doesn't give much info

I'm seeing this in the network response:

```
[{"errors":[{"message":"Internal server error","category":"internal","locations":[{"line":2,"column":3}],"path":["deleteEspressoDatetime"]}],"data":{"deleteEspressoDatetime":null}}]
```

and this is what's sent in the request:

```
input: {clientMutationId: "DELETE_DATETIME", id: "RGF0ZXRpbWU6NQ==", deletePermanently: true}
clientMutationId: "DELETE_DATETIME"
deletePermanently: true
id: "RGF0ZXRpbWU6NQ=="
```

PR #2737, opened 2020-04-28
Fix priority for filters and sorters | Closes #2736

### LINKED ISSUE #2736: Priority for entity list filters and sorters doesn't seem to be wired up
It looks like  we missed the priority arg for filters and sorters due to which it's possible that the result may not be as desired. For example, we need to make sure that sales filter for dates should run after status filter even if later is registered first.

manzoorwanijk:
This PR adds support for priority argument for filters and sorters.

### DIFF
--- a/assets/src/application/ui/layout/entityList/filterBar/filterState/useEntityFilterService.ts
+++ b/assets/src/application/ui/layout/entityList/filterBar/filterState/useEntityFilterService.ts
@@ -3,6 +3,7 @@ import { Entity } from '@appServices/apollo/types';
 import { EntityFilterService, EntityListFilterStateManager } from './types';
 import { useFilterBarService, FilterBarServiceCbArgs } from '../subscription';
 import { SubscriptionCallback } from '@appServices/subscription';
+import { sortBy, pathOr } from 'ramda';
 
 type ELFSM = EntityListFilterStateManager;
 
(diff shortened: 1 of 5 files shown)


assets/src/application/ui/layout/entityList/filterBar/filterState/useEntityFilterService.ts
```diff
@@ -17,7 +18,8 @@ const useEntityFilterService = <D extends string, L extends string, E extends En
 	const getCallbackList = (
 		mappedCallbackList: ReturnType<typeof getFilters>
 	): Array<SubscriptionCallback<FilterBarServiceCbArgs<E, ELFSM>, E[]>> => {
-		return Object.values(mappedCallbackList).map(({ callback }) => callback);
+		const subscriptions = sortBy(pathOr(10, ['options', 'priority']), Object.values(mappedCallbackList));
```
nicolad, line 21: nice extraction :+1:

assets/src/domain/eventEditor/services/filterState/tickets/useTicketsFilterBarService.ts
```diff
@@ -39,7 +39,7 @@ const useTicketsFilterBarService = (): void => {
 		// Register isChained filter
 		const unSubscribeIsChainedFilter = registerTicketsFilter(({ entityList, filterState }) => {
 			return isChainedFilter({ isChained: filterState.isChained, tickets: entityList });
-		});
+		}, 9); // we want isChained to run first
```
nicolad, line 42: :+1:

PR #2739, opened 2020-04-28
Fix stale entities for menu items | Closes #2738

### LINKED ISSUE #2738: Entity actions menu items use state entities
If you make changes to an entity and then create a copy of it, then latest changes won't be copied.

manzoorwanijk:
This PR fixes the stale entities of entity actions menu items.

### DIFF
--- a/assets/src/domain/eventEditor/services/apollo/queries/tickets/useTickets.ts
+++ b/assets/src/domain/eventEditor/services/apollo/queries/tickets/useTickets.ts
@@ -7,7 +7,7 @@ import { getCacheIds } from '@appServices/predicates';
 
 const useTickets = (): Array<Ticket> => {
 	const options = useTicketQueryOptions();
-	const { data } = useCacheQuery<TicketsList>({ ...options, fetchPolicy: 'cache-first' });
+	const { data } = useCacheQuery<TicketsList>(options);
 
 	const nodes = data?.espressoTickets?.nodes || [];
 
--- a/assets/src/domain/eventEditor/ui/datetimes/datesList/actionsMenu/dropdown/DateMainMenu.tsx
+++ b/assets/src/domain/eventEditor/ui/datetimes/datesList/actionsMenu/dropdown/DateMainMenu.tsx
@@ -9,16 +9,16 @@ import TrashDate from './TrashDate';
 
 import { DateMainMenuProps } from './types';
 
-const DateMainMenu: React.FC<DateMainMenuProps> = ({ datetime }) => {
+const DateMainMenu: React.FC<DateMainMenuProps> = ({ id }) => {
 	const toggleProps: DropdownToggleProps = {
 		tooltip: __('event date main menu'),
 	};
 
 	return (
 		<DropdownMenu toggleProps={toggleProps}>
-			<EditDate datetime={datetime} />
-			<CopyDate datetime={datetime} />
-			<TrashDate datetime={datetime} />
+			<EditDate id={id} />
+			<CopyDate id={id} />
+			<TrashDate id={id} />
 		</DropdownMenu>
 	);
 };
--- a/assets/src/domain/eventEditor/ui/datetimes/datesList/actionsMenu/dropdown/DateMainMenu.tsx
+++ b/assets/src/domain/eventEditor/ui/datetimes/datesList/actionsMenu/dropdown/DateMainMenu.tsx
@@ -6,19 +6,23 @@ import { DropdownMenu, DropdownToggleProps } from '@application/ui/layout';
 import CopyDate from './CopyDate';
 import EditDate from './EditDate';
 import TrashDate from './TrashDate';
+import { useDatetimeItem } from '@edtrServices/apollo';
 
 import { DateMainMenuProps } from './types';
 
-const DateMainMenu: React.FC<DateMainMenuProps> = ({ id }) => {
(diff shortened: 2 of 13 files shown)


assets/src/domain/eventEditor/services/apollo/queries/tickets/useTickets.ts
```diff
@@ -7,7 +7,7 @@ import { getCacheIds } from '@appServices/predicates';
 
 const useTickets = (): Array<Ticket> => {
 	const options = useTicketQueryOptions();
-	const { data } = useCacheQuery<TicketsList>({ ...options, fetchPolicy: 'cache-first' });
+	const { data } = useCacheQuery<TicketsList>(options);
```
manzoorwanijk, line 10: This caused refetchiing of tickets when deleting or creating a datetime.

tn3rb: oh wasn't that added to fix another issue though?

assets/src/domain/eventEditor/ui/datetimes/datesList/actionsMenu/dropdown/DateMainMenu.tsx
```diff
@@ -6,10 +6,14 @@ import { DropdownMenu, DropdownToggleProps } from '@application/ui/layout';
...
 import TrashDate from './TrashDate';
+import { useDatetimeItem } from '@edtrServices/apollo';
 
 import { DateMainMenuProps } from './types';
 
-const DateMainMenu: React.FC<DateMainMenuProps> = ({ datetime }) => {
+const DateMainMenu: React.FC<DateMainMenuProps> = ({ datetime: entity }) => {
+	// Make sure to subscribe to Apollo cache
+	// to avoid stale data
+	const datetime = useDatetimeItem({ id: entity.id });
```
tn3rb, line 16: another possible way to do this would be to use `Entity.cacheId` to key the components that receive an entity as a prop.

So instead of having to have every component that needs an entity have to go through the context provider to get it, and assuming the entity can be passed down from a parent component that is also using the entity, we would instead do something like:

```jsx
const ComponentThatGetsEntityFromContext = () => {
	const entity = useEntityContext();
	return (
		<div>
			<ComponentA key={ entity.cacheId } entity={ entity } />
			<SomeOtherComponent />
			<ComponentB key={ entity.cacheId } entity={ entity } />
		</div>
	);
}
```

so instead of ComponentA and ComponentB also having to retrieve the entity from the context provider, they will be guaranteed to get refreshed when the `cacheID` changes because that is being used to key the component. This might be more performant than having to constantly go through the context provider. That said, we don't want to courier an entity through components that don't need it themselves, so when there is no direct connection to the entity then **that's** when we use the context provider. But if the parent is already using the same entity, then let's avoid always relying on the context.

Just got switched over to EDTR/master and looked at the code and this is definitely a case that needs to use the context provider, so nothing I wrote above applies here 😕

manzoorwanijk: I have been thinking about that for quite a while now. My thought was to create something like  `useDatetime()` and `useTicket()` which would fetch the entity fed to card context.

tn3rb: ya that's fine, but I think that should only be used if the parent component does not already receive the entity via props or via the context provider (or custom hook like you are suggesting).

Consider the following hierarchy:

```jsx
	<TopLevelComponent> // <== uses entity
		<FirstLevel> // <== uses entity
			<SecondLevel> // <== uses entity
				<ThirdLevel>
					<FourthLevel>
						<FifthLevel>
							<SixthLevel></SixthLevel> // <== uses entity
						</FifthLevel>
					</FourthLevel>
				</ThirdLevel>
			</SecondLevel>
		</FirstLevel>
	</TopLevelComponent>
```

If the `TopLevelComponent` is pulling the entity from the Context Provider and uses it in some way, and then the `FirstLevel`, `SecondLevel`, and `SixthLevel` components also require the entity, then I think the First and Second level ones should receive the entity via a prop because it can be passed directly, but the `SixthLevel` component should use the context provider because its parent component has no need for the entity.

So imho, the entity access should be like this:

```jsx
	<TopLevelComponent> // <== uses Context Provider
		<FirstLevel> // <== entity passed via prop
			<SecondLevel> // <== entity passed via prop
				<ThirdLevel>
					<FourthLevel>
						<FifthLevel>
							<SixthLevel></SixthLevel> // <== uses Context Provider
						</FifthLevel>
					</FourthLevel>
				</ThirdLevel>
			</SecondLevel>
		</FirstLevel>
	</TopLevelComponent>
```

Then for the components that receive the entity via prop, they should use the entity cacheID as a key which will guarantee a rerender when that entity changes.

ANOTHER potential approach is to use the Data Container => JSX approach and keep all of the data management aspects of our logic in data containers and then have dumb UI components that only render JSX. The data containers could utilize memoization and what not for cache control and the templates could use cacheIDs for keys.

As well, the data containers could either be wrapper components that would render the JSX component, OR they could instead be custom hooks that are simply called within the JSX component, although I suspect this may result in more unnecessary rerenders.

nicolad: +1 for data containers :+1:

PR #2745, opened 2020-04-29
Fix Drag-n-Drop when not sorting | Closes #2743

### LINKED ISSUE #2743: Table View Rows Can Be "Dragged n Dropped" Even If Sorting Has Not Been Enabled
Title says it all:

- click on the Table View button in the Entity List filter bar

- "grab" a row

- "drag" a row

- "drop" a row (doesn't actually result in order being changed)

- works even when there is only ONE row in the Entity List

manzoorwanijk:
This PR disables drag and drop when sorting is not enabled.

### DIFF
--- a/assets/src/application/ui/layout/entityList/EntityTable.tsx
+++ b/assets/src/application/ui/layout/entityList/EntityTable.tsx
@@ -28,6 +28,7 @@ const EntityTable = <E extends Entity, FS extends ELFSM>({
 		tableId,
 		tableCaption,
 	};
+	const onDragEnd = filterState.sortingEnabled ? onSort : null;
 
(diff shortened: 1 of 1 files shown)


assets/src/application/ui/layout/entityList/EntityTable.tsx
```diff
@@ -28,14 +28,15 @@ const EntityTable = <E extends Entity, FS extends ELFSM>({
 		tableId,
 		tableCaption,
 	};
+	const onDragEnd = filterState.sortingEnabled ? onSort : null;
```
nicolad, line 31: makes sense :+1:

PR #2746, opened 2020-04-29
Button styles refactor; partially closes #2734.

nicolad:
- Partially closes: https://github.com/eventespresso/event-espresso-core/issues/2725


nicolad:
This is default appearance of button w/o the `CSSReset` from `ThemeProvider`

image
***
but with reset some of the noise is removed:

image

nicolad:
Button is picking default styles from here:

image

***
hope that's ok, because those default button styles overrides those recommended 
 here: https://github.com/eventespresso/event-espresso-core/issues/2725#issuecomment-618629015:

image

assets/src/application/ui/layout/entityActionsMenu/style.scss
```diff
@@ -0,0 +1,26 @@
...
+	}
+
+	.ee-btn-default.ee-btn:focus,
+	.ee-btn-primary.ee-btn:focus,
+	.ee-btn-secondary.ee-btn:focus,
+	.ee-btn-accent.ee-btn:focus {
+		border: 1px solid var(--ee-color-primary-low-contrast);
+		box-shadow: 0 0 0 1px var(--ee-color-primary-low-contrast);
+	}
+}
```
nicolad, line 26: move these here since it's more close to the source where these are used, but anyway will have to adjust to look like this:
image

***
and more likely that those changes will be applied in dropdown menu item component

assets/src/domain/eventEditor/ui/datetimes/datesList/actionsMenu/DateActionsMenu.tsx
```diff
@@ -1,18 +1,14 @@
...
 
-const DateActionsMenu: React.FC<ActionsMenuComponentProps<Datetime>> = ({ entity, ...menuProps }) => {
+const DateActionsMenu: React.FC<ActionsMenuComponentProps<Datetime>> = ({ entity, ...props }) => {
 	const menuItems = useDatesActionMenuItems(entity);
 
-	return (
-		<div className='ee-entity-menu' {...menuProps}>
-			{menuItems}
-		</div>
-	);
```
nicolad, line 15: duplicated by ticket actions menu

nicolad:
next related things to do after this PR:
- update icon button
- refactor `EntityActionsMenu` since it also uses button

assets/src/application/ui/input/Button/Button.tsx
```diff
@@ -17,28 +17,19 @@ type BtnType = React.ComponentType<ButtonProps>;
...
 			'ee-btn-default': buttonType === ButtonType.DEFAULT,
+			'ee-btn-huge': buttonSize === ButtonSize.HUGE,
 			'ee-btn-primary': buttonType === ButtonType.PRIMARY,
 			'ee-btn-minimal': buttonType === ButtonType.MINIMAL,
 			'ee-btn-secondary': buttonType === ButtonType.SECONDARY,
-			'ee-btn-tiny': buttonSize === ButtonSize.TINY,
 			'ee-btn-small': buttonSize === ButtonSize.SMALL,
-			'ee-btn-big': buttonSize === ButtonSize.BIG,
-			'ee-btn-huge': buttonSize === ButtonSize.HUGE,
+			'ee-btn-tiny': buttonSize === ButtonSize.TINY,
```
tn3rb, line 32: i get that you want to alphabetize things but imho it is worse to break up the logical groupings, which makes it more difficult to differentiate `ButtonType`s from `ButtonSize`es.

That said, what would have been better is if I made the values for those enums lowercase so that we could use them directly to generate these classes, ie:

```ts
const className = classNames('ee-btn-base ee-btn', `ee-btn--${buttonSize}`, `ee-btn--${buttonType}`, props.className);
```

what do you think about doing that instead?

assets/src/application/ui/layout/entityActionsMenu/EntityActionsMenu.tsx
```diff
@@ -1,32 +1,30 @@
...
-	className = classNames({
-		[className]: className,
-		'ee-entity-actions-menu': true,
+	const className = classNames(props.className, 'ee-entity-actions-menu', {
 		'ee-entity-actions-menu--horizontal': layout === EntityActionsMenuLayout.Horizontal,
 		'ee-entity-actions-menu--vertical': layout === EntityActionsMenuLayout.Vertical,
 	});
 	return (
-		<div className={className} {...menuProps}>
+		<div className={className} {...props}>
```
tn3rb, line 27: 👍 

ya we should standardize this and always use `...props` instead of having it change from component to component

nicolad:
@tn3rb these buttons are using `Button` from `infrastructure` layer:

image
***
in order to make it look like the other default one's I need to use the logic from the `Button` from app layer, so I'd have to import app `Button` into infra `Modal`, but not sure if this approach is ok

assets/src/application/ui/layout/entityActionsMenu/style.scss
tn3rb: yup totally agree

assets/src/domain/eventEditor/ui/datetimes/datesList/AddNewDateButton.tsx
```diff
@@ -25,11 +18,13 @@ const AddNewDateButton: React.FC = () => {
 	};
 
 	return (
-		<div style={btnRowStyle}>
-			<Button buttonText={__('Add New Date')} icon={Calendar} mr={2} onClick={onClickAddNew} variant='outline' />
-			<Button buttonText={__('Ticket Assignments')} icon={Ticket} onClick={onOpen} variant='outline' />
+		<>
+			<ButtonGroup marginBottom='2rem'>
```
tn3rb, line 22: is this the same as `mb={2}` ?

assets/src/infrastructure/ui/inputs/Button/Button.tsx
```diff
@@ -2,16 +2,16 @@ import React from 'react';
...
 type ButtonType = React.ComponentType<ButtonProps>;
 
 const Button = React.forwardRef<ButtonType, ButtonProps>(({ children, buttonText, icon, ...props }, ref) => {
-	const className = classNames('ee-btn-base', props.className);
+	const className = classNames('ee-btn-base ee-btn', props.className);
 	const text = children || buttonText;
 
 	return (
-		<ChakraButton {...props} className={className} leftIcon={icon} ref={ref}>
+		<ChakraButton {...props} className={className} leftIcon={icon} ref={ref} variant='unstyled'>
```
tn3rb, line 14: 👍

assets/src/application/ui/layout/entityActionsMenu/EntityActionsMenu.tsx
nicolad: I thought about this, it make sense to have it differently when there are multiple sources of props, like `toggleProps` and for example `buttonProps` in the same component

assets/src/domain/eventEditor/ui/datetimes/datesList/AddNewDateButton.tsx
nicolad: yes

nicolad: no, tested wrongly

PR #2747, opened 2020-04-29
Remove relations on permanent delete | Closes #2742

### LINKED ISSUE #2742: Remove Relations From Entities to Enable Permanent Deletion and User Against Proof Last Datetime Deletion
see also: #2735 

The EE db model system does not allow entities to be permanently deleted if that would result in orphaned entities remaining in the database. So if a ticket has a relation to a datetime, the datetime can NOT be deleted because then the ticket would be orphaned with no way to remove that record other than editing the database directly.

This unfortunately requires some extra complexity in order to protect users from themselves and maintain data integrity.

Here's what needs to happen:

## trashing dates

- when the user clicks to trash a datetime we need to first look at the existing relations and then display a confirmation popup:

- if there are no relations then the confirmation notice should be something like
```"Are you sure you want to trash this datetime?"```
then proceed with trashing the datetime if OK is pressed

- if there are existing relations then the confirmation notice should be something like
```"Existing ticket assignments will be removed and any tickets with no other relations to existing datetimes will also be trashed. Are you sure you want to trash this datetime, remove its ticket assignments, and trash any tickets with no other relations?"```
then if OK is pressed, proceed with trashing the datetime as well as any tickets with no other relations, then remove relations for tickets that still have assignments to other dates.

So for example, given the following dates and tickets, where **`X`** represents a relation:

 . | ticket A | ticket B
--- | --- | ---
**date A** | **`X`** | **`X`**
**date B** |   | **`X`**

trashing **date A** should result in the following:

- removal of relation between **date A** and **ticket B**
- `isTrashed` flag for **ticket A** set to `true`, effectively removing it from the UI unless trashed tickets are viewed

and NOT result in:

- removal of relation between **date A** and **ticket A** because that would orphan **ticket A**
- **ticket B** getting trashed since it still has a relation to **date B**

## permanently deleting dates

- similar to the above we need to first check if there any existing relations and then display a confirmation popup:

- if there are no relations then the confirmation notice should be something like
```"This action is permanent and can not be undone! Are you sure you want to delete this datetime?"```
then if OK is pressed, proceed with removing the trashed datetime from the database

- if there are existing relations then the confirmation notice should be something like
```"This action is permanent and can not be undone! Existing trashed tickets assigned to this datetime will also be permanently deleted. Are you sure you want to delete this datetime and its related tickets?"```
then if OK is pressed, proceed with removing the trashed datetime from the database as well as any related trashed tickets

### PR DESCRIPTION
**Pull request by manzoorwanijk**
This PR refactors permanent deletions to handle relations.
- Refactors permanent delete mutations
- Fixes TPC issue of base price not being created
- Fixes query args for connection resolvers

### DIFF
--- a/assets/src/domain/eventEditor/ui/tickets/ticketPriceCalculator/data/useInitialState.ts
+++ b/assets/src/domain/eventEditor/ui/tickets/ticketPriceCalculator/data/useInitialState.ts
@@ -34,21 +34,12 @@ const useInitialState = ({ ticketId }: BaseProps): StateInitializer => {
 
 	// get all related prices
 	const unSortedPrices = useTicketPrices(ticketId);
-
 	//sort'em
 	const sortedPrices = sortByPriceOrderIdAsc(unSortedPrices);
 
-	const getPrices = useCallback(() => {
-		const hasBasePrice = sortedPrices.filter(isBasePrice).length;
-
-		if (hasBasePrice) return sortedPrices;
-
-		return [basePrice, ...sortedPrices];
-	}, [basePrice, sortedPrices]);
-
 	// convert to TPC price objects by adding
 	// "priceType" and "priceTypeOrder"
-	const prices = getPrices().map<TpcPriceModifier>((price) => {
+	let prices = sortedPrices.map<TpcPriceModifier>((price) => {
 		const priceTypes = getRelations({
 			entity: 'prices',
 			entityId: price.id,
@@ -58,6 +49,14 @@ const useInitialState = ({ ticketId }: BaseProps): StateInitializer => {
 		const priceTypeId = priceTypes[0];
 		return { ...price, priceType: priceTypeId, priceTypeOrder: priceTypeIdOrder[priceTypeId] };
 	});
+
+	const hasBasePrice = prices.filter(isBasePrice).length;
+	// if there is no basePrice
+	if (!hasBasePrice) {
+		// add the base price with `isNew` flag to make sure it's created on submit
+		// `order` as 1 to make sure it remains at the top
+		prices = [{ ...basePrice, order: 1, isNew: true }, ...prices];
+	}
 
 	return useCallback<StateInitializer>(
 		(initialState) => {
--- a/core/domain/services/graphql/mutators/DatetimeDelete.php
+++ b/core/domain/services/graphql/mutators/DatetimeDelete.php
@@ -35,7 +35,41 @@ class DatetimeDelete extends EntityMutator
                 $entity = EntityMutator::getEntityFromInputData($model, $input);
 
                 // Delete the entity
-                $result = ! empty($input['deletePermanently']) ? $entity->delete_permanently() : $entity->delete();
+                if (! empty($input['deletePermanently'])) {
+                    // all related tickets
+                    $tickets = $entity->tickets();
+                    foreach ($tickets as $ticket) {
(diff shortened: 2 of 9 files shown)

### CODE REVIEW

File: assets/src/domain/eventEditor/ui/tickets/ticketPriceCalculator/data/useInitialState.ts [hunk 1]
```diff
@@ -59,6 +50,14 @@ const useInitialState = ({ ticketId }: BaseProps): StateInitializer => {
...
 	});
 
+	const hasBasePrice = prices.filter(isBasePrice).length;
+	// if there is no basePrice
+	if (!hasBasePrice) {
+		// add the base price with `isNew` flag to make sure it's created on submit
+		// `order` as 1 to make sure it remains at the top
+		prices = [{ ...basePrice, order: 1, isNew: true }, ...prices];
+	}
+
```
nicolad, line 60: nice comments and extraction :+1:

core/domain/services/graphql/mutators/DatetimeDelete.php
```diff
@@ -35,7 +35,41 @@ public static function mutateAndGetPayload(EEM_Datetime $model, Datetime $type)
...
+                    foreach ($tickets as $ticket) {
+                        // if the ticket is related to a maximum of one datetime
+                        if ($ticket->count_related('Datetime') <= 1) {
+                            // trash the ticket
+                            $ticket->delete();
+                        }
+                    }
+                    // trash the datetime
+                    $result = $entity->delete();
+                }
```
tn3rb, line 72: would be nicer / more readable to extract these blocks into separate functions like:

```php
 if (! empty($input['deletePermanently'])) {
	$result = DatetimeDelete::deleteDatetimeAndRelations($entity);
} else {
	$result = DatetimeDelete::trashDatetimeAndRelations($entity);
}
```

as well, the logic within the tickets loop is nearly identical to this code from `services/graphql/mutators/TicketDelete.php`

```php
// Remove related prices for the ticket
$entity->delete_related_permanently('Price');
// Remove relation with datetimes
$entity->_remove_relations('Datetime');
// Now delete the ticket permanently
$result = $entity->delete_permanently();
```

so that could be extracted into a public function and called from here.

```php
if ($ticket->count_related('Datetime') === 1) {
    TicketDelete::deleteTicketAndRelations($ticket);
}
```

Then if for some reason, this logic isn't quite right and needs to be updated, you only have to make a change in that one place:

manzoorwanijk: Done, thank you for the suggestion.

I don't why I don't feel confident about doing such things in PHP although I did think about doing that extraction 🙄

core/domain/services/graphql/mutators/PriceDelete.php
```diff
@@ -33,8 +38,14 @@ public static function mutateAndGetPayload(EEM_Price $model, Price $type)
             try {
                 /** @var EE_Price $entity */
                 $entity = EntityMutator::getEntityFromInputData($model, $input);
+
                 // Delete the entity
-                $result = ! empty($input['deletePermanently']) ? $entity->delete_permanently() : $entity->delete();
+                if (! empty($input['deletePermanently'])) {
+                    $result = PriceDelete::deletePriceAndRelations($entity);
+                } else {
```
tn3rb, line 45: 👍 good thinking extracting this too... never know when we might need it somewhere else

PR #2755, opened 2020-04-30
Stop setting extra meta for every message

Pebblo:
There is an 'error' set in extra meta each time a message changes status to EEM_Message::status_messenger_executing.

I think the intention is that if you see that error often you need to update your server config etc. So its set incase the request doesn't complete etc.

But it's on every messenger when it changes the status to 'executing' and then isn't removed. I know I could add some code to remove when finished but I don't remember ever getting a report of that error on the forums and if there is instances of the message queue going crazy we can just add `EE_DEBUG_MESSAGES` to `wp-config.php` for stuff like this and work from there.

Trigger a registration message within Event Espresso, either by registering onto an Event or manually triggering a message.

Then check the `esp_extra_meta` table and find the last entry (sort but row ID and find the last row) with something like:

`28401 | 5034 | Message | MSG_error | A message with this status indicates that there wa..`

Each time a message is generated and sent it adds another row of the above.

Now, switch to this branch and add a couple of registrations so you trigger messages again.

Check `esp_extra_meta` again and confirm there are no additional entries for the above additional registrations and that you actually received the messages.

Enable `EE_DEBUG_MESSAGES` by adding `define( 'EE_DEBUG_MESSAGES', true);` to `wp-config.php`

Trigger more messages and check `esp_extra_meta` again, this time, because EE_DEBUG_MESSAGES is enabled you **should** see those DB entries being added to the table each time a message is generated again.

## Checklist

* [x] My code is tested.

### DIFF
--- a/core/db_classes/EE_Message.class.php
+++ b/core/db_classes/EE_Message.class.php
@@ -869,15 +869,17 @@ class EE_Message extends EE_Base_Class implements EEI_Admin_Links
     public function set_messenger_is_executing()
     {
         $this->set_STS_ID(EEM_Message::status_messenger_executing);
-        $this->set_error_message(
-            esc_html__(
-                'A message with this status indicates that there was a problem that occurred while the message was being
-                processed by the messenger.  It is still possible that the message was sent successfully, but at some
-                point during the processing there was a failure.  This usually is indicative of a timeout issue with PHP 
-                or memory limits being reached.  If you see this repeatedly you may want to consider upgrading the memory 
-                available to PHP on your server.',
-                'event_espresso'
-            )
-        );
+        if (EEM_Message::debug()) {
+            $this->set_error_message(
+                esc_html__(
+                    'A message with this status indicates that there was a problem that occurred while the message was being
+                    processed by the messenger.  It is still possible that the message was sent successfully, but at some
+                    point during the processing there was a failure.  This usually is indicative of a timeout issue with PHP 
+                    or memory limits being reached.  If you see this repeatedly you may want to consider upgrading the memory 
+                    available to PHP on your server.',
+                    'event_espresso'
+                )
+            );
+        }
     }
 }


garthkoyle:
@joaovictorangeline can you test this? Accessing the db table and such is a little beyond my comfort level for testing.

joaovictorangeline:
@Pebblo @tn3rb I need more details on how to reproduce it. 
Could you provide it?

tn3rb:
mmm.. sry... i think you'll need to ask @Pebblo for more info about that
he's the messages guru these days

garthkoyle:
@Pebblo I wonder if this meta would be useful if there are times we have to look at the database to determine if a message had been sent? If something goes wrong sending these messages, would it be useful to have the meta recorded? cc @tn3rb

tn3rb:
I'm guessing from the original comment for this PR that it is not useful and just fills up the db unnecessarily

Pebblo:
Sorry guys, i missed the notifications for this ticket.

@joaovictorangeline

The opening post of this ticket had details on how reproduce and check this but I've broken it down a little more if that doesn't help please let me know what you are stuck on.

@garthkoyle wrt this:

> I wonder if this meta would be useful if there are times we have to look at the database to determine if a message had been sent? If something goes wrong sending these messages, would it be useful to have the meta recorded?

Every single message that is generated adds the above to extra_meta, so if it works or it fails we still have that same entry in the DB, meaning it doesn't really help.

Imo it's just cluttering up the DB and so far I've not found any real world benefit for it doing so. As with a lot of the message system its not always clear cut which is why I used the `EE_DEBUG_MESSAGES` constant for this one rather than just completely removing it. If we 'need' it, we can add the above to `wp-config.php`

garthkoyle:
**For context, I set the messages to send on the same request.**

I see this on Master:

image


This branch:
image

This branch using the define( 'EE_DEBUG_MESSAGES', true); in the wp-config.php file.

image


I think that looks right @Pebblo 

Ready to merge.

Pebblo:
Yep, thats exactly what I expect. Thanks @garthkoyle 

I've also retested this and got the same result so merging in

PR #2760, opened 2020-04-30
Fix Default Editor Data (and Refactor for Better SRP) closes #2718

### LINKED ISSUE #2718: Default Ticket Loaded for New Events Belongs to Existing Events
When a user creates a new Event, it needs to be populated with at least one datetime and one ticket that should also have a relation set up between them (ie: a record added to `wp_esp_datetime_ticket`).

However, an error like the following currently appears when you attempt to save a new event:

```
Fatal error: Uncaught DomainException:
An EE_Ticket for Event 9 was supplied while calculating event space availability for Event 13.
in /core/domain/services/event/EventSpacesCalculator.php:222

Stack trace:
#0 /core/domain/services/event/EventSpacesCalculator.php(186): EventSpacesCalculator->validateTicket(Object(EE_Ticket))
#1 /core/db_classes/EE_Event.class.php(852): EventSpacesCalculator->setActiveTickets(Array)
#2 /core/db_classes/EE_Event.class.php(895): EE_Event->spaces_remaining(Array)
#3 /core/db_classes/EE_Event.class.php(980): EE_Event->perform_sold_out_status_check()
#4 /admin_pages/event
```

Here is a screenshot of a new event showing the default datetime and ticket prior to saving the event:

as you can see the datetime has an ID of 2 and the ticket has an ID of 1.

Here is a screenshot of the `wp_esp_datetime` table showing that Datetime 1 belongs to Event 9 and Datetime 2 belongs to Event 13

Here is a screenshot of the `wp_esp_datetime_ticket` table showing that both Datetime 1 and Datetime 2 have a relation with Ticket 1

This is the reason for the Fatal Error that occurs in the `EventSpacesCalculator`.

The solution is to create a copy of all default tickets when generating a new event instead of loading in the original tickets. Then when those are saved, the new IDs can be used with new relations.

tn3rb:
see #2718 

Previously the Event Editor assigned default tickets and prices to a new datetime created for new events. This would cause an error when saved because the ticket default used would of course also be used by another datetime for another event, but tickets can not be shared between events.

So this PR fixes the issues simply by cloning default entities prior to assigning them to the new event.

Also took the opportunity to extract a bunch of logic into separate classes to better satisfy SRP, etc


core/domain/entities/admin/GraphQLData/GeneralSettings.php
```diff
@@ -0,0 +1,47 @@
...
+            'query'          => $query,
+        ];
+
+        if (! empty($where_params)) {
+            $data['variables'] = [];
+            $data['variables']['where'] = $where_params;
+        }
+
+        $responseData = $this->makeGraphQLRequest($data);
+        return ! empty($responseData[ $field_key ]) ? $responseData[ $field_key ] : null;
```
nicolad, line 45: `CurrentUser`, `Datetimes` and `GeneralSettings` have this block being duplicated. It's ok for 2-3 instances to have duplicate, if this will be used in more classes, than it's worth to extract somewhere.

core/domain/entities/admin/GraphQLData/PriceTypes.php
```diff
@@ -0,0 +1,60 @@
...
+            ],
+            'query'          => $query,
+        ];
+
+        if (! empty($where_params)) {
+            $data['variables']['where'] = $where_params;
+        }
+
+        $responseData = $this->makeGraphQLRequest($data);
+        return ! empty($responseData[ $field_key ]) ? $responseData[ $field_key ] : null;
```
nicolad, line 58: just scrolled a bit :) 
yes, this block seems a good candidate for an extraction :)

core/domain/services/admin/events/editor/EventEditor.php
```diff
@@ -0,0 +1,159 @@
...
+     * @since $VID:$
+     */
+    public function loadScriptsStyles()
+    {
+        if ($this->admin_config->useAdvancedEditor()) {
+            $eventId = $this->event instanceof EE_Event ? $this->event->ID() : 0;
+            if (! $eventId) {
+                global $post;
+                $eventId = isset($_REQUEST['post']) ? absint($_REQUEST['post']) : 0;
+                $eventId = $eventId === 0 && $post instanceof WP_Post && $post->post_type === 'espresso_events'
```
nicolad, line 110: this `0` make me think that another intermediary variable might be helpful here because it doesn't make a lot of sense to set `$eventId` to zero and the check `$eventId === 0`
also, there is a big chance that I might not see the big picture here

core/domain/services/admin/events/editor/EventEditor.php
```diff
@@ -0,0 +1,159 @@
...
+
+        $currentUser = $this->current_user->getData();
+
+        $generalSettings = $this->general_settings->getData();
+
+        $i18n = $this->jed_locale->getData('event_espresso');
+
+        $assetsUrl = EE_PLUGIN_DIR_URL . 'assets/dist/';
+
+        return compact('event', 'graphqlEndpoint', 'currentUser', 'generalSettings', 'i18n', 'assetsUrl');
```
nicolad, line 157: have no idea what this `compact` is about, but this entire block looks very clean :+1:

core/domain/services/admin/events/editor/EventEntityRelations.php
```diff
@@ -0,0 +1,165 @@
...
+        ];
+        // Get the IDs of all datetime tickets.
+        $ticketIds = $this->ticket_model->get_col([
+            [ 'Datetime.DTT_ID' => ['IN', $datetimeIds] ],
+            'default_where_conditions' => 'minimum',
+        ]);
+        foreach ($ticketIds as $ticketId) {
+            $GID = $this->convertToGlobalId($this->ticket_model->item_name(), $ticketId);
+
+            foreach ($related_models as $key => $model) {
```
nicolad, line 100: can this be done w/o nested `foreach`?

caffeinated/admin/extend/events/Extend_Events_Admin_Page.core.php
```diff
@@ -1,7 +1,7 @@
 <?php
 
 use EventEspresso\core\domain\services\admin\events\default_settings\AdvancedEditorAdminFormSection;
-use EventEspresso\core\domain\services\admin\events\editor\AdvancedEditorData;
+use EventEspresso\core\domain\services\admin\events\editor\EventEditor;
```
manzoorwanijk, line 4: Yeeaaaaaahhhhhh! It's just an editor, nothing advanced! Why to scare away people? 😄

core/domain/entities/admin/GraphQLData/GeneralSettings.php
manzoorwanijk, line 45: I agree with @nicolad. we could extract that block to parent class
```suggestion
        return $this->getGraphQLData($data, $field_key);
```

Also, this `$data` variable is better if named as `$params`

core/domain/services/admin/events/editor/EventEditor.php
tn3rb: that's Grand Chef Manzoor's handy work

core/domain/entities/admin/GraphQLData/PriceTypes.php
tn3rb: will do

core/domain/services/admin/events/editor/EventEditor.php
tn3rb: will do

core/domain/services/admin/events/editor/EventEntityRelations.php
tn3rb: mmmm... not in any way that would be more straightforward

core/domain/entities/admin/GraphQLData/CurrentUser.php
```diff
@@ -0,0 +1,49 @@
...
+            }
+        }
+QUERY;
+        $this->setParams([
+            'operation_name' => 'GET_CURRENT_USER',
+            'query'          => $query,
+        ]);
+
+        return $this->getQueryResponse($field_key, $where_params);
+    }
```
manzoorwanijk, line 48: Cool 👍

core/domain/entities/admin/GraphQLData/GraphQLData.php
```diff
@@ -0,0 +1,77 @@
...
+     * @since $VID:$
+     */
+    protected function getQueryResponse($field_key, array $where_params = [])
+    {
+        if (! empty($where_params)) {
+            if (! array_key_exists('variables', $this->params)) {
+                $this->params['variables'] = [];
+            }
+            $this->params['variables']['where'] = $where_params;
+        }
```
manzoorwanijk, line 52: 👍

core/domain/services/admin/events/editor/EventEditorGraphQLData.php
```diff
@@ -0,0 +1,124 @@
...
+        $tickets = null;
+        $prices = null;
+
+        if (empty($datetimes['nodes']) || (isset($_REQUEST['action']) && $_REQUEST['action'] === 'create_new')) {
+            $this->default_entities->getData($eventId);
+            $datetimes = $this->datetimes->getData(['eventId' => $eventId]);
+        }
+
+        if (! empty($datetimes['nodes'])) {
+            $datetimeIn = wp_list_pluck($datetimes['nodes'], 'id');
```
manzoorwanijk, line 102: Nice 🥇

tn3rb: LOLZ  you did that !!! 😆

PR #2768, opened 2020-05-04
Refactor Datepicker and Timepicker; partially closes #2725.

### LINKED ISSUE #2725: Refactor styles

nicolad:
- Partially closes https://github.com/eventespresso/event-espresso-core/issues/2725

### DIFF
--- a/assets/src/infrastructure/ui/inputs/dateTime/Datepicker/style.scss
+++ b/assets/src/infrastructure/ui/inputs/dateTime/Datepicker/style.scss
@@ -1,0 +1,5 @@
+.ee-date-picker {
+	[data-baseweb='base-input'] {
+		border: 0;
+	}
+}
--- a/assets/src/infrastructure/ui/inputs/dateTime/Timepicker/style.scss
+++ b/assets/src/infrastructure/ui/inputs/dateTime/Timepicker/style.scss
@@ -5,10 +5,10 @@
 		align-items: center;
 		display: flex;
 		min-height: 1.875rem;
-		padding: 0 0.5rem;
 
 		> div {
 			background-color: transparent;
+			padding: 0 0.5rem;
 
 			@extend .base-inputs-styles;
 			&:hover {
@@ -19,6 +19,9 @@
 			}
 			> div {
 				padding: 0;
+				> div {
+					display: contents;
+				}
 			}
 		}
 	}
--- a/assets/src/infrastructure/ui/inputs/dateTime/Timepicker/style.scss
+++ b/assets/src/infrastructure/ui/inputs/dateTime/Timepicker/style.scss
@@ -1,0 +1,25 @@
+@import '../../style.scss';
+
+.ee-time-picker {
+	[data-baseweb='select'] {
+		align-items: center;
+		display: flex;
+		min-height: 1.875rem;
+		padding: 0 0.5rem;
+
+		> div {
+			background-color: transparent;
+
+			@extend .base-inputs-styles;
+			&:hover {
+				@extend .input-select-textarea-hover;
+			}
+			&:focus {
+				@extend .input-select-textarea-focus;
+			}
+			> div {
+				padding: 0;
+			}
+		}
+	}
+}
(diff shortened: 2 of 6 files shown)


nicolad:
image

nicolad:
Peek 2020-05-04 16-57
***
baseui made it tricky to impossible to have control over focus state of timepicker :disappointed: 
will try again in another PR
Peek 2020-05-04 17-00

nicolad:
zoomed:

image

nicolad:
FF:

image

assets/src/infrastructure/ui/inputs/dateTime/Timepicker/style.scss
```diff
@@ -0,0 +1,28 @@
...
+			&:hover {
+				@extend .input-select-textarea-hover;
+			}
+			&:focus {
+				@extend .input-select-textarea-focus;
+			}
+			> div {
+				padding: 0;
+				> div {
+					display: contents;
```
nicolad, line 23: items are not aligned w/o this prop:

image

nicolad: didn't knew about this property of display until today
***
```
display: contents;—a magical new display value that essentially makes the container disappear, making the child elements children of the element the next level up in the DOM.
```
https://css-tricks.com/get-ready-for-display-contents/

assets/src/infrastructure/ui/inputs/dateTime/Datepicker/style.scss
```diff
@@ -0,0 +1,5 @@
+.ee-date-picker {
+	[data-baseweb='base-input'] {
```
tn3rb, line 2: booooo  

...not your fault... i get why you had to do this...

but boooo... like why release something as public but make it difficult to modify?!!?

assets/src/infrastructure/ui/inputs/dateTime/Timepicker/style.scss
tn3rb: ooo i didn't know about that either... COOL

assets/src/infrastructure/ui/inputs/style.scss
```diff
@@ -7,20 +7,50 @@
...
-		border-color: var(--ee-border-color);
-		border-style: solid;
-		border-width: 2px;
-		&:focus {
-			/* Visible in Windows high-contrast themes */
-			outline-color: transparent;
-			outline-width: 2px;
-			outline-style: solid;
-		}
+.base-inputs-styles {
```
tn3rb, line 16: if this is a new class added by us, then it should be namespaced with `ee-`
I'm guessing it is not since I'm not seeing it applied to a component anywhere

assets/src/infrastructure/ui/inputs/style.scss
```diff
@@ -7,20 +7,50 @@
...
+	}
+	&:focus {
+		@extend .input-select-textarea-focus;
+	}
+}
+
+// this approach is needed in cases when classname cannot be applied directly on input
+// and we need to style that input based on it's parent
+.ee-input-base-wrapper {
+	input[type='text'],
```
tn3rb, line 34: 👍

assets/src/infrastructure/ui/inputs/dateTime/Datepicker/style.scss
nicolad: yeah, it's a bit disappointing to see this from a solution backed by such a big company as Uber

assets/src/infrastructure/ui/inputs/style.scss
nicolad: good observation, that one is used only to be extended, I also was thinking what is the best approach since in sass docs I saw that they have this prefix for stuff that is extended, but not used standalone:

image

nicolad: lemme know if you want to use that `%` prefix to mark this as being used only for extension

nicolad: link from sass docs where screenie is from: https://sass-lang.com/guide

PR #2769, opened 2020-05-05
Load React && ReactDOM from CDN closes #2762

### LINKED ISSUE #2762: Load external deps
To make all our domains work smoothly and to reduce bundle size (especially for blocks, we need to load some packages externally. Here are the details:

1. **React and ReactDOM**
React and ReactDOM should be loaded for EDTR set as as dependency for the main EDTR JS bundle. IMHO there should be different builds for development and production. If `SCRIPT_DEBUG` is defined it should load the development build of React and ReactDOM. More info [here](https://reactjs.org/docs/add-react-to-a-website.html) and [here](https://reactjs.org/docs/cdn-links.html).
I found something related to this in `registerWpAssets` inside `core/domain/services/assets/CoreAssetManager.php`
2. **WP I18N**
I guess this is already available and can simply be bundled by adding `wp-i18n` as a dependency to EDTR js bundle in `wp_enqueue_script`

So, after this EDTR JS bundle will have 3 more deps - `react`, `react-dom`, `wp-i18n`

**Note:** We also need to ensure if we want to support WP < 5.0, because `wp-i18n` will not be available.

tn3rb:
- adds new `ReactAssetManager` for loading React externals from CDN

- also extracts logic from `CoreAssetManager` into new `LegacyEditorAssetManager`

- adds ability to set attributes on asset script tags

- BUT WAIT!

- THERE'S MORE!!!

- it slices it dices

- it fixes some path issues in the Assets Registry 

- it removes old EDTRv1 assets we are no longer using (left some in place in case they are still needed)

- it cleans up some of the EE_Admin page classes and removes unsightly pet stains


core/domain/services/assets/EspressoEditorAssetManager.php
```diff
@@ -45,7 +45,17 @@ public function addAssets()
...
+                ReactAssetManager::JS_HANDLE_REACT_DEV,
+                ReactAssetManager::JS_HANDLE_REACT_DOM_DEV,
+            ]
+            : [
+                ReactAssetManager::JS_HANDLE_REACT_PROD,
+                ReactAssetManager::JS_HANDLE_REACT_DOM_PROD,
+            ];
+        $this->addJs(EspressoEditorAssetManager::JS_HANDLE_EDITOR, $dependencies)
+             ->setRequiresTranslation();
     }
```
manzoorwanijk, line 59: Did you forget to add `wp-i18n` to deps as explained in the issue? 😄

core/domain/services/assets/ReactAssetManager.php
```diff
@@ -0,0 +1,58 @@
...
+ */
+class ReactAssetManager extends AssetManager
+{
+    const JS_HANDLE_REACT_DEV = 'react-development';
+    const JS_HANDLE_REACT_PROD = 'react-production';
+    const JS_HANDLE_REACT_DOM_DEV = 'react-dom-development';
+    const JS_HANDLE_REACT_DOM_PROD = 'react-dom-production';
+    // base URLs
+    const URL_BASE_UMD_REACT = 'https://unpkg.com/react@16/umd/';
+    const URL_BASE_UMD_REACT_DOM = 'https://unpkg.com/react-dom@16/umd/';
```
manzoorwanijk, line 28: What if we want to work offline sometimes? May be bundle it instead?

core/domain/services/assets/EspressoEditorAssetManager.php
```diff
@@ -45,7 +45,17 @@ public function addAssets()
      */
     private function registerJavascript()
     {
-        $this->addJs(self::JS_HANDLE_EDITOR, [CoreAssetManager::JS_HANDLE_JS_CORE])->setRequiresTranslation();
+        $dependencies = defined('SCRIPT_DEBUG')
```
manzoorwanijk, line 48: I am a kind of person who sets `SCRIPT_DEBUG` to false when I want to test things for production 🙈 
```suggestion
        $dependencies = defined('SCRIPT_DEBUG') && SCRIPT_DEBUG
```

core/domain/services/assets/ReactAssetManager.php
```diff
@@ -0,0 +1,58 @@
...
+
+    /**
+     * @throws InvalidDataTypeException
+     * @throws InvalidEntityException
+     * @throws DuplicateCollectionIdentifierException
+     * @throws DomainException
+     */
+    public function addAssets()
+    {
+        if (defined('SCRIPT_DEBUG')) {
```
manzoorwanijk, line 38:
```suggestion
        if (defined('SCRIPT_DEBUG') && SCRIPT_DEBUG) {
```

core/domain/services/assets/EspressoEditorAssetManager.php
manzoorwanijk: I see that `wp-i18n` is already loaded somewhere for event edit page, can you please check if that's some EDTRv1 js still there?

core/domain/services/assets/EspressoEditorAssetManager.php
```diff
@@ -45,7 +45,17 @@ public function addAssets()
...
+            ? [
+                ReactAssetManager::JS_HANDLE_REACT_DEV,
+                ReactAssetManager::JS_HANDLE_REACT_DOM_DEV,
+            ]
+            : [
+                ReactAssetManager::JS_HANDLE_REACT_PROD,
+                ReactAssetManager::JS_HANDLE_REACT_DOM_PROD,
+            ];
```
manzoorwanijk, line 56:
```suggestion
            ];
            $dependencies[] = 'wp-i18n';
```

core/domain/services/assets/EspressoEditorAssetManager.php
tn3rb: we're calling `->setRequiresTranslation()` which results in the scripts being run through `wp.i18n.setLocaleData()` when loaded...

hmmm... I also saw that `CoreAssetManager` defines `JS_HANDLE_I18N` which I assumed would have `wp-i18n` set as a dependency but that's not actually the case... will have to look into that more

core/domain/services/assets/ReactAssetManager.php
tn3rb: I thought the idea was to use the CDN?

but ya, I can copy those scripts to the vendor folder

core/domain/services/assets/EspressoEditorAssetManager.php
tn3rb: oh oops... I swear that's how it was in my original version that got lost, but I forgot to do that this time around

core/domain/values/assets/JavascriptAsset.php
```diff
@@ -98,6 +98,11 @@ public function requiresTranslation()
     public function setRequiresTranslation($requires_translation = true)
     {
         $this->requires_translation = filter_var($requires_translation, FILTER_VALIDATE_BOOLEAN);
+        $dependencies = $this->dependencies();
+        if (! in_array('wp-i18n', $dependencies, true)){
+            $dependencies[] = 'wp-i18n';
+            $this->setDependencies($dependencies);
+        }
```
manzoorwanijk, line 105: Cool 👍

PR #2770, opened 2020-05-05
Replace babel-plugin-makepot; closes #2666.

### LINKED ISSUE #2666: Replace babel-plugin-makepot
@wordpress has moved @wordpress/babel-plugin-makepot to WP CLI

https://github.com/WordPress/gutenberg/issues/14780#issuecomment-479410142

nicolad:

- Closes: https://github.com/eventespresso/event-espresso-core/issues/2666

This is how dependency has been added:

image

### DIFF
--- a/bin/make-pot.sh
+++ b/bin/make-pot.sh
@@ -1,0 +1,2 @@
+#!/bin/bash
+wp i18n make-pot . --include"included-*.php" --skip-js
--- a/composer.json
+++ b/composer.json
@@ -3,7 +3,7 @@
 	"type": "wordpress-plugin",
 	"license": "GPL-2.0-or-later",
 	"description": "Event Espresso 4 Core for WordPress: Build an Event Ticketing Website Today!",
-	"authors" : [
+	"authors": [
 		{
 			"name": "Event Espresso Teem",
 			"email": "developers@eventspresso.com",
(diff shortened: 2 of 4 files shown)


composer.json
```diff
@@ -32,7 +38,8 @@
 	},
 	"require": {
 		"composer/installers": "~1.0",
-		"php": ">=5.5"
+		"php": ">=5.5",
+		"wp-cli/i18n-command": "^2.2"
```
nicolad, line 42: https://github.com/wp-cli/i18n-command/

nicolad: https://packagist.org/packages/wp-cli/i18n-command

bin/make-pot.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/bash
+wp i18n make-pot . --include"included-*.php" --skip-js
```
nicolad, line 2: I saw this article and thought this can be executed as a bash script: https://n8finch.com/wrote-first-bash-script-implement-wp-cli-managed-sites

nicolad: open to smth better

nicolad:
I am not sure what to do with that `composer-setup.php`, it was added during executing `composer require "wp-cli/i18n-command"`
should this be in gitignore?

bin/make-pot.sh
nicolad: related: https://developer.wordpress.org/block-editor/developers/internationalization/#create-translation-file

Pebblo:
Hi guys,

Just noting that we currently do not generate the POT file used by core (and all add-ons) from within core itself.

As it stands we us [BuildMachine](https://github.com/eventespresso/grunt-wp-plugin-buildmachine) to generate the POT for all repos in a single step using make-pot so I think the changes this would actually need to be done there.

bin/make-pot.sh
manzoorwanijk: `npm` script should be enough if WP-CLI is installed.

tn3rb:
created the following issue for confirming JS file translations are being done on the buildmachine:

https://github.com/eventespresso/grunt-wp-plugin-buildmachine/issues/8

and putting this PR and its Issue on hold

PR #2772, opened 2020-05-05
Prepare GQL for attendees queries | Closes #2771

### LINKED ISSUE #2771: Prepare GQL for Attendees List
We need to query attendees list for blocks.

manzoorwanijk:
This PR adds support for attendees list query for GQL.

### DIFF
--- a/core/domain/services/graphql/types/Attendee.php
+++ b/core/domain/services/graphql/types/Attendee.php
@@ -3,10 +3,13 @@
 namespace EventEspresso\core\domain\services\graphql\types;
 
 use EEM_Attendee;
+use EE_Attendee;
 use EventEspresso\core\services\graphql\fields\GraphQLFieldInterface;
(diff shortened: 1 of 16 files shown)


core/domain/services/graphql/types/Attendee.php
```diff
@@ -0,0 +1,203 @@
...
+            ),
+            new GraphQLOutputField(
+                'address2',
+                'String',
+                'address2',
+                esc_html__('Address Part 2', 'event_espresso'),
+                null,
+                null,
+                ['ee_edit_contacts']
+            ),
```
tn3rb, line 90: that middle output field is a mix of avatar and address 1
maybe fix its description then place it after the address 2 field

manzoorwanijk: Oops! 🤦‍♂️

PR #2791, opened 2020-05-08
Imrpove filterbar subscription service and its usage | Closes #2749

### LINKED ISSUE #2749: Entity lists are not sorted by date when loaded.
By default, `sortBy` is set to `date` for dates and tickets. It doesn't seem to work when the list loads for the first time. It only happens if you either change the filters or update an entity from the list.
Solution: Instead of setting the default value in `useReducer`, we should use
```js
useEffect(() => setSortBy('date'), []);
```

manzoorwanijk:
This PR:
- Improves `useSubscriptionService`, `useFilterBarService` and `useEntityFilterService`
- Updates deps in `useMemo` calls in `useFilteredEntities`

### DIFF
--- a/assets/src/application/ui/layout/entityList/filterBar/filterState/useEntityFilterService.ts
+++ b/assets/src/application/ui/layout/entityList/filterBar/filterState/useEntityFilterService.ts
@@ -1,4 +1,4 @@
-import { useCallback } from 'react';
+import { useCallback, useMemo } from 'react';
 import { Entity } from '@dataServices/types';
 import { EntityFilterService, EntityListFilterStateManager } from './types';
 import { useFilterBarService, FilterBarServiceCbArgs } from '../subscription';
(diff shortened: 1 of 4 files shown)


assets/src/application/ui/layout/entityList/filterBar/filterState/useFilteredEntities.ts
```diff
@@ -34,13 +34,13 @@ const useFilteredEntities = <D extends string, L extends string, E extends Entit
...
 		return applySearches(filteredEntities, filterState);
-	}, [cacheIds, searchText]);
+	}, [applySearches, cacheIds, searchText]);
 
 	// sort it
 	cacheIds = entityListCacheIdString(searchResults);
 	const sortedEntities = useMemo<Array<E>>(() => {
 		return applySorters(searchResults, filterState);
-	}, [cacheIds, sortBy]);
+	}, [applySorters, cacheIds, sortBy]);
```
manzoorwanijk, line 43: Absence of this dependency here was causing that issue, but then adding that dep caused too many re-renders. That's why I wrapped all the callbacks in `useCallback`

assets/src/application/ui/layout/entityList/filterBar/filterState/useEntityFilterService.ts
```diff
@@ -15,55 +15,63 @@ const useEntityFilterService = <D extends string, L extends string, E extends En
...
 
-	const applyFilters: EFS['applyFilters'] = useCallback(
+	// avoid the callback being affected by change in other callbacks
+	const filterIdsStr = Object.keys(getFilters()).join(':');
+	const applyFilters = useCallback<EFS['applyFilters']>(
 		(entityList, filterState) => {
 			return applyCallbacks(entityList, filterState, getFilters());
 		},
-		[domain, listId]
+		[applyCallbacks, filterIdsStr]
```
tn3rb, line 46: what the?!?! those previous values were not even used in the callback !?!?!

is there any chance that `getFilters` will change?
I can't see those changing dynamically...

assets/src/application/ui/layout/entityList/filterBar/filterState/useFilteredEntities.ts
tn3rb: ya you pretty much have to memoize all callbacks else React generates a new instance every render which triggers further rendering of components downstream

assets/src/application/ui/layout/entityList/filterBar/subscription/useFilterBarService.ts
```diff
@@ -18,52 +19,79 @@ const useFilterBarService: FBShook = (domain, entityListId) => {
...
+	);
+
+	const getCallbacks = useCallback(
+		({ listId = entityListId, type }) => {
+			invariant(listId, 'No `listId` provided');
+			const allSubscriptions = getSubscriptions();
+
+			const isForList = pathEq(['options', 'listId'], listId);
+			const isOfType = pathEq(['options', 'type'], type);
+			const isOfTypeAndForList = allPass([isForList, isOfType]);
```
tn3rb, line 37: can't say I understand how this is working, but that's ok 🤷‍♂️

assets/src/application/ui/layout/entityList/filterBar/filterState/useEntityFilterService.ts
manzoorwanijk: Yes, it will. For example if your register a sorter, it changes the subscriptions hash thus affection all (filters, sorters and searches). What we do here is that we update the dep to only change when a new filter is registered or updated.

assets/src/application/ui/layout/entityList/filterBar/subscription/useFilterBarService.ts
manzoorwanijk: 😄 
Let me explain:
- `listId` can be `'dates-list' | 'tickets-list'`
- `type` can be `'filter' | 'sort' | 'search'`

So `isOfTypeAndForList ` checks if it's a filter for dates-list or a sorter for tickets-list and so on

PR #2800, opened 2020-05-12
Entity List UI Issues; closes #2783.

### LINKED ISSUE #2783: EDTR Entity List UI Issues
#### long titles and descriptions

As you can see from the above image, really long entity names mess up the card layout.

In EDTR v1 the following things were done:

- names under 25 characters in length had a "short-title" css class added to the heading tag that increased the font-size and line-height

- names over 25 characters but under 50 characters in length had a "long-title" css class added to the heading tag that decreased the font-size and line-height

- names over 50 characters in length were truncated so that they do not span more than 2 (or 3?) lines of text and an ellipsis was added. It would be fantastic if this logic could be added into a configurable component that could simply wrap text elements to provide this functionality.

- we don't want to change the font-size and line-height for entity descriptions, but we do want to truncate those entries so that the content is truncated and an ellipsis is appended. Again, would be ideal if a component could be created for doing this and then used for both the datetime and ticket, names and descriptions. This should be done in a separate PR.

---

#### entity status label

in the image above you can also see that the status label in the bottom of the sidebar is way too small so let's add the following to what's there already:

```css
  font-weight: bolder;
  letter-spacing: 1px;
  text-transform: uppercase;
```

should look something like:

ScreenShot_20200506154312

---

#### timezone info

- currently the timezone info icon button is placed between the year and time (see image for long titles and descriptions). Let's move it after the time so that it does not break the continuity of the date time display (see image for the entity status label).

- the `box-shadow` for the timezone info modal is barely noticable and appears flat but also above the cards which have a more pronounced box-shadow, which looks weird. Let's change the timezone info modal to use `box-shadow: var(--ee-box-shadow-big);`. As well, the blue box-shadow that is applied when the modal is focused either needs to be removed, or changed to be a border.

---

#### pagination

- %cough%cough%fugly%cough%cough%

- add `padding: var(--ee-padding-huge) var(--ee-padding-default)` to `.ee-entity-pagination`

- convert to a "button group" **kinda** like this: https://bootstrapshuffle.com/classes/button-groups/btn-group but following the same kind of style as our other buttons:

which is basically using a mix of the following applied to what's there already:

```css
  background: var(--ee-background-color);
  border: 2px solid var(--ee-color-grey-10);
  border-radius: var(--ee-border-radius-default);
  height: 3rem;
  width: 3rem;
```

but it gets tricky when you want to apply hover effects

---

#### ticket price

- picture is worth a thousand words:

---

#### item counts

item counts for relations look fine when the number is positive:

but is off center when the value is zero:

- but instead of displaying `0`, let's use `!` with a bold font (but make sure it is centered)

nicolad:
- Closes: https://github.com/eventespresso/event-espresso-core/issues/2783

## How has this been tested
Peek 2020-05-12 17-16


nicolad:
Peek 2020-05-12 17-54

nicolad:
image

nicolad:
Peek 2020-05-12 23-17

nicolad:
made `PerPage` more consistent:

image

nicolad:
image

nicolad:
image

nicolad:
FF:

image

assets/src/application/ui/display/TimezoneTimeInfo/style.scss
```diff
@@ -1,11 +1,15 @@
-.ee-timezone-info {
+.ee-timezone-info.ee-timezone-info {
```
nicolad, line 1: added for specificity

assets/src/application/ui/input/InlineEditInput/InlineEditText.tsx
```diff
@@ -1,10 +1,48 @@
 import React from 'react';
+import Dotdotdot from 'react-dotdotdot';
+
+import { Textfit } from 'react-textfit';
```
nicolad, line 4: I've added these 2 deps because I wanted to focus first on the result and later replace with our custom-built solution or even better - clone the already existing package

nicolad:
Both name and desc clamped:

image

assets/src/application/ui/input/InlineEditInput/InlineEditText.tsx
```diff
@@ -1,10 +1,48 @@
...
 import { TextProps } from './types';
-import { InlineEdit } from '@infraUI/inputs';
+import { InlineEdit, InlineEditPreviewProps } from '@infraUI/inputs';
+
+const Preview: React.FC<InlineEditPreviewProps> = ({ fitText, isEditing, onRequestEdit, value }) => {
+	if (isEditing) {
+		return null;
+	}
+
+	if (value.length > 50) {
```
nicolad, line 14: noticed that 50 char limit isn't fitting in one line:

image

nicolad: changed to 40:

image

assets/src/application/ui/input/InlineEditInput/InlineEditText.tsx
```diff
@@ -1,10 +1,49 @@
...
 import { TextProps } from './types';
-import { InlineEdit } from '@infraUI/inputs';
+import { InlineEdit, InlineEditPreviewProps } from '@infraUI/inputs';
+
+const Preview: React.FC<InlineEditPreviewProps> = ({ fitText, isEditing, onRequestEdit, value }) => {
+	if (isEditing) {
+		return null;
+	}
+
+	if (value.length > 40) {
```
nicolad, line 14: 40 chars title:

image

assets/src/application/ui/layout/EntityCard/styles.scss
```diff
@@ -78,14 +92,21 @@
...
-				> span {
+				span {
 					color: var(--ee-default-text-color-super-low-contrast);
-					font-size: var(--ee-font-size-bigger);
+					font-size: var(--ee-font-size-big);
 					font-weight: bolder;
+
+					&.ee-currency-input__before-amount {
+						padding-bottom: 2px; // added for vertical alignment
+					}
```
nicolad: image

assets/src/domain/eventEditor/ui/datetimes/datesList/editable/EditableName.tsx
```diff
@@ -14,6 +14,8 @@ interface EditableNameProps extends DateItemProps {
 const EditableName: React.FC<EditableNameProps> = ({ className, entity: datetime, view = 'card' }) => {
 	const { updateEntity } = useDatetimeMutator(datetime.id);
 
+	const dateName = datetime.name ?? __('Edit title...');
```
manzoorwanijk, line 17: 👍

assets/src/application/ui/input/InlineEditInput/InlineEditText.tsx
tn3rb: The titles don't need to be crammed into one line. Form #2783:

> names over 50 characters in length were truncated so that they do not span more than 2 (or 3?) lines of text

I think two lines is good and we don't want the font size to get too small as the title should still be the most prominent datum in the main card body. The above looks like the title font is smaller than the desctiprion font.

nicolad: made 18px(16px is the desc font-size) the minimum font-size and 30 chars the lowest limit before it moves into 2 lines


Peek 2020-05-13 21-06

nicolad: 2 lines is max:

image

assets/src/application/ui/layout/entityList/pagination/style.scss
```diff
@@ -7,14 +7,7 @@
...
 	justify-content: flex-end;
-
-	&.hidePagination {
-		.ant-pagination-prev,
-		.ant-pagination-item,
-		.ant-pagination-next {
-			display: none;
-		}
-	}
+	padding: var(--ee-padding-huge) var(--ee-padding-default);
```
tn3rb, line 10: YAY 😄

assets/src/infrastructure/ui/display/pagination/style.scss
```diff
@@ -0,0 +1,94 @@
...
+
+.ee-pagination {
+	display: flex;
+	flex-direction: row;
+	padding: 1em;
+
+	&__item {
+		align-items: center;
+		background: var(--ee-background-color);
+		border: $border-style;
```
tn3rb, line 12: cool 👍

assets/src/application/ui/input/InlineEditInput/InlineEditText.tsx
```diff
@@ -1,10 +1,49 @@
...
+				max={24} // based on --ee-font-size-bigger: 1.5rem;
+				min={18}
+				mode='single'
+			>
+				<span onClick={onRequestEdit}>{value}</span>
+			</Textfit>
+		);
+	}
+
+	return <span onClick={onRequestEdit}>{value}</span>;
```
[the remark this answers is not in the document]
nicolad: thank you, yes, good observation

Issue #2808, opened 2020-05-12
Add base price to a ticket on server-side when it's created


manzoorwanijk:
When a ticket is created, we need to make sure that it has at least one base price apart from the default prices.
Currently, we add prices to a ticket in:
- `addDefaultPrices` in `EventEspresso\core\domain\services\graphql\data\mutations`
- `getData` in `EventEspresso\core\domain\services\admin\events\editor`

Also, the prices added in `addDefaultPrices` need to be cloned instead of just adding a relation.

PR #2811, opened 2020-05-12
Default Entity Generators | Closes #2811

tn3rb:
see: #2808

- extracts default entity generation into separate classes for SRP

- clones default entities and sets `ID`s and `is_default` flags to false

- ensures that base price is set for tickets


core/domain/services/admin/entities/DefaultPrices.php
```diff
@@ -0,0 +1,109 @@
...
+     * @throws EE_Error
+     * @throws ReflectionException
+     * @since $VID:$
+     */
+    protected function createNewBasePrice(EE_Ticket $ticket)
+    {
+        $new_base_price = $this->price_model->get_new_price();
+        $base_price_type = $this->price_model->get_one([
+            'PRT_ID' => EEM_Price_Type::base_type_base_price
+        ]);
```
manzoorwanijk, line 94: This query fails, it has to be
```suggestion
        $base_price_type = $this->price_model->get_one([[
            'PRT_ID' => EEM_Price_Type::base_type_base_price
        ]]);
```

core/domain/services/admin/entities/DefaultPrices.php
```diff
@@ -0,0 +1,109 @@
...
+        if (! $base_price_type instanceof EE_Price_Type) {
```
manzoorwanijk, line 95:
```suggestion
        if (! $base_price_type instanceof EE_Price) {
```

core/domain/services/admin/entities/DefaultPrices.php
```diff
@@ -0,0 +1,109 @@
...
+     * @param EE_Ticket $ticket
+     * @return EE_Price
+     * @throws EE_Error
+     * @throws ReflectionException
+     * @since $VID:$
+     */
+    protected function createNewBasePrice(EE_Ticket $ticket)
+    {
+        $new_base_price = $this->price_model->get_new_price();
+        $base_price_type = $this->price_model->get_one([
```
manzoorwanijk, line 92:
```suggestion
        $base_price = $this->price_model->get_one([
```

core/domain/services/admin/entities/DefaultPrices.php
```diff
@@ -0,0 +1,109 @@
...
+                esc_html__(
+                    'A valid base price type could not be retrieved from the database.',
+                    'event_espresso'
+                )
+            );
+        }
+        $new_base_price->set('PRT_ID', $base_price_type->ID());
+        $new_base_price->set('PRC_is_default', false);
+        $new_base_price->save();
+        $new_base_price->_add_relation_to($ticket, 'Ticket');
```
manzoorwanijk, line 106: Where is the price `amount` set/calculated?

core/domain/services/admin/entities/DefaultPrices.php
```diff
@@ -0,0 +1,109 @@
...
+        ]);
+        if (! $base_price_type instanceof EE_Price_Type) {
+            throw new RuntimeException(
+                esc_html__(
+                    'A valid base price type could not be retrieved from the database.',
+                    'event_espresso'
+                )
+            );
+        }
+        $new_base_price->set('PRT_ID', $base_price_type->ID());
```
manzoorwanijk, line 103:
```suggestion
        $new_base_price->set('PRT_ID', EEM_Price_Type::base_type_base_price);
```

core/domain/services/admin/entities/DefaultPrices.php
```diff
@@ -0,0 +1,111 @@
...
+            ]
+        ]);
+        if (! $base_price instanceof EE_Price) {
+            throw new RuntimeException(
+                esc_html__(
+                    'A valid base price could not be retrieved from the database.',
+                    'event_espresso'
+                )
+            );
+        }
```
manzoorwanijk, line 104: @tn3rb what if there are no prices in the database? Shouldn't we still create a base price?

core/domain/services/admin/entities/DefaultPrices.php
tn3rb: DOH 🤦‍♂️  good catch

core/domain/services/admin/entities/DefaultPrices.php
tn3rb: another DOH 🤦‍♂️

core/domain/services/admin/entities/DefaultPrices.php
tn3rb: default price is normally zero, but are you suggesting we set it based on the ticket price?

### 🤔 

hmmm ya we could do that...

core/domain/services/admin/entities/DefaultPrices.php
tn3rb: no this is not correct

`PRT_ID` is the database ID for the actual price type record, whereas `EEM_Price_Type::base_type_base_price` corresponds to the `PBT_ID` 


PRT_ID | PRT_name | PBT_ID
-- | -- | --
1 | Base Price | 1
2 | Percent Discount | 2
3 | Dollar Discount | 2
4 | Percent Surcharge | 3
5 | Dollar Surcharge | 3
6 | Regional Tax | 4
7 | Federal Tax | 4

so although by default... on a fresh install of EE, `PRT_ID` === `PBT_ID` (`EEM_Price_Type::base_type_base_price`) 

if someone deletes those records and/or mucks with this table in their db, then `PRT_ID` for the base price type can end up being something completely different. This is why I was actually pulling the base price type record from the db and then using its actual `PRT_ID`.

We've had this issue happen in the past where a user had manually deleted these records from the db, which were then repopulated on the next EE request but with new higher auto-incremented IDs. Needless to say, things went 💥 !

manzoorwanijk: Thanks 👍

manzoorwanijk: Then we first need to get the pricetype object and use its ID

core/domain/services/admin/entities/DefaultPrices.php
tn3rb: hmmmm...
maybe... although my concern at this point is that something else is happening that is preventing that from occurring. There's other logic in place already that checks for default system data and creates it if need be. So I'm inclined to just leave things as is and force the user to contact support to figure out what the real root issue is.

manzoorwanijk: Cool 👍

Issue #2834, opened 2020-05-14
Fix Tooltips For Buttons


tn3rb:
tooltips are not necessary when they contain the same text as the item they are wrapping:

![ScreenShot_20200514125307.png](https://images.zenhubusercontent.com/55c8c0f14f7a08fbb7f431cd/a001e752-0f5c-4711-b73d-8d905c66a862)

PR #2838, opened 2020-05-15
fix tooltips

tn3rb:
see: #2834

- standardizes prop names to "tooltip" and "tooltipProps" (for use with composed elements)

- standardizes tooltip placement for similar elements

- removes tooltips that duplicate button text or labels

- fixes styles

### DIFF
--- a/assets/src/application/ui/display/withLabel/withLabel.tsx
+++ b/assets/src/application/ui/display/withLabel/withLabel.tsx
@@ -4,8 +4,10 @@ import { LabelPosition, withLabelProps } from './types';
 import './style.scss';
 
 const withLabel = <P extends withLabelProps>(WrappedComponent: React.ComponentType<P>) => {
+
 	type Ref = React.Ref<typeof WrappedComponent>;
-	type refProps = { forwardedRef: Ref };
+	type refProps = { forwardedRef: Ref; };
+
 	const WithLabel: React.FC<P & refProps> = ({
 		forwardedRef,
 		id,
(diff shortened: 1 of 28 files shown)


assets/src/infrastructure/ui/display/Tooltip/index.tsx
```diff
@@ -1,16 +1,22 @@
...
 
 import type { TooltipProps } from '../types';
 import './style.scss';
 
-const Tooltip: React.FC<TooltipProps> = ({ children, ...props }) => {
-	const ariaLabel = props.title || props['aria-label'];
-	const label = props.title;
+const Tooltip: React.FC<TooltipProps> = ({ children, tooltip, ...props }) => {
+	const ariaLabel = tooltip || props['aria-label'];
+	const className = classNames(props.className, 'ee-tooltip');
```
manzoorwanijk, line 10: Nice change 👍

assets/src/application/ui/display/withLabel/withLabel.tsx
```diff
@@ -14,35 +16,25 @@ const withLabel = <P extends withLabelProps>(WrappedComponent: React.ComponentTy
...
-			'ee-input-label__wrapper--bottom-center': labelPosition === LabelPosition.BOTTOM_CENTER,
-			'ee-input-label__wrapper--bottom-left': labelPosition === LabelPosition.BOTTOM_LEFT,
-			'ee-input-label__wrapper--left-bottom': labelPosition === LabelPosition.LEFT_BOTTOM,
-			'ee-input-label__wrapper--left-middle': labelPosition === LabelPosition.LEFT_MIDDLE,
-			'ee-input-label__wrapper--left-top': labelPosition === LabelPosition.LEFT_TOP,
-		});
+		const className = classNames(
+			labelClassName,
+			'ee-input-label__wrapper',
+			labelPosition && `ee-input-label__wrapper--${labelPosition}`
```
nicolad, line 22: :sunglasses:

assets/src/application/ui/display/withTooltip/withTooltip.tsx
```diff
@@ -35,40 +31,39 @@ const withTooltip = <P extends withTooltipProps>(WrappedComponent: React.Compone
 				'ee-mobile-help-text--short': tooltip.length < 25,
 				'ee-mobile-help-text--long': tooltip.length > 50,
 			});
-			tooltipProps = { ...props.tooltipProps, overlayClassName: 'ee-mobile-help-text__tooltip' };
+			tooltipProps = { ...tooltipProps, className: 'ee-mobile-help-text__tooltip' };
 			toolTipped = (
 				<div className='ee-mobile-help-text__btn-wrap'>
 					<WrappedComponent
-						{...(props as P)}
-						aria-label={ariaLabel}
```
nicolad, line 43: I've added `aria-label` here because some of the chakra's components had it required.
I suppose that now it should be passed explicitly instead of being derived from `tooltip`?

tn3rb: ya that's correct, the tooltip should be thought of more as "help text" which is not the same thing as an aria-label anyways
a little more work for us, but end result should be better a11y

PR #2842, opened 2020-05-16
ton-o-ui-fixes

tn3rb:
- simplifies and standardizes borders, colours, classes, plus hover and focus states for inputs and buttons

- improves tooltip readability and aligns styles with rest of EDTR

- fixes that damn fugly uber datepicker

- tweaks sidebar calendar elements


assets/src/application/ui/forms/espressoForm/renderers/FieldRenderer.tsx
```diff
@@ -27,23 +27,27 @@ const FieldRenderer: React.FC<FieldRendererProps> = (props) => {
...
-			<FormLabel htmlFor={ props.input.name }>
-				{ label }
-				{ info && (
-					<Tooltip placement='right' tooltip={ info }>
-						<InfoCircleOutlined />
+			<FormLabel htmlFor={props.input.name}>
+				{label}
+				{info && (
+					<Tooltip placement='right' tooltip={info}>
+						<InfoCircleOutlined className='ee-form-item__help-info' />
```
manzoorwanijk, line 42: 👍

assets/src/infrastructure/ui/layout/menu/Menu.tsx
```diff
@@ -4,8 +4,12 @@ import { Menu as ChakraMenu } from '@chakra-ui/core';
 
 import { MenuProps } from './types';
 
-const Menu: React.FC<MenuProps> = ({ children, ...props }) => {
-	return <ChakraMenu {...props}>{children}</ChakraMenu>;
+const Menu: React.FC<MenuProps> = ({ closeOnBlur = true, children, ...props }) => {
```
nicolad, line 7: set `closeOnBlur` explicitly to `true`, but it's still not closing when it's clicked outside of the area

assets/src/application/ui/display/TimezoneTimeInfo/style.scss
```diff
@@ -1,11 +1,30 @@
...
+		height: 30px;
+		margin: var(--ee-margin-micro);
+		min-width: 30px;
+		padding: 4px;
+		width: 30px;
+	}
+
+	&__icon {
+		height: 16px;
+		margin: 0;
```
nicolad, line 18: doesn't seem to have any effects, is this fixing smth?

assets/src/application/ui/display/TimezoneTimeInfo/Trigger.tsx
```diff
@@ -8,19 +8,17 @@ interface TriggerProps {
...
-		/>
-	)
-);
+const Trigger = forwardRef<typeof IconButton, TriggerProps>(({ tooltip, ...props }, ref) => (
+	<IconButton
+		{...props}
+		borderless
+		color='white'
+		className='ee-timezone-info__button'
+		icon={() => <GlobalOutlined className='ee-timezone-info__icon' />}
```
nicolad: I think it's better to lift these kind of styles at higher leve, had created `withSize` HOC, if any more props will be needed, can rename this differently.

assets/src/application/ui/display/withLabel/withLabel.tsx
```diff
@@ -17,20 +17,21 @@ const withLabel = <P extends withLabelProps>(WrappedComponent: React.ComponentTy
 		...props
 	}) => {
 		const className = classNames(
-			labelClassName,
-			'ee-input-label__wrapper',
-			labelPosition && `ee-input-label__wrapper--${labelPosition}`
+			'ee-input__wrapper',
+			label && labelClassName,
```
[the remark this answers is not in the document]
nicolad: or I might miss smth here: `label && labelClassName` ?

assets/src/application/ui/forms/espressoForm/renderers/FieldRenderer.tsx
nicolad: it's ok, but maybe we can centralize this with a prop and apply needed styles at higher level,
the intention being a less fragile code

assets/src/application/ui/forms/espressoForm/styles.scss
```diff
@@ -75,16 +75,17 @@
...
-			width: 50%;
-			display: inline-flex;
-			padding-right: var(--ee-margin-default);
-			margin: var(--ee-margin-default) 0;
-			label {
-				display: flex;
-				align-items: center;
+
+			&__help-info {
+				color: var(--ee-color-primary-low-contrast);
```
nicolad, line 87: IMO, it's better to centralize color, to make it more consistent with other icons styles;
when doing previous refactor I noticed that is a bit hard to maintain these kinds of local patched styles

assets/src/infrastructure/ui/layout/menu/MenuToggle.tsx
```diff
@@ -6,6 +6,7 @@ import { MenuToggleProps } from './types';
 const MenuToggle = React.forwardRef<typeof ChakraMenuButton, MenuToggleProps>(
 	({ children, variant = 'unstyled', ...props }, ref) => {
 		return (
+			// @ts-ignore
 			<ChakraMenuButton {...props} variant={variant} ref={ref}>
```
manzoorwanijk: Better to add it only for that prop in a separate line
```suggestion
			<ChakraMenuButton
				{...props}
				// @ts-ignore
				variant={variant}
				ref={ref}
			>
```

nicolad: good catch, updated :+1:

assets/src/infrastructure/ui/layout/menu/Menu.tsx
nicolad: @manzoorwanijk I had reverted this

assets/src/infrastructure/ui/layout/menu/MenuToggle.tsx
tn3rb: thnx guys

assets/src/application/ui/forms/espressoForm/renderers/FieldRenderer.tsx
tn3rb: would be nice if our icons had a class applied to them by default, ex: `ee-icon--help-info`

nicolad: yes, a generated one based on the component name would be smth interesting

assets/src/application/ui/display/TimezoneTimeInfo/style.scss
tn3rb: there _was_ a margin being applied to that icon that this was removing... ???

assets/src/application/ui/display/withLabel/withLabel.tsx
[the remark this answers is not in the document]
tn3rb: no you're right... i didn't realize that about that lib

assets/src/application/ui/forms/espressoForm/styles.scss
tn3rb: hmmm  ya this is case is different than the others since it is on the coloured background of the sidebar, but you're right that that could be encoded somehow maybe with an `inverse` prop or something

assets/src/application/ui/display/icons/svgs/withSize.tsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import classNames from 'classnames';
+import { withSizeProps, IconSize } from './types';
+import './style.scss';
+
+const withSize = <P extends withSizeProps>(WrappedComponent: React.ComponentType<P>) => {
+	type ISize = { size: IconSize };
+
+	const WithSize: React.FC<P & ISize> = ({ size, ...props }) => {
+		const className = classNames(props.className, size && `ee__icon--${size}`);
```
tn3rb, line 10: can this not be baked right into our icon component?
or will that not work because those are dynamically generated?

I think of HOCs as enhancing a component with additional logic that doesn't belong in the component, whereas an icon's size seems like it should be a property of the icon itself, but there might be a technical reason why you have done this.

assets/src/infrastructure/ui/layout/menu/Menu.tsx
tn3rb: I moved the tooltip out of the dropdown menu and then wrapped that component in the `witTooltip` HOC:

https://github.com/eventespresso/event-espresso-core/pull/2838/commits/c5e220e23329a6160d484282c88897089125ad0e#diff-1b4e6999f94ed2e1742f9f0aff62127eR8-R33

so that might have caused this change, but we should jump back to a commit before the above was merged to confirm

assets/src/application/ui/forms/espressoForm/renderers/FieldRenderer.tsx
```diff
@@ -39,7 +39,7 @@ const FieldRenderer: React.FC<FieldRendererProps> = (props) => {
 				{label}
 				{info && (
 					<Tooltip placement='right' tooltip={info}>
-						<InfoCircleOutlined className='ee-form-item__help-info' />
+						<InfoCircleOutlined />
```
nicolad, line 42: image

assets/src/application/ui/forms/espressoForm/styles.scss
nicolad: yes, `inverse` prop sounds like an interesting approach

assets/src/application/ui/display/icons/svgs/withEnhance.tsx
```diff
@@ -0,0 +1,38 @@
...
+		.join('-');
+};
+
+const withEnhance = <P extends withSizeProps>(WrappedComponent: React.ComponentType<P>) => {
+	type ISize = { noMargin?: boolean; size: IconSize };
+
+	// @ts-ignore
+	const displayName = WrappedComponent?.render?.name;
+
+	const generatedIconClassName = displayName && 'ee-icon--' + camelToSnakeCase(displayName.replace('Svg', ''));
```
tn3rb, line 19: cool 👍

assets/src/application/ui/display/icons/svgs/withEnhance.tsx
```diff
@@ -0,0 +1,38 @@
+import React from 'react';
+import classNames from 'classnames';
+import { withSizeProps, IconSize } from './types';
+import './style.scss';
+
+const camelToSnakeCase = (str: string) => {
```
nicolad, line 6: @tn3rb let me know if this is smth that can be reused, such that it can be moved into a cross-cutting folder

assets/src/application/ui/forms/espressoForm/renderers/FieldRenderer.tsx
nicolad: done

assets/src/application/ui/display/icons/svgs/withEnhance.tsx
```diff
@@ -0,0 +1,38 @@
...
+	const WithSize: React.FC<P & ISize> = ({ noMargin, size, ...props }) => {
+		const className = classNames(
+			'ee-svg',
+			size && `ee__icon--${size}`,
+			noMargin && 'ee__icon--no-margin',
+			generatedIconClassName,
+			props.className
+		);
+
+		console.log('props.className', props.className);
```
tn3rb, line 30: oops... need to remove this...

assets/src/application/ui/display/icons/svgs/withEnhance.tsx
```diff
@@ -0,0 +1,38 @@
...
+	const generatedIconClassName = displayName && 'ee-icon--' + camelToSnakeCase(displayName.replace('Svg', ''));
+
+	const WithSize: React.FC<P & ISize> = ({ noMargin, size, ...props }) => {
+		const className = classNames(
+			'ee-svg',
+			size && `ee__icon--${size}`,
+			noMargin && 'ee__icon--no-margin',
+			generatedIconClassName,
+			props.className
+		);
```
tn3rb, line 28: HOCception !!!
can this not just be combined with the logic inside `withEnhance` ???

assets/src/application/ui/display/withLabel/withLabel.tsx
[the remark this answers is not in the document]
nicolad: ok, will update that

tn3rb:
> let me know if this is smth that can be reused

oh ya I think so... maybe extract to `/application/services/utilities/text/`

assets/src/application/ui/display/icons/svgs/withSize.tsx
nicolad: yes, we can pass directly height and width:
```
<GlobalOutlined heght="1rem" noMargin width="1rem"/>
```
but I thought that having an universal approach is easier to maintain, rather than explicit settings

assets/src/application/ui/display/icons/svgs/withEnhance.tsx
nicolad: splitted, thx

assets/src/application/ui/display/icons/svgs/withEnhance.tsx
nicolad: thx

assets/src/application/ui/display/TimezoneTimeInfo/style.scss
nicolad: you are right, that has been applied via `ee-svg`, my bad

assets/src/application/ui/display/icons/svgs/index.ts
```diff
@@ -43,6 +47,6 @@ export { ReactComponent as WarningTriangle } from './warningTriangle.svg';
 
 // Enhanced icons
 
-export const InfoCircleOutlined = withEnhance(InfoCircleOutlinedSVG);
+export const InfoCircleOutlined = pipe(withGeneratedClassName, withSize)(InfoCircleOutlinedSVG);
 
-export const GlobalOutlined = withEnhance(GlobalOutlinedSVG);
+export const GlobalOutlined = pipe(withGeneratedClassName, withSize)(GlobalOutlinedSVG);
```
nicolad: the order of HOCs that are applied matters, tried the other way and it wasn't working as expected

assets/src/application/ui/display/icons/svgs/index.ts
```diff
@@ -1,4 +1,8 @@
-import withEnhance fro./withEnhancenhance';
+import { pipe } from 'ramda';
```
nicolad, line 1: I know only this function from `ramda`

assets/src/application/ui/display/icons/svgs/index.ts
```diff
@@ -39,3 +44,9 @@ export { ReactComponent as UpCircleFilled } from './upCircleFilled.svg';
 export { ReactComponent as Warning } from './warning.svg';
 
 export { ReactComponent as WarningTriangle } from './warningTriangle.svg';
+
+// Enhanced icons
+
+export const InfoCircleOutlined = pipe(withGeneratedClassName, withSize)(InfoCircleOutlinedSVG);
+
+export const GlobalOutlined = pipe(withGeneratedClassName, withSize)(GlobalOutlinedSVG);
```
tn3rb, line 52: sry I don't think I was very clear in my comments before...

when I said: 

> can this not just be combined with the logic inside withEnhance ???

what I meant was that having a separate HOC **just** for applying the icon name as a css class to an icon...
as well as **another** separate HOC for applying size and margin related classes to an icon...
seems kind of redundant...

They are both just applying classes to an icon, therefore it seems more logical to me to combine these together into one HOC that applies icon classes.
To go further I **actually** think that typical modifier classes for a component should be part of the component itself and not an HOC which imho should be reserved for applying unrelated and/or atypical changes to a component.

In both of these cases however I think that the enhancements should apply to ALL of our icons:

- adding the icon name as a css class should be applied to ALL of our icons, because that will undoubtedly be handy and allow any icon to be targeted via css. This is similar to how all of the WP dashicons have their icon name as part of their css classes like `dashicons dashicons-calendar`.

- being able to change the size of an icon also seems like something that all icons should have the capacity to do. AFAIK this can be achieved indirectly by modifying the font size for the container, but imho it owul dbe better to have some standardized sizes that could be controlled more centrally

- removing margins also seems pretty generic and could likely be a nice to have for any icon

So bottom line, let's just have a single HOC for adding css modifiers
(and I'm assuming that the HOC is used because the icon components are dynamically generated and there isn't a single `Icon` component where that logic can be centralized?

assets/src/application/ui/forms/espressoForm/renderers/FieldRenderer.tsx
tn3rb: love it 👍 great job

assets/src/application/ui/display/icons/svgs/index.ts
nicolad: there are custom templates from `svgr` which would apply that HOC to each of the SVG component: https://react-svgr.com/docs/custom-templates/
***
need to think of an easier way to achieve this

nicolad: tried dynamically to apply those HOC, but got no luck:
```
import { pipe } from 'ramda';

import withGeneratedClassName from './svgs/withGeneratedClassName';
import withSize from './svgs/withEnhance';

import * as icons from './svgs';

const ehancedIcons = Object.entries(icons).reduce((acc, [name, component]) => {
	return (acc[name] = pipe(withGeneratedClassName, withSize)(component));
}, {});

export default ehancedIcons;
```

assets/src/application/ui/display/icons/svgs/withEnhance.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+		return <WrappedComponent {...(props as P)} className={className} />;
+	};
+
+	return WithEnhance;
+};
```
tn3rb, line 30: ya that's so much better imho

Issue #2797, opened 2020-05-11
EDTR Data Handling Checklist


tn3rb:
on latest `EDTR/master` branch:

- confirm that all CRUD functionality works as expected:

     - [x] entity creation

     - [x] entity editing

     - [x] entity deletion

     - [x] dates and times update correctly with no offset shenanigans

- confirm that all TAM relation logic works as expected

     - [x] existing relations

     - [x] adding new relations

     - [x] removing new relations

     - [x] removing existing relations

     - [x] error states for missing relations

- confirm that all Price Calculator logic works as expected

     - [x] existing prices and price modifiers

     - [x] adding new price modifiers

     - [x] removing new price modifiers

     - [x] removing existing price modifiers

     - [x] reverse calculation -> ticket toal and/or base price

     - [x] base price is always first in list

     - [x] default base price is added if none exists

     - [x] only one base price can be added and does not appear in price type options list for other modifiers

     - [x] new calculated ticket total price is reflected in ticket card or ticket edit form when calculator is closed

     - [x] editing ticket total price from ticket card or ticket edit form is reflected when calculator is opened

- review existing unit test code coverage for data management

     - [x] add unit tests where possible to increase coverage as much as possible (ie: 90% +)

manzoorwanijk:
- For a new and existing event
    - [ ] New Date/Ticket
    - [ ] Inline Edit of fields
    - [ ] Editing fields in modal
    - [ ] ❌ **Date and time has unexpected change, #2753**
    - [ ] Inline date capacity change updates related ticket quantity
    - [ ] ❌ **Date capacity change in modal does NOT update related ticket quantity**
    - [ ] Trashing date/ticket
    - [ ] Trashing date, trashes lone related tickets
    - [ ] Permanent deletion of date deletes lone related tickets
    - [ ] Sorting of dates/tickets

manzoorwanijk:
- TAM (For new and exiting event)
    - [ ] New relations added
    - [ ] Existing relations retained
    - [ ] Existing relations deleted
    - [ ] Newly added relations deleted
    - [ ] Error for orphan dates and tickets

manzoorwanijk:
Some miscellaneous issues
- When publishing a new event
`[12-May-2020 08:57:24 UTC] PHP Notice:  Undefined index: edit_event_datetimes in wp-content/plugins/event-espresso-core/admin_pages/events/Events_Admin_Page.core.php on line 1182`

[...]

PR #2844, opened 2020-05-18
Fix isTaxable

manzoorwanijk:
see [#2797](https://github.com/eventespresso/event-espresso-core/issues/2797#issuecomment-630101968)

This PR fixes the `isTaxable` issue of tickets


assets/src/domain/eventEditor/ui/tickets/ticketPriceCalculator/data/useDataReducer.ts
```diff
@@ -13,6 +13,7 @@ export const initialState: DataState = {
 const useDataReducer = (initializer: StateInitializer): DataStateReducer => {
 	const dataReducer: DataStateReducer = (state, action) => {
 		const { type, id, index, fieldValues, ticketPrice, price, prices } = action;
+		let isTaxable: boolean;
```
nicolad, line 16: what was the reason it was pulled at a higher level?

manzoorwanijk: because it's use in many switch cases.

Issue #2847, opened 2020-05-18
Entity Action Main Menu Dropdown Issues


tn3rb:
![ScreenShot_20200518142407.png](https://images.zenhubusercontent.com/55c8c0f14f7a08fbb7f431cd/24cd77c5-00f2-4138-90c1-daf8d3f337ad)

- dropdown position seems to be odd and not location aware (ie: will appear off screen instead of repositioning in order to be visible)

- dropdown will ONLY close if you click the menu button a second time

nicolad:
positioning depends on `popper.js`, let's see if updating to v2 in chakra will fix this issue;
asked [here](https://github.com/chakra-ui/chakra-ui/issues/469#issuecomment-630858126) a more exact time of when upgrade will happen

PR #2849, opened 2020-05-19
Entity Action Main Menu Dropdown Issues.

nicolad:
- Closes: https://github.com/eventespresso/event-espresso-core/issues/2847

### DIFF
--- a/assets/src/application/ui/input/Button/IconButton/index.tsx
+++ b/assets/src/application/ui/input/Button/IconButton/index.tsx
@@ -23,13 +23,13 @@ const IconButton = forwardRef<typeof IconButtonAdapter, IconButtonProps>(
 
 		return (
 			<IconButtonAdapter
-				aria-label={ ariaLabel }
-				className={ className }
-				icon={ icon }
-				onClick={ onClick }
-				tabIndex={ 0 }
-				ref={ ref }
-				{ ...props }
+				{...props}
+				aria-label={ariaLabel}
+				className={className}
+				icon={icon}
+				onClick={onClick}
+				tabIndex={0}
+				ref={ref}
 			/>
 		);
 	}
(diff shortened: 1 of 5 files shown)


nicolad:
Tried to unwrap `withTooltip` from `DropdownToggle` , doesn't had any effects:

Peek 2020-05-19 14-06

nicolad:
Used menu component directly and it seems to work:

Peek 2020-05-19 14-29

nicolad:
don't understand how this is related, but noticed that while I had fixed that issue with closing upon clicking outside the positioning changed a little bit:

Peek 2020-05-19 17-32

assets/src/application/ui/input/Button/IconButton/index.tsx
```diff
@@ -23,13 +23,13 @@ const IconButton = forwardRef<typeof IconButtonAdapter, IconButtonProps>(
...
 		return (
 			<IconButtonAdapter
-				aria-label={ ariaLabel }
-				className={ className }
-				icon={ icon }
-				onClick={ onClick }
-				tabIndex={ 0 }
-				ref={ ref }
-				{ ...props }
+				{...props}
```
nicolad, line 26: this has overridden some props

assets/src/application/ui/input/Button/style.scss
```diff
@@ -18,7 +18,7 @@
 	box-shadow: 0 0 0 1px var(--ee-color-primary-low-contrast);
 }
 
-.ee-btn-base {
+.ee-btn-base.ee-btn-base {
```
nicolad, line 21: this was added to fight some of the default chakra's styles

nicolad:
> do we know what the definitive cause was?

my educated guess is inclined towards the fact that chakra's menu did not liked smth from the `withLabel` HOC from the `IconButton` under the `@application/ui/input`, that's why I switched to the other one from `@infraUI/inputs`; didn't dug deeper because it already took some time to figure out

PR #2165, opened 2020-01-02
Combine DOM dump data

manzoorwanijk:
This PR updates/combines the DOM data required for editor. The new shape of the data can be seen in `/domain/eventEditor/types.ts`

```ts
interface EventData {
	dbId: number;
	datetimes?: DatetimeEdge;
	tickets?: TicketEdge;
	prices?: PriceEdge;
	priceTypes?: PriceTypeEdge;
	relations?: RelationalData;
}

interface EEEditorData {
	event: EventData;
	graphqlEndpoint?: string;
	currentUser?: CurrentUserProps;
	generalSettings?: GeneralSettings;
}
```
The reason for:
- Creating `EventData ` and moving its related data into `EEEditorData.event` is to make room for future event related data.
- Using `dbId` instead of just `id` or `eventId` is to make it consistent with current shape of EDTR entities and leave `id` for GUID if we ever need that.

Completes #2144 

## Checklist

* [ ] I have read the documentation relating to systems affected by this pull request, see https://github.com/eventespresso/event-espresso-core/tree/master/docs
* [ ] User input is adequately validated and sanitized
* [ ] all publicly displayed strings are internationalized (usually using `esc_html__()`, see https://codex.wordpress.org/I18n_for_WordPress_Developers)
* [ ] My code is tested.
* [ ] My code follows the Event Espresso code style.
* [ ] My code has proper inline documentation.
* [ ] My code accounts for when the site is in Maintenance Mode (MM2 especially disallows usage of models)

PR #2850, opened 2020-05-19
Dynamic entry points

### LINKED ISSUE #2825: Add Dynamic Entry Point Routing
- currently webpack uses a hardcoded entry point in `config/webpack.config.js`

- this is fine atm because we are currently only using React for the event editor

- eventually we will begin to convert other admin pages (as well as frontend elements) to use React + AGQL

- before this happens we will need to know how we are going to handle routing

- ideally, we will: 

     - dynamically route all requests through a single webpack entry to the appropriate UI

     - lazy load components accordingly

     - dynamically fetch data that has been preloaded into the DOM (likely nothing any more special that what we are doing already is needed here, but some logic could potentially be abstracted and shared better)

manzoorwanijk:
This PR creates dynamic entry points for all domains. Here is how it works
- Name of the main entry point is `eventespresso-main` which is loaded for all domains
- This root entry point check for `window.eeDomain` which is a string name of the domain to load
- Each domain must have `main.ts` in the root directory to import/init the domain

#### Why `main.ts`?
If we use `index.ts`, webpack matches all the subdirectories and creates separate tons of bundles for each directory. Also, it slow down the build process quite considerably. So, `main.ts` is to make the pattern a bit explicit for faster builds.

### What's next?
- Create/standardize DOM data for `eeDomain` and `assetsUrl` #2165 
- Centralize loading of main file for each route
- It no longer needs CSS to be enqueued, it's handled by JS
- Webpack config  clean up

Closes #2825


assets/src/domain/eventEditor/interfaces/types.ts
```diff
@@ -29,5 +29,6 @@ declare global {
 	interface Window {
 		eeEditorData: EEEditorData;
 		eejsdata: EEJSData;
+		eeDomain: string;
```
nicolad, line 32: noticed that this is in `eventEditor` domain (`assets/src/domain/eventEditor/interfaces/types.ts`),
this will be moved later?

assets/src/domain/index.ts
```diff
@@ -0,0 +1,14 @@
...
+if (domain) {
+	import(
+		/* webpackExclude: /(shared)/ */
+		/* webpackChunkName: "[request]" */
+		`./${domain}/main.ts`
+	).catch(console.error);
+	console.log('domain', domain);
+} else {
+	console.error('No domain supplied');
+}
```
tn3rb, line 14: LOLZ... that was easy 😆

assets/src/domain/shared/services/publicPath.ts
```diff
@@ -0,0 +1,21 @@
...
+no-var
+*/
+/* global __webpack_public_path__ */
+declare var __webpack_public_path__: string;
+let assetsUrl = window?.eeEditorData?.assetsUrl;
+
+// TODO add assetUrl to global data for all domains
+// temporary workaround
+if (!assetsUrl) {
+	assetsUrl = window.location.origin + '/wp-content/plugins/event-espresso-core/assets/dist/';
```
tn3rb, line 16:
```
'/wp-content/plugins/event-espresso-core/assets/dist/'
```

we'll have to get this from the DOM because sometimes people rename the EE folder while FTPing files to their server

config/webpack.config.js
```diff
@@ -132,20 +131,14 @@ module.exports = function (webpackEnv) {
...
 	const externals = {
 		react: 'React',
 		'react-dom': 'ReactDOM',
-		jquery: 'jQuery'
+		jquery: 'jQuery',
 	};
 
 	// Define WordPress dependencies
-	const wpPackages = ['components', 'element', 'blocks', 'i18n',  'block-editor'];
+	const wpPackages = ['components', 'element', 'blocks', 'i18n', 'block-editor'];
```
tn3rb, line 141: maybe at some point we can split this so that the block related dependencies only load for the blocks domain

assets/src/domain/index.ts
```diff
@@ -0,0 +1,14 @@
+import './shared/services/publicPath';
+
+const domain = window?.eeDomain;
+
+if (domain) {
+	import(
+		/* webpackExclude: /(shared)/ */
+		/* webpackChunkName: "[request]" */
+		`./${domain}/main.ts`
```
tn3rb, line 9: `main.ts` is good, but I'm thinking we could make it 100% explicitly clear by using something like `entry-point.ts`

assets/src/domain/index.ts
```diff
@@ -0,0 +1,14 @@
...
+	).catch(console.error);
+	console.log('domain', domain);
```
nicolad, line 11: what about:
```
	import(
		/* webpackExclude: /(shared)/ */
		/* webpackChunkName: "[request]" */
		`./${domain}/main.ts`
	)
.then(() => {
	console.log('domain', domain);
})
.catch(console.error);
```
?
it is a bit uncommon to see a promise with just a catch

assets/src/domain/shared/services/publicPath.ts
```diff
@@ -0,0 +1,21 @@
+/* eslint-disable 
+no-unused-vars,
+no-native-reassign,
+no-global-assign,
+@typescript-eslint/camelcase,
+@typescript-eslint/no-unused-vars,
+no-var
+*/
```
nicolad, line 8: these are similar to magic comments?
https://webpack.js.org/api/module-methods/#magic-comments

assets/src/domain/shared/services/publicPath.ts
manzoorwanijk: Yes

assets/src/domain/index.ts
manzoorwanijk: Agreed

assets/src/domain/shared/services/publicPath.ts
manzoorwanijk: ESLint and TS don't like that global variable being undefined and modified.

assets/src/domain/index.ts
manzoorwanijk: Done

assets/src/domain/index.ts
manzoorwanijk: Done

manzoorwanijk:
@tn3rb It loads everything well except blocks for which the assets are not being loaded at all.

config/paths.js
```diff
@@ -74,23 +74,13 @@ module.exports = {
...
-	entries: {
-		'eventespresso-editor': resolveModule(resolveApp, ASSETS_SOURCE_FOLDER + '/domain/eventEditor/index'),
-		'eventespresso-blocks': resolveModule(resolveApp, ASSETS_SOURCE_FOLDER + '/domain/blocks/index'),
-		'eventespresso-wp-plugins-page': resolveModule(resolveApp, ASSETS_SOURCE_FOLDER + '/domain/wpPluginsPage/index'),
-		// To be done as part of: https://github.com/eventespresso/event-espresso-core/issues/2250
-		// 'ee-wp-plugins-page': resolveModule(resolveApp, assetsPath + 'wp-plugins-page/index'),
-		// 'eventespresso-core-blocks': resolveModule(resolveApp, assetsPath + 'blocks/index'),
-		// 'eventespresso-core-blocks-frontend': resolveModule(resolveApp, assetsPath + 'blocks/index-frontend'),
-	},
-
```
nicolad, line 93: :sunglasses:

assets/src/domain/eventEditor/interfaces/types.ts
manzoorwanijk: Yes, when global assets are combined, those 3 will be removed and replaced with that global object.

PR #2852, opened 2020-05-19
TAM close btn fix.

nicolad:
- The issue was caused by increased specificity.

## How has this been tested
Peek 2020-05-19 19-44

### DIFF
--- a/assets/src/application/ui/input/Button/style.scss
+++ b/assets/src/application/ui/input/Button/style.scss
@@ -18,7 +18,7 @@
 	box-shadow: 0 0 0 1px var(--ee-color-primary-low-contrast);
 }
 
-.ee-btn-base.ee-btn-base {
+.ee-btn-base {
 	border-style: solid;
 	border-width: 2px;
 	cursor: pointer;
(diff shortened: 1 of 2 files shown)


assets/src/infrastructure/ui/layout/modal/modalCloseButtonProps/styles.scss
```diff
@@ -5,7 +5,6 @@
 		position: absolute;
 		top: 8px;
 		right: 12px;
-		min-width: auto;
```
nicolad, line 8: related to this width:

image

assets/src/application/ui/input/Button/style.scss
```diff
@@ -18,7 +18,7 @@
 	box-shadow: 0 0 0 1px var(--ee-color-primary-low-contrast);
 }
 
-.ee-btn-base.ee-btn-base {
+.ee-btn-base {
```
tn3rb, line 21: uh oh.. is this going to cause issues elsewhere?
do you remember what you added this in to fix?

nicolad: yes, I have added that recently because there was mismatch between these 2 buttons:

image

nicolad: image

nicolad: but then added a helper css class name and that specificity wasn't needed, but had to research the impact of that specificity

nicolad: was added here: https://github.com/eventespresso/event-espresso-core/pull/2849/files#r427355084

PR #2854, opened 2020-05-20
fix loading for block assets

tn3rb:
### in this PR

- changes primary entry point name from 'eventespresso-main' to 'eventespresso-core-app'

- sets 'blocks' domain and loads assets for routes requiring GB blocks

- fixes SSR for Event Attendees block by parsing DBIDs from GUIDs

### outstanding issue

The Event Espresso block category appears in the GB editor and the Event Attendees block can be loaded, configured, saved, and rendered on the frontend correctly. However, refreshing the editor page (even after having saved the post) and/or leaving the editor and returning to the same post, results in the following error appearing where the Event Attendees block was previously:

> Your site doesn’t include support for the "eventespresso/event-attendees" block. You can leave this block intact or remove it entirely.

without doing anything else, you can re-add the block and have everything work as before, but the previously saved block is now completely broken. If you refresh at this point you will now have two of the above errors.

It appears that the problem may be caused by incorrect loading order in the JS, according to similar issues I found online:

- https://github.com/WordPress/gutenberg/issues/13793
- https://wordpress.stackexchange.com/questions/355536/your-site-doesn-t-include-support-for-the-block-after-registering-a-block

I attempted to apply a fix based on that second link, but it did not work. I've left that code in for the time being.

### DIFF
--- a/assets/src/application/services/utilities/dom/domReady.ts
+++ b/assets/src/application/services/utilities/dom/domReady.ts
@@ -1,0 +1,17 @@
+/**
+ * TypeScript copy of @wordpress/dom-ready
+ * @see https://github.com/WordPress/gutenberg/blob/master/packages/dom-ready/src/index.js
+ */
+const domReady = (callback: VoidFunction): void => {
+	if (
+		document?.readyState === 'complete' || // DOMContentLoaded + Images/Styles/etc loaded, so we call directly.
+		document?.readyState === 'interactive' // DOMContentLoaded fires at this point, so we call directly.
+	) {
+		return callback();
+	}
+
+	// DOMContentLoaded has not fired yet, delay callback until then.
+	document?.addEventListener('DOMContentLoaded', callback);
+};
+
+export default domReady;
--- a/assets/src/domain/blocks/event-attendees/index.tsx
+++ b/assets/src/domain/blocks/event-attendees/index.tsx
@@ -1,29 +1,27 @@
 import { registerBlockType } from '@wordpress/blocks';
 import { __ } from '@wordpress/i18n';
 
+import { canUseDOM, domReady } from '@appServices/utilities/dom';
+
 import attributes from './attributes';
 import EventAttendeesEdit from './edit';
 import type { EventAttendeesAttributes } from './types';
 import './style.scss';
 
-registerBlockType<EventAttendeesAttributes>('eventespresso/event-attendees', {
-	title: __('Event Attendees', 'event_espresso'),
+const registerEventAttendeesBlock: VoidFunction = () => {
+	registerBlockType<EventAttendeesAttributes>('eventespresso/event-attendees', {
+		title: __('Event Attendees', 'event_espresso'),
+		description: __('Displays a list of people that have registered for the specified event', 'event_espresso'),
+		icon: 'groups',
+		category: 'event-espresso',
+		keywords: [__('event', 'event_espresso'), __('attendees', 'event_espresso'), __('list', 'event_espresso')],
+		attributes,
+		edit: EventAttendeesEdit,
+		save() {
+			return null;
+		},
+		// TODO migrate attributes to use GUID
+	});
+};
 
-	description: __('Displays a list of people that have registered for the specified event', 'event_espresso'),
-
-	icon: 'groups',
-
-	category: 'event-espresso',
-
-	keywords: [__('event', 'event_espresso'), __('attendees', 'event_espresso'), __('list', 'event_espresso')],
-
-	attributes,
-
-	edit: EventAttendeesEdit,
-
-	save() {
-		return null;
-	},
-
-	// TODO migrate attributes to use GUID
-});
+canUseDOM && domReady(registerEventAttendeesBlock);
--- a/assets/src/domain/blocks/event-attendees/index.tsx
+++ b/assets/src/domain/blocks/event-attendees/index.tsx
@@ -1,27 +1,21 @@
 import { registerBlockType } from '@wordpress/blocks';
 import { __ } from '@wordpress/i18n';
 
-import { canUseDOM, domReady } from '@appServices/utilities/dom';
-
 import attributes from './attributes';
 import EventAttendeesEdit from './edit';
 import type { EventAttendeesAttributes } from './types';
 import './style.scss';
 
-const registerEventAttendeesBlock: VoidFunction = () => {
-	registerBlockType<EventAttendeesAttributes>('eventespresso/event-attendees', {
-		title: __('Event Attendees', 'event_espresso'),
-		description: __('Displays a list of people that have registered for the specified event', 'event_espresso'),
-		icon: 'groups',
-		category: 'event-espresso',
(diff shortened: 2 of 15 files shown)


assets/src/domain/blocks/event-attendees/index.tsx
```diff
@@ -1,29 +1,27 @@
...
+		attributes,
+		edit: EventAttendeesEdit,
+		save() {
+			return null;
+		},
+		// TODO migrate attributes to use GUID
+	});
+};
+
+canUseDOM && domReady(registerEventAttendeesBlock);
```
manzoorwanijk, line 27: I don't think we need to use `domReady` if the script is loaded in correct order.

assets/src/domain/index.ts
```diff
@@ -3,15 +3,12 @@ import './shared/services/publicPath';
...
+	console.log(`%c importing: /domain/${domain}/entryPoint.ts`, 'color: SkyBlue;');
 	import(
 		/* webpackExclude: /(shared)/ */
 		/* webpackChunkName: "[request]" */
 		`./${domain}/entryPoint.ts`
-	)
-		.then(() => {
-			console.log('domain', domain);
-		})
-		.catch(console.error);
```
manzoorwanijk, line 14: Vadim won't be happy with this 😄

core/domain/entities/editor/blocks/EventAttendees.php
```diff
@@ -128,6 +127,9 @@ public function initialize()
     private function getAttributesMap()
     {
         return array(
+            'event'             => 'sanitize_text_field',
+            'datetime'          => 'sanitize_text_field',
+            'ticket'            => 'sanitize_text_field',
             'eventId'           => 'absint',
             'datetimeId'        => 'absint',
             'ticketId'          => 'absint',
```
manzoorwanijk, line 135: These could be removed, or they are to be left there for existing/saved blocks?

core/domain/services/blocks/EventAttendeesBlockRenderer.php
```diff
@@ -53,6 +54,34 @@ public function render(array $attributes)
...
+        $ticket = isset($attributes['ticket']) ? $attributes['ticket'] : '';
+        $datetime = isset($attributes['datetime']) ? $attributes['datetime'] : '';
+        $event = isset($attributes['event']) ? $attributes['event'] : '';
+        if ($ticket !== '') {
+            $ticketId = $this->parseGUID($ticket);
+            $attributes['ticketId'] = $ticketId;
+        } elseif ($datetime !== '') {
+            $datetimeId = $this->parseGUID($datetime);
+            $attributes['datetimeId'] = $datetimeId;
+        } elseif ($event !== '') {
```
manzoorwanijk, line 75: I think it's better to use `if (! empty($ticket))` to take `null` into consideration.

assets/src/application/services/utilities/dom/domReady.ts
```diff
@@ -0,0 +1,17 @@
+/**
+ * TypeScript copy of @wordpress/dom-ready
+ * @see https://github.com/WordPress/gutenberg/blob/master/packages/dom-ready/src/index.js
```
nicolad, line 3: :+1:

assets/src/domain/index.ts
nicolad: just checked and found this answer on stackoverflow of this exact situation and it's possible to use `.catch` w/o `.then`: https://stackoverflow.com/a/36213268
***
not a biggie but the `console.log` from the beginning will be printed even if we'd have errors, but as it was before it was printing only after successful import

core/domain/entities/editor/blocks/EventAttendees.php
```diff
@@ -185,8 +187,10 @@ private function sanitizeAttributes(array $attributes)
...
     {
         $attributes = $this->sanitizeAttributes($attributes);
-        return (is_archive() || is_front_page() || is_home()) && ! $attributes['displayOnArchives']
-            ? ''
-            : $this->renderer->render($attributes);
+        if (! (bool) $attributes['displayOnArchives'] && (is_archive() || is_front_page() || is_home())) {
+            return '';
+        }
+        $this->loadGraphQLRelayAutoloader();
+        return $this->renderer->render($attributes);
```
nicolad, line 194: this is simpler, than the ternary :+1:

manzoorwanijk:
> After debugging that issue with blocks as mentioned by <@U03T6AM1N>, I finally found the issue. The issue here is that the dynamic entry point is loaded lazily, not as main thread as expected by GB. Since, for blocks we are dependent upon GB, it's an issue we need to find a solution for.
<sub>[View in Slack](https://eventespresso.slack.com/archives/C080DDVLM/p1589971290494700)</sub>

manzoorwanijk:
Potential solutions:
- Set [`webpackMode`](https://webpack.js.org/api/module-methods/#magic-comments) to `eager` in dynamic import
    - Pros
        - Dynamic entry point will work as it is
    - Cons
        - Separate chunks won't be created for each domain, which means that JS of all the domains will be in the main file, thus a huge bundle size with unused code

- Create a separate entry point for blocks without dynamic import
    - Pros
        - Bundle size will be minimal
    - Cons
        - Requires a separate assets management

assets/src/domain/blocks/event-attendees/index.tsx
tn3rb: ya that was only added as an attempt to fix that issue, more than happy to remove it

assets/src/domain/index.ts
tn3rb: ya I changed it because when trying to work on this issue, I wasn't getting anything printed at first, because it wasn't resolving successfully but there also wasn't an error

I'm fine with this being changed back to something else, but I DO agree with that SO comment that a `.then()` clause is not needed

core/domain/services/blocks/EventAttendeesBlockRenderer.php
tn3rb: that was already handled by the ternaries that are before this block of code.

we first resolved the variables from the array like so:

```php
        $ticket = isset($attributes['ticket']) ? $attributes['ticket'] : '';
        $datetime = isset($attributes['datetime']) ? $attributes['datetime'] : '';
        $event = isset($attributes['event']) ? $attributes['event'] : '';
```

so with php `isset($attributes['ticket'])` will return `false` if that value is `null`
(because `null` is the default value for a variable that has not been set)
which would then set the value of `$attributes['ticket']` to `''`
which can then be quickly and directly compared against using `if ($ticket !== '') {`

tn3rb:
> Set webpackMode to eager in dynamic import

nope nope nope

> Add script to head instead of footer in wp_enqueue_script

an additional con is that this increases initial page load time (ie: [time to first byte](https://en.wikipedia.org/wiki/Time_to_first_byte))

> Create a separate entry point for blocks without dynamic import
> Pros
> Bundle size will be minimal
> Cons
> Requires a separate assets management

well we already have separate assets management in place as is, or that was the path we were previously going down anyways, so changing from  separate asset management for ALL routes to only having to manage two is a big improvement imho

I think separate entry points is the only viable solution at least if we want to maintain some dynamic importing

PR #2857, opened 2020-05-20
TPC inputs styles fixes.

nicolad:

image
***
image
***

## How has this been tested
image
***
image

### DIFF
--- a/assets/src/domain/eventEditor/ui/tickets/ticketPriceCalculator/fields/BaseField.tsx
+++ b/assets/src/domain/eventEditor/ui/tickets/ticketPriceCalculator/fields/BaseField.tsx
@@ -18,7 +18,11 @@ const BaseField: React.FC<BaseFieldProps> = ({
 	value,
 	...props
 }) => {
-	const className = classNames('ee-input-base ee-input', { 'ee-select': component === 'select' }, props.className);
+	const className = classNames(
+		props.className,
+		'ee-input-base ee-input',
+		component === 'select' && 'ee-select',
+	);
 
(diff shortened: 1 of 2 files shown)


assets/src/domain/eventEditor/ui/tickets/ticketPriceCalculator/fields/BaseField.tsx
```diff
@@ -15,8 +16,10 @@ const BaseField: React.FC<BaseFieldProps> = ({
 	getValue,
 	setValue,
 	value,
-	...rest
+	...props
 }) => {
+	const className = classNames('ee-input-base ee-input', { 'ee-select': component === 'select' }, props.className);
```
tn3rb, line 21: what do you guys think about this approach:

```suggestion
	const className = classNames(
		props.className,
		'ee-input-base ee-input',
		component === 'select' && 'ee-select',
	);
```

- put `props.className` first because custom classes would be the most specific

- instead of using the classNames `key => value` / `className => conditional` pairing inside an object, use `conditional && className` format because the className won't get added if condition is not true, and conditional format is more typical of regular JS

can't remember which one of you guys did this first, but I saw it and liked it alot, because it appears more clean to me

nicolad: NOICEEE, thank you for deciding this, had some doubts sometimes when I had to add `props.className`, wasn't sure if it should be at the beginning or the end, agree with your suggestion
***
I am happy with that short circuit evaluation approach :+1:

PR #2862, opened 2020-05-21
Added more useCallbacks and updated exhaustive-deps.

nicolad:
- Title says it all.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -32,7 +32,7 @@
 		"jest/valid-describe": "off",
 		"no-case-declarations": "warn",
 		"react-hooks/rules-of-hooks": "error",
-		"react-hooks/exhaustive-deps": "off",
+		"react-hooks/exhaustive-deps": "error",
 		"react/prop-types": "off",
 		"react/static-property-placement": "off",
 		"react/display-name": "off",
--- a/assets/src/application/services/hooks/useMemoStringify.ts
+++ b/assets/src/application/services/hooks/useMemoStringify.ts
@@ -1,11 +1,11 @@
 import { useMemo } from 'react';
-import { is } from 'ramda';
+import { is, toString } from 'ramda';
 
 const useMemoStringify = (data: any, deps?: any[]): any => {
 	let dataStringified: string;
 
 	if (deps && Array.isArray(deps)) {
-		dataStringified = deps.join(':');
+		dataStringified = deps.map(toString).join(':');
 	}
 
 	if (is(Object, data)) {
--- a/assets/src/application/services/hooks/useMemoStringify.ts
+++ b/assets/src/application/services/hooks/useMemoStringify.ts
@@ -1,0 +1,19 @@
+import { useMemo } from 'react';
+import { is } from 'ramda';
+
+const useMemoStringify = (data: any, deps?: any[]): any => {
+	let dataStringified: string;
+
+	if (deps && Array.isArray(deps)) {
+		dataStringified = deps.join(':');
+	}
+
+	if (is(Object, data)) {
+		dataStringified = JSON.stringify(data);
+	}
+
+	// eslint-disable-next-line react-hooks/exhaustive-deps
+	return useMemo(() => data, [dataStringified]);
+};
+
+export default useMemoStringify;
--- a/assets/src/domain/eventEditor/hooks/filterBar/useFilterBarUISubscription.ts
+++ b/assets/src/domain/eventEditor/hooks/filterBar/useFilterBarUISubscription.ts
@@ -18,6 +18,8 @@ const useFilterBarUISubscription = (): void => {
 			unsubscribeDatesListFilterBar();
 			unsubscribeTicketsListFilterBar();
 		};
+
+		// eslint-disable-next-line react-hooks/exhaustive-deps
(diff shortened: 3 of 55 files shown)


assets/src/domain/eventEditor/hooks/filterBar/useFilterBarUISubscription.ts
```diff
@@ -18,6 +18,8 @@ const useFilterBarUISubscription = (): void => {
 			unsubscribeDatesListFilterBar();
 			unsubscribeTicketsListFilterBar();
 		};
+
+		// eslint-disable-next-line react-hooks/exhaustive-deps
```
nicolad, line 22: maybe I should replace this with `useIfMounted`?

nicolad:
I've added `client` in few `useCallback`s just to please `react-hooks/exhaustive-deps`

assets/src/application/services/hooks/useMemoStringify.ts
```diff
@@ -0,0 +1,19 @@
+import { useMemo } from 'react';
+import { is } from 'ramda';
+
+const useMemoStringify = (data: any, deps?: any[]): any => {
```
nicolad, line 4: I tried to pass the second argument as an object with `deps` property, but still think that is't maybe better to split the logic here into 2 hooks

nicolad: also those `any` types also hinders me a bit;
I have mixed feelings about this hook

nicolad: it doesn't solve a lot whole of stuff, but it makes the end components a bit cleaner w/o that `stringify` or `join` logic

.eslintrc.json
```diff
@@ -32,7 +32,7 @@
 		"jest/valid-describe": "off",
 		"no-case-declarations": "warn",
 		"react-hooks/rules-of-hooks": "error",
-		"react-hooks/exhaustive-deps": "off",
+		"react-hooks/exhaustive-deps": "error",
```
nicolad, line 35: I think we should have this set as `error`, this will serve as safe guard

nicolad: as @manzoorwanijk have noticed few weeks ago we had few `useCallback` deps array empty which causes some issues

nicolad: and I think is important to keep this rule as error since we will start memoizing more aggresively

manzoorwanijk: Nice change 👍

assets/src/application/services/hooks/useMemoStringify.ts
```diff
@@ -0,0 +1,19 @@
...
+
+	if (is(Object, data)) {
+		dataStringified = JSON.stringify(data);
+	}
+
+	// eslint-disable-next-line react-hooks/exhaustive-deps
+	return useMemo(() => data, [dataStringified]);
+};
+
+export default useMemoStringify;
```
manzoorwanijk, line 19: I am sorry but I don't understand the idea behind `deps`. I think when the data is being stringified, that should be enough for dependency check.

nicolad: Initially, I've created this hook to handle cases like these:

image

nicolad: that was needed for 2 reasons:
- to hide that `.join(':')` that was repeated multiple times
- to avoid using `// eslint-disable-next-line react-hooks/exhaustive-deps` every time a dependency is stringified.

nicolad: and after dealing with those arrays as dependency, I noticed that we have cases when object is stringified, like here:

image

***
and decided to abstract this case also in `useMemoStringify`, by checking first if that is object and then applying `JSON.stringify`

manzoorwanijk: Thank you for the explanation 👍 
I have one last concern regarding this
```js
if (deps && Array.isArray(deps)) {
	dataStringified = deps.join(':');
}
```
it's also possible that `deps` may not always be an array of `strings`, it can be an array of objects. So it may be better to do this to make sure we are joining strings:
```js
import { toString } from 'ramda';

if (deps && Array.isArray(deps)) {
	dataStringified = deps.map(toString).join(':');
}
```

assets/src/domain/eventEditor/services/apollo/mutations/datetimes/useDatetimeMutator.ts
```diff
@@ -1,3 +1,4 @@
+// import { useMemo } from 'react';
```
manzoorwanijk, line 1: May be remove this?

assets/src/application/services/hooks/useMemoStringify.ts
nicolad: thx, updated

assets/src/domain/eventEditor/services/apollo/mutations/datetimes/useDatetimeMutator.ts
nicolad: yes, cleaned that up

PR #2866, opened 2020-05-22
Alert & banner components update.

nicolad:
- Replaces existing `Alert` with `Banner` and made it easy to extend
- Adds `Alert` component, based on: https://chakra-ui.com/alert#status

### DIFF
--- a/assets/src/infrastructure/ui/display/Banner.tsx
+++ b/assets/src/infrastructure/ui/display/Banner.tsx
@@ -1,0 +1,21 @@
+import React from 'react';
+import { Alert as ChakraAlert, AlertIcon, AlertTitle, AlertDescription, AlertProps, IconProps } from '@chakra-ui/core';
+
+interface Props extends AlertProps {
+	description: string;
+	iconProps?: IconProps;
+	title: string;
+}
+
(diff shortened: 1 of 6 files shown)


assets/src/infrastructure/ui/display/Banner.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+interface Props extends AlertProps {
+	description: string;
+	iconProps?: IconProps;
+	title: string;
+}
+
+const Banner: React.FC<Props> = ({ children, description, iconProps, title, ...props }) => (
+	<ChakraAlert flexDirection='column' justifyContent='center' textAlign='center' height='200px' {...props}>
+		<AlertIcon size='40px' mr={0} {...iconProps} />
```
tn3rb, line 12: only suggestion is maybe to use rem instead of px as it supposedly scales better

```suggestion
		<AlertIcon size='2.5rem' mr={0} {...iconProps} />
```

2.5rem should equal 40px at the standard font scaling

nicolad: good one, thx :bow:

PR #2870, opened 2020-05-25
Plinko Routing

tn3rb:
On the JS application side of things we recently switched to utilizing a single entry point with dynamic routing to the appropriate domain (use case) instead of having to add a new entry point for each individual domain. This means that the PHP side of things can be greatly simplified because we will no longer have to add asset managers for every domain, but instead just have to include a bit of extra domain/routing information in the DOM data dump, which we would have to do anyways.

Internal routing within EE on the server side has not been centralized or handled by any specific class and has instead been managed primarily through hooks and/or EE_Admin, and for the most part has done things inline with the typical "WordPress Way", ie: load EVERYTHING and let the hook callbacks figure it all out for themselves.

This won't necessarily work well for the dynamic routing we are planning JS side because we need to ensure that only a single domain (use case) is loaded and has its data dumped to the DOM. Because of this, a better more focused approach to routing is needed.

This PR adds some classes for defining, collecting, and managing "routes" which essentially represent a particular point in the EE core code execution path. Unlike routing in MVC applications where a URL and/or request parameter(s) can be parsed in order to determine the correct controller to load, which typically have a 1:1 relationship between a route and a controller.

I've done things a little bit differently and allowed routes to define any particular location along the full path that code may take while executing. This is because EE handles many different kinds of requests (HTTP, REST API, AJAX, iFrame, GQL, etc) but has many aspects of our boilerplate logic that is shared between two or more of those request types. So for an example, ALL admin requests will share a certain small amount of boilerplate code, but AJAX and REST API requests don't require javascript assets (although currently those kinds of assets may very well be getting loaded).

In this branch, a Route can be defined that detects any base admin request and can then load JUST the boilerplate that is common for ALL admin requests. This may include the loading of additional routes for further request detection. Then another Route can detect whether the current request is for AJAX and execute any additional logic required at that point to help handle the AJAX request. So during the course of a request, several Routes may execute so as to only load the logic required from that point onwards during the request.

In the following example, Routes could be loaded at any one of the following "nodes":
routing
but a Route loaded at point 2 would only execute logic that is common for points 5 and 6, and point 5 might load another Route that only executes logic common for points 9 and 10. This allows code to be extracted into granular chunks and avoids the kind of duplication that can occur when you try to connect top level nodes directly with the target end node. (ie: jumping directly from point 1 to 9 or point 1 to 10, requires 9 and 10 to each handle the logic loaded at points 2 and 5). I call this Plinko Routing because the requests zig zag their way from their starting point at the top to their destination. Routes are therefore only responsible for directing traffic from their point onwards.

Routes currently only do three primary things:

- detect whether they match the current request (this can be done by directly examining the incoming Request or via a Route Match Specification class for better SRP)

- register dependencies for any classes that are required from that point forwards in the request

- load additional classes for handling the request


tn3rb:
Final changes to get Routes working, sry for it still being a pretty big PR, but at least's it's not 3800 LOC changes anymore 😆

core/admin/EE_Admin_Page_CPT.core.php
```diff
@@ -944,76 +932,6 @@ public function load_global_scripts_styles()
...
-     * @return void
-     * @throws DomainException
-     */
-    public function enqueueLegacyEditorAssets()
-    {
-        $this->asset_manager->enqueueBrowserAssets();
-        wp_enqueue_script('ee-accounting');
-    }
-
-
```
nicolad, line 1016: have no idea what was the purpose of this piece, but seeing code being removed is pleasing, it might mean that some of the obsolete or overengineered

core/domain/entities/routing/handlers/shared/RoutingRequests.php
```diff
@@ -0,0 +1,88 @@
...
+
+    /**
+     * returns true if the current request matches this route
+     *
+     * @return bool
+     * @since   $VID:$
+     */
+    public function matchesCurrentRequest()
+    {
+        return ! $this->request->isActivation() || $this->request->isUnitTesting();
```
manzoorwanijk, line 27: 👍

core/domain/services/assets/EspressoLegacyAdminAssetManager.php
```diff
@@ -3,18 +3,16 @@
...
- * Class EspressoAdminAssetManager
- * Description
+ * Class EspressoLegacyAdminAssetManager
  *
  * @package EventEspresso\core\domain\services\assets
  * @author  Brent Christensen
  * @since   $VID:$
  */
-class EspressoAdminAssetManager extends AssetManager
+class EspressoLegacyAdminAssetManager extends AssetManager
```
manzoorwanijk, line 15: Yeah, that's legacy now 😎

core/libraries/form_sections/strategies/display/EE_Admin_File_Uploader_Display_Strategy.strategy.php
```diff
@@ -76,8 +102,7 @@ protected function src_exists($src)
         $results = wp_remote_head($src);
         if (is_array($results) && ! $results instanceof WP_Error) {
             return strpos($results['headers']['content-type'], "image") !== false;
-        } else {
-            return false;
         }
+        return false;
```
manzoorwanijk, line 106: 👍

tests/mocks/core/domain/entities/routing/handlers/shared/GQLRequestsMock.php
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    public function registerDependencies() {
+        parent::registerDependencies();
+    }
+
+    public function requestHandler() {
+        parent::requestHandler();
+    }
+}
```
manzoorwanijk, line 41: Needs EOL

tests/testcases/core/services/dependencies/DependencyResolverTest.php
```diff
@@ -33,27 +38,29 @@ class DependencyResolverTest extends EspressoPHPUnitFrameworkTestCase
...
      */
     protected function setUp()
     {
         $this->loader = LoaderFactory::getLoader();
-        $this->request_params = array(
-            'oof' => 'OOF!',
+        $this->request_params = [
+            'oof'  => 'OOF!',
             'ouch' => 'OUCH!',
             'owie' => 'OWIE!',
```
manzoorwanijk, line 56: Nice expressions 😄

PR #2876, opened 2020-05-26
Route && RouteInterface

tn3rb:
see: https://github.com/eventespresso/event-espresso-core/pull/2870

the PR adds Route class && RouteInterface

 - class for detecting and matching with incoming requests
(this can be done by directly examining the incoming Request
or via a Route Match Specification class for better SRP and sharing)

 - registers dependencies for any classes that are required from that point forwards in the request

 - loads additional classes for handling the request

### DIFF
--- a/core/domain/entities/routing/handlers/Route.php
+++ b/core/domain/entities/routing/handlers/Route.php
@@ -1,0 +1,155 @@
+<?php
+
+namespace EventEspresso\core\domain\entities\routing\handlers;
+
+use DomainException;
+use EE_Dependency_Map;
+use EventEspresso\core\domain\entities\routing\specifications\RouteMatchSpecificationInterface;
+use EventEspresso\core\services\loaders\LoaderInterface;
+use EventEspresso\core\services\request\RequestInterface;
+
+/**
+ * Class Route
+ * - class for detecting and matching with incoming requests
+ * (this can be done by directly examining the incoming Request
+ * or via a Route Match Specification class for better SRP and sharing)
+ * - registers dependencies for any classes that are required from that point forwards in the request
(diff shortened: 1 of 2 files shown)


core/domain/entities/routing/handlers/Route.php
```diff
@@ -0,0 +1,155 @@
...
+     *
+     * @return bool
+     * @since   $VID:$
+     */
+    public function matchesCurrentRequest()
+    {
+        return $this->specification instanceof RouteMatchSpecificationInterface
+            ? $this->specification->isMatchingRoute()
+            : false;
+    }
```
manzoorwanijk, line 106: We can't use null coalescing operator yet... :(

core/domain/entities/routing/handlers/Route.php
```diff
@@ -0,0 +1,155 @@
...
+ * @author  Brent Christensen
+ * @since   $VID:$
+ */
+abstract class Route implements RouteInterface
+{
+
+    // /** will most likely need this for routing so pencilled it in for now
+    //  * @type CommandBusInterface $command_bus
+    //  */
+    // protected $command_bus;
```
manzoorwanijk, line 29: May be remove this?

core/domain/entities/routing/handlers/Route.php
tn3rb: not until WP and/or EE requires minimum PHP version of 7

core/domain/entities/routing/handlers/Route.php
tn3rb: i'm like 99.99% certain I will be implementing that... so I could remove it... but will most definitely be adding it back in

PR #2877, opened 2020-05-26
Admin Routes

tn3rb:
see: https://github.com/eventespresso/event-espresso-core/pull/2870

adds the following admin Route classes:

 - EspressoEventEditor

 - EspressoLegacyAdmin

 - PersonalDataRequests

 - PueRequests

 - WordPressPluginsPage


core/domain/entities/routing/handlers/admin/WordPressPluginsPage.php
```diff
@@ -0,0 +1,72 @@
...
+    /**
+     * returns true if the current request matches this route
+     *
+     * @return bool
+     * @since   $VID:$
+     */
+    public function matchesCurrentRequest()
+    {
+        return $this->request->isAdmin()
+               && strpos($this->request->requestUri(), 'wp-admin/plugins.php') !== false;
```
manzoorwanijk, line 29: Is this always guaranteed? What about the folks who rename `wp-admin` to something else to protect it from being hammered?

manzoorwanijk: I think a better way may be to use [WP Screen](http://codex.wordpress.org/Class_Reference/WP_Screen) API

tn3rb: oh great point.. ya will fix

core/domain/entities/routing/handlers/admin/EspressoEventEditor.php
```diff
@@ -0,0 +1,133 @@
...
+            [
+                'EEM_Price'      => EE_Dependency_Map::load_from_cache,
+                'EEM_Price_Type' => EE_Dependency_Map::load_from_cache,
+            ]
+        );
+    }
+
+
+    /**
+     * implements logic required to run during request
```
nicolad, line 124: it's a middleware, kinda?

tn3rb: mmm... no it's not middleware it's more like undefined logic that's been spread out all over the place that's been extracted and moved into a formally defined interface that can be better managed by the system.

Previously the request detection and subsequent handling of said requests were being done directly inside `EE_System` (and other places) with a ton of other logic all mixed up together. Dependencies for any classes being loaded were just ALL hardcoded directly in `EE_Dependency_Map` regardless of whether or not they were actually needed for the current request.

So now, that logic has been moved into these `Route` classes, which currently only have the following three methods for their public interface:

- `matchesCurrentRequest()` returns true if the current request matches this route

- `handleRequest()` logic to run when above method returns `true`

- `isHandled()` true if `handleRequest()` has already run

and the new `RouteHandler` class can execute that logic for any incoming request that has been added to its internal collection which is done simply by providing the FQCN to the Route class:

```php
 $this->router->addRoute('EventEspresso\core\domain\entities\routing\handlers\admin\EspressoEventEditor');
```

So all of the previous logic inside that Route class which was previously in `EE_System` has been replaced by that one line of code in `EE_System`, and the dependencies that previously were ALWAYS loaded in `EE_Dependency_Map` are now only registered if they are needed for the current request.

core/domain/entities/routing/handlers/admin/WordPressPluginsPage.php
tn3rb: couldn't use `WP_Screen` because that is created too late, but I still fixed the issue

manzoorwanijk: But here it's still the same, may be you forgot to make changes to this file?

tn3rb: DOH... i fixed the route match specification but not this class..

core/domain/entities/routing/handlers/admin/WordPressPluginsPage.php
```diff
@@ -0,0 +1,72 @@
...
+     * returns true if the current request matches this route
+     *
+     * @return bool
+     * @since   $VID:$
+     */
+    public function matchesCurrentRequest()
+    {
+        global $pagenow;
+        return $this->request->isAdmin() && $pagenow && $pagenow === 'plugins.php';
+    }
```
manzoorwanijk, line 30: Cool 👍

core/domain/entities/routing/handlers/admin/EspressoEventEditor.php
nicolad: thank you @tn3rb for such an elaborate explanation :bow: 

it's interesting to understand some of the underlying server-side logic, not just the client-side of things, I never know from where I can get some inspiration for some patterns

Issue #2874, opened 2020-05-26
ButtonGroup improvements


nicolad:
- Based on this [comment](https://github.com/eventespresso/event-espresso-core/pull/2869#discussion_r430630359) from the previous PR there some improvements that needs to be done
- Also currently ButtonGroup is handling cases when there are 2 items, but it isn't designed to handle more than that.

nicolad:
material design:

![image](https://user-images.githubusercontent.com/1477453/83034907-71283400-a041-11ea-8fd3-ffa4962694ee.png)

nicolad:
material design:

![image](https://user-images.githubusercontent.com/1477453/83035035-a16fd280-a041-11ea-9a0a-680745c13325.png)

nicolad:
KendoReact:

![image](https://user-images.githubusercontent.com/1477453/83035106-b9475680-a041-11ea-92f2-3dbf9bed4721.png)

nicolad:
semantic ui:

![image](https://user-images.githubusercontent.com/1477453/83035155-c8c69f80-a041-11ea-8e57-2a5885e632d0.png)

nicolad:
Links:
- https://react-bootstrap.github.io/components/button-group/
- https://material-ui.com/components/button-group/
- https://www.telerik.com/kendo-react-ui/components/buttons/buttongroup/
- https://react.semantic-ui.com/elements/button/

PR #2869, opened 2020-05-25
Filter button and other buttons update.

nicolad:
## Problem this Pull Request solves
- Closes: https://github.com/eventespresso/event-espresso-core/issues/2865 and https://github.com/eventespresso/event-espresso-core/issues/2867

PR #2880, opened 2020-05-27
ButtonGroup improvs.

nicolad:
- Makes it possible for `ButtonGroup` to wrap more than 2 buttons
- Receives button size at the parent level, w/o the need to pass to each items individually
- Closes https://github.com/eventespresso/event-espresso-core/issues/2874

## How has this been tested
image

### DIFF
--- a/assets/src/application/ui/input/Button/ButtonGroup/index.tsx
+++ b/assets/src/application/ui/input/Button/ButtonGroup/index.tsx
@@ -1,16 +1,36 @@
-import React from 'react';
+import React, { cloneElement } from 'react';
 import classNames from 'classnames';
 
-import { ButtonGroup as ButtonGroupAdapter, ButtonGroupProps } from '@infraUI/inputs';
+import { ButtonGroup as ButtonGroupAdapter } from '@infraUI/inputs';
+import { ButtonGroupProps } from './types';
+import { cleanChildren } from './utils';
 
 import './style.scss';
 
-const ButtonGroup: React.FC<ButtonGroupProps> = ({ children, ...props }) => {
+const ButtonGroup: React.FC<ButtonGroupProps> = ({ buttonSize, children, ...props }) => {
 	const className = classNames(props.className, 'ee-btn-group');
+	const validChildren = cleanChildren(children);
+	const clones = validChildren.map((child, index) => {
+		const isFirst = index === 0;
+		const isLast = index === validChildren.length - 1;
+
+		// @ts-ignore
+		return cloneElement(child, {
+			// @ts-ignore
+			buttonSize: buttonSize || child.props.buttonSize,
+
+			// @ts-ignore
+			...(isFirst && { className: classNames(child.props.className, 'ee-btn--first') }),
+			// @ts-ignore
+			...(isLast && { className: classNames(child.props.className, 'ee-btn--last') }),
+			// @ts-ignore
+			...(!isFirst && !isLast && { className: classNames(child.props.className, 'ee-btn--no-border-radius') }),
+		});
+	});
 
 	return (
 		<ButtonGroupAdapter {...props} className={className}>
-			{children}
+			{clones}
 		</ButtonGroupAdapter>
 	);
 };
--- a/assets/src/application/ui/input/Button/ButtonGroup/index.tsx
+++ b/assets/src/application/ui/input/Button/ButtonGroup/index.tsx
@@ -18,13 +18,6 @@ const ButtonGroup: React.FC<ButtonGroupProps> = ({ buttonSize, children, ...prop
 		return cloneElement(child, {
 			// @ts-ignore
 			buttonSize: buttonSize || child.props.buttonSize,
-
-			// @ts-ignore
-			...(isFirst && { className: classNames(child.props.className, 'ee-btn--first') }),
-			// @ts-ignore
-			...(isLast && { className: classNames(child.props.className, 'ee-btn--last') }),
-			// @ts-ignore
-			...(!isFirst && !isLast && { className: classNames(child.props.className, 'ee-btn--no-border-radius') }),
 		});
 	});
 
--- a/assets/src/application/ui/input/Button/ButtonGroup/index.tsx
+++ b/assets/src/application/ui/input/Button/ButtonGroup/index.tsx
@@ -10,13 +10,8 @@ import './style.scss';
 const ButtonGroup: React.FC<ButtonGroupProps> = ({ buttonSize, children, ...props }) => {
 	const className = classNames(props.className, 'ee-btn-group');
 	const validChildren = cleanChildren(children);
-	const clones = validChildren.map((child, index) => {
-		const isFirst = index === 0;
-		const isLast = index === validChildren.length - 1;
-
-		// @ts-ignore
+	const clones = validChildren.map((child: any) => {
 		return cloneElement(child, {
-			// @ts-ignore
 			buttonSize: buttonSize || child.props.buttonSize,
 		});
 	});
--- a/assets/src/application/ui/input/Button/ButtonGroup/style.scss
+++ b/assets/src/application/ui/input/Button/ButtonGroup/style.scss
@@ -1,20 +1,21 @@
 .ee-btn-group {
-	.ee-btn {
-		margin-right: 0;
+	margin-right: var(--ee-margin-default);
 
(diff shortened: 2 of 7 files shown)


nicolad:
After https://github.com/eventespresso/event-espresso-core/pull/2869 will be merged, the base branch will be changed to `EDTR/master`

assets/src/application/ui/input/Button/ButtonGroup/index.tsx
```diff
@@ -1,16 +1,36 @@
...
+			buttonSize: buttonSize || child.props.buttonSize,
+
+			// @ts-ignore
+			...(isFirst && { className: classNames(child.props.className, 'ee-btn--first') }),
+			// @ts-ignore
+			...(isLast && { className: classNames(child.props.className, 'ee-btn--last') }),
+			// @ts-ignore
+			...(!isFirst && !isLast && { className: classNames(child.props.className, 'ee-btn--no-border-radius') }),
+		});
+	});
```
nicolad, line 29: this approach was borrowed from here: https://github.com/chakra-ui/chakra-ui/blob/master/packages/chakra-ui/src/ButtonGroup/index.js

assets/src/application/ui/input/Button/ButtonGroup/style.scss
```diff
@@ -1,20 +1,21 @@
 .ee-btn-group {
-	.ee-btn {
-		margin-right: 0;
+	margin-right: var(--ee-margin-default);
 
-		&:first-child {
+	.ee-btn.ee-btn {
```
nicolad, line 4: specificity war

assets/src/application/ui/layout/entityList/filterBar/buttons/CardViewFilterButton.tsx
```diff
@@ -13,7 +13,6 @@ const CardViewFilterButton: React.FC<CardViewFilterButtonProps> = ({ listId, set
 	return (
 		<Button
 			active={view === 'card'}
-			buttonSize={ButtonSize.SMALL}
```
nicolad, line 16: less boilerplate

assets/src/application/ui/input/Button/ButtonGroup/utils.ts
```diff
@@ -0,0 +1,5 @@
+import { Children, isValidElement } from 'react';
+
+export function cleanChildren(children) {
+	return Children.toArray(children).filter((child) => isValidElement(child));
+}
```
nicolad, line 5: this utility is very generic and I wasn't sure where to put it

nicolad: @tn3rb your suggestion on where this file should be is needed here

assets/src/application/ui/input/Button/ButtonGroup/style.scss
```diff
@@ -1,20 +1,21 @@
...
+		margin: 0;
+
+		&--no-border-radius {
+			border-radius: 0;
+		}
```
tn3rb, line 9: no need for an explicit class

```suggestion
		&:not(:first-child),
		&:not(:last-child) {
			border-radius: 0;
		}
```

assets/src/application/ui/input/Button/ButtonGroup/style.scss
```diff
@@ -1,20 +1,21 @@
...
+
+		&--first {
```
tn3rb, line 11:
```suggestion
		&:first-child {
```

assets/src/application/ui/input/Button/ButtonGroup/style.scss
```diff
@@ -1,20 +1,21 @@
...
 			border-top-right-radius: 0;
 			border-bottom-right-radius: 0;
 		}
 
-		&:last-child {
+		&--last {
```
tn3rb, line 16:
```suggestion
		&:last-child {
```

assets/src/application/ui/input/Button/ButtonGroup/index.tsx
tn3rb, line 29: cloning is completely unnecessary imho, this can all be achieved using css pseudo selectors

assets/src/application/ui/input/Button/ButtonGroup/utils.ts
tn3rb, line 5: handy little function...

but does this work?

```suggestion
	return Children.toArray(children).filter(isValidElement);
```

assets/src/application/ui/layout/entityList/filterBar/buttons/EntityListViewButtonGroup.tsx
```diff
@@ -15,7 +15,7 @@ const EntityListViewButtonGroup: React.FC<EntityListViewButtonGroupProps> = ({
 	view,
 }) => {
 	return (
-		<ButtonGroup>
+		<ButtonGroup buttonSize={ButtonSize.SMALL}>
```
tn3rb, line 18: oh i see... you're passing additional props through to the children...

ya ok so cloning is the way to go... I thought initially it was just being done to add that `--no-border-radius` modifier...

did you see the discussion Manzoor and I had yesterday about different methods for adding review feedback?
here's a perfect example of where I didn't realize why something was done a particular way until after already providing feedback.
normally I would go back and delete those irrelevant comments, but I'll let them stand this time just for fun 😆

assets/src/application/ui/input/Button/ButtonGroup/style.scss
nicolad: great suggestion, applied, thx

nicolad: that was applying the property to all buttons:

image

nicolad: applied a slightly different selector: 

image

assets/src/application/ui/input/Button/ButtonGroup/style.scss
nicolad: done, thx

assets/src/application/ui/input/Button/ButtonGroup/style.scss
nicolad: done, thx

assets/src/application/ui/input/Button/ButtonGroup/utils.ts
nicolad: updated based on suggestion

nicolad: this is copied from chakra: https://github.com/chakra-ui/chakra-ui/blob/master/packages/chakra-ui/src/utils/index.js#L184

nicolad: > but does this work?

what exactly seems tricky in this function?

assets/src/application/ui/layout/entityList/filterBar/buttons/EntityListViewButtonGroup.tsx
nicolad: yes, let it leave here for fun
***
`cloneElement` this is an interesting pattern, the only thing I don't like is that it implies a bit of magic, because prop is passed less explicit

assets/src/application/ui/input/Button/ButtonGroup/utils.ts
tn3rb: i didn't know if there was a reason you had chosen to do `filter((child) => isValidElement(child))` over `filter(isValidElement)`

so i wasn't 100% sure if there was something i was missing for why that latter suggestion wouldn't work

nicolad: I had a case in the past when it was important to keep this way:
```
filter((child) => isValidElement(child))
```
because when passed implicitly it will pass also the second argument of the filter which is index and if that function accepts the second argument, than it might mess things up

nicolad: checked `isValidElement` and it receives only 1 argument so we are good:

https://github.com/facebook/react/blob/master/packages/react/src/jsx/ReactJSXElementValidator.js#L66

PR #2888, opened 2020-05-28
Added ticket price amount labels.

nicolad:
- Adds background to ticket price amount labels


nicolad:
Next thing to do:
- improve error state, like smth:
image

assets/src/domain/eventEditor/ui/tickets/ticketPriceCalculator/fields/MoneyField/updateInputsSizes.tsx
```diff
@@ -0,0 +1,39 @@
...
+		if (currency) {
+			const after = currency.getElementsByClassName('ee-money-field__after')[0]?.getBoundingClientRect();
+			const before = currency.getElementsByClassName('ee-money-field__before')[0]?.getBoundingClientRect();
+			const input = currency.getElementsByClassName('ee-input__price-field')[0]?.getBoundingClientRect();
+			const labelWidth = after?.width ?? before?.width;
+
+			updatedPercentSignInput({ beforeWidth: labelWidth, inputWidth: input?.width, signB4 });
+		}
+	}
+};
```
nicolad, line 37: all this mumbo-jumbo is hacky thing, before doing this I had considered:
- using `findDOMNode` from `ReactDOM`, but `It has been deprecated in StrictMode.`
- using refs, but it would create more mess

assets/src/domain/eventEditor/ui/tickets/ticketPriceCalculator/fields/MoneyField/updateInputsSizes.tsx
```diff
@@ -0,0 +1,39 @@
...
+const updatedPercentSignInput = ({ beforeWidth, inputWidth, signB4 }: updatedPercentSignInputProps): void => {
+	document
+		.querySelectorAll('.ee-money-field--with-percent-sign .ee-input__price-field')
+		.forEach((element: HTMLElement) => {
+			if (signB4) {
+				element.style.marginLeft = `${beforeWidth}px`;
+			}
+
+			element.style.width = `${inputWidth}px`;
+		});
```
tn3rb, line 16: whoah... why are you adding styles like this?
why not within the component when it renders?

assets/src/domain/eventEditor/ui/tickets/ticketPriceCalculator/fields/MoneyField/updateInputsSizes.tsx
```diff
@@ -0,0 +1,39 @@
...
+};
+
+interface updateInputsSizesProps {
+	isPercent: boolean;
+	signB4: boolean;
+}
+
+const updateInputsSizes = ({ isPercent, signB4 }: updateInputsSizesProps): void => {
```
manzoorwanijk, line 24: May be just name it `updateInputSizes` without `s` after `Input`

assets/src/domain/eventEditor/ui/tickets/ticketPriceCalculator/fields/MoneyField/updateInputsSizes.tsx
```diff
@@ -0,0 +1,39 @@
...
+	document
+		.querySelectorAll('.ee-money-field--with-percent-sign .ee-input__price-field')
+		.forEach((element: HTMLElement) => {
+			if (signB4) {
+				element.style.marginLeft = `${beforeWidth}px`;
+			}
+
+			element.style.width = `${inputWidth}px`;
+		});
+};
```
manzoorwanijk, line 17: No sure if I understand the need for accessing DOM here.

assets/src/domain/eventEditor/ui/tickets/ticketPriceCalculator/fields/MoneyField/updateInputsSizes.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+const updateInputsSizes = ({ isPercent, signB4 }: updateInputsSizesProps): void => {
+	if (isPercent) {
+		const currency = document.getElementsByClassName('ee-money-field--with-currency-sign')[0];
+
+		if (currency) {
+			const after = currency.getElementsByClassName('ee-money-field__after')[0]?.getBoundingClientRect();
+			const before = currency.getElementsByClassName('ee-money-field__before')[0]?.getBoundingClientRect();
+			const input = currency.getElementsByClassName('ee-input__price-field')[0]?.getBoundingClientRect();
+			const labelWidth = after?.width ?? before?.width;
```
tn3rb, line 32: this can likely be accomplished using css `calc()` no?

ex:

```css
calc( var(--ee-mony-field-input-width) - var(--ee-mony-field-label-3));
```

that's not quite what you are doing here, but still seems unnecessary to use the DOM like this

assets/src/domain/eventEditor/ui/tickets/ticketPriceCalculator/fields/MoneyField/updateInputsSizes.tsx
nicolad: done, thx

assets/src/domain/eventEditor/ui/tickets/ticketPriceCalculator/fields/MoneyField/updateInputsSizes.tsx
nicolad: the end result was to avoid these zig zags:

image

assets/src/domain/eventEditor/ui/tickets/ticketPriceCalculator/fields/MoneyField/updateInputsSizes.tsx
nicolad: the data flow is a bit tricky because first it should check sizes from a different row and then pass back down to input levels the sizes

assets/src/domain/eventEditor/ui/tickets/ticketPriceCalculator/fields/MoneyField/updateInputsSizes.tsx
nicolad: it was dynamically checking the size and then applying that based on results, hardcoding isn't quite an option because label can have different length

nicolad:
image

nicolad:
Since I had to refactor a bit some of the related fields, the zig zag issue will be fixed separately.

assets/src/domain/eventEditor/ui/tickets/ticketPriceCalculator/fields/MoneyField/updateInputsSizes.tsx
nicolad: have removed the DOM manipulations

assets/src/application/ui/input/InputWithLabel/index.tsx
```diff
@@ -0,0 +1,22 @@
...
+import './style.scss';
+
+interface InputWithLabel {
+	label: React.ReactNode;
+	labelPosition: 'left' | 'right';
+}
+
+const InputWithLabel: React.FC<InputWithLabel> = ({ children, label, labelPosition, ...props }) => {
+	const leftLabel = labelPosition === 'left' && label;
+	const rightLabel = labelPosition === 'right' && label;
```
manzoorwanijk, line 13: May be set a default value for `labelPosition`?

assets/src/domain/eventEditor/ui/tickets/ticketPriceCalculator/fields/useBaseField.tsx
```diff
@@ -0,0 +1,50 @@
...
+		}
+	} else {
+		fieldValue = format(fieldValue, name);
+	}
+
+	if (fieldValue === null) {
+		fieldValue = '';
+	}
+
+	return { handlers, fieldValue };
```
manzoorwanijk, line 47: `useMemo`

assets/src/domain/eventEditor/ui/tickets/ticketPriceCalculator/fields/usePrice.tsx
```diff
@@ -0,0 +1,26 @@
+import { useCallback } from 'react';
+
+import { useDataState } from '../data';
+import { BaseFieldProps } from './types';
+
+type BFP = BaseFieldProps;
+
+const usePrice = ({ field, price }: any): any => {
```
manzoorwanijk, line 8: Must need types here, it's not clear what `price` is, is it amount (number) or price object?

assets/src/domain/eventEditor/ui/tickets/ticketPriceCalculator/fields/usePrice.tsx
```diff
@@ -0,0 +1,26 @@
...
+		(value) => {
+			updatePrice({ id: price.id, fieldValues: { [field]: value } });
+		},
+		[updatePrice, field]
+	);
+
+	return {
+		getValue,
+		setValue,
+	};
```
manzoorwanijk, line 23: memo

assets/src/application/ui/input/InputWithLabel/index.tsx
nicolad: yes, good one, let's have right side as default

assets/src/domain/eventEditor/ui/tickets/ticketPriceCalculator/fields/useBaseField.tsx
nicolad: thanks, done

assets/src/domain/eventEditor/ui/tickets/ticketPriceCalculator/fields/BaseNumberInputField.tsx
```diff
@@ -0,0 +1,47 @@
...
+	const [fieldValue, setFieldValue] = React.useState(Number(value ?? getValue()));
+
+	const className = classNames(props.className, 'ee-input-base ee-input');
+	const onChange = useCallback(
+		(value) => {
+			setFieldValue(value);
+			setValue(parse(value, name));
+		},
+		[name, parse, setValue]
+	);
```
manzoorwanijk, line 32: Maintaining a local state may fail in case it's a base price and `reverseCalculate` is `true`. Changing of ticket total won't reflect the change in base price.

assets/src/domain/eventEditor/ui/tickets/ticketPriceCalculator/fields/usePrice.tsx
nicolad: sure, I was trying to make it work and forgot about proper types

assets/src/domain/eventEditor/ui/tickets/ticketPriceCalculator/fields/usePrice.tsx
nicolad: thx

assets/src/domain/eventEditor/ui/tickets/ticketPriceCalculator/fields/MoneyField/updateInputsSizes.tsx
tn3rb: that shouldn't be necessary because the offset is based on the currency sign and not the inputs
technically we should be able to determine the offset before the TPC even loads.

there are basically three scenarios to account for, but the first won't ever happen, so only the last two scenarios from the following cases will actually exist:

- currency sign before amount : prepend inputs with space for currency sign

- currency sign before + percentage modifiers : same as above, but need to also append inputs with space for % sign. The previous point won't actually exist because we will almost always need to account for %, so may as well just do so by default

- currency sign after amount : append inputs with space for currency sign. the same spacer will be used for % signs as well, so if the currency sign is 3 characters, then the spacer for the % sign will also be the same width as if it had 3 characters.

So to describe those two real scenarios again but in a different way:

- currency sign before amount: add spacer before all inputs with width defined by number of characters in currency sign PLUS add single width spacer after all inputs for % sign:  `[$$$$][ input ][%]`

- - currency sign after amount: add spacer after all inputs with width defined by number of characters in currency sign, same spacer will be used for % sign:  `[ input ][$%$%]`

assets/src/domain/eventEditor/ui/tickets/ticketPriceCalculator/fields/BaseNumberInputField.tsx
nicolad: updated: https://github.com/eventespresso/event-espresso-core/pull/2888/commits/d501db03e4fdb4d653b3e2cfcec8fa4c68ab53b8

assets/src/domain/eventEditor/ui/tickets/ticketPriceCalculator/fields/MoneyField/updateInputsSizes.tsx
nicolad: yes, after I've done that direct DOM mumbo-jumbo also was thinking that it's simpler to derive the space needed from sign length; was focused on doing that refactor related to BaseField and chakra's inputs, but will focus now on the aspect you've described

assets/src/domain/eventEditor/ui/tickets/ticketPriceCalculator/fields/BaseNumberInputField.tsx
nicolad: removed local state, had added this before because of that string to number coercion, but wrapped `fieldValue` with number and it's good to go

assets/src/domain/eventEditor/ui/tickets/ticketPriceCalculator/fields/BaseNumberInputField.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+	return (
+		<NumberInput
+			// because it can affect other tickets that have this price
+			// default price amount should not be changeable
+			isDisabled={disabled}
```
nicolad, line 33: based on: https://github.com/chakra-ui/chakra-ui/blob/master/packages/chakra-ui/src/NumberInput/index.js#L36

nicolad:
image

nicolad:
image

nicolad:
image

nicolad:
image

nicolad:
image

nicolad:
image

nicolad:
image

nicolad:
image

nicolad:
image

assets/src/domain/eventEditor/ui/tickets/ticketPriceCalculator/fields/BaseNumberInputField.tsx
```diff
@@ -0,0 +1,41 @@
...
+	placeholder,
+	getValue,
+	setValue,
+	value,
+}) => {
+	const fieldValue = Number(value ?? getValue());
+
+	const onChange = useCallback(
+		(value) => {
+			setValue(parse(value, name));
```
manzoorwanijk, line 22: `value` is already defined in upper scope

assets/src/domain/eventEditor/ui/tickets/ticketPriceCalculator/fields/usePrice.tsx
```diff
@@ -0,0 +1,29 @@
...
+import { BaseFieldProps, PriceFieldProps, UsePrice } from './types';
+
+type BFP = BaseFieldProps;
+
+const usePrice = ({ field, price }: PriceFieldProps): UsePrice => {
+	const { updatePrice } = useDataState();
+
+	const getValue: BFP['getValue'] = useCallback(() => price[field], [price[field]]);
+
+	const setValue: BFP['setValue'] = useCallback(
```
manzoorwanijk, line 13: To set better type safety, it's good to use `useCallback` generic to set types for internal callback and the variable will also infer the type from it.
```suggestion
	const getValue = useCallback<BFP['getValue']>(() => price[field], [price[field]]);

	const setValue = useCallback<BFP['setValue']>(
```

assets/src/domain/eventEditor/ui/tickets/ticketPriceCalculator/fields/usePrice.tsx
```diff
@@ -0,0 +1,29 @@
+import { useCallback, useMemo } from 'react';
+
+import { useDataState } from '../data';
+import { BaseFieldProps, PriceFieldProps, UsePrice } from './types';
+
+type BFP = BaseFieldProps;
+
+const usePrice = ({ field, price }: PriceFieldProps): UsePrice => {
```
manzoorwanijk, line 8: Since this hook only deals with price amount, it's better to name it something like `usePriceAmount`

assets/src/domain/eventEditor/ui/tickets/ticketPriceCalculator/fields/BaseNumberInputField.tsx
nicolad: thx, updated: https://github.com/eventespresso/event-espresso-core/pull/2888/commits/6e0783ab85a24ffcf09e1e3e8ab885d0c5115089

assets/src/domain/eventEditor/ui/tickets/ticketPriceCalculator/fields/usePrice.tsx
nicolad: great suggestion, updated

PR #2891, opened 2020-05-28
No object literals in props

manzoorwanijk:
This PR memoizes all the object literal props to avoid re-renders

### DIFF
--- a/assets/src/application/ui/display/EmptyState.tsx
+++ b/assets/src/application/ui/display/EmptyState.tsx
@@ -2,6 +2,7 @@ import React from 'react';
 import classNames from 'classnames';
 
 import { Banner } from '@infraUI/display';
+import { useMemoStringify } from '@application/services/hooks';
 
 interface EmptyStateProps {
 	children?: React.ReactNode;
@@ -12,11 +13,11 @@ interface EmptyStateProps {
 
 const EmptyState: React.FC<EmptyStateProps> = ({ children, description, title, ...props }) => {
 	const className = classNames(props.className, 'ee-empty-state');
-	const iconProps = {
+	const iconProps = useMemoStringify({
 		color: 'gray.100',
 		name: 'question-outline',
 		size: '96px',
-	};
+	});
 
 	return (
 		<Banner
--- a/assets/src/application/ui/display/TimezoneTimeInfo/Trigger.tsx
+++ b/assets/src/application/ui/display/TimezoneTimeInfo/Trigger.tsx
@@ -2,23 +2,28 @@ import React, { forwardRef } from 'react';
 
 import { GlobalOutlined } from '@appDisplay/icons/svgs';
 import { IconButton } from '@application/ui/input';
+import { useMemoStringify } from '@application/services/hooks';
+import { TooltipProps } from '@infraUI/display';
 
 interface TriggerProps {
 	tooltip: string;
 	onClick?: VoidFunction;
 }
 
-const Trigger = forwardRef<typeof IconButton, TriggerProps>(({ tooltip, ...props }, ref) => (
-	<IconButton
-		{...props}
-		borderless
-		color='white'
-		className='ee-timezone-info__button'
-		icon={() => <GlobalOutlined noMargin size='small' />}
-		tooltip={tooltip}
-		tooltipProps={{ placement: 'top' }}
-		ref={ref}
-	/>
-));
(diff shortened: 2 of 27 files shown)


assets/src/application/ui/display/TimezoneTimeInfo/Trigger.tsx
```diff
@@ -2,23 +2,28 @@ import React, { forwardRef } from 'react';
...
-));
+const Trigger = forwardRef<typeof IconButton, TriggerProps>(({ tooltip, ...props }, ref) => {
+	const tooltipProps = useMemoStringify<TooltipProps>({ placement: 'top' });
+	return (
+		<IconButton
+			{...props}
+			borderless
+			color='white'
+			className='ee-timezone-info__button'
+			icon={() => <GlobalOutlined noMargin size='small' />}
```
tn3rb, line 21: what about callbacks like this?
they can be wrapped in `useCallback` no?

we can do that in another PR though

assets/src/domain/eventEditor/ui/datetimes/datesList/filterBar/controls/SalesControl.tsx
```diff
@@ -5,39 +5,34 @@ import { SelectInput } from '@appInputs/SelectInput';
...
+		{
+			value: DatetimeSales.above50Capacity,
+			label: __('dates above 50% capacity'),
+		},
+		{
+			value: DatetimeSales.below50Capacity,
+			label: __('dates below 50% capacity'),
+		},
+	]);
+	return <SelectInput label={__('sales')} value={sales} options={options} onChangeValue={setSales} />;
```
tn3rb, line 35: NOICE

assets/src/application/ui/display/EmptyState.tsx
```diff
@@ -12,11 +13,11 @@ interface EmptyStateProps {
 
 const EmptyState: React.FC<EmptyStateProps> = ({ children, description, title, ...props }) => {
 	const className = classNames(props.className, 'ee-empty-state');
-	const iconProps = {
+	const iconProps = useMemoStringify({
 		color: 'gray.100',
 		name: 'question-outline',
 		size: '96px',
-	};
+	});
```
nicolad, line 20: this was already extracted wouldn't be simpler if we'd pull this out of component ?

assets/src/application/ui/display/TimezoneTimeInfo/Trigger.tsx
nicolad: or even better we might update `IconButton` such that it receives just JSX;
I think I read in one of KCD articles that passing JSX from the parent is more performant than rendering that at child level

assets/src/infrastructure/ui/inputs/dateTime/Datepicker/index.tsx
```diff
@@ -22,9 +23,11 @@ const Datepicker: React.FC<DatepickerProps> = ({ value, onChange, onChangeValue,
 		[onChange, onChangeValue]
 	);
 
+	const dateValue = useMemoStringify([value]);
+
 	return (
 		<div className={className}>
-			<BaseUIDatepicker {...props} onChange={onChangeHandler} value={[value]} />
+			<BaseUIDatepicker {...props} onChange={onChangeHandler} value={dateValue} />
```
nicolad, line 30: good one :+1:

assets/src/domain/eventEditor/ui/tickets/ticketsList/filterBar/controls/StatusControl.tsx
```diff
@@ -5,51 +5,49 @@ import { SelectInput } from '@appInputs/SelectInput';
...
+			{
+				value: TicketsStatus.expiredOnly,
+				label: __('expired tickets only'),
+			},
+			{
+				value: TicketsStatus.trashedOnly,
+				label: __('trashed tickets only'),
+			},
+		],
+		[isChained]
```
nicolad, line 48: nice

assets/src/application/ui/display/TimezoneTimeInfo/Trigger.tsx
manzoorwanijk: No they cannot be wrapped in `useCallback` because `icon` expects a React Component. May be `React.memo` instead.

PR #2906, opened 2020-06-01
Fix duplicate key warning in TPC

### LINKED ISSUE #2905: Children with the same key warning when adding default prices in TPC
Steps:
1. Make sure you have at least 2 default prices
2. Create a new ticket and goto TPC step
3. Click Add Default prices
4. Check console.

image

manzoorwanijk:
This PR adds `uuid` to the cloned prices in TPC to avoid keys being duplicated when price id is empty string.
Closes #2905


assets/src/domain/eventEditor/ui/tickets/ticketPriceCalculator/hooks/useAddDefaultPrices.ts
```diff
@@ -34,7 +35,7 @@ const useAddDefaultPrices = (): VoidFunction => {
 			return {
 				...priceModifier,
 				// clone it
-				id: '',
+				id: uuidv4(),
```
nicolad, line 38: good catch :+1:

Issue #2903, opened 2020-05-29
Review EDTR for Mobile, i18n, RTL,  and a11y issues


tn3rb:
need to test EDTR and look for issues pertaining to:

### i18n

- priority 1

- ensuring all public facing strings are translatable (try PigLatin plugin)

- ensure RTL languages are not impeded by UI

### a11y

- priority 2
- ensure EDTR is navigable via keyboard

- employ a11y tools to test for issues:

    - https://www.softwaretestinghelp.com/accessibility-testing-tools/

    - https://chrome.google.com/webstore/detail/wave-evaluation-tool/jbbplnpkjmmeebjpijfedlgcdilocofh

    - https://chrome.google.com/webstore/detail/axe-web-accessibility-tes/lhdoppojpmngadmnindnejefpokejbdd

### Mobile/Responsiveness

- priority 3

- ensure EDTR is navigable via touch

PLZ document all issues discovered here so that we can discuss solutions and make actionable items

nicolad:
great idea with prioritization :+1:

manzoorwanijk:
Translations not working for blocks. GB needs special loading of translation files. More info [here](https://developer.wordpress.org/block-editor/developers/internationalization/#load-translation-file).

manzoorwanijk:
> <@U03T6AM1N> <@UPHV02MC1> Re. a11y, I disconnected my mouse from my computer to check keyboard a11y of EDTR. I checked almost every component and it worked perfectly :heart_eyes:
<sub>[View in Slack](https://eventespresso.slack.com/archives/C080DDVLM/p1591021253291800)</sub>

manzoorwanijk:
## Mobile/Responsiveness
I tested on my OnePlus 6 (Android 10) with `chrome://inspect` to make sure there are no console errors. Here are some style issues I found:

manzoorwanijk:
#### 1. Needs a fix I guess
![image](https://user-images.githubusercontent.com/18226415/83419208-5808f400-a442-11ea-974d-2d117035882f.png)

manzoorwanijk:
#### 2. Some misplacement?
![image](https://user-images.githubusercontent.com/18226415/83419437-b03ff600-a442-11ea-95fd-6b710a85ba45.png)

manzoorwanijk:
#### 3. TPC button seems to be angry with the rest
[...]

PR #2909, opened 2020-06-02
More responsive fixes.

nicolad:
- Apply responsive fixes to issues from: https://github.com/eventespresso/event-espresso-core/issues/2903


nicolad:
image

nicolad:
image

nicolad:
image

nicolad:
image

nicolad:
image

nicolad:
image

nicolad:
image

nicolad:
image

nicolad:
image

nicolad:
image

nicolad:
image

nicolad:
image
image

assets/src/application/ui/input/Button/IconButton/style.scss
```diff
@@ -30,6 +32,12 @@
 		padding: var(--ee-padding-micro);
 		top: 0;
 		width: var(--ee-icon-button-size);
+
+		@include max782px {
```
nicolad, line 36: @tn3rb I am open to a better naming of these breakpoint mixins

assets/src/application/ui/input/InputWithLabel/style.scss
```diff
@@ -8,6 +10,10 @@
 			height: 42px;
 			margin: 0;
 			padding: 0;
+
+			@include max782px {
+				height: 40px;
```
nicolad, line 15: ideally, this value should derive from the input size, but that would mean that size needs to be passed to `InputWithLabel`, didn't want to slide through yak-shaving path here for now

assets/src/application/ui/layout/entityList/filterBar/style.scss
```diff
@@ -158,37 +165,10 @@ $filter-btn: '.ee-filter-bar__btn';
...
-	}
-
-	/* WordPress Admin Media Query Breakpoint: one column on the post write/edit screen */
-	@media only screen and (max-width: 850px) {
-	}
-
-	/* WordPress Admin Media Query Breakpoint: Tablet */
-	@media screen and (max-width: 782px) {
-	}
-
```
nicolad, line 191: hope this is ok, will prevent a lot of scrolling around
and is more maintanable

assets/src/application/ui/layout/espressoTable/style.scss
```diff
@@ -28,6 +30,10 @@
 	&__is-scrollable .ee-rspnsv-table__inner-wrapper {
 		overflow-x: scroll !important;
 		overflow-y: visible;
+
+		@include max782px {
+			overflow-x: unset !important;
```
nicolad, line 35: fix for the undesired scroll on mobile:

image

assets/src/domain/eventEditor/ui/datetimes/dateForm/multiStep/ContentBody.tsx
```diff
@@ -33,29 +32,33 @@ const ContentBody: React.FC = ({ children }) => {
 					<div>
 						<DateFormSteps current={current} />
 						{/* RFF fields */}
-						{current === 0 && children}
+						{current === 0 && (
+							<>
+								{children}
+								<ButtonRow rightAligned>
```
nicolad, line 38: doesn't really make sense to have a button row with a single button, had to do this to align that button on the right side

assets/src/application/ui/layout/espressoTable/style.scss
tn3rb: ya that looks a **LOT** better 🏅 

I have a few requests/suggestions based on the above image:

- decrease the space (margin or padding) on either side of the TAM table. Maybe half of what there would be good. Don't need that much white space on mobile when real estate is so expensive.

- button height is fine, but they don't need to be that wide, so maybe make them square, push them all the way to the right, and then use all of the gained space for the ticket info.

- for the ticket info, let's try the following:

	- move the sale date with the coloured status background to the far left of the cell, and increase the height to 100% of the cell interior (ie: keep the same spacing that exists now between the date block and the cell outline)

	- don't know if this is possible as is, (might have to add some additional markup like span tags) but put the month, day, and year portions for the dates on their own lines and center align them 

end result for layout should look kinda like:

```
**********
*   Jun  *					ID: 117	   * * * *  
*   03   *		Test Test Test Test	   * btn *  
*  2020  *					  $ 115	   * * * *  
**********
```

- finally, let's make the ID text a little bit lighter in color, and then maybe bold the ticket title

assets/src/domain/eventEditor/ui/datetimes/dateForm/multiStep/ContentBody.tsx
tn3rb: > doesn't really make sense to have a button row with a single button

why? it's just a layout element that separates the buttons from the other elements on the page.

How does the `rightAligned` prop affect things with regards to RTL ?

assets/src/domain/eventEditor/ui/ticketAssignmentsManager/components/table/styles.scss
```diff
@@ -3,14 +3,21 @@
...
+
+.ticket-assignment-manager.ticket-assignment-manager.ee-rspnsv-table__outer_wrapper {
+	.ee-rspnsv-table.ee-rspnsv-table {
 		background: var(--ee-background-color);
 
 		&__inner-wrapper {
 			background: var(--ee-background-color);
 			margin: 0 0 var(--ee-margin-smaller) 300px;
-			width: calc(100% - 300px) !important;
+			width: calc(100% - 300px);
```
tn3rb, line 15: 👍 👍  I have a bad habit of falling back on this when I get stuck instead of spending the extra time to solve things properly

assets/src/application/ui/input/InputWithLabel/style.scss
nicolad: there was an issue that I tried to fix, but don't know which one

nicolad: updated: 			
image

assets/src/application/ui/layout/espressoTable/style.scss
nicolad: relation button adjusted:

image

nicolad: hope this is close to what was expected:

image
***
- first, I've applied overflow hidden and ellipsis for name because multi line title would have created problems with space occupied by price
- I positioned price absolutely because it was tricky to arrange those 3 elements as expected

assets/src/domain/eventEditor/ui/datetimes/dateForm/multiStep/ContentBody.tsx
nicolad: I've used `justify-content: flex-end`, seems ok on RTL:

image

assets/src/application/ui/display/DebugInfo.tsx
```diff
@@ -41,7 +41,7 @@ const DebugInfo: React.FC<DebugInfoProps> = ({ data, asJson = true, asCollapse =
 
 	return (
 		<>
-			<Button buttonText={buttonText} onClick={handleToggle} size='md' mt='2em' />
+			<Button className='ee-debug-info-btn' buttonText={buttonText} onClick={handleToggle} size='md' mt='2em' />
```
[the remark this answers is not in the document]
nicolad: good observation, removed these altogether because font-size is used from the generic btn styles and because margin top doesn't feel quite right, instead added a bit of padding to the div above the button - `.sections-wrapper`

assets/src/application/ui/display/DebugInfo.tsx
```diff
@@ -41,7 +41,7 @@ const DebugInfo: React.FC<DebugInfoProps> = ({ data, asJson = true, asCollapse =
 
 	return (
 		<>
-			<Button buttonText={buttonText} onClick={handleToggle} size='md' mt='2em' />
+			<Button className='ee-debug-info-btn' buttonText={buttonText} onClick={handleToggle} />
```
tn3rb, line 44: great 👍

assets/src/application/ui/layout/espressoTable/style.scss
tn3rb: it's looking better... don't like the offset text alignments though...

let's just left align all of the ticket info so that all of that text is consistent and we don't have to deal with absolute positioning, which can cause unexpected issues

assets/src/domain/eventEditor/ui/ticketAssignmentsManager/components/table/styles.scss
```diff
@@ -158,39 +158,34 @@
...
 						&__name {
 							@include max782px {
 								font-weight: bold;
-								padding: 20px 10px;
 								overflow: hidden;
 								text-overflow: ellipsis;
 								text-align: left;
 								white-space: nowrap;
-								max-width: 210px;
+								max-width: 430px;
```
nicolad, line 179: image

Issue #2911, opened 2020-06-02
Email Validation Exception Thrown by Default Payment Method Setup During Activation


tn3rb:
The `EEM_Payment_Method` model has a relation with the `WP_User` table but does NOT have the foreign key (FK) set up the way we set up all of our other foreign keys, which would normally be done by adding the FK as a field on the other model (like how registrations have a `TXN_ID` field). Instead,  `EEM_Payment_Method`  has a `PMD_wp_user` field, but the `WP_User` table's primary key (PK) is just `ID` cuz it's a WP User object.

Anyways, the bug is that during activation when we try to create default payment methods, despite `PMD_wp_user` having a default value set for the admin super user (1), `EEM_Base` does not detect that because it doesn't see the  `PMD_wp_user` field as the FK for the `WP_User`  table. So then `EEM_Base` tries to use the default values for the  `WP_User`  table fields, which includes `null` for the email address,
which then of course fails our email validation and throws an exception... during activation :confused:

This is interfering with some work I'm doing for EDTR, so will need to try and figure out how to inform `EEM_Base` that the `PMD_wp_user` field can be used as the FK for the `WP_User` table.

PR #2916, opened 2020-06-03
Add Foreign Key Alias

tn3rb:
see #2911

this PR:

- adds a `$foreign_key_aliases` property to `EEM_Base`

- sets values for that array  in`EEM_WP_User`

- then in `EEM_Base::_deduce_fields_n_values_from_cols_n_values()`, if no Primary Key value is found for a table in the incoming `$cols_n_values` data, we check if any values exist in the `$foreign_key_aliases` for the table. If any of those values match the keys in the incoming `$cols_n_values` data, then we use that value for `$table_pk_value`

- corrects a null pointer exception (golden rule violation) in `EE_Email_Field`

- fixes property name typo, phpdocs, and formatting in `EEM_Payment_Method`

### testing notes

I'm not sure of how to reproduce the bug via normal plugin operation as I discovered it while coding and the error was interfering with the code I was working on. This was happening during regular plugin activation and I'm guessing that WP normally just ate the error and allowed the plugin to continue to activate. So here are some **suggested** steps to try although I don't know whether these would have revealed the issue before:

- perform standard new activation on a fresh site

- [ ] confirm (possibly via directly examining the db) that the default payment methods have been associated with your user account. Bonus points if your user ID is NOT `1` but it is correctly assigned.

- the changes in this PR should only affect logic that involves relations with the `WP_User` model, so test any logic you can think of where the current WP User matters, for example:

     - test the functionality of the EE WP User add-on

     - test various capabilities such as event ownership and edit permissions

- [ ] confirm that the registration process functions as expected

### DIFF
--- a/core/db_models/EEM_Base.model.php
+++ b/core/db_models/EEM_Base.model.php
@@ -5436,7 +5436,7 @@ abstract class EEM_Base extends EE_Base implements ResettableInterface
             // then check if that alias exists in the incoming data
             // AND that the actual PK the $FK_alias represents matches the $qualified_column (full PK)
             foreach ($this->foreign_key_aliases as $FK_alias => $PK_column) {
-                if (isset($cols_n_values[ $FK_alias ]) && $PK_column === $qualified_column) {
+                if ($PK_column === $qualified_column && isset($cols_n_values[ $FK_alias ])) {
                     $value = $cols_n_values[ $FK_alias ];
                     break;
                 }
--- a/core/db_models/EEM_Base.model.php
+++ b/core/db_models/EEM_Base.model.php
@@ -271,6 +271,12 @@ abstract class EEM_Base extends EE_Base implements ResettableInterface
     protected $_has_primary_key_field = null;
 
     /**
+     * array in the format:  [ FK alias => full PK ]
+     * where keys are local column name aliases for foreign keys
+     * and values are the fully qualified column name for the primary key they represent
+     *  ex:
+     *      [ 'Event.EVT_wp_user' => 'WP_User.ID' ]
+     *
      * @var array $foreign_key_aliases
      */
     protected $foreign_key_aliases = [];
@@ -5382,13 +5388,6 @@ abstract class EEM_Base extends EE_Base implements ResettableInterface
                 $table_obj->get_fully_qualified_pk_column(),
                 $table_obj->get_pk_column()
             );
-            // no PK?  ok check if there is a foreign key alias set for this table
-            if ($table_pk_value === null && ! empty($this->foreign_key_aliases)) {
-                // then check if that alias exists in the incoming data
-                foreach ($this->foreign_key_aliases as $FK_alias) {
-                    $table_pk_value = isset($cols_n_values[ $FK_alias ]) ? $cols_n_values[ $FK_alias ] : null;
-                }
-            }
             // there is a primary key on this table and its not set. Use defaults for all its columns
             if ($table_pk_value === null && $table_obj->get_pk_column()) {
                 foreach ($this->_get_fields_for_table($table_alias) as $field_name => $field_obj) {
@@ -5432,6 +5431,16 @@ abstract class EEM_Base extends EE_Base implements ResettableInterface
             $value = $cols_n_values[ $qualified_column ];
         } elseif (isset($cols_n_values[ $regular_column ])) {
             $value = $cols_n_values[ $regular_column ];
+        } elseif (! empty($this->foreign_key_aliases)) {
+            // no PK?  ok check if there is a foreign key alias set for this table
+            // then check if that alias exists in the incoming data
+            // AND that the actual PK the $FK_alias represents matches the $qualified_column (full PK)
+            foreach ($this->foreign_key_aliases as $FK_alias => $PK_column) {
+                if (isset($cols_n_values[ $FK_alias ]) && $PK_column === $qualified_column) {
+                    $value = $cols_n_values[ $FK_alias ];
+                    break;
+                }
+            }
         }
         return $value;
     }
--- a/core/db_models/EEM_Base.model.php
+++ b/core/db_models/EEM_Base.model.php
@@ -6,7 +6,6 @@ use EventEspresso\core\exceptions\InvalidInterfaceException;
 use EventEspresso\core\exceptions\ModelConfigurationException;
 use EventEspresso\core\exceptions\UnexpectedEntityException;
 use EventEspresso\core\interfaces\ResettableInterface;
-use EventEspresso\core\services\orm\ModelFieldFactory;
 use EventEspresso\core\services\loaders\LoaderFactory;
 
 /**
@@ -272,6 +271,11 @@ abstract class EEM_Base extends EE_Base implements ResettableInterface
     protected $_has_primary_key_field = null;
 
     /**
+     * @var array $foreign_key_aliases
+     */
+    protected $foreign_key_aliases = [];
+
+    /**
      * Whether or not this model is based off a table in WP core only (CPTs should set
      * this to FALSE, but if we were to make an EE_WP_Post model, it should set this to true).
      * This should be true for models that deal with data that should exist independent of EE.
@@ -288,7 +292,7 @@ abstract class EEM_Base extends EE_Base implements ResettableInterface
      * null until initialized by hasPasswordField()
      */
     protected $has_password_field;
-    
+
     /**
      * @var EE_Password_Field|null Automatically set when calling getPasswordField()
      */
@@ -5378,6 +5382,13 @@ abstract class EEM_Base extends EE_Base implements ResettableInterface
                 $table_obj->get_fully_qualified_pk_column(),
                 $table_obj->get_pk_column()
             );
+            // no PK?  ok check if there is a foreign key alias set for this table
+            if ($table_pk_value === null && ! empty($this->foreign_key_aliases)) {
+                // then check if that alias exists in the incoming data
+                foreach ($this->foreign_key_aliases as $FK_alias) {
+                    $table_pk_value = isset($cols_n_values[ $FK_alias ]) ? $cols_n_values[ $FK_alias ] : null;
+                }
+            }
             // there is a primary key on this table and its not set. Use defaults for all its columns
             if ($table_pk_value === null && $table_obj->get_pk_column()) {
                 foreach ($this->_get_fields_for_table($table_alias) as $field_name => $field_obj) {
--- a/core/db_models/EEM_Event_Message_Template.model.php
+++ b/core/db_models/EEM_Event_Message_Template.model.php
@@ -1,15 +1,13 @@
 <?php
 
- /**
+/**
  *  EEM_Event_Message_Template
  *  Model for relation table between EEM_Message_Template_Group and EEM_Event
  *
- * @package     Event Espresso
- * @subpackage  models
- * @since           4.3.0
+ * @package          Event Espresso
+ * @subpackage       models
+ * @since            4.3.0
  * @author           Darren Ethier
- *
- * ------------------------------------------------------------------------
  */
 class EEM_Event_Message_Template extends EEM_Base
 {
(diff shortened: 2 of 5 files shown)


core/db_models/EEM_Payment_Method.model.php
```diff
@@ -27,82 +24,97 @@ class EEM_Payment_Method extends EEM_Base
...
      * @Constructor
      * @access   protected
-     * @return EEM_Payment_Method
+     * @param null $timezone
+     * @throws EE_Error
      */
     protected function __construct($timezone = null)
     {
-        $this->singlular_item = __('Payment Method', 'event_espresso');
+        $this->singular_item = __('Payment Method', 'event_espresso');
```
tn3rb, line 42: singlular ???

nicolad: singlular from single

core/db_models/EEM_Base.model.php
```diff
@@ -5421,6 +5431,16 @@ protected function _get_column_value_with_table_alias_or_not($cols_n_values, $qu
...
+            // then check if that alias exists in the incoming data
+            // AND that the actual PK the $FK_alias represents matches the $qualified_column (full PK)
+            foreach ($this->foreign_key_aliases as $FK_alias => $PK_column) {
+                if ($PK_column === $qualified_column && isset($cols_n_values[ $FK_alias ])) {
+                    $value = $cols_n_values[ $FK_alias ];
+                    break;
+                }
+            }
         }
         return $value;
```
tn3rb, line 5445: this is essentially where the bug was...
The fully qualified primary key for the `EEM_WP_User` model is `WP_User.ID` but other models like `EEM_Event` specify that key like `Event.EVT_wp_user`.
So the incoming data might include a value for `Event.EVT_wp_user, but we were only checking for `WP_User.ID` or `ID` which of course would not exist.

core/db_models/EEM_Payment_Method.model.php
tn3rb: point taken... English often makes little to no sense.

core/db_models/EEM_Payment_Method.model.php
```diff
@@ -268,25 +358,24 @@ public function verify_button_urls($payment_methods = null)
...
                     $payment_method->save(
                         [
-                            'PMD_button_url' => $payment_method->type_obj()->default_button_url()
+                            'PMD_button_url' => $payment_method->type_obj()->default_button_url(),
                         ]
                     );
                 }
             } catch (EE_Error $e) {
-                $payment_method->set_active(false);
+                $payment_method->deactivate();
```
tn3rb, line 373: the `set_active()` method is not a regular setter and you can't just pass it `flase` because it does NOT even have any parameters!!!
so if you want to deactivate a payment method you need to call the `deactivate()` method

core/db_models/fields/EE_Email_Field.php
```diff
@@ -42,8 +43,8 @@ public function __construct($table_column, $nice_name, $nullable, $default_value
     public function prepare_for_set($email_address)
     {
         try {
-            $email_address = EmailAddressFactory::create($email_address);
-            return $email_address->get();
+            $email_address_obj = EmailAddressFactory::create($email_address);
+            return $email_address_obj instanceof EmailAddress ? $email_address_obj->get() : '';
```
tn3rb, line 47: Golden Rule: if it can be null, it will be null!

core/db_models/EEM_Payment_Method.model.php
knazart: Same thing on https://github.com/eventespresso/event-espresso-core/blob/97c97dfe9e55dd6115abb6026f9ba30164a5cbea/core/db_models/EEM_Event_Message_Template.model.php#L28 ain't it ?

core/db_models/EEM_Payment_Method.model.php
```diff
@@ -27,82 +24,148 @@ class EEM_Payment_Method extends EEM_Base
...
+                ),
+                'PMD_button_url'      => new EE_Plain_Text_Field(
+                    'PMD_button_url',
+                    __('Button URL', 'event_espresso'),
+                    true,
+                    ''
+                ),
+                'PMD_scope'           => new EE_Serialized_Text_Field(
+                    'PMD_scope',
+                    __('Usable From?', 'event_espresso'),
```
knazart, line 116: While here, should we also update this (and all the above) to:

```suggestion
                    esc_html__('Usable From?', 'event_espresso'),
```

?

core/db_models/EEM_Payment_Method.model.php
```diff
@@ -111,112 +174,129 @@ public function is_valid_scope($scope)
...
+
     /**
      * Counts all active gateways in the specified scope
+     *
      * @param string $scope one of EEM_Payment_Method::scope_*
-     * @param array $query_params
+     * @param array  $query_params
      * @return int
+     * @throws EE_Error
+     * @throws EE_Error
```
knazart, line 206: Looks like these doubled. Or one of these had to be of a different error type ?
```suggestion
     * @throws EE_Error
```

core/db_models/EEM_Payment_Method.model.php
```diff
@@ -111,112 +174,129 @@ public function is_valid_scope($scope)
...
      * argument to get all active for a given scope
-     * @param string $scope one of the constants EEM_Payment_Method::scope_*
-     * @param array $query_params @see https://github.com/eventespresso/event-espresso-core/tree/master/docs/G--Model-System/model-query-params.md
-     * @return array @see https://github.com/eventespresso/event-espresso-core/tree/master/docs/G--Model-System/model-query-params.md
+     *
+     * @param string $scope        one of the constants EEM_Payment_Method::scope_*
+     * @param array  $query_params
+     * @return array
      * @throws EE_Error
+     *@see https://github.com/eventespresso/event-espresso-core/tree/master/docs/G--Model-System/model-query-params.md
```
knazart, line 256:
```suggestion
     * @see https://github.com/eventespresso/event-espresso-core/tree/master/docs/G--Model-System/model-query-params.md
```

core/db_models/EEM_Payment_Method.model.php
```diff
@@ -111,112 +174,129 @@ public function is_valid_scope($scope)
...
 
     /**
      * Gets one active payment method. see @get_all_active for documentation
+     *
      * @param string $scope
      * @param array  $query_params
-     * @return EE_Payment_Method
+     * @return EE_Base_Class|EE_Payment_Method|EE_Soft_Delete_Base_Class|NULL
+     * @throws EE_Error
+     * @throws EE_Error
```
knazart, line 271: Once more here
```suggestion
     * @throws EE_Error
```

core/db_models/EEM_Payment_Method.model.php
```diff
@@ -254,10 +340,14 @@ public function ensure_is_ID($base_obj_or_id_or_slug)
...
     /**
-     * Verifies the button urls on all the passed payment methods have a valid button url. If not, resets them to their default.
-     * @param EE_Payment_Method[] $payment_methods. If NULL is provided defaults to all payment methods active in the cart
+     * Verifies the button urls on all the passed payment methods have a valid button url.
+     * If not, resets them to their default.
+     *
+     * @param EE_Payment_Method[] $payment_methods if NULL, defaults to all payment methods active in the cart
+     * @throws EE_Error
+     * @throws ReflectionException
+     * @throws EE_Error
```
knazart, line 350:
```suggestion
     * @throws ReflectionException
     * @throws EE_Error
```

core/db_models/EEM_Payment_Method.model.php
```diff
@@ -360,23 +452,26 @@ public function get_all_for_transaction($transaction, $scope)
...
-     * Note: if an offline payment method was selected on the related transaction then this will have no payment methods returned.
-     * It will ONLY return a payment method for a PAYMENT recorded against the registration.
-     *
-     * @param EE_Registration|int $registration_or_reg_id  Either the EE_Registration object or the id for the registration.
+     * @param EE_Registration|int $registration_or_reg_id Either the EE_Registration object or the id for the
+     *                                                    registration.
      * @return EE_Payment|null
+     * @throws EE_Error
+     * @throws EE_Error
+     * @throws EE_Error
```
knazart, line 463: Same ?

core/db_models/EEM_WP_User.model.php
```diff
@@ -100,6 +100,16 @@ protected function __construct($timezone = null, ModelFieldFactory $model_field_
...
         );
+        $this->foreign_key_aliases = [
+            'Event.EVT_wp_user'          => 'WP_User.ID',
+            'Payment_Method.PMD_wp_user' => 'WP_User.ID',
+            'Price.PRC_wp_user'          => 'WP_User.ID',
+            'Price_Type.PRT_wp_user'     => 'WP_User.ID',
+            'Question.QST_wp_user'       => 'WP_User.ID',
+            'Question_Group.QSG_wp_user' => 'WP_User.ID',
+            'Venue.TKT_wp_user'          => 'WP_User.ID',
+            'Ticket.VNU_wp_user'         => 'WP_User.ID',
```
knazart, line 111: Should this be:

```suggestion
            'Ticket.TKT_wp_user'          => 'WP_User.ID',
            'Venue.VNU_wp_user'         => 'WP_User.ID',
```
?

core/db_models/fields/EE_Email_Field.php
knazart: Yup, yup, yup.

core/db_models/EEM_Payment_Method.model.php
tn3rb: yes we should... I totally missed that 

thnx, will update

core/db_models/EEM_Payment_Method.model.php
tn3rb: yup... magical IDE updates not turning out well

thnx will update

core/db_models/EEM_Payment_Method.model.php
tn3rb: weird... never happened in the past (that I'm aware of), so must have just started after an update

core/db_models/EEM_WP_User.model.php
tn3rb: ya me likey, thnx will update

core/db_models/EEM_Payment_Method.model.php
tn3rb: good catch... will fix

tn3rb:
@knazart 
thnx for the great review, have made all of the requested changes

core/db_models/EEM_Payment_Method.model.php
```diff
@@ -215,12 +212,11 @@ public function count_active($scope = null, $query_params = [])
...
-     * @param string $scope        one of the constants EEM_Payment_Method::scope_*
-     * @param array  $query_params @see
-     *                             https://github.com/eventespresso/event-espresso-core/tree/master/docs/G--Model-System/model-query-params.md
-     * @return array @see
-     *               https://github.com/eventespresso/event-espresso-core/tree/master/docs/G--Model-System/model-query-params.md
+     * @param string $scope one of the constants EEM_Payment_Method::scope_*
+     * @param array  $query_params
+     * @return array
      * @throws EE_Error
+     * @see https://github.com/eventespresso/event-espresso-core/tree/master/docs/G--Model-System/model-query-params.md
```
knazart: 👍

core/db_models/EEM_Event_Message_Template.model.php
```diff
@@ -1,95 +1,127 @@
...
         }
-
-        if (!empty($EVT_IDs)) {
-            $where['EVT_ID'] = array( 'IN', (array) $EVT_IDs );
+        if (! empty($EVT_IDs)) {
+            $where['EVT_ID'] = ['IN', (array) $EVT_IDs];
         }
 
-        return $this->delete(array( $where ), false);
+        return $this->delete([$where], false);
```
knazart, line 125: 👍

core/db_models/EEM_Payment_Method.model.php
```diff
@@ -347,11 +342,12 @@ public function ensure_is_ID($base_obj_or_id_or_slug)
...
      * @throws EE_Error
      * @throws ReflectionException
-     * @throws EE_Error
      */
     public function verify_button_urls($payment_methods = null)
     {
-        $payment_methods = is_array($payment_methods) ? $payment_methods : $this->get_all_active(EEM_Payment_Method::scope_cart);
+        $payment_methods = is_array($payment_methods)
+            ? $payment_methods
+            : $this->get_all_active(EEM_Payment_Method::scope_cart);
```
knazart, line 350: 👍

Pebblo:
> confirm (possibly via directly examining the db) that the default payment methods have been associated with your user account. Bonus points if your user ID is NOT 1 but it is correctly assigned.

Whilst no errors are thrown etc, the PMD_wp_user field is set to 1 regardless of what my User account ID is.

This is a clean activation using a second admin account:

https://monosnap.com/file/H83y0tR7dW1G9wZAOIfUNtdaqLjBxk

The last row is from me enabling the payment methods pro add-on and creating a second Invoice payment method with the same user account used to install/activate EE. 

> the changes in this PR should only affect logic that involves relations with the WP_User model, so test any logic you can think of where the current WP User matters, for example:
> 
> test the functionality of the EE WP User add-on
> 
> test various capabilities such as event ownership and edit permissions

Tested the WP User account with this branch, everything worked as expected.

I then used a plugin I share with users to create a couple of roles on the site 'Event Organizer' and 'Events Manager' (Organizer only has access to Events they are set as the author of and event data for those events, managers have full access to EE but not WP)

Again, worked as expected.

> confirm that the registration process functions as expected

Test multiple registrations paid and free, also linked WP accounts using the EE WP Users add-on, all good.

tn3rb:
> Whilst no errors are thrown etc, the PMD_wp_user field is set to 1 regardless of what my User account ID is.

**FINALLY** figured this out...

So based on the code I've looked at, what you experienced above is the expected behaviour, although I'm not sure why it was set up the way it is.

During activation, the 4.6 DMS script runs and calls `EE_DMS_Core_4_6_0::add_default_admin_only_payments()` which gets the User ID saved to the db for all of the payment methods from `EEH_Activation::get_default_creator_id()`;

The phpdoc comments on that method state:

> Tries to find the oldest admin for this site.  If there are no admins for this site then return NULL.

which is achieved via a query that looks something like:

```sql
SELECT user_id
FROM wp_usermeta
WHERE meta_key = 'wp_capabilities'
	AND meta_value LIKE '{/*big-number*/*}administrator{/*big-number*/}'
ORDER BY user_id ASC
LIMIT 0,1
```

which is almost guaranteed to return 1 for the initial super user, unless that record has been deleted from the db at some point.

Now the million dollar question is **WHY** are we not using the current user ID?

Possible reasons I can think of include:

- the WP current user is not set up until the `init` hook point is run and the DMS script is executing before that?

- there was some permissions issue that has happened in the past with EE activation and it was concluded that Payment Methods should be assigned to the WP super user?

- the User ID was originally hardcoded as 1 on the assumption that the person activating EE would always be the super user, but then that caused an issue once upon a time when user 1 had been deleted on a site, and so somebody changed the logic to simply find the earliest admin account?

Considering that this has not caused any issues that I am aware of, I'm inclined to leave as is.

thoughts?

garthkoyle:
Was there aa reason this was done for Event Smart?

tn3rb:
> Was there aa reason this was done for Event Smart?

I guess it's possible although each ES site has its own tables so whoever the primary account holder is (not sure what term you guys use for that) would almost certainly have a User ID of 1. It's possible that ES can't use the WP current user global because maybe the site is not being installed and activated directly by the user that is opening the account. But even if that is the case, it would be unnecessary to search for "the oldest" admin since there would only ever be one record in the users table for a newly created site and its ID would be 1.

Pebblo:
Looks like this was added for multisite compatibility:

https://events.codebasehq.com/projects/event-espresso/tickets/6845

Used by the ee4-multisite add-on:

https://events.codebasehq.com/projects/event-espresso/repositories/ee4-multisite/blob/9ecd5e6a45/EED_Multisite.module.php#L263

So I'd say leave this as is.

Just to note it looks like `EEH_Activation::get_default_creator_id();` is used everywhere that a user_id is set within EE so seems odd that we only get an error with payment methods.

PR #2939, opened 2020-06-08
Fix fields to sync from TPC to ticket edit form in multi-step

### LINKED ISSUE #2937: Ticket name not changed in multistep edit form if you use TPC step
Steps:
1. Click on **edit ticket** from dropdown menu
2. change the name in the form
3. Click on **Add ticket prices** button
4. Then complete to submit

Expected:
Ticket name should change.

Actual:
Ticket name remains the same it was before

manzoorwanijk:
This fields fixes the issue of TPC name being synced to edit form in multi-step.

Closes #2937

### DIFF
--- a/assets/src/domain/eventEditor/ui/tickets/ticketForm/multiStep/useDataListener.ts
+++ b/assets/src/domain/eventEditor/ui/tickets/ticketForm/multiStep/useDataListener.ts
@@ -1,8 +1,14 @@
 import { useEffect } from 'react';
 import { useForm } from 'react-final-form';
+import { pick } from 'ramda';
 
 import { useDataState as useTAMDataState } from '@edtrUI/ticketAssignmentsManager/data';
 import { useDataState as useTPCDataState } from '@edtrUI/tickets/ticketPriceCalculator/data';
+import { Ticket } from '@edtrServices/apollo';
+
+// The fields that need to be synced from TPC to tikcet edit form
+// This is to avoid ticket.name being synced
+const TPC_TICKET_FIELDS_TO_SYNC: Array<keyof Ticket> = ['isTaxable', 'price', 'reverseCalculate'];
 
 /**
  * A custom hook which subscribes to TAM and TPC data and updates
@@ -21,7 +27,8 @@ const useDataListener = () => {
 
 	const { deletedPrices, prices, ticket } = useTPCDataState();
 	useEffect(() => {
-		Object.entries(ticket).forEach(([key, value]) => {
+		const fieldsToSync = pick(TPC_TICKET_FIELDS_TO_SYNC, ticket);
+		Object.entries(fieldsToSync).forEach(([key, value]) => {
 			// update value of each field in RFF state
 			mutators.updateFieldValue(key, value);
 		});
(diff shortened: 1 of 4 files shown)


assets/src/domain/eventEditor/ui/tickets/ticketForm/multiStep/useDataListener.ts
```diff
@@ -1,8 +1,14 @@
...
 import { useForm } from 'react-final-form';
+import { pick } from 'ramda';
 
 import { useDataState as useTAMDataState } from '@edtrUI/ticketAssignmentsManager/data';
 import { useDataState as useTPCDataState } from '@edtrUI/tickets/ticketPriceCalculator/data';
+import { Ticket } from '@edtrServices/apollo';
+
+// The fields that need to be synced from TPC to tikcet edit form
+// This is to avoid ticket.name being synced
+const TPC_TICKET_FIELDS_TO_SYNC: Array<keyof Ticket> = ['isTaxable', 'price', 'reverseCalculate'];
```
tn3rb, line 11: i'm wondering whether it would be better to locate this somewhere within the TPC folder and then import it here?
that could work well if it was part of a file where these fields are handled in TPC, that way if something ever gets added or changed, this constant can be updated at the source.

if there isn't a good spot in TPC to locale this logic, then it's fine as is.

thoughts?

assets/src/domain/eventEditor/ui/tickets/ticketForm/multiStep/useDataListener.ts
```diff
@@ -21,7 +27,8 @@ const useDataListener = () => {
 
 	const { deletedPrices, prices, ticket } = useTPCDataState();
 	useEffect(() => {
-		Object.entries(ticket).forEach(([key, value]) => {
+		const fieldsToSync = pick(TPC_TICKET_FIELDS_TO_SYNC, ticket);
+		Object.entries(fieldsToSync).forEach(([key, value]) => {
```
tn3rb, line 31: don't you hate when things seem totally obvious in retrospect ?!?!

assets/src/domain/eventEditor/ui/tickets/ticketForm/multiStep/useDataListener.ts
manzoorwanijk: There already is a constant in `/domain/eventEditor/ui/tickets/ticketPriceCalculator/utils/constants.ts` named `ticketFieldsToUse` which inlcudes `name` but I deliberately didn't use that to avoid any future issue if we add any other field to TPC. But I think your suggestion is better, I can just remove `name` from  `ticketFieldsToUse` and use it to sync fields. Then in future if other fields are added to that array, we can accordigly update the code here.

tn3rb: let's keep things as is then
# 🐑

assets/src/domain/eventEditor/ui/tickets/ticketForm/multiStep/useDataListener.ts
```diff
@@ -1,8 +1,14 @@
 import { useEffect } from 'react';
 import { useForm } from 'react-final-form';
+import { pick } from 'ramda';
 
 import { useDataState as useTAMDataState } from '@edtrUI/ticketAssignmentsManager/data';
 import { useDataState as useTPCDataState } from '@edtrUI/tickets/ticketPriceCalculator/data';
+import { Ticket } from '@edtrServices/apollo';
+
+// The fields that need to be synced from TPC to tikcet edit form
```
nicolad, line 9: me being nitpicky: `ticket`

Issue #2935, opened 2020-06-05
EDTR UI Tweaks


tn3rb:
# Edit Entity Forms

### ⋆⋆⋆ required input indicator ⋆⋆⋆

there is currently nothing to indicate when an input is required, and yet the form submit button will be disabled.

![ScreenShot_20200605160203](https://user-images.githubusercontent.com/1751030/83928822-f39dab80-a745-11ea-9e7d-76fcbd2b35df.png)

typically we have appended an `*` to the input label to indicate that it is a required field, so let's do that and give it the following styles (will need to wrap asterisk in a span with a class like `ee-required-asterisk`):

```css
  color: var(--ee-color-accent);
  font-weight: 900;
  margin: var(--ee-margin-micro);
```

It is only when you focus out of an input that a validation notice appears

![ScreenShot_20200605160151](https://user-images.githubusercontent.com/1751030/83928826-fa2c2300-a745-11ea-885d-861263972e75.png)

let's also add a class to those validation error notices and give them the following styles:

```css
  color: var(--ee-color-white);
  background: var(--ee-color-accent);
  border-radius: var(--ee-border-radius-small);
  display: inline-flex;
  letter-spacing: 2px;
  padding: var(--ee-padding-tiny) var(--ee-padding-default) var(--ee-padding-tiny) var(--ee-padding-small);
```

which should look something like:

![ScreenShot_20200605161034](https://user-images.githubusercontent.com/1751030/83929117-23997e80-a747-11ea-93fb-88f69ef7def1.png)

.

### ⋆⋆⋆ button colors and icons ⋆⋆⋆

plz adjust modal form buttons as follows:

- add `ee-btn--primary` class to all submit buttons (including the TAM modal submit)

- add `ee-btn--secondary` class to the "add ticket prices" button but **ONLY** when on the first step of the edit ticket form

![ScreenShot_20200605164645](https://user-images.githubusercontent.com/1751030/83930390-3498be80-a74c-11ea-9b6a-08337822013f.png)

- add the same "save" icon that appears on the TAM form to all submit buttons

![ScreenShot_20200605164158](https://user-images.githubusercontent.com/1751030/83930271-8b51c880-a74b-11ea-90de-c933535f3703.png)

- as shown in the previous two images, add a second arrow icon to any buttons that jump two steps. They will need to have a negative margin of `-0.5rem` to the side of the icon bordering its twin in order to pull the two icons together

---

# Ticket Price Calculator

[...]

PR #2945, opened 2020-06-09
Added styles to FormErrorMessage.

nicolad:
- Partially closes: https://github.com/eventespresso/event-espresso-core/issues/2935

## How has this been tested
image
***
image


assets/src/application/ui/forms/espressoForm/renderers/ErrorMessage/style.scss
```diff
@@ -0,0 +1,15 @@
+.ee-form-error-message.ee-form-error-message {
```
nicolad, line 1: this is a war, not with North Korea, but still

PR #2949, opened 2020-06-11
Added RemoveTaxesButton.

nicolad:
## Problem this Pull Request solves
- Partially closes: https://github.com/eventespresso/event-espresso-core/issues/2935

***
Original requirements:

### ⋆⋆⋆ add a "remove taxes" button ⋆⋆⋆

after clicking the "add default taxes" button, taxes have to be removed manually by clicking the delete button for each one. This is fine, but could be expedited by changing the "add default taxes" button to say "remove taxes" once taxes have been added, and upon clicking:

 - remove ALL tax prices related to the ticket

 - delete tax prices that are NOT defaults

 - set `isTaxable` on ticket to `false`. This should also occur if taxes are deleted manually using the trash button

PR #2950, opened 2020-06-11
Fix TPC tax deletion issue

### LINKED ISSUE #2934: TPC Issues: Price Order Not Set Correctly && Taxes Applied After Deleting Prices
### Steps to Reproduce

- use https://brent.eventespresso.com/

- create a new ticket and on the TPC form step click Add Default Prices

- set base ticket price to $10 and leave tax as is

- save, assign a date, save new ticket, and wait for mutations to complete

- click TPC icon to edit prices

- delete all prices and submit

- click TPC icon again to edit prices

- **???** default tax appears again despite having been deleted

- if you monkey around with variations of the above steps you can sometimes get the default base price to appear AFTER the default tax, but this can not be reproduced reliably and seems to happen almost randomly, but essentially, the PRC_order is not getting set. I've looked directly in the db on my testing site and most of the prices have an order of 0. You can use this tool by going to https://brent.eventespresso.com/wp-admin/tools.php?page=databasebrowser

So looks like two things need to happen

- `isTaxable` needs to be set back to `false` when ALL prices are deleted

- `PRC_order` field needs to get set upon save. This will be a bit tricky because the order of the default taxes can not be modified, so all prices and modifiers prior to taxes can simply be ordered sequentially (1,2,3...) but any that appear after taxes will need to start from the highest order of the last tax (ie, if the last tax has an order of 70, then the next modifier needs to start at 71, although it might be wise to put a bigger gap between taxes and any additional prices (like start at 80)

manzoorwanijk:
This PR fixes the issue of default tax not being removed from a ticket with the following changes:
- Allows empty `prices` array in GQL schema to be handled in ticket update mutation
- Creates tax removal hooks - `useRemoveAllTaxes` (for #2949) and `useRemoveDefaultTax`
- Wires up price deletion buttons
Closes #2934

### DIFF
--- a/assets/src/domain/eventEditor/ui/tickets/ticketPriceCalculator/buttons/DeleteAllPricesButton.tsx
+++ b/assets/src/domain/eventEditor/ui/tickets/ticketPriceCalculator/buttons/DeleteAllPricesButton.tsx
@@ -5,24 +5,31 @@ import { ConfirmDelete } from '@appDisplay/confirm';
 import { useDataState } from '../data';
 import { ButtonType } from '@application/ui/input';
 import { useMemoStringify } from '@application/services/hooks';
+import { useRemoveDefaultTax } from '../hooks';
 
 const DeleteAllPricesButton: React.FC = () => {
-	const { prices, deletePrice, updateTicketPrice } = useDataState();
+	const { prices, deletePrice, ticket, updateTicketPrice } = useDataState();
+	const removeDefaultTax = useRemoveDefaultTax(ticket?.id);
 
 	const buttonProps = useMemoStringify({
 		buttonText: __('Delete all prices'),
 		buttonType: ButtonType.ACCENT,
 	});
 	const onConfirm = useCallback(() => {
-		prices.forEach(({ id, isNew, isDefault }) => {
-			deletePrice(id, isNew || isDefault);
+		prices.forEach((price) => {
+			// delete the price from TPC state
+			deletePrice(price.id, price.isNew || price.isDefault);
+			// Remove default tax from relations
+			removeDefaultTax(price);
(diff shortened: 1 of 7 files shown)


assets/src/domain/eventEditor/ui/tickets/ticketPriceCalculator/buttons/DeleteAllPricesButton.tsx
```diff
@@ -5,24 +5,31 @@ import { ConfirmDelete } from '@appDisplay/confirm';
...
 		buttonType: ButtonType.ACCENT,
 	});
 	const onConfirm = useCallback(() => {
-		prices.forEach(({ id, isNew, isDefault }) => {
-			deletePrice(id, isNew || isDefault);
+		prices.forEach((price) => {
+			// delete the price from TPC state
+			deletePrice(price.id, price.isNew || price.isDefault);
+			// Remove default tax from relations
+			removeDefaultTax(price);
```
nicolad, line 23: nice

assets/src/domain/eventEditor/ui/tickets/ticketPriceCalculator/hooks/useRemoveAllTaxes.ts
```diff
@@ -0,0 +1,22 @@
+import { useCallback } from 'react';
+
+import { useDataState } from '../data';
+import useRemoveDefaultTax from './useRemoveDefaultTax';
+import { isTax } from '@sharedEntities/priceTypes/predicates';
+
+const useRemoveAllTaxes = (): VoidFunction => {
```
nicolad, line 7: this is cool :sunglasses:

PR #2952, opened 2020-06-11
Move expiredEventMessage conditional so that it is displayed the ticket selector for expired events

Pebblo:
Add the ticket selector shortcode for an expired event to a page and view it.

In master you'll see something like this: https://monosnap.com/file/XgnJd3D2X8lgr7gMeWgX98zSo3bq5Q

With this branch:
https://monosnap.com/file/EICn58Qts3NECbzd0BZx5uNHExPPmG

Check that the  'ticket sales closed' message still displays as expected (datetime(s) on the event aren't all expired but all ticket sell until dates have expired). It should show:

>We're sorry, but ticket sales have been closed at this time. Please check back again later.

## Checklist

* [ ] My code is tested.

### DIFF
--- a/modules/ticket_selector/DisplayTicketSelector.php
+++ b/modules/ticket_selector/DisplayTicketSelector.php
@@ -203,6 +203,11 @@ class DisplayTicketSelector
         if (! $this->setEvent($event)) {
             return false;
         }
+        // is the event expired ?
+        $template_args['event_is_expired'] = ! is_admin() ? $this->event->is_expired() : false;
+        if ($template_args['event_is_expired']) {
+            return $this->expiredEventMessage();
+        }
         // begin gathering template arguments by getting event status
         $template_args = array('event_status' => $this->event->get_active_status());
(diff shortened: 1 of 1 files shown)


modules/ticket_selector/DisplayTicketSelector.php
```diff
@@ -259,6 +259,33 @@ public function display($event = null, $view_details = false)
...
+                   ));
+        d((
+                       $_event_active_status !== EE_Datetime::active
+                       && $_event_active_status !== EE_Datetime::upcoming
+                       && $_event_active_status !== EE_Datetime::sold_out
+                       && ! (
+                           $_event_active_status === EE_Datetime::inactive
+                           && is_user_logged_in()
+                       )));
+
```
tn3rb, line 288: what's this?  is this from the Kint debugger?

Pebblo: Oops! My bad.

Yeah that Kint and I completely forgot I had that code in there from when I was figuring out which conditional was 'failing'.

PR #2953, opened 2020-06-11
TPC input height fix.

nicolad:

image

## How has this been tested
image


nicolad:
the problem was a subtle inconsistency between global input height and button height;
for example input height here is also 40px, but since it's not side-by-side it's not easy to spot the inconsistency, this is why it was caught in TPC context:

image

nicolad:
> wonder if we should change the inputs and buttons to all use the same var for height?

this is tricky because we have button with different sizes, but we don't have size prop and related logic for inputs, need to think of a way this could be accomplished;
***
is there a case where, for example if we'd set on form level the prop size to be small(for) example and than use that to control input and button sizes ?

PR #2959, opened 2020-06-12
Fix default dates constants

### LINKED ISSUE #2957: Default date for new entities is wrong
Steps:
1. Add new date or ticket
2. Check  the date in date picker

Expected: It should be one month from current date

Actual: It's not exactly one month from current date.

manzoorwanijk:
This PR fixes the bug in default date constants which wrongly used day of the week as day of the month.

Closes #2957

### DIFF
--- a/assets/src/domain/shared/constants/defaultDates.ts
+++ b/assets/src/domain/shared/constants/defaultDates.ts
@@ -1,8 +1,8 @@
 export const NOW = new Date();
 export const CURRENT_YEAR = NOW.getFullYear();
 export const CURRENT_MONTH = NOW.getMonth();
-export const CURRENT_DAY = NOW.getDay();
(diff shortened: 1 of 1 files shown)


assets/src/domain/shared/constants/defaultDates.ts
```diff
@@ -1,8 +1,8 @@
 export const NOW = new Date();
 export const CURRENT_YEAR = NOW.getFullYear();
 export const CURRENT_MONTH = NOW.getMonth();
-export const CURRENT_DAY = NOW.getDay();
-export const PLUS_ONE_MONTH = new Date(CURRENT_YEAR, CURRENT_MONTH + 1, CURRENT_DAY);
-export const PLUS_TWO_MONTHS = new Date(CURRENT_YEAR, CURRENT_MONTH + 2, CURRENT_DAY);
+export const CURRENT_DATE = NOW.getDate(); // day of the month
```
nicolad, line 4: :+1:

PR #2986, opened 2020-06-23
Control icon horizontal margins and color via prop.

nicolad:
- Initially within the icon's `withEnhance` HOC I've added possibility to generate icon classname based on it's display name. Although this works, this approach is enhancing patchy-way of applying styles, leading in less maintainable CSS and a bit of guesswork, ie.: how did `ee-icon--info-circle-outlined` got generated.
- In this PR I've made a quick refactor to be able to pass `horizontalMargin` prop

### DIFF
--- a/assets/src/application/ui/display/icons/svgs/style.scss
+++ b/assets/src/application/ui/display/icons/svgs/style.scss
@@ -1,7 +1,8 @@
+@import "application/ui/styles/margins.scss";
+
 svg {
 	&.ee-svg.ee-svg {
 		display: inline-flex;
-		margin: 0 var(--ee-margin-micro);
 
 		&.ee-icon {
 			height: 1.25rem;
--- a/assets/src/application/ui/display/icons/svgs/types.ts
+++ b/assets/src/application/ui/display/icons/svgs/types.ts
@@ -4,3 +4,6 @@ export interface IconProps {
 }
 
 export type IconSize = 'big' | 'bigger' | 'small' | 'smaller' | 'tiny';
+
+export type HorizontalMargin = 'extreme' | 'huge' | 'big' | 'bigger' | 'small' | 'smaller' | 'tiny' | 'micro' | 'nano';
+
--- a/assets/src/application/ui/forms/espressoForm/renderers/FieldRenderer.tsx
+++ b/assets/src/application/ui/forms/espressoForm/renderers/FieldRenderer.tsx
@@ -40,7 +40,7 @@ const FieldRenderer: React.FC<FieldRendererProps> = (props) => {
 				{label}
 				{info && (
 					<Tooltip placement='right' tooltip={info}>
-						<InfoCircleOutlined />
+						<InfoCircleOutlined color="var(--ee-color-primary-low-contrast)" horizontalMargin="tiny" />
 					</Tooltip>
 				)}
 			</FormLabel>
--- a/assets/src/application/ui/forms/espressoForm/styles.scss
+++ b/assets/src/application/ui/forms/espressoForm/styles.scss
@@ -1,4 +1,4 @@
-@import 'application/ui/styles/mixins/breakpoints.scss';
+@import "application/ui/styles/mixins/breakpoints.scss";
 
 .ee-form {
 	.form-wrapper {
@@ -212,11 +212,6 @@
 				@include max360px {
 					width: 100%;
 				}
-			}
-
-			.ee-icon--info-circle-outlined {
-				color: var(--ee-color-primary-low-contrast);
-				margin: 0 var(--ee-margin-tiny);
 			}
 		}
 
--- a/assets/src/application/ui/input/InlineEditInput/style.scss
+++ b/assets/src/application/ui/input/InlineEditInput/style.scss
@@ -14,5 +14,13 @@
 }
 
 .ee-tabbable-text {
+	border-radius: var(--ee-border-radius-small);
+	border-width: 3px;
 	padding: var(--ee-padding-micro) var(--ee-padding-small);
+
+	&:active,
+	&:focus {
+		border-width: 3px;
+		border-color: var(--ee-color-primary);
+	}
 }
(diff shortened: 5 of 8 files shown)


assets/src/application/ui/display/icons/svgs/style.scss
```diff
@@ -1,7 +1,8 @@
+@import "application/ui/styles/margins.scss";
+
 svg {
 	&.ee-svg.ee-svg {
 		display: inline-flex;
-		margin: 0 var(--ee-margin-micro);
```
nicolad, line 4: I've set default `horizontalMargin` to `micro` in `withEnhance`

assets/src/application/ui/display/icons/svgs/types.ts
```diff
@@ -4,3 +4,6 @@ export interface IconProps {
 }
 
 export type IconSize = 'big' | 'bigger' | 'small' | 'smaller' | 'tiny';
+
+export type HorizontalMargin = 'extreme' | 'huge' | 'big' | 'bigger' | 'small' | 'smaller' | 'tiny' | 'micro' | 'nano';
```
nicolad, line 8: later this can be moved into a more centralized place or derived from smth like `MarginSizes`

assets/src/application/ui/styles/margins.scss
```diff
@@ -0,0 +1,51 @@
+.ee-horizontal-margin {
+    &--nano {
+        margin-left: var(--ee-margin-nano);
+        margin-right: var(--ee-margin-nano);
+    }
```
nicolad, line 5: this can be improved with SASS foreach, but didn't wanted to spend too much time on this now

assets/src/application/ui/display/icons/svgs/types.ts
manzoorwanijk: Not sure if an absolute value may be better here

nicolad: yes, this is a good idea, the reason I've went with these values is that these are matching with the CSS variables, which have absolute values set in `:root`;
need to hear what preference @tn3rb also have

tn3rb: yes please use variables for things like this because then the values can be changed via a theme.
Eventually the goal is to create the ability for users to create their own themes, which will really only be control over a dozen or so parameters, but enough to tweak EE's display to better match their site/preferences.

assets/src/application/ui/forms/espressoForm/renderers/FieldRenderer.tsx
```diff
@@ -40,7 +40,7 @@ const FieldRenderer: React.FC<FieldRendererProps> = (props) => {
 				{label}
 				{info && (
 					<Tooltip placement='right' tooltip={info}>
-						<InfoCircleOutlined />
+						<InfoCircleOutlined color="var(--ee-color-primary-low-contrast)" horizontalMargin="tiny" />
```
tn3rb, line 43: wiat... why is this necessary?

assets/src/application/ui/forms/espressoForm/styles.scss
```diff
@@ -213,11 +213,6 @@
 					width: 100%;
 				}
 			}
-
-			.ee-icon--info-circle-outlined {
-				color: var(--ee-color-primary-low-contrast);
-				margin: 0 var(--ee-margin-tiny);
-			}
```
tn3rb, line 220: i don't understand why this isn't the better solution !?!?

assets/src/application/ui/input/InlineEditInput/style.scss
```diff
@@ -14,5 +14,13 @@
...
 .ee-tabbable-text {
+	border-radius: var(--ee-border-radius-small);
+	border-width: 3px;
 	padding: var(--ee-padding-micro) var(--ee-padding-small);
+
+	&:active,
+	&:focus {
+		border-width: 3px;
+		border-color: var(--ee-color-primary);
+	}
```
nicolad, line 25: context: https://github.com/eventespresso/event-espresso-core/pull/2987#discussion_r444296484

nicolad:
> my apologies but I'm totally not grokking the reason for this change... ??? ...
> 
> > this approach is enhancing patchy-way of applying styles, leading in less maintainable CSS and a bit of guesswork, ie.: how did ee-icon--info-circle-outlined got generated
> 
> Is the issue simply that the css classes are being dynamically generated from the svg names? As long as those don't change then I really don't see what the issue is. IMHO it's perfectly acceptable to not have those hardcoded anywhere, especially if there is a prominent comment somewhere in the icon component code that explains this. Better yet, we could place a `README.md` file in the icon component folder that explains this, as well as provides details on props, etc.
> 
> Personally I think the code you added for generating the icons was a really smart thing to do since it cut out a lot of repetitious boilerplate. Dynamically generated css classes is just a product of that "factory" process and is totally acceptable. I mean, isn't having to "guess" where/how the `ee-icon--info-circle-outlined` css class was generated the exact same thing as having to "guess" where/how the `InfoCircleOutlined` component was generated?
> 
> The only potential issue I can see with using the dynamically generated icons is if we ever want/need to change them down the road. For example, what if we want to change the info icon to have a solid background instead of an outline? This of course wouldn't be any different than if we were using completely hard coded icons, as we would still have to go through and change any of the client code where those icons are used, so this is really kind of a non-issue.
> 
> So I'm back to not grokking why this change is necessary, but maybe I have just completely missed the point (wouldn't be the first time).

The approach I've made with those generated names isn't very maintainable because it's not aligning with the design system way, in which the focus is on providing an API that meets the needs of different scenarios, opposed to having exceptions and applying changes/overwrites locally. 
In this specific example, the difference isn't that sizable because there is only one prop, but I've made this change earlier, than having later a lot of customizations scattered in different `scss` for each icon that needs customization.
I see this as moving low-level CSS properties to a higher-level, where this can be shareable in the future.

Issue #2985, opened 2020-06-22
Inline Edit Inputs Need TabIndex and Tooltips


tn3rb:
### steps to reproduce

- click on "card view" button in one of the entity list filter bars just to place tabbing start position

- start clicking TAB on keyboard and advance through filter bar options and into Entity Card

- tabbing skips right over inline edit inputs like name and description

- now use mouse to hover over inline edit inputs

- no feedback is provided to indicate that these elements are clickable / editable

### solutions

- add tab indexes to ALL inline edit inputs so that they can be accessed via keyboard

- add simple default tooltip like "click to edit" to inline edits and/or change text color to `--ee-input-secondary` on hover

PR #2987, opened 2020-06-23
Added TabbableText; js_only.

nicolad:
- Closes https://github.com/eventespresso/event-espresso-core/issues/2985

## How has this been tested
Peek 2020-06-23 13-34

### DIFF
--- a/assets/src/application/ui/input/InlineEditInput/style.scss
+++ b/assets/src/application/ui/input/InlineEditInput/style.scss
@@ -12,3 +12,7 @@
 		vertical-align: middle;
 	}
 }
+
+.ee-tabbable-text {
+	padding: var(--ee-padding-micro) var(--ee-padding-small);
+}
--- a/assets/src/application/ui/input/InlineEditInput/types.ts
+++ b/assets/src/application/ui/input/InlineEditInput/types.ts
@@ -8,6 +8,6 @@ export interface TextProps extends Omit<InlineEditProps, 'inputType'> {
 }
 
 export interface TabbableTextProps {
-	onRequestEdit: VoidFunction;
+	onClick: VoidFunction;
 	text: string;
 }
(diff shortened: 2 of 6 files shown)


assets/src/application/ui/input/InlineEditInput/types.ts
```diff
@@ -1,8 +1,13 @@
...
-export interface TextAreaProps extends Omit<InlineEditProps, 'inputType'> {}
+export interface TextAreaProps extends Omit<InlineEditProps, 'inputType'> { }
 
 export interface TextProps extends Omit<InlineEditProps, 'inputType'> {
 	fitText?: boolean;
 	tag?: React.ElementType;
 }
+
+export interface TabbableTextProps {
+	onClick: VoidFunction;
```
manzoorwanijk, line 11: `onClick` is too generic. May be name it something very specific - related to what it actually does, like `onEdit`, `onRequestEdit` etc.

nicolad: this is funny, that's what I did in previous commit, but thought it wouldn't be clear why I am passing `onRequestEdit` prop, will update now

assets/src/application/ui/input/InlineEditInput/style.scss
```diff
@@ -12,3 +12,7 @@
 		vertical-align: middle;
 	}
 }
+
+.ee-tabbable-text {
+	padding: var(--ee-padding-micro) var(--ee-padding-small);
+}
```
tn3rb, line 18: needs same `:active` state border styles as other elements

nicolad: Peek 2020-06-23 18-24

nicolad: update applied here: https://github.com/eventespresso/event-espresso-core/pull/2986/commits/e996f49a1134afd0ec1a959cdea39203e26148e1

Issue #2984, opened 2020-06-22
Form Info Icon Tooltips Don't Work on Touch Devices


tn3rb:
### steps to reproduce

- open any entity edit form (or add new entity form)

- find any Info Icon such as the blue icon in the following image:
![ScreenShot_20200622095322](https://user-images.githubusercontent.com/1751030/85337880-28279c00-b496-11ea-9b4e-61edcfb1c46b.png)

- hover over icon to display tooltip

- now use your browser's dev tools to switch to responsive/mobile emulator and choose any device that utilizes a touch screen

- use touch emulator and "click" on the Info Icon

- no tooltip appears

### potential solutions

- would be fantastic if icons could be made to respond to click/touch as well as hover (or maybe convert to click/touch only)

- if click/touch is not possible, then hide info icons on touch/mobile and display help text after/below/beside inputs

### secondary issue

- some styles do not appear in production builds (for example, the icon color)

- compare info icons on local test site using dev build vs NK dev site using production build

PR #2988, opened 2020-06-24
Added clickable tooltips.

nicolad:
- Closes: https://github.com/eventespresso/event-espresso-core/issues/2984

## How has this been tested
Peek 2020-06-24 14-32

### DIFF
--- a/assets/src/application/ui/display/Tooltip/ClickableIconWithTooltip.tsx
+++ b/assets/src/application/ui/display/Tooltip/ClickableIconWithTooltip.tsx
@@ -1,0 +1,25 @@
+import React from 'react';
+import { useDisclosure } from '@chakra-ui/hooks';
+
+import { Tooltip } from '@infraUI/display';
+import { ClickableIconWithTooltipProps } from './types';
+
+import './style.scss';
+
+const ClickableIconWithTooltip: React.FC<ClickableIconWithTooltipProps> = ({ icon: Icon, tooltipText }) => {
+	const { isOpen, onOpen, onClose } = useDisclosure();
+
+	const onClick = () => {
+		return isOpen ? onClose() : onOpen();
+	};
+
+	const icon = <Icon className='ee-clickable-tooltip' onClick={onClick} />;
+
+	return (
+		<Tooltip isOpen={isOpen} tooltip={tooltipText}>
+			{icon}
+		</Tooltip>
+	);
+};
+
+export default ClickableIconWithTooltip;
--- a/assets/src/application/ui/display/Tooltip/ClickableIconWithTooltip.tsx
+++ b/assets/src/application/ui/display/Tooltip/ClickableIconWithTooltip.tsx
@@ -1,4 +1,5 @@
 import React from 'react';
+import classNames from 'classnames';
 import { useDisclosure } from '@chakra-ui/hooks';
 
 import { Tooltip } from '@infraUI/display';
@@ -6,14 +7,15 @@ import { ClickableIconWithTooltipProps } from './types';
 
 import './style.scss';
 
-const ClickableIconWithTooltip: React.FC<ClickableIconWithTooltipProps> = ({ icon: Icon, tooltipText }) => {
+const ClickableIconWithTooltip: React.FC<ClickableIconWithTooltipProps> = ({ icon: Icon, tooltipText, ...props }) => {
 	const { isOpen, onOpen, onClose } = useDisclosure();
+	const className = classNames('ee-clickable-tooltip', props.className);
 
 	const onClick = () => {
 		return isOpen ? onClose() : onOpen();
 	};
 
-	const icon = <Icon className='ee-clickable-tooltip' onClick={onClick} />;
+	const icon = <Icon className={className} onClick={onClick} />;
 
 	return (
 		<Tooltip isOpen={isOpen} tooltip={tooltipText}>
--- a/assets/src/application/ui/display/Tooltip/types.ts
+++ b/assets/src/application/ui/display/Tooltip/types.ts
@@ -1,6 +1,7 @@
-import HelpIcon from '@appDisplay/icons/HelpIcon';
+import { InfoCircleOutlined } from '@appDisplay/icons/svgs';
 
 export interface ClickableIconWithTooltipProps {
-	icon: typeof HelpIcon;
+	className?: string;
+	icon: typeof InfoCircleOutlined;
 	tooltipText: string;
 }
--- a/assets/src/application/ui/display/Tooltip/types.ts
+++ b/assets/src/application/ui/display/Tooltip/types.ts
@@ -1,0 +1,6 @@
+import HelpIcon from '@appDisplay/icons/HelpIcon';
+
+export interface ClickableIconWithTooltipProps {
+	icon: typeof HelpIcon;
+	tooltipText: string;
+}
--- a/assets/src/application/ui/display/confirm/useConfirmationDialog.tsx
+++ b/assets/src/application/ui/display/confirm/useConfirmationDialog.tsx
@@ -1,5 +1,5 @@
 import React, { useCallback, useMemo } from 'react';
-import { useDisclosure } from '@chakra-ui/core';
+import { useDisclosure } from '@chakra-ui/hooks';
 import { __ } from '@wordpress/i18n';
 
 import { AlertDialog } from '@infraUI/display';
--- a/assets/src/application/ui/display/icons/HelpIcon/index.tsx
+++ b/assets/src/application/ui/display/icons/HelpIcon/index.tsx
@@ -1,15 +1,26 @@
 import React, { forwardRef } from 'react';
 import classNames from 'classnames';
 
-import { InfoCircleOutlined } from '../svgs/index'
-import { IconProps } from '../types'
+import { ClickableIconWithTooltip } from '@appDisplay/Tooltip';
+import { InfoCircleOutlined } from '../svgs/index';
+import { IconProps } from '../types';
 
-import './style.scss'
+import './style.scss';
 
-const HelpIcon = forwardRef<IconProps, any>((props, ref) => {
+const HelpIcon = forwardRef<IconProps, any>(({ clickable, tooltipText, ...props }, ref) => {
 	const className = classNames('ee-help-icon', props.className);
 
-	return <InfoCircleOutlined {...props} className={className} ref={ref} />
+	if (clickable) {
+		return (
+			<ClickableIconWithTooltip
+				className={className}
+				icon={InfoCircleOutlined}
+				tooltipText={tooltipText}
+			/>
+		);
+	};
+
+	return <InfoCircleOutlined {...props} className={className} ref={ref} />;
 });
 
 export default HelpIcon;
--- a/assets/src/application/ui/display/icons/HelpIcon/index.tsx
+++ b/assets/src/application/ui/display/icons/HelpIcon/index.tsx
@@ -1,0 +1,15 @@
+import React, { forwardRef } from 'react';
+import classNames from 'classnames';
+
+import { InfoCircleOutlined } from '../svgs/index'
+import { IconProps } from '../types'
+
+import './style.scss'
+
+const HelpIcon = forwardRef<IconProps, any>((props, ref) => {
+	const className = classNames('ee-help-icon', props.className);
+
+	return <InfoCircleOutlined {...props} className={className} ref={ref} />
+});
+
+export default HelpIcon;
--- a/assets/src/application/ui/display/icons/HelpIcon/style.scss
+++ b/assets/src/application/ui/display/icons/HelpIcon/style.scss
@@ -1,0 +1,4 @@
+.ee-help-icon {
+	color: var(--ee-color-primary-low-contrast);
+	margin: 0 var(--ee-margin-tiny);
+}
--- a/assets/src/application/ui/display/withTooltip/style.scss
+++ b/assets/src/application/ui/display/withTooltip/style.scss
@@ -1,4 +1,4 @@
-@import 'application/ui/styles/mixins/breakpoints.scss';
+@import "application/ui/styles/mixins/breakpoints.scss";
 
 :root {
 	--ee-tooltip-background-color: #cdeaff;
@@ -17,7 +17,7 @@
 	position: relative;
 	width: auto;
 	word-spacing: 0.15rem;
-	z-index: 1401;
+	z-index: 100000;
 }
 
 .ee-mobile-help-text {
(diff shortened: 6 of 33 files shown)


assets/src/application/ui/input/InlineEditInput/style.scss
```diff
@@ -12,3 +12,19 @@
...
+	&:hover {
+		color: var(--ee-input-secondary);
+	}
+
+	&:active,
+	&:focus {
+		border-width: 3px;
+		border-color: var(--ee-color-primary);
+	}
+}
```
nicolad, line 30: context: https://github.com/eventespresso/event-espresso-core/pull/2987#discussion_r444296484

assets/src/application/ui/display/withTooltip/style.scss
```diff
@@ -17,7 +17,7 @@
 	position: relative;
 	width: auto;
 	word-spacing: 0.15rem;
-	z-index: 1401;
+	z-index: 100000;
```
nicolad, line 20: I tried different values and only starting with this one it started to appear above the underlying modal.

assets/src/application/ui/display/icons/HelpIcon/style.scss
```diff
@@ -0,0 +1,4 @@
+.ee-help-icon {
+	color: var(--ee-color-primary-low-contrast);
+	margin: 0 var(--ee-margin-tiny);
```
nicolad, line 3: I was able to reproduce the icon w/o the color using `yarn prod`:
image
***
after extracting that icon styles it worked on prod the same as on dev

image

assets/src/application/ui/display/withTooltip/style.scss
nicolad: the proof:

Peek 2020-06-24 16-45

assets/src/domain/eventEditor/ui/datetimes/datesList/actionsMenu/dropdown/DateMainMenu.tsx
```diff
@@ -20,11 +20,11 @@ const DateMainMenu: React.FC<DateMainMenuProps> = ({ datetime }) => {
...
 		? __(
-				'Are you sure you want to permanently delete this datetime? This action is permanent and can not be undone.'
-		  )
+			'Are you sure you want to permanently delete this datetime? This action is permanent and can not be undone.'
+		)
 		: __(
-				'Are you sure you want to move this datetime to the trash? You can "untrash" this datetime later if you need to.'
-		  );
+			'Are you sure you want to move this datetime to the trash? You can "untrash" this datetime later if you need to.'
+		);
```
nicolad, line 27: I don't understand why prettier is behaving this way, I didn't changed any configs locally

package.json
```diff
@@ -54,6 +54,7 @@
 	"dependencies": {
 		"@apollo/react-hooks": "^3.1.5",
 		"@chakra-ui/core": "^0.8.0",
+		"@chakra-ui/hooks": "^1.0.0-next.2",
```
nicolad, line 57: I noticed that in chakra they are using lately this package separately: https://github.com/chakra-ui/chakra-ui/blob/master/packages/tooltip/src/use-tooltip.ts#L1

assets/src/application/ui/display/confirm/useConfirmationDialog.tsx
```diff
@@ -1,5 +1,5 @@
 import React, { useCallback, useMemo } from 'react';
-import { useDisclosure } from '@chakra-ui/core';
+import { useDisclosure } from "@chakra-ui/hooks"
```
manzoorwanijk, line 2: When did we switch to double quotes? 🤔

assets/src/application/ui/display/Tooltip/types.ts
```diff
@@ -0,0 +1,7 @@
+import PopperJs from "popper.js";
```
manzoorwanijk, line 1: This needs to be in deps

assets/src/application/ui/display/icons/HelpIcon/index.tsx
```diff
@@ -0,0 +1,15 @@
...
+
+import './style.scss'
+
+const HelpIcon = forwardRef<IconProps, any>((props, ref) => {
+	const className = classNames('ee-help-icon', props.className);
+
+	return <InfoCircleOutlined {...props} className={className} ref={ref} />
+});
+
+export default HelpIcon;
```
manzoorwanijk, line 15: EOL

assets/src/application/ui/display/confirm/useConfirmationDialog.tsx
nicolad: my prettier is not working as expected, will reinstall VS Code and it's plugins

assets/src/application/ui/display/Tooltip/types.ts
nicolad: chakra depends on this and I was thinking that this is the same

assets/src/application/ui/display/confirm/useConfirmationDialog.tsx
manzoorwanijk: Oh OK

assets/src/application/ui/display/Tooltip/types.ts
nicolad: btw that `right` placement cause button being overlapped and I've removed that, maybe it's better to let `popper` to choose automatically the best placement:

image

nicolad: w/o placement:

image

assets/src/application/ui/display/icons/HelpIcon/index.tsx
nicolad: updated

assets/src/application/ui/display/Tooltip/types.ts
```diff
@@ -0,0 +1,4 @@
+export interface ClickableIconWithTooltipProps {
+    icon: React.ReactNode;
```
manzoorwanijk, line 2: Here is that TS issue. You have set it as `ReactNode` rather than `ComponentType`

nicolad: updated there

nicolad: TS is still showing an error in `ClickableIconWithTooltip`, but we will can deal with that when we will move to monorepo, because there will be some changes related to icons

assets/src/application/ui/display/Tooltip/types.ts
```diff
@@ -0,0 +1,4 @@
+export interface ClickableIconWithTooltipProps {
+	icon: React.ComponentType;
+	tooltipText: string;
+}
```
manzoorwanijk, line 4:
```suggestion
import HelpIcon from '@appDisplay/icons/HelpIcon';

export interface ClickableIconWithTooltipProps {
	icon: typeof HelpIcon;
	tooltipText: string;
}
```

assets/src/application/ui/display/Tooltip/types.ts
```diff
@@ -1,4 +1,6 @@
+import HelpIcon from '@appDisplay/icons/HelpIcon';
+
 export interface ClickableIconWithTooltipProps {
-	icon: React.ComponentType;
+	icon: typeof HelpIcon;
```
nicolad, line 4: interesting, so here other icons also should be added

manzoorwanijk: You can also use `React.ComponentType<IconProps>`

nicolad: Thanks for the suggestion :bow:

assets/src/application/ui/input/InlineEditInput/style.scss
nicolad: already fixed here: https://github.com/eventespresso/event-espresso-core/pull/3002

assets/src/application/ui/display/icons/HelpIcon/index.tsx
```diff
@@ -0,0 +1,15 @@
...
+import { InfoCircleOutlined } from '../svgs/index'
+import { IconProps } from '../types'
+
+import './style.scss'
+
+const HelpIcon = forwardRef<IconProps, any>((props, ref) => {
+	const className = classNames('ee-help-icon', props.className);
+
+	return <InfoCircleOutlined {...props} className={className} ref={ref} />
+});
```
tn3rb, line 13: 👍 👍

assets/src/application/ui/forms/espressoForm/renderers/FieldRenderer.tsx
```diff
@@ -39,9 +39,7 @@ const FieldRenderer: React.FC<FieldRendererProps> = (props) => {
 			<FormLabel htmlFor={props.input.name}>
 				{label}
 				{info && (
-					<Tooltip placement='right' tooltip={info}>
-						<InfoCircleOutlined />
-					</Tooltip>
+					<ClickableIconWithTooltip icon={HelpIcon} tooltipText={info} />
```
tn3rb, line 42: I'm thinking it might be better to convert `ClickableIconWithTooltip` into an HOC and then wrap the export of `HelpIcon` with it.

Then this client code would only need to import and use `HelpIcon` and not have to deal with configuration issues pertaining to `ClickableIconWithTooltip`.

IMHO that will make things more readable and will also make it easier to make changes to the tooltip aspect of `HelpIcon` since it will only require updates to the one file.

But I'm open to debate on this in case there is a good reason to keep things as is.

thoughts?

assets/src/application/ui/input/InlineEditInput/InlineEditText.tsx
```diff
@@ -33,14 +33,14 @@ const Preview: React.FC<InlineEditPreviewProps> = ({ fitText, isEditing, onReque
...
-const InlineEditText: React.FC<TextProps> = ({ fitText = true, placeholder = '', tag: as, ...props }) => {
+const InlineEditText: React.FC<TextProps> = ({ fitText = true, placeholder = '', tag: as, tooltip, ...props }) => {
 	return (
 		<InlineEdit
 			{...props}
 			as={as}
 			inputType='text'
 			placeholder={placeholder}
-			Preview={(previewProps) => <Preview {...previewProps} fitText={fitText} />}
+			Preview={(previewProps) => <Preview {...previewProps} fitText={fitText} tooltip={tooltip} />}
```
tn3rb, line 43: is this prop name on `InlineEdit` supposed to be Capitalized?

should it not be:

```suggestion
			preview={(previewProps) => <Preview {...previewProps} fitText={fitText} tooltip={tooltip} />}
```

or is it because it gets rendered inside `InlineEdit` as if it was an imported component?

[EDIT] ya I jsut saw that the other inputs are doing the same thing, so that must be the case

assets/src/application/ui/forms/espressoForm/renderers/FieldRenderer.tsx
nicolad: Just today @manzoorwanijk pointed out that one of the HOC doesn't have typed return and applying that kind of types is tricky/ complicated. Then I noticed that all of ours HOCs doesn't have return type and created an issue to solve this problem later: https://github.com/eventespresso/barista/projects/1#card-40726511

assets/src/application/ui/input/InlineEditInput/InlineEditText.tsx
nicolad: that is from the type:
```

export interface InlineEditProps extends Partial<ChakraEditableProps>, CommonInputProps<HTMLInputElement> {
	inputType?: InputType;
	Preview?: React.ComponentType<InlineEditPreviewProps>;
}
```

assets/src/application/ui/forms/espressoForm/renderers/FieldRenderer.tsx
tn3rb: really?
I'm seeing this:

```
const withLabel: <P extends withLabelProps>(WrappedComponent: React.ComponentType<P>) => React.ForwardRefExoticComponent<React.PropsWithoutRef<P> & React.RefAttributes<React.ComponentType<P>>>
```

which looks to me like it's returning `React.ForwardRefExoticComponent`

tn3rb: basically the return types are being inferred from:

```js
return React.forwardRef(ForwardedComponentWithTooltip);
```

so the return types are not hard coded, but they are still there nonetheless

nicolad: I think @manzoorwanijk can have something to say here since he originally raised the issue

manzoorwanijk: Some of our HOCs (not all) do not have the return type defined. Though TS infers it most of the time but we should be explicit wherever possible.

manzoorwanijk: If a HOC uses generics and then uses those generics in its underlying types, then it's difficult to set the return type. In such cases, we can leave it to TS to infer.

tn3rb: is `React.ForwardRefExoticComponent` sufficient though or was your original comment because we should be more explicit so that the return types show the specific component being used (ie: via generics)?

assets/src/application/ui/display/Tooltip/ClickableIconWithTooltip.tsx
```diff
@@ -1,19 +1,21 @@
...
 
 import { Tooltip } from '@infraUI/display';
 import { ClickableIconWithTooltipProps } from './types';
 
 import './style.scss';
 
-const ClickableIconWithTooltip: React.FC<ClickableIconWithTooltipProps> = ({ icon: Icon, tooltipText }) => {
+const ClickableIconWithTooltip: React.FC<ClickableIconWithTooltipProps> = ({ icon: Icon, tooltipText, ...props }) => {
 	const { isOpen, onOpen, onClose } = useDisclosure();
+	const className = classNames('ee-clickable-tooltip', props.className);
```
tn3rb, line 12: 👍

assets/src/application/ui/forms/espressoForm/renderers/FieldRenderer.tsx
```diff
@@ -38,9 +37,7 @@ const FieldRenderer: React.FC<FieldRendererProps> = (props) => {
 		<FormControl className={className} isInvalid={Boolean(errorMessage)} isRequired={required}>
 			<FormLabel htmlFor={props.input.name}>
 				{label}
-				{info && (
-					<ClickableIconWithTooltip icon={HelpIcon} tooltipText={info} />
-				)}
+				{info && <HelpIcon clickable tooltipText={info} />}
```
tn3rb, line 40: ya that works

Issue #2996, opened 2020-06-24
Entity Card GUID


tn3rb:
### issue posted by @garthkoyle 

What is this number in the datetimes card `20:U6MjA=` and will it be visible when distributed? 

https://www.screencast.com/t/Jw92Vj47tOOn

tn3rb:
Those are the entity GUIDs (Globally Unique ID)

and yes I was planning on leaving those in place to benefit us with any troubleshooting, but those could be moved to a more discrete location, (like into the Entity Action Menu dropdowns)

sethshoultes:
I agree with moving that data to more discrete location. Maybe you add another button link that shows details about the ticket, such as the ID and GUID:
![Events ‹ North Korea Best Korea — WordPress 2020-06-24 17-35-32(1)](https://user-images.githubusercontent.com/235315/85637769-4ed60600-b641-11ea-86e8-ca7d50c38257.jpg)

tn3rb:
I want to reserve that space in the sidebar for more important things, so we can move that info into the dropdown menu

sethshoultes:
Sounds good.

garthkoyle:
What about putting the GUID inside the hamberger menu?

tn3rb:
yes that's what I meant by:

> we can move that info into the dropdown menu

I had intended the existing location to be both discrete and yet still easily accessible for the benefit of troubleshooting, but I can move it into the dropdown/hamberder menu if its current location is really that bad.

I really don't think it is going to clean up the UI in any dramatic way, by removing it from where it is, but at least if it is moved somewhere with some more screen real estate available we can add labels so that it won't be so mysterious.

garthkoyle:
@tn3rb can you run it by your team and see what they think? cc @Pebblo

Pebblo:
Howdy Brent, can you explain a little how these will help with troubleshooting?

The reason I ask is I'm wondering if it would be better to leave them where they are but only display them if WP_DEBUG is enabled.

Here's what I suspect would happen from some of our users (I think most would simply question what it is and then ignore it or complain about it being distracting 😞 ).

1. Uses new editor and notices the 'funky text' above the cards.
2. Posts up on the forum asking what they are and why they have them. (Gets a response explaining they are the GUID and are there to help troubleshooting)
3. Adds 'i-Break-EE-Plugin' to the site which throws 27million JS errors on our editor and the now the user can't edit the cards (if they display).
[...]

Issue #2992, opened 2020-06-24
Combine "# Sold" & "Reg List" Entity Details


tn3rb:
### issue posted by @garthkoyle 

Some rough thinking here...Could we combine the "# Sold" with the "Reg List" action, so if they click on the number sold they would also get the registration list? Showing the button and the quantity sold could simplify what people might expect from the button and give us a cleaner UI???https://www.screencast.com/t/fb6jGzG6oGg

Perhaps the issue is the definition of "Sold" vs. the "Reg List"?

tn3rb:
Just to clarify things, you are suggesting to:

- remove the reg list icon button link

- make the quantity sold count text into a link that goes to the registration list (ie do what the reg list icon button link did)

I personally don't think that will make a big difference with the UI in terms of space (since those things are not currently crowded) and would actually be worse in terms of UI best practices, and here's why:

For links or actionable items that do something when you click on them, you want the UI to be one of the following things:

- an icon representation on a button - ie: a trash can to represent delete

- a text link where the text clearly states the action - ie: text that says "delete item"

- combination of the above two - ie: button with trash can icon AND "delete item" text

Your suggestion would be to remove the icon button, and make the sold count into a text link (if I'm understanding you correctly) which would mean having an actionable link with text like "123". This is not clear at all and can not benefit from adding a legend item like we have for the icon buttons, because the text will obviously change every time a ticket is sold.

So unless I am misinterpreting your suggestion I am inclined to not make that change since it breaks UI best practices

garthkoyle:
I think you understand correctly. But IMO, seeing a number of **123 Sold** that is blue to indicate a link is obvious that it will give you a registration list of the "sold" registrations/tickets/people.

Pebblo:
Personally I think both of those details/icons are valid and would like to throw in another option, keep both but make sold clickable.

This is what you get when combining the reg list into sold and making it a link: https://monosnap.com/file/YuB7xcylZYJm7P7NjPlxvVEpJzuWmH

So make '### sold' a link which lists only Approved registrations linked to that datetime/ticket.

Leave the reg list icon as is as it links to all registrations linked to the datetime/ticket.

I'm not a big fan of combining reg link (as it is now) into the sold section as the reg list is not a 'sold' list and our users need access to both.

[...]

PR #3022, opened 2020-07-06
UI tweaks.

nicolad:
- Closes: https://github.com/eventespresso/event-espresso-core/issues/2996
- Closes: https://github.com/eventespresso/event-espresso-core/issues/2992


assets/src/application/ui/display/EntityIDs/index.tsx
```diff
@@ -17,15 +16,14 @@ export interface EntityIdProps {
...
 		'ee-align-lft': align === EntityIdAlignment.LEFT,
 		'ee-align-rgt': align === EntityIdAlignment.RIGHT,
 	});
+
 	return (
-		<div className={htmlClass}>
+		<div className={className}>
 			<span className={'ee-entity-dbid'}>{dbid}</span>
-			<span className={'ee-entity-id-separator'}>{':'}</span>
-			<span className={'ee-entity-guid'}>{shortenGuid(guid)}</span>
```
nicolad, line 28: experimented with this approach as well:
image

assets/src/application/ui/display/entityDetailsPanel/EntityDetail.tsx
```diff
@@ -7,10 +7,11 @@ const EntityDetail: React.FC<EntityDetailProps> = ({ render: Component, ...props
 		return <Component {...props} />;
 	}
 	const { id, label, value, className = '' } = props;
+
 	return (
-		<div className={`ee-entity-details-div ${id}-div ${className}`}>
-			<div className={`ee-entity-details-label-div ${id}-label`}>{label}</div>
-			<div className={`ee-entity-details-value-div ${id}-value`}>{value}</div>
+		<div className={`ee-entity-details ${id}-div ${className}`}>
```
nicolad, line 12: not sure what was the intention behind `${id}-div`, but left that there for the time being

assets/src/application/ui/display/entityDetailsPanel/Sold.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+import './style.scss';
+
+interface Props {
+	dbId: EntityDbId;
+	sold?: number;
+	type: 'date' | 'ticket';
+}
+
+const Sold: React.FC<Props> = ({ sold = 0, type, ...props }) => {
```
nicolad, line 21: before going with simple `Sold`, this was named `EntitySoldDetail`, but that prefix and suffix was just plain noise and wasn't aiding in any way

assets/src/application/ui/input/Button/Link.tsx
```diff
@@ -5,20 +5,30 @@ import { Tooltip } from '@infraUI/display';
 import { LinkProps } from './types';
 import './style.scss';
 
-const Link: React.FC<LinkProps> = ({ external, href, icon, tooltip, tooltipProps, ...props }) => {
+const Link: React.FC<LinkProps> = ({ children, external, href, icon, tooltip, tooltipProps, ...props }) => {
 	const className = classNames('ee-btn-base ee-icon-button ee-icon-button--borderless', props.className);
 
-	if (external && icon) {
+	if (external && !icon && tooltip) {
```
nicolad, line 11: my personal preference is avoiding nesting logic, this is why I didn't place this in the conditional below

assets/src/application/ui/input/Button/Link.tsx
```diff
@@ -5,20 +5,20 @@ import { Tooltip } from '@infraUI/display';
 import { LinkProps } from './types';
 import './style.scss';
 
-const Link: React.FC<LinkProps> = ({ external, href, icon, tooltip, tooltipProps, ...props }) => {
+const Link: React.FC<LinkProps> = ({ children, external, href, icon, tooltip, tooltipProps, ...props }) => {
 	const className = classNames('ee-btn-base ee-icon-button ee-icon-button--borderless', props.className);
 
-	if (external && icon) {
+	if (external && tooltip) {
```
tn3rb, line 11: not grokking what `external` is intending to do the way things are set up now...
currently it appears to do nothing but control whether the link renders or not.

I'm wondering if it was maybe originally intended to control the display of the `Link` icon
and would be used to indicate that a link will take the user to another website.

As well, if the link does not have a tooltip then nothing is rendered, which also doesn't seem right.

Maybe this component should be something like this instead:

```js
const Link: React.FC<LinkProps> = ({ children, external, href, icon, tooltip, tooltipProps, ...props }) => {
	const className = classNames('ee-btn-base ee-icon-button ee-icon-button--borderless', props.className);
	const linkOnly = (
		<a href={href} className={className} target='_blank' rel='noopener norefferer'>
			{icon ? icon : children}
			{external ? <LinkIcon /> : null}
		</a>
	);
	if (tooltip) {
		return (
			<Tooltip tooltip={tooltip} {...tooltipProps}>
				{linkOnly}
			</Tooltip>
		);
	}
	return { linkOnly };
};
```

but ya know... done correctly (since that was whipped together quickly)

assets/src/application/ui/display/entityDetailsPanel/Sold.tsx
tn3rb: ya I think that's a nice change 👍

assets/src/application/ui/display/EntityIDs/index.tsx
tn3rb: I had done that too at some point, but I've also had some other feature ideas that may require an extremely prominent spot in the UI like that. For example, I have a branch that adds a `locked` boolean property to the dates and tickets, and displays a lock icon in that same top corner of the sidebar where you put the ID. when toggled on, the lock  prevents those entities from being edited via automated processes like REM. I think it's a good chance that we will end up adding that but I didn't feel it was critical for REM + EDTR and so I left it out for now.

assets/src/application/ui/display/entityDetailsPanel/Sold.tsx
```diff
@@ -0,0 +1,44 @@
...
+	} = useConfig();
+	const adminUrl = getAdminUrl({ adminSiteUrl: admin, page: ADMIN_ROUTES.REGISTRATIONS });
+	const eventId = useEventId();
+	const dbId = type === 'date' ? { datetime_id: props.dbId } : { ticket_id: props.dbId };
+
+	const regListUrl = addQueryArgs(adminUrl, {
+		event_id: eventId,
+		return: 'edit',
+		...{ dbId },
+	});
```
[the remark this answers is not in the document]
nicolad: thank you for suggestion

assets/src/application/ui/display/entityDetailsPanel/style.scss
```diff
@@ -57,47 +57,54 @@
...
+				text-align: center;
+				vertical-align: middle;
+				width: 25%;
+			}
+
+			.entity-sold {
+				align-items: center;
+				display: inline-flex;
+				justify-content: center;
+				margin-bottom: -10px;
```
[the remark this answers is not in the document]
nicolad: removed, thx

assets/src/application/ui/display/entityDetailsPanel/style.scss
```diff
@@ -57,47 +57,54 @@
...
-	.ee-entity-details-value-div {
-		font-weight: bold;
-		margin-bottom: var(--ee-border-radius-default);
 
-		.dashicon {
-			margin: 0;
-			position: relative;
+		&__value {
+			font-weight: bold;
+			margin-bottom: var(--ee-border-radius-default);
```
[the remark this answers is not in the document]
nicolad: done: https://github.com/eventespresso/event-espresso-core/pull/3022/commits/aba297bc7aa8f028724f71b7b4cc9d755c7da6eb
thanks

assets/src/application/ui/input/Button/Link.tsx
nicolad: originally, I thought that we might abstract the `Link` from `react-router` also in this component, but later found out that we won't use that library, and that's why I left that comment that it might be extended later, because I wasn't sure what other cases it should also handle

nicolad: since that anchor element has `target='_blank' rel='noopener norefferer'` that means that it always will be external, so I've removed external prop;
your suggestion looks great, updated with some changes: https://github.com/eventespresso/event-espresso-core/pull/3022/commits/f45bedc09117fec7979a36cb30e445b9fee952bd

nicolad: I did not added this
`{external ? <LinkIcon /> : null}`
because that icon needs to be customized, as in `RegistrationsLink` with `Groups` icon

assets/src/application/ui/display/entityDetailsPanel/style.scss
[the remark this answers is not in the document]
nicolad: `margin-bottom: -0.4rem` is a bit too close to the text from the bottom:

image
***
0.3 is still too close:
image
***
0.2 is ok, IMHO
image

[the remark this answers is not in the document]
tn3rb: great... I should have zoomed in more 👓

assets/src/application/ui/input/Button/Link.tsx
tn3rb: > since that anchor element has target='_blank' rel='noopener norefferer' that means that it always will be external

well `target='_blank'` just means the link will open in another tab or window, but my understanding of that "external" icon that appears after some links ios that it is supposed to indicate that the link takes you to another site altogether


> I did not added this `{external ? <LinkIcon /> : null}`

ya no worries, I didn't expect those suggested changes to be the final code and fully expected additional changes and/or fixes

nicolad: > "external" icon that appears after some links ios that it is supposed to indicate that the link takes you to another site altogether
yes, you are right, but currently we are not covering cases when we open a link of another site, nor do we have that external icon I posted yesterday, that's why I remove `external` prop for the moment to avoid any confusion and inconsistencies

PR #3026, opened 2020-07-07
i18n for blocks and plugins page

### LINKED ISSUE #3025: i18n doesn't seem to work for blocks and wpPluginsPage
May be this should be handled after #2858

There is an issue that needs to be dealt with. It seems like `i18n` functions will not work without text domain in blocks. So it's better to create a wrapper (or barista package) that adds the text domain.

manzoorwanijk:
This PR wire up i18n for blocks and plugins page domain.

Closes #3025

### DIFF
--- a/assets/src/domain/blocks/i18nInit.ts
+++ b/assets/src/domain/blocks/i18nInit.ts
@@ -1,0 +1,5 @@
+import { setLocaleData } from '@wordpress/i18n';
+
+// init i18n for all blocks
+const localeData = window?.eventEspressoData?.i18n;
+setLocaleData(localeData, 'event_espresso');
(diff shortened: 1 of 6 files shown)


assets/src/domain/blocks/i18nInit.ts
```diff
@@ -0,0 +1,4 @@
+import { setLocaleData } from '@wordpress/i18n';
+
+const localeData = window?.eventEspressoData?.i18n;
+setLocaleData(localeData, 'event_espresso');
```
manzoorwanijk, line 4: In blocks, we need to add text-domain for i18n in order to differentiate between text domains because there are other plugins/themes who register their i18n text domains.

nicolad: cool, didn't knew this

Issue #2989, opened 2020-06-24
Datetime Sorting Lost on Update.


tn3rb:
### issues posted by @Pebblo 

My datetimes: https://monosnap.com/file/dRWRXJcUACwhFqLFBVPNSIGaCcEdv3

Sorted datetimes:
https://monosnap.com/file/f1P9xa4CzxnsYrr4x5P9oNncGvfN52

- clicking on 'enable sorting'
- setting my order
- switching back to card view

Update the event and the sorting is lost (It goes back to the first screenshot).

EDIT - Same issue with Ticket sorting.

tn3rb:
unfortunately at the moment, there is no mechanism for saving your personal filter bar preferences.

we **could** change the entity lists to default to custom order sorting instead of using start dates

Pebblo:
Just to note, I'm not concerned about the filter bar preferences here, in fact, I didn't even set any filters for this other than to enable custom sorting which automatically switches to table view.

However, the custom sorting I've set by dragging the datetimes into the order I want them in should take priority over any default ordering. If I drag them to be ordered Datetime 1, 2, 3 that would be what I expect them to load in when I next open/updated the event regardless.

tn3rb:
> If I drag them to be ordered Datetime 1, 2, 3 that would be what I expect them to load in when I next open/updated the event regardless.

yes that will require the change that I suggested.

@nicolad 

let's change the default sort order for the entity lists to custom order plz

nicolad:
sorting can be saved now in session storage because in the meantime [this reducer](https://github.com/eventespresso/barista/pull/231/files#diff-1bea31ecf27216cb5368db9e0dcc89a2be35e2dd55385b2acd22ebc5a8aa3c28R6) was added

PR #3028, opened 2020-07-07
Changed default entity list sort order.

nicolad:
- Closes: https://github.com/eventespresso/event-espresso-core/issues/2989

## How has this been tested
Peek 2020-07-07 17-16


nicolad:
was thinking to avoid duplication and to have default order set in `useEntityListFilterStateManager`, but that isn't that trivial to change;
any thoughts?

tn3rb:
> any thoughts?

nope I think your solution is the correct one

PR #3037, opened 2020-07-09
New Date Picker

tn3rb:
In https://github.com/eventespresso/event-espresso-core/projects/10#card-41154086 Tony documented some issues with the existing datepicker. I've been loathing that Uber datepicker for some time so all it took was any excuse to swap it out with something else.

The good news is that our use of an adapter layer made the swap effortless, on the react/js side of things at least (styling the datepicker to match our UI took most of the time)

I tackled this on my own because I didn't want to subject you guys to yet ANOTHER datepicker change 😆 

Here's what things look like atm:

ScreenShot_20200708181318

datepicker in action:

ScreenShot_20200708181400

timepicker in action:

ScreenShot_20200708181419

this one appears to work a little better with the keyboard.

I'm going to work some more on this, because I haven't fully explored all of the options the libraries have available, which are here:

- https://github.com/wojtekmaj/react-date-picker
- https://github.com/wojtekmaj/react-time-picker

I've also added 

- https://github.com/wojtekmaj/react-datetimerange-picker

and plan on adding a button in the entity card sidebars where the timezone info button is now (and move that into the edit forms) so that dates can be edited directly from the cards without having to go through the entire edit form.


assets/src/application/ui/display/icons/svgs/clockOutlined.svg
```diff
@@ -0,0 +1,40 @@
...
+	height="1.25em"
+	width="1.25em"
+>
+	<path d="M12,0A12,12,0,1,0,24,12,12.014,12.014,0,0,0,12,0Zm0,22A10,10,0,1,1,22,12,10.011,10.011,0,0,1,12,22Z">
+	</path>
+	<path
+		d="M17.134,15.81,12.5,11.561V6.5a1,1,0,0,0-2,0V12a1,1,0,0,0,.324.738l4.959,4.545a1.01,1.01,0,0,0,1.413-.061A1,1,0,0,0,17.134,15.81Z"
+	>
+	</path>
+</svg>
```
nicolad, line 40: nice

assets/src/infrastructure/ui/inputs/dateTime/Datepicker/style.scss
```diff
@@ -1,32 +1,216 @@
...
+
+				&__divider {
+					color: var(--ee-color-grey-10);
+					font-size: var(--ee-font-size-big);
+					font-weight: 400;
+					margin: 0 var(--ee-margin-micro);
+					padding: 0;
+				}
+
+				// the following style rules do the same thing, but need to be kept separate, else they no workie
```
nicolad, line 72: :+1:

assets/src/infrastructure/ui/inputs/dateTime/Datepicker/style.scss
```diff
@@ -1,32 +1,216 @@
...
+				&:active {
+					@extend .ee-date-picker-active;
+				}
+
+				&:focus {
+					@extend .ee-date-picker-focus;
+				}
+
+				&:hover {
+					@extend .ee-date-picker-hover;
```
nicolad, line 91: nice

assets/src/infrastructure/ui/inputs/dateTime/Datepicker/style.scss
```diff
@@ -1,32 +1,216 @@
...
+					border: 5px solid var(--ee-background-color);
+					box-shadow: var(--ee-box-shadow-big);
+					font-family: Montserrat, 'Segoe UI', Tahoma, sans-serif;
+					font-size: var(--ee-font-size-default);
+
+					&__navigation {
+						height: var(--ee-icon-button-size);
+						line-height: var(--ee-icon-button-size);
+
+						&__arrow {
```
nicolad, line 146: cool

assets/src/infrastructure/ui/inputs/dateTime/Datepicker/style.scss
```diff
@@ -1,32 +1,216 @@
...
+								margin: 0;
+								width: var(--ee-icon-button-size);
+
+								&--neighboringMonth {
+									color: var(--ee-default-text-color-low-contrast);
+								}
+
+								&--weekend {
+									color: var(--ee-color-primary-low-contrast);
+								}
```
nicolad, line 189: nice that they followed BEM

assets/src/infrastructure/ui/inputs/dateTime/Timepicker/index.tsx
```diff
@@ -1,32 +1,46 @@
...
-import { TimepickerProps } from '../types';
-import withBaseProvider from '../withBaseProvider';
+import { CloseOutlined, ClockOutlined } from '@appDisplay/icons/svgs';
+import { TimePickerProps } from '../types';
 
 import './style.scss';
 
-const Timepicker: React.FC<TimepickerProps> = ({ onChange, onChangeValue, ...props }) => {
+const Timepicker: React.FC<TimePickerProps> = ({ onChange, onChangeValue, ...props }) => {
+	const newDate = props?.value;
```
nicolad, line 12: maybe ?
```suggestion
	const newDate = props?.value && new Date(props.value);
```

assets/src/infrastructure/ui/inputs/dateTime/Timepicker/index.tsx
```diff
@@ -1,32 +1,46 @@
...
-	const onChangeHandler: TimepickerProps['onChange'] = useCallback(
-		(date) => {
+	const onChangeHandler: TimePickerProps['onChange'] = useCallback(
+		(time) => {
+			// incoming value from timepicker is 24hr time like "17:00"
+			const timeParts: string[] = time.split(':');
+			const hours = parseInt(timeParts[0]);
+			const minutes = parseInt(timeParts[1]);
+			newDate.setHours(hours);
+			newDate.setMinutes(minutes);
```
nicolad, line 21: as it is now this will mutate `value` and will cause re-render

assets/src/infrastructure/ui/inputs/dateTime/Timepicker/style.scss
```diff
@@ -1,26 +1,158 @@
...
+					color: var(--ee-text-on-primary);
+				}
+				select::selection {
+					background-color: var(--ee-color-primary-low-contrast);
+					color: var(--ee-text-on-primary);
+				}
+				select::-moz-selection {
+					background-color: var(--ee-color-primary-low-contrast);
+					color: var(--ee-text-on-primary);
+				}
```
nicolad, line 106: image
 

```suggestion
input[type='number']::selection,
input[type='number']::-moz-selection,
select::selection,
select::-moz-selection {
  background-color: red;
	color: green;
}
				 

```

package.json
```diff
@@ -60,6 +60,7 @@
 		"@emotion/core": "^10.0.28",
 		"@emotion/styled": "^10.0.27",
 		"@typeform/embed": "^0.22.0",
+		"@wojtekmaj/react-datetimerange-picker": "^3.0.0",
```
nicolad, line 63: if we don't use baseui for anything else than date and timepicker maybe we should remove from deps?

assets/src/infrastructure/ui/inputs/dateTime/Timepicker/index.tsx
```diff
@@ -1,32 +1,46 @@
...
+	const newDate = props?.value;
 	const className = classNames('ee-input-base-wrapper ee-time-picker', props.className);
-	const onChangeHandler: TimepickerProps['onChange'] = useCallback(
-		(date) => {
+	const onChangeHandler: TimePickerProps['onChange'] = useCallback(
+		(time) => {
+			// incoming value from timepicker is 24hr time like "17:00"
+			const timeParts: string[] = time.split(':');
+			const hours = parseInt(timeParts[0]);
+			const minutes = parseInt(timeParts[1]);
```
manzoorwanijk, line 19: It may not be good to use `.split()`, it's better to use `parse` from `date-fns` and pass the given format from `constants` package.

manzoorwanijk: Here you go
image

manzoorwanijk: Click on that close icon on TimePicker to see that error.

assets/src/infrastructure/ui/inputs/dateTime/Datepicker/index.tsx
```diff
@@ -1,35 +1,40 @@
...
 		[onChange, onChangeValue]
 	);
 
-	const dateValue = useMemoStringify([value]);
-
 	return (
 		<div className={className}>
-			<BaseUIDatepicker {...props} onChange={onChangeHandler} value={dateValue} />
+			<ReactDatepicker
+				format={'y-MM-dd'}
```
manzoorwanijk, line 28: better to use a format from `constants` package.

assets/src/infrastructure/ui/inputs/dateTime/Datepicker/style.scss
tn3rb: copied from work you did 😄

assets/src/infrastructure/ui/inputs/dateTime/Datepicker/style.scss
tn3rb: kinda sorta... they use multiple elements in the class names though, like:

```
react-calendar__month-view__days__day
```

so might be more accurate to say they use BEEEEEM 😆

assets/src/infrastructure/ui/inputs/dateTime/Timepicker/index.tsx
tn3rb: thnx done

assets/src/infrastructure/ui/inputs/dateTime/Timepicker/style.scss
tn3rb: doesn't work that way, that's why I added this comment:

```
the following style rules do the same thing, but need to be kept separate, else they no workie
```

see this article for an explanation (near the end of the article):
https://css-tricks.com/overriding-the-default-text-selection-color-with-css/

package.json
tn3rb: oh right, thnx 👍

assets/src/infrastructure/ui/inputs/dateTime/Timepicker/style.scss
nicolad: oh, I see it's related to those prefixes, later we can improve this with some mixins:
https://css-tricks.com/snippets/sass/mixin-prefix-properties/
***
not related to this PR

assets/src/infrastructure/ui/inputs/dateTime/Timepicker/index.tsx
tn3rb: good catch thnx 👍

assets/src/infrastructure/ui/inputs/dateTime/Datepicker/index.tsx
tn3rb: kk will do

Issue #3042, opened 2020-07-23
Make 'Edit Datetime/Ticket dates' button icon more obvious?


Pebblo:
I completely missed the fact that the calendar icon under the date was a button that allowed you to edit the date/time of the datetime/ticket.

![Entity Sidebar edit date button](https://user-images.githubusercontent.com/1751030/88406233-3dd5fd00-cd85-11ea-9696-9363a8ec356b.png)

I was about to create a ticket requesting something like clicking the date took you to the 'full' editor for easier access or something and just happened to hover the mouse over that icon. 

I really like having those details editable there but given the fact that I didn't even realise it was a button, should we make that a little more obvious?

Or maybe add that icon to the legend?

tn3rb:
@nicolad 

any ideas about how to do this without wrecking the existing UI design too much?

potential solutions I can think of:

- add a border to that button (may want to consider that for ALL icons for consistency then)

- change ALL button backgrounds to default background color: `var(--ee-background-color)`

- change ALL button backgrounds to default button background color: `var(--ee-button-background)`

- change ALL button backgrounds to a semi-opaque background color like `rgba( 192, 192, 192, 0.25)` which would potentially work on ALL background colors whether dark or light. We could simply change the `--ee-button-background` var to this value 


I kinda like that last option, because it requires the least changes and works ok with the existing styles.

Here's a screenie after haxxoring things via dev tools:

![ScreenShot_20200724085134](https://user-images.githubusercontent.com/1751030/88410408-34e82a00-cd8b-11ea-8b3f-3dffc540c062.png)

I'm open to other ideas though

Pebblo:
Noice! This is muchos better.

Thanks guys.

PR #3051, opened 2020-07-27
Adjusted EditDateRangeButton noticeability.

nicolad:
Closes: https://github.com/eventespresso/event-espresso-core/issues/3042

### DIFF
--- a/assets/src/application/ui/calendars/dateDisplay/EditDateRangeButton.tsx
+++ b/assets/src/application/ui/calendars/dateDisplay/EditDateRangeButton.tsx
@@ -17,6 +17,7 @@ const EditDateRangeButton: React.FC<EditDateButtonProps> = ({ header, onEditHand
 		},
 		[onClose, onEditHandler]
 	);
+
 	return (
 		<Popover
 			className={'ee-edit-calendar-date-range'}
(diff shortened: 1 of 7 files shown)


nicolad:
I tried to add border to the button, but it still feels a bit awkward:

image

nicolad:
Then I thought, that maybe the problem is that the button is placed in the sidebar where all the elements are read-only and this is the only exception and thought that maybe it should be located somewhere where user will intuitively expect this to be - near the other clickable buttons.
So I've tried to place that near the other buttons from `DateDetailsPanel`.
***
image

***
this is an experiment, so I need more feedback

tn3rb:
code looks good but I want to keep the button for editing the dates as close to the dates as possible, ie: keep it in the sidebar. Writing out that comment makes me realize that I should have swapped the buttons location with the date status text label so that again, it's as close as it can be to the thing it is editing.

I just don't think it will be as intuitive for people when it is separated from the date and pretty certain there will be some ppl that won't realize what date they are editing if the button is in the entity details panel. Having the button **immediately** below the date it is editing is a little bit more intuitive, ie: here's a date, and a button for editing it.

Next time plz just do a quick mock-up for an idea like this instead of making code changes and pushing a PR, ie: just use your browser dev tools to cut the code for the button out of the sidebar, then paste it into the new location, do whatever other quick edits are needed, then take a screen shot. Then you've only spent 5 minutes or so.

We'll keep this PR open for now and I'll monkey around with the button location and style and see if I can get something that I'll be happy with and then maybe let support choose if I'm still unsure.

nicolad:
Obsolete due: https://github.com/eventespresso/event-espresso-core/pull/3056

Issue #3045, opened 2020-07-23
Datetime/Ticket with titles longer than 30 chars vanish in table view


Pebblo:
To reproduce create a Datetime or Ticket with a title longer than 30 chars.

I created them in card view and switched but as an example, this vanishes from the table view:
>How long do we need to make the ticket titles

This does not:
>How long do we need to make th

The do however appear if you enable sorting.

##Screenshots

Missing titles:
![tickets with long titles in table view](https://user-images.githubusercontent.com/1751030/88406575-b937ae80-cd85-11ea-88b9-a672cef0e15e.png)
(Ticket 11 and 12 have long titles)

Enable sorting:
![tickets with long titles in sorting view](https://user-images.githubusercontent.com/1751030/88406718-ebe1a700-cd85-11ea-8961-a2c0ab39eec1.png)

In card view, I removed the 'e' from the end of title 12:
![tickets with long titles in table view 2](https://user-images.githubusercontent.com/1751030/88406731-f13ef180-cd85-11ea-838f-16298f449eb8.png)

PR #3052, opened 2020-07-27
Table view: fixed name field overflow issue.

nicolad:
- Closes: https://github.com/eventespresso/event-espresso-core/issues/3045


nicolad:
image

nicolad:
Longer titles do not break the layout, do we need to limit the visible length anyway?

image

assets/src/application/ui/input/InlineEditInput/types.ts
```diff
@@ -1,8 +1,11 @@
 import { InlineEditProps } from '@infraUI/inputs';
 
-export interface TextAreaProps extends Omit<InlineEditProps, 'inputType'> { tooltip?: string; }
+export interface TextAreaProps extends Omit<InlineEditProps, 'inputType'> {
+	tooltip?: string;
+}
 
 export interface TextProps extends Omit<InlineEditProps, 'inputType'> {
+	clampable?: boolean;
```
manzoorwanijk, line 8: 🤔

nicolad: I also wasn't sure what is the best name, but check it before applying: https://en.wiktionary.org/wiki/clampable

tn3rb: what does the 'clamp" property on the `Dotdotdot` component even do?
does that restrict the line length or the number of lines? (i'm guessing the latter)

if it simply cuts the line length, then maybe `truncate` would be a better name, but I suspect that it does more than that.

...

Just looked at the code for the `Dotdotdot` component and ya, the `clamp` property controls the number of lines displayed. 

What might be a better solution is instead of doing this:

```js
	if (clampable && value.length > 25) {
		return <Dotdotdot clamp={2}>{textInput}</Dotdotdot>;
	}
```

do something like:

```js
	if (lineCount && value.length > 25) {
		return <Dotdotdot clamp={lineCount}>{textInput}</Dotdotdot>;
	}
```

so instead of having a `clampable` boolean property we have a `lineCount` integer property that allows for greater control over how the text is displayed, which might be needed for other use cases.

@nicolad 

another thing that might be a good idea is to extract and consolidate the logic for the `Preview` component that is used in the `InlineEditText` and `InlineEditTextArea` components and then import back into those. Then this fix for number of lines displayed can cover both use cases.

nicolad: thx for the feedback, updated

Issue #3041, opened 2020-07-21
Use a different time input?


tn3rb:
08:00 is invalid (option for 24hour?)

Backspace doesn't work so appears as though the field isn't working.

All times must be selected.
Type 7:00 PM and it'll appear valid, it shows the time below for you to select. Just press enter and nothing happens, tab away and the time will switch back to whatever was set before. 

To select a time you enter it and either click the exact same time you just entered or press enter. We've had many users commenting on the current editor and the fact that they prefer to enter dates/times manually which is fine if they want to but with the new EDTR is frustrating.

tn3rb:
@Pebblo 

I've replaced the date and time pickers from Uber with some new ones that hopefully work better. I really didn't like how those other ones worked either so was happy to swap them out. That said, there's not really a huge selection of date and time pickers available that work well, are accessible, and compatible with the libraries we are using, so hopefully these ones will fair better than the last ones.

I've pushed the changes to the `EDTR/production` branch which you can try out locally, or via my dev site: https://brent.eventespresso.com/wp-admin/admin.php?page=espresso_events

Pebblo:
Thanks, Brent.

I think this new one fits better! It does have its own little quirks but at least doesn't appear broken like before (backspace doing nothing and then times getting lost when you typed it in and didn't select it).

Times seem fine, small gripe is that I don't like the clock face popup and think we could get away with nothing at all there but that's more personal preference really. 

I did find an issue that I think will need some attention on the datepicker though:

Edit a datetime and click on the Start Date month....

Select a month and then select an option on the datepicker (notice how the datepicker updates as you select a month).

Now click the month dropdown again, selecting another month, this time the datepicker itself doesn't update with your selection and shows whatever was set last time.

---

The month dropdown appears over the date picker, I suspect this is expected and is to give easy access to changing the months? Not a big issue but I think it looks a little funky like this:

![image](https://user-images.githubusercontent.com/4345417/88289938-e9337300-cced-11ea-8092-dbd89cb54969.png)

tn3rb:
it may be possible to hide the clock popup, as it really does seem pointless doesn't it?

those other issues are internal to this new datepicker component and we won't have any control over fixing them. That said, I can see if there is something funny about the way I've configured it that is causing the problems.

With regards to the month options appearing over top of the calendar... which did you click on last? If you had just clicked on the month selector, wouldn't you expect the month options to appear above the calendar?
[...]

PR #3053, opened 2020-07-28
DateTimeRangePicker improvements.

nicolad:
- Closes: https://github.com/eventespresso/event-espresso-core/issues/3041

### DIFF
--- a/assets/src/application/ui/styles/mixins/screen-reader.scss
+++ b/assets/src/application/ui/styles/mixins/screen-reader.scss
@@ -1,0 +1,28 @@
+// stylelint-disable declaration-no-important
+
+// Only display content to screen readers
+//
+// See: https://a11yproject.com/posts/how-to-hide-content/
+// See: https://hugogiraudel.com/2016/10/13/css-hide-and-seek/
+
+@mixin visually-hidden {
+	position: absolute !important;
+	width: 1px !important;
+	height: 1px !important;
+	padding: 0 !important;
+	margin: -1px !important; // Fix for https://github.com/twbs/bootstrap/issues/25686
+	overflow: hidden !important;
+	clip: rect(0, 0, 0, 0) !important;
+	white-space: nowrap !important;
+	border: 0 !important;
+}
+
+// Use to only display content when it's focused.
+//
+// Useful for "Skip to main content" links; see https://www.w3.org/TR/2013/NOTE-WCAG20-TECHS-20130905/G1
+
+@mixin visually-hidden-focusable {
+	&:not(:focus) {
+		@include visually-hidden();
+	}
+}
--- a/assets/src/infrastructure/ui/inputs/dateTime/DateTimeRangePicker/index.tsx
+++ b/assets/src/infrastructure/ui/inputs/dateTime/DateTimeRangePicker/index.tsx
@@ -4,7 +4,6 @@ import { default as ReactDateTimeRangePicker } from '@wojtekmaj/react-datetimera
 import { __ } from '@wordpress/i18n';
 
 import { DateTimeRangePickerProps } from '../types';
-import { ClockOutlined } from '@appDisplay/icons/svgs';
 import useConfig from '@appServices/config/useConfig';
 import { convertWordPressDateFormat, convertWordPressTimeFormat } from '../utilities';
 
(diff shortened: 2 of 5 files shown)


nicolad:
The problem occurs only when month picker is displayed as a dropdown/ select, because when it's showed as text, it's working fine:

Peek 2020-07-28 13-06

nicolad:
The hacky fix I've applied is hiding select's options and working with month selector from the top section:

Peek 2020-07-28 13-08

assets/src/infrastructure/ui/inputs/dateTime/DateTimeRangePicker/style.scss
```diff
@@ -0,0 +1,11 @@
...
+	.react-datetimerange-picker {
+		&__inputGroup {
+			&__month {
+				option {
+					display: none;
+				}
+			}
+		}
+	}
+}
```
tn3rb, line 11: maybe we should use the css tricks for hiding content except for screen readers, like:

```css
      .sr-only {
            position: absolute;
            width: 1px;
            height: 1px;
            padding: 0;
            margin: -1px;
            overflow: hidden;
            clip: rect(0, 0, 0, 0);
            white-space: nowrap; /* added line */
            border: 0;
      }
```

tn3rb:
i also noticed in the screenies you posted that when you select a month via the calendar, the month options look **REALLY** big... can you tighten those up a bit plz? looks like the widths are ok but the heights are too big. Would be nice if the overall calendar modal size doesn't change much when switching between month selction mode and day selection mode (regular calendar)

assets/src/infrastructure/ui/inputs/dateTime/DateTimeRangePicker/style.scss
```diff
@@ -1,6 +1,16 @@
 @import 'application/ui/styles/mixins/screen-reader.scss';
 
 .ee-date-time-range-picker.ee-date-time-range-picker {
+	.react-calendar {
+		&__year-view {
+			.react-calendar {
+				&__tile {
+					padding: var(--ee-padding-small) var(--ee-padding-smaller);
```
nicolad, line 8: image

nicolad: image

nicolad: image

assets/src/infrastructure/ui/inputs/dateTime/DateTimeRangePicker/style.scss
nicolad: @tn3rb I've tried the approach you've posted above and the latest bootstrap mixin to visually hidden, but got no luck;
also, tried to find some insights on stackoverflow and such and it doesn't seem smth straightforward to deal with in this case of `select` element and the `a11y` implications :disappointed: 
since we don't have much control through the API, an ultimate solution might be to clone and adapt `react-datetimerange-picker`, but that might fall outside of the scope of this PR

assets/src/application/ui/styles/mixins/screen-reader.scss
```diff
@@ -0,0 +1,28 @@
...
+	position: absolute !important;
+	width: 1px !important;
+	height: 1px !important;
+	padding: 0 !important;
+	margin: -1px !important; // Fix for https://github.com/twbs/bootstrap/issues/25686
+	overflow: hidden !important;
+	clip: rect(0, 0, 0, 0) !important;
+	white-space: nowrap !important;
+	border: 0 !important;
+}
```
tn3rb, line 18: ya this will be handy 👍

assets/src/infrastructure/ui/inputs/dateTime/DateTimeRangePicker/index.tsx
```diff
@@ -4,34 +4,28 @@ import { default as ReactDateTimeRangePicker } from '@wojtekmaj/react-datetimera
...
-	...props
-}) => {
+const DateTimeRangePicker: React.FC<DateTimeRangePickerProps> = ({ endDate, onChange, startDate, ...props }) => {
 	const {
 		dateTimeFormats: { dateTimeFormat },
 		locale: { user },
 	} = useConfig();
 
-	// convert date format to accepatble values for react-datetimerange-picker
+	// convert date format to acceptable values for react-datetimerange-picker
```
tn3rb, line 19: 👍  that previous spelling was literally NOT acceptable 😆

assets/src/infrastructure/ui/inputs/dateTime/DateTimeRangePicker/style.scss
tn3rb: the month picker looks MUCH better

PR #3059, opened 2020-07-31
Fix date validation bug

### LINKED ISSUE #3058: Date and time validation bug
While working on REM validation in barista, I realized that there is a bug in EDTR date and ticket dates validation bug.
Steps:
- Add or Edit a date or ticket
- Select Start and End dates on same date of two different months (e.g. start 15th August and end 15th September)
- Set End time less than start time

You will see the validation error even though it should not be there.

manzoorwanijk:
This PR fixes date validation bug in entity edit forms

Closes #3058

### DIFF
--- a/assets/src/domain/shared/services/utils/dateAndTime.ts
+++ b/assets/src/domain/shared/services/utils/dateAndTime.ts
@@ -4,7 +4,19 @@ import {
 	DEFAULT_DATE_FORMAT,
 	DEFAULT_TIME_FORMAT,
 } from '@appConstants/dateFnsFormats';
-import { formatISO, isValid, parse, parseISO, format } from 'date-fns';
+import {
+	formatISO,
+	isValid,
+	parse,
+	parseISO,
+	format,
+	getYear,
+	getMonth,
+	getDate,
+	setYear,
+	setMonth,
+	setDate,
+} from 'date-fns';
 import { __ } from '@wordpress/i18n';
 import * as yup from 'yup';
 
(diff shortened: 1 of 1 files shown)


assets/src/domain/shared/services/utils/dateAndTime.ts
```diff
@@ -60,6 +72,33 @@ const transformTime = (value: any, originalValue: any): Date => {
 	return transformDateTime(value, originalValue, MOMENT_TIME_FORMAT);
 };
 
+/**
+ * sets year, month and date for a time value from the given date object.
+ */
+export const updateTimeFromDate = (time: Date, date: Date): Date => {
+	// clone to avoid mutating the actual object
```
nicolad, line 79: perfect

nicolad: not only it prevents obsolete re-renders, but this way some of the unexpected bugs are prevented :+1:

assets/src/domain/shared/services/utils/dateAndTime.ts
```diff
@@ -68,33 +107,22 @@ export const dateAndTimeSchema = yup.object({
...
-							__('End Date & Time must be set later than the Start Date & Time')
-						);
-					}
-				}
-				return schema;
-			}
-		),
+		.when(['startTime', 'endDate'], (startTime: Date, endDate: Date, schema: yup.DateSchema) => {
+			return transformTimeByDate(endDate, schema).min(startTime, dateErrorMessage);
+		}),
```
nicolad, line 125: great job

PR #3069, opened 2020-08-03
Don't Apply Default Taxes to Default Free Tickets

### LINKED ISSUE #3047: Don't apply taxes to tickets by default
Taxes are applied to tickets by default, so even if you leave the price as 0.00 and publish you'll see the `*` and `* price includes taxes` on the ticket selector.

To reproduce, create a new event, set the title and datetime title and ticket title as you would.

Then view the event.

The legacy editor has taxes disabled by default so I'm guessing those shouldn't be applied here.

Also, should free tickets ever show as being taxable?

tn3rb:
closes #3047 

this PR checks that a default ticket is not free before applying taxes, even if those are marked as a default price

### DIFF
--- a/core/domain/services/admin/entities/DefaultPrices.php
+++ b/core/domain/services/admin/entities/DefaultPrices.php
@@ -60,7 +60,9 @@ class DefaultPrices implements DefaultEntityGeneratorInterface
         if (! $entity instanceof EE_Ticket) {
             throw new InvalidEntityException($entity, 'EE_Ticket');
         }
+        $taxes = [];
         $new_prices = [];
+        $is_free = true;
         $has_base_price = false;
         $default_prices = $this->price_model->get_all_default_prices(false, true);
         if (is_array($default_prices)) {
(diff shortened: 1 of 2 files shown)


core/domain/services/admin/entities/DefaultPrices.php
```diff
@@ -60,21 +60,35 @@ public function create(EE_Base_Class $entity)
...
+                ) ||(
+                    // or has fixed amount surcharge
+                    $default_price_clone->is_surcharge()
+                    && ! $default_price_clone->is_percent()
+                )) {
+                    $is_free = false;
+                }
                 $default_price_clone->set('PRC_ID', null);
                 $default_price_clone->set('PRC_is_default', false);
                 $default_price_clone->save();
```
manzoorwanijk, line 94: This means that the base price or other non-tax default prices will still be added regardless of the ticket being free or not, right?

tn3rb: yes correct

PR #3070, opened 2020-08-04
Toolbar icon fix.

nicolad:
Fixes the positioning of this element:

image

***
As seen in the screencast from below applying the font-size change on `ee-icon-size-20` has no effect on icon size, changes on icon size are taking place only when it is applied on `ee-icon-size-20:before`, but even that is overridden by `#wp-admin-bar-espresso-toolbar .ee-icon-ee-cup-thick:before` which has higher priority and is also using `important`

Peek 2020-08-04 10-32

### DIFF
--- a/core/admin/assets/admin-menu-styles.css
+++ b/core/admin/assets/admin-menu-styles.css
@@ -473,43 +473,32 @@ div.TB-ee-frame a {
     color: var(--ee-status-color-purple);
 }
 
-.ee-icon-size-8,
 .ee-icon-size-8:before { font-size: 8px !important; }
 
-.ee-icon-size-12,
 .ee-icon-size-12:before { font-size: 12px !important; }
 
-.ee-icon-size-14,
 .ee-icon-size-14:before { font-size: 14px !important; }
 
-.ee-icon-size-16,
 .ee-icon-size-16:before {
 	font-size: 16px !important;
 	margin-right: -.002em;
 }
 
-.ee-icon-size-18,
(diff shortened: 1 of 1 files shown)


core/admin/assets/admin-menu-styles.css
```diff
@@ -496,10 +496,13 @@ div.TB-ee-frame a {
 
 .ee-icon-size-20,
 .ee-icon-size-20:before {
-	font-size: 20px !important;
```
nicolad, line 499: result

image

core/admin/assets/admin-menu-styles.css
```diff
@@ -496,10 +496,13 @@ div.TB-ee-frame a {
 
 .ee-icon-size-20,
 .ee-icon-size-20:before {
```
nicolad, line 498: @tn3rb I see that these class name is used in few files, I need to know if the change/fix I've introduced won't conflict or break smth:

image

tn3rb:
totally baffled by those changes in that PR...
I remember looking at those icon-size classes and seeing that they were duplicated, ie:

```css
.ee-icon-size-8 { font-size: 8px !important; }

.ee-icon-size-8:before { font-size: 8px !important; }
```

and so i just combined them together...

but I specifically remember comparing the rules that each class was applying to make sure that they **could** be combined.


but now looking at the diff...
i don't see the original duplications  i removed !!! 

Might be best to remove ALL of the classes **without** the `:before` from those rules

PR #3075, opened 2020-08-06
Load Barista

tn3rb:
This PR basically comments out the EDTR domain entry point and loads assets required for Barista to take over


tn3rb:
# 🤦‍♂️ 
DOH... don't need to nor do we want to merge this into EDTR master

PR #3076, opened 2020-08-07
EBulk delete GQL schema

manzoorwanijk:
This PR adds bulk delete schema to GQL

### DIFF
--- a/core/domain/services/graphql/mutators/TicketDelete.php
+++ b/core/domain/services/graphql/mutators/TicketDelete.php
@@ -100,8 +100,6 @@ class TicketDelete extends EntityMutator
     public static function trashTicket($entity)
     {
         // trash the ticket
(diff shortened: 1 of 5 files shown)


core/domain/services/graphql/mutators/TicketDelete.php
```diff
@@ -85,4 +85,23 @@ public static function deleteTicketAndRelations($entity)
...
+     * @throws InvalidInterfaceException
+     * @throws InvalidDataTypeException
+     * @throws EE_Error
+     */
+    public static function trashTicket($entity)
+    {
+        // trash the ticket
+        $result = $entity->delete();
+
+        return $result;
```
tn3rb, line 105: can just return directly

```suggestion
        // trash the ticket
        return $entity->delete();
```

Issue #3082, opened 2020-08-13
Add `permission_callback` to `register_rest_route()` args


tn3rb:
from: https://developer.wordpress.org/rest-api/extending-the-rest-api/adding-custom-endpoints/#permissions-callback

> As of WordPress 5.5, if a permission_callback is not provided, the REST API will issue a _doing_it_wrong notice.

EE's REST API currently handles permissions via our own model system. Will need to investigate whether this can easily be integrated into the REST API route registration. If not, then might have to use `__return_true` which is the solution for public endpoints.

Issue #2790, opened 2020-05-08
wp_make_content_images_responsive() deprecated in WP 5.5


tn3rb:
`EE_Post_Content_Field::_setup_the_content_wp_core_only_filters()`  contains the following code:

```php
        if (function_exists('wp_make_content_images_responsive')) {
            add_filter('the_content_wp_core_only', 'wp_make_content_images_responsive', 10);
        }
```

but `wp_make_content_images_responsive()` has been changed to:

```php
/**
 * Filters 'img' elements in post content to add 'srcset' and 'sizes' attributes.
 *
 * @since 4.4.0
 * @deprecated 5.5.0
 *
 * @see wp_image_add_srcset_and_sizes()
 *
 * @param string $content The raw post content to be filtered.
 * @return string Converted content with 'srcset' and 'sizes' attributes added to images.
 */
function wp_make_content_images_responsive( $content ) {
	_deprecated_function( __FUNCTION__, '5.5.0', 'wp_filter_content_tags()' );

	// This will also add the `loading` attribute to `img` tags, if enabled.
	return wp_filter_content_tags( $content );
}
```

just need to change that hook to use the new filter and test that nothing else breaks

PR #3083, opened 2020-08-13
Fix Issues Caused by WP 5.5 Update

tn3rb:
This PR fixes the following two issues:

- #2790 
- #3082 

for the first issue, I simply added a check for the new post content filter callback and used that if it existed else continued to use the deprecated callback.

for the second issue, I simply injected a default callback used for public API endpoints if nothing was set. 
The EE REST API uses capabilities set via our db model system, so this was unnecessary for our purposes.


espresso.php
```diff
@@ -61,7 +61,7 @@ function espresso_duplicate_plugin_error()
     }
     add_action('admin_notices', 'espresso_duplicate_plugin_error', 1);
 } else {
-    define('EE_MIN_PHP_VER_REQUIRED', '5.4.0');
+    define('EE_MIN_PHP_VER_REQUIRED', '5.6.2');
```
manzoorwanijk, line 64: 👍

modules/core_rest_api/EED_Core_Rest_Api.module.php
```diff
@@ -214,6 +214,13 @@ public static function register_routes()
                     } else {
                         $single_endpoint_args['callback'] = $data_for_single_endpoint['callback'];
                     }
+                    // As of WordPress 5.5, if a permission_callback is not provided,
+                    // the REST API will issue a _doing_it_wrong notice.
+                    // Since the EE REST API defers capabilities to the db model system,
+                    // we will just use the generic WP callback for public endpoints
+                    if (! isset($single_endpoint_args['permission_callback'])) {
+                        $single_endpoint_args['permission_callback'] = '__return_true';
+                    }
```
manzoorwanijk, line 223: Nice 👍

espresso.php
tn3rb: oh ya, forgot I meant to post a link showing that that is WP's minimum now:
https://wordpress.org/support/article/requirements/

assets/src/eejs/validators/instance-of.js
```diff
@@ -7,12 +7,17 @@
...
-	return objectInstance.constructor.name === expectedName;
+	if (objectInstance.constructor) {
+		if (objectInstance.constructor.name && objectInstance.constructor.name === expectedName) {
+			return true;
+		}
+		if (objectInstance.constructor.displayName && objectInstance.constructor.displayName === expectedName) {
+			return true;
+		}
+	}
+	return objectInstance.hasOwnProperty('displayName') && objectInstance.displayName === expectedName;
```
tn3rb, line 22: could have done this much more compact but it really reduced the readability so i split things up like this to make it easier to grok

assets/src/vo/currency.js
```diff
@@ -69,18 +84,21 @@ export class Currency {
...
 	/**
 	 * Constructor
+	 *
 	 * @param {{}} currencyConfig An object containing the configuration for
 	 * this currency value object.  On construction, the Currency object is
 	 * frozen so that it becomes immutable.
 	 */
 	constructor( currencyConfig ) {
 		Currency.validateCurrencyConfig( currencyConfig );
+		this.displayName = 'Currency';
```
tn3rb, line 101: this is duplicated so that the `object.constructor.displayName` checks work

.travis.yml
```diff
@@ -119,11 +121,6 @@ jobs:
     - php: 7.2
       env: WP_MULTISITE=1
       if: NOT commit_message =~ js_only AND NOT commit_message =~ dependabot AND NOT sender =~ dependabot
-    # wp 4.5 builds
-    - php: 5.5
-      dist: trusty
-      env: WP_VERSION=4.5
-      if: NOT commit_message =~ js_only AND NOT commit_message =~ dependabot AND NOT sender =~ dependabot
```
tn3rb, line 126: shaved like 12-15 minutes off of the tests times by removing this :)

assets/src/vo/date-time/datetime.js
```diff
@@ -340,18 +361,14 @@ export default class DateTime {
 			! isUndefined( momentInstance.tz() ) &&
 			momentInstance.tz() !== 'UTC' ?
 			new this(
-				...this[ privateMethods.normalizeArguments ](
-					momentInstance.toISOString(),
-					momentInstance.tz(),
-					momentInstance.locale()
-				)
```
tn3rb, line 347: the sole purpose of `normalizeArguments()` was to swap the order of the properties depending on whether you were creating an instance of `Datetime` or `ServerDateTime`

assets/src/vo/date-time/server-date-time.js
```diff
@@ -25,34 +25,35 @@ import moment from 'moment-timezone';
...
 		// value timezone (HAS_TIMEZONE_STRING is just a shortcut check).
 		if (
 			HAS_TIMEZONE_STRING ||
 			( !! timezone && timezone !== 'UTC' )
-		) {
-			super( iso8601DateString, timezone, locale );
-		} else {
-			const datetime = !! iso8601DateString ?
+			) {
+				super( iso8601DateString, timezone, locale, 'ServerDateTime' );
```
tn3rb, line 48: had to pass the class name to the parent class because it uses `Object.freeze` which prevents setting properties after calling super... and you JS doesn't like when you set properties before you call super, soooo.... here we are!

.travis.yml
manzoorwanijk: cool 👍

assets/src/vo/date-time/datetime.js
manzoorwanijk: I hate overkill-ing the simple needs.

assets/src/vo/date-time/server-date-time.js
manzoorwanijk: JS is not for OOP, 😄

assets/src/eejs/validators/instance-of.js
manzoorwanijk: 👍

assets/src/vo/date-time/datetime.js
```diff
@@ -270,15 +290,14 @@ export default class DateTime {
 		}
 	}
 
-	static [ privateMethods.normalizeArguments ]( dateValue, timezone, locale ) {
-		return this.name === 'ServerDateTime' ?
-			[ dateValue, locale, timezone ] :
-			[ dateValue, timezone, locale ];
+	static [privateMethods.normalizeArguments](dateValue, timezone, locale) {
+		return [ dateValue, timezone, locale ];
```
manzoorwanijk, line 294: 👍

tn3rb:
### testing

- experiment with REST API, specifically with accessing data that requires authentication

- [ ] confirm that everything works as expected

- edit an EE event description and save

- [ ] confirm that everything works as expected

- edit/save a WP post

- [ ] confirm that everything works as expected

- add/edit/configure Event Attendees Gutenberg block

- [ ] confirm that everything works as expected

sethshoultes:
Everything seems to be working for me with this branch installed.

PR #3085, opened 2020-08-17
Don't Double Tax

tn3rb:
Previously, taxes were always applied globally by simply tagging a ticket as being taxable. During registration when line items were being generated this flag would be copied to the `LIN_is_taxable` field on the line item. Then when calculating totals, any ticket line item with that flag would have the global taxes calculated for it based on its unit price. 

In the new Event EDTR, taxes are added as price modifiers to tickets just like any other modifiers. This was done in order to facilitate reverse calculation where the ticket base price is calculated from the ticket total as opposed to the other way around. This also gives more control over taxes to users and solves the rare issue where some items are subject to some taxes and not others.

This PR solves the issue by only flagging ticket line items as taxable if:

- the ticket has NO price modifiers that are taxes

- is flagged as being taxable

this results in the "global" taxes only being applied where applicable.

### DIFF
--- a/core/db_classes/EE_Line_Item.class.php
+++ b/core/db_classes/EE_Line_Item.class.php
@@ -1236,8 +1236,7 @@ class EE_Line_Item extends EE_Base_Class implements EEI_Line_Item
             return 0;
         }
         // ensure all non-line items and non-sub-line-items have a quantity of 1 (except for Events)
-        if (! $this->is_line_item() && ! $this->is_sub_line_item() && ! $this->is_cancellation()
-        ) {
+        if (! $this->is_line_item() && ! $this->is_sub_line_item() && ! $this->is_cancellation()) {
             if ($this->OBJ_type() !== EEM_Line_Item::OBJ_TYPE_EVENT) {
                 $this->set_quantity(1);
             }
@@ -1290,7 +1289,11 @@ class EE_Line_Item extends EE_Base_Class implements EEI_Line_Item
         $subtotal_quantity = 0;
         // get the total of all its children
         foreach ($my_children as $child_line_item) {
-            if ($child_line_item instanceof EE_Line_Item && ! $child_line_item->is_cancellation()) {
+            if ($child_line_item instanceof EE_Line_Item) {
+                // skip line item if it is cancelled or is a tax
+                if ($child_line_item->is_cancellation() || $child_line_item->is_tax()) {
+                    continue;
+                }
                 // percentage line items are based on total so far
                 if ($child_line_item->is_percent()) {
                     // round as we go so that the line items add up ok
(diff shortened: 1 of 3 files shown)


core/db_classes/EE_Ticket.class.php
```diff
@@ -72,8 +72,9 @@ class EE_Ticket extends EE_Soft_Delete_Base_Class implements EEI_Line_Item_Objec
      *                                        date_format and the second value is the time format
      * @return EE_Ticket
      * @throws EE_Error
+     * @throws ReflectionException
      */
-    public static function new_instance($props_n_values = array(), $timezone = null, $date_formats = array())
+    public static function new_instance($props_n_values = [], $timezone = null, $date_formats = [])
```
tn3rb, line 77: most of the changes in this file were just formatting

core/db_classes/EE_Ticket.class.php
```diff
@@ -393,42 +416,66 @@ public function base_price($return_array = false)
...
+     */
+    public function tax_price_modifiers()
+    {
+        $query_params = [
+            0 => [
+                'Price_Type.PBT_ID' => EEM_Price_Type::base_type_tax,
+            ],
+        ];
         return $this->prices($query_params);
     }
```
tn3rb, line 451: this is the only real code change in this file

core/helpers/EEH_Line_Item.helper.php
```diff
@@ -309,13 +309,15 @@ public static function create_ticket_line_item(EE_Line_Item $total_line_item, EE
         $event = sprintf(_x('(For %1$s)', '(For Event Name)', 'event_espresso'), $first_datetime_name);
         // get event subtotal line
         $events_sub_total = self::get_event_line_item_for_ticket($total_line_item, $ticket);
+        $taxes = $ticket->tax_price_modifiers();
+        $is_taxable = empty($taxes) ? $ticket->taxable() : false;
```
tn3rb, line 313: so essentially, if the taxes have already been applied, which is obvious by the existence of tax price modifiers, then set `is_taxable` to `false`, otherwise, use the setting for the global flag

core/db_classes/EE_Line_Item.class.php
```diff
@@ -1351,7 +1354,11 @@ protected function _recalculate_pretax_total_for_line_item($calculated_total_so_
         $quantity_for_total = 1;
         // get the total of all its children
         foreach ($my_children as $child_line_item) {
-            if ($child_line_item instanceof EE_Line_Item && ! $child_line_item->is_cancellation()) {
+            if ($child_line_item instanceof EE_Line_Item) {
+                // skip line item if it is cancelled or is a tax
+                if ($child_line_item->is_cancellation() || $child_line_item->is_tax()) {
+                    continue;
+                }
```
manzoorwanijk, line 1361: `$my_children` 😄

tn3rb:
all of the failing unit tests are because of the issue fixed in #3083

manzoorwanijk (approved):
Great! Nobody should pay double taxes 😄

PR #3104, opened 2020-08-26
Add session_write_close() directly after any time we call start_session()

Pebblo:
See: https://core.trac.wordpress.org/ticket/47320

WP5.5 added a 'session check' to Site Health, users are asking why they are getting this notice now:

https://monosnap.com/file/GwlHOkbNUAc8851HLpfdr5pI7uvR66

As we don't actually write to the session and just use the ID I've just added `session_write_close()` directly after any session_start() call.

## How has this been tested
In master go to Tools -> Site Health.
With EE active you should see a notice about how an active PHP session has been detected.

Switch to this branch and confirm that no longer shows.

With this branch confirm that adding registrations on the front end and admin works, and using MER you can add various tickets to your cart from different events and then checkout without an issue.

## Checklist

* [x] My code is tested.

### DIFF
--- a/core/services/session/SessionStartHandler.php
+++ b/core/services/session/SessionStartHandler.php
@@ -63,6 +63,7 @@ class SessionStartHandler
                 $this->checkCustomSessionSaveHandler();
             } else {
                 session_start();
+                session_write_close();
             }
         }
     }
@@ -95,6 +96,7 @@ class SessionStartHandler
         // on a previous request then just start the session
         if ($this->sessionSaveHandlerIsValid()) {
             session_start();
+            session_write_close();
             return;
         }
         // If not, then attempt to deal with any errors,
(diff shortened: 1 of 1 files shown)


garthkoyle:
This passes the test. The session was detected on master, but is not on this branch.

Pebblo:
Just to dot some i's and cross some t's here, did you also check that registration worked with and without MER?

Afaik it's only the session ID used and we don't write anything else to _SESSION itself, so locking it is fine and we'd know pretty quick if it broke as adding events to SPCO would fail, right @tn3rb ?

tn3rb:
that's correct we only use the session for its ID

garthkoyle:
Yes, a registration WITH MER and WITHOUT MER worked as expected.

Pebblo:
kk, thanks guys! Merging in.

PR #3120, opened 2020-09-09
If the user can read private events then include the private status when pulling upcoming events

Pebblo:
Reported: https://eventespresso.com/topic/possible-bug-filter-by-category-shows-private-categoryupcoming-does-not/?view=all

To reproduce set an event to have a status of private and set a category on the event.

Then in the event list table (Event Espresso -> Events), filter the category.

You'll see your private event listed.

Now filter the status to show Upcoming events, the private event disappears.

Switch to this branch and repeat, it should not.

This just using a capability check that is used all over but if you want to confirm, remove `ee_read_private_events` from your user account and repeat.

## Checklist

* [x] My code is tested.

### DIFF
--- a/core/db_models/EEM_Event.model.php
+++ b/core/db_models/EEM_Event.model.php
@@ -624,7 +624,13 @@ class EEM_Event extends EEM_CPT_Base
         }
         // let's add specific query_params for active_events
         // keep in mind this will override any sent status in the query AND any date queries.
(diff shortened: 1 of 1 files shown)


tn3rb (approved):
looks great 💯

garthkoyle:
Confirmed according to the checklist.

PR #3128, opened 2020-09-18
Simplify JS i18n Process

tn3rb:
This PR removes a bunch of code related to the previous asset build process that is no longer necessary due to assets now being built and deployed from the Barista mono-repo.


core/domain/Domain.php
```diff
@@ -20,7 +20,9 @@ class Domain extends DomainBase implements CaffeinatedInterface
      */
     const API_NAMESPACE = 'ee/v';
 
-    const ASSET_NAMESPACE = 'eventespresso-';
+    const ASSET_NAMESPACE = 'eventespresso';
+
+    const TEXT_DOMAIN = 'event_espresso';
```
tn3rb, line 25: this is NOT used within any WP i18n functions and only for secondary text domain identification.

core/services/assets/I18nRegistry.php
```diff
@@ -2,15 +2,14 @@
...
 /**
- * I18nRegistry
- * Handles registering translations for registered script handles.
- *
  * @package EventEspresso\core\services\assets
  * @author  Darren Ethier
  * @since   4.9.62.p
+ * @deprecated $VID:$
  */
 class I18nRegistry
```
tn3rb, line 14: totally nerfed this entire class 😆 because none of it is needed now

core/domain/services/assets/CoreAssetManager.php
```diff
@@ -107,51 +107,6 @@ public function addJavascriptFiles()
...
-        );
-        /** site locale (user locale if user logged in) */
-        $this->registry->addData(
-            'locale',
-            [
-                'user' => get_user_locale(),
-                'site' => get_locale(),
-            ]
-        );
-
```
manzoorwanijk, line 154: 👍

core/services/assets/I18nRegistry.php
manzoorwanijk: Cool 🚀

manzoorwanijk (approved):
It is always great when lesser code can do the job.

**AWESOME** 🚀

PR #3132, opened 2020-09-28
Don't update the Transaction/Registrations within the cron check when the primary registrant is Not Approved

Pebblo:
....they most likely finalized using a free ticket.

Messages can be triggered multiple times if you use a DRS of Not Approved and register onto a free ticket.

This branch simply checks that the Primary Registration is NOT set to Not Approved when a Transaction status is 'Complete' and the expired transaction cron runs.

To reproduce set up an event to use a DRS of 'Not Approved' and set it to have a free ticket.

Register onto that ticket.

EE will trigger the 'Not Approved' message.

Now us [WP Crontol](https://en-gb.wordpress.org/plugins/wp-crontrol/) (or whatever else you prefer, to run the `AHEE__EE_Cron_Tasks__expired_transaction_check` job 

(note that function is passed the TXN ID in the additional params for the cron, make sure you are running the cron that matches the TXN ID for the above registration)

Each time it runs it will retrigger the Not Approved message so it means that all Not Approved messaged registrations on free tickets will trigger 2 messages (once when they select and finalize and another when the session expires and the above cron runs, by default an hour late).

This branch should prevent that.

## Checklist

* [x] My code is tested.

### DIFF
--- a/core/EE_Cron_Tasks.core.php
+++ b/core/EE_Cron_Tasks.core.php
@@ -420,7 +420,7 @@ class EE_Cron_Tasks extends EE_Base
                     // Completed TXNs
                     case EEM_Transaction::complete_status_code:
                         // Don't update the transaction/registrations if the Primary Registration is Not Approved.
-                        $transaction->primary_registration();
+                        $primary_registration = $transaction->primary_registration();
                         if ($primary_registration instanceof EE_Registration
                             && $primary_registration->status_ID() == EEM_Registration::status_id_not_approved
                         ) {
--- a/core/EE_Cron_Tasks.core.php
+++ b/core/EE_Cron_Tasks.core.php
@@ -419,16 +419,22 @@ class EE_Cron_Tasks extends EE_Base
                 switch ($transaction->status_ID()) {
                     // Completed TXNs
                     case EEM_Transaction::complete_status_code:
-                        /** @type EE_Transaction_Processor $transaction_processor */
-                        $transaction_processor = EE_Registry::instance()->load_class('Transaction_Processor');
-                        $transaction_processor->update_transaction_and_registrations_after_checkout_or_payment(
-                            $transaction,
-                            $transaction->last_payment()
-                        );
-                        do_action(
-                            'AHEE__EE_Cron_Tasks__process_expired_transactions__completed_transaction',
-                            $transaction
-                        );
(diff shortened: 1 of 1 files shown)


core/EE_Cron_Tasks.core.php
```diff
@@ -419,16 +419,22 @@ public static function process_expired_transactions()
...
-                            $transaction,
-                            $transaction->last_payment()
-                        );
-                        do_action(
-                            'AHEE__EE_Cron_Tasks__process_expired_transactions__completed_transaction',
-                            $transaction
-                        );
+                        // Don't update the transaction/registrations if the Primary Registration is Not Approved.
+                        $transaction->primary_registration();
+                        if ($primary_registration instanceof EE_Registration
```
tn3rb, line 424: did you mean to capture this as a variable???
ie:

```suggestion
                        $primary_registration = $transaction->primary_registration();
                        if ($primary_registration instanceof EE_Registration
```

core/EE_Cron_Tasks.core.php
```diff
@@ -419,16 +419,22 @@ public static function process_expired_transactions()
...
+                            && $primary_registration->status_ID() == EEM_Registration::status_id_not_approved
```
tn3rb, line 425: wait... what?

the title of the PR is:

> Don't update the Transaction/Registrations within the cron check when the primary registrant is Not Approved

but this looks like it would ONLY do so when the primary registrant **IS** "Not Approved"

shouldn't this be:

```suggestion
                            && $primary_registration->status_ID() !== EEM_Registration::status_id_not_approved
```

either way this should use type safe comparison, ie: `===` or `!==`

tn3rb (dismissed):
ummm... a couple of things don't quite look right

otherwise looks like it should be a good fix

Pebblo:
Oh, my, word! I gotta chalk that up to some Monday madness or something 🤦‍♂️ 

I moved the code around a little for better readability after testing and I think I just completely mashed it up.

Committed your changes and retesting again just to be sure.

lorenzocaum:
I ran through testing and this checks out.

With the existing version, a not approved email is triggered after registration. Then it is triggered when the cron runs. (I used WP crontrol)

On the ENH/dont-trigger-not-approved-messages-on-cron branch, a single not approved registration is triggered after registration. It doesn't send extras when the cron runs. (I also used WP crontrol here and ran the cron several times.)

Please merge.

PR #3138, opened 2020-09-29
Don't Load Bundled WPGraphQL if PHP < 7 closes #3135

### LINKED ISSUE #3135: WP GraphQL is loaded for PHP < 7
While working on #3134, I tested the branch with PHP 5.6, it resulted in fatal error. The reason being that WP GraphQL is loaded or used for PHP < 7

The two files that load it without a version check are :
- `core/domain/entities/routing/data_nodes/core/CurrentUser.php`
- `core/domain/entities/routing/handlers/shared/GQLRequests.php`

tn3rb:
see #3135

In the `\EventEspresso\core\domain\entities\routing\data_nodes\core\CurrentUser` class constructor, we were loading the bundled WPGraphQL plugin if it was not already active, but this causes errors if the PHP version is less than 7. This PR simply checks the PHP version before doing so.

Plz note that in `\EventEspresso\core\domain\entities\routing\handlers\shared\GQLRequests::requestHandler()` we also load the bundled WPGraphQL plugin, but that method is not called if the conditional in `matchesCurrentRequest()` returns false, and that conditional already performs a PHP version check.

### DIFF
--- a/core/domain/entities/routing/data_nodes/core/CurrentUser.php
+++ b/core/domain/entities/routing/data_nodes/core/CurrentUser.php
@@ -24,7 +24,7 @@ class CurrentUser extends JsonDataNode
      */
     public function __construct(JsonDataNodeValidator $validator)
     {
-        if (! class_exists('WPGraphQL')) {
+        if (PHP_VERSION_ID > 70000 && ! class_exists('WPGraphQL')) {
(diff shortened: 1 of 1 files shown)


manzoorwanijk (approved):
Great! Thank you for the fix

PR #3140, opened 2020-09-29
When an event is expired, show both the expired event message an the view details button ...

Pebblo:
... unless on a single event page.

So in https://github.com/eventespresso/event-espresso-core/pull/2952

I moved the conditional to display the expired event message to be before the view details button, this means the ticket selector no longer shows the view details button and just displays a message. Turns out some people are using that button:

https://eventespresso.com/topic/view-details-button-not-displaying-on-expired-events/

See the events after the first 3 here: https://monosnap.com/file/eHu68FaO1sUdvy7Y9r1FR9odyTpoR3

So I think it's worth adding the view details back in if is_single() isn't true.

So event lists show like this: https://monosnap.com/file/7CIZbO9jg5K1n1lHXcIZrFy1PuzoR9

With both the expired event message and the view details button.

The single event page, doesn't show the view details: 

https://monosnap.com/file/7nWPqVPCvAMZ20CZce81zIqmoG7xyM

Then the ticket selector shows both:

https://monosnap.com/file/xgacaUB7wCxbtQ3iklDHnyzQr5T93O

## How has this been tested
Use `[ESPRESSO_EVENTS show_expired=true]` and view your expired events.

Check that on the event list it shows a view details button (master will not)

On the event page itself, check is doesn't show the view details button.

Use `[ESPRESSO_TICKET_SELECTOR event_id=X]` on another page and confirm it shows the message and view details button.
(X being the ID of one of your expired events)

## Checklist

* [x] My code is tested.


tn3rb (approved):
Looks great 💯 

sry for taking so long to review this... don't know how I missed this 🤷‍♂️ 

next time give me a kick if I don't respond in a reasonable amount of time

garthkoyle:
I managed to confused myself but once Tony straightened me out this is working great.

PR #3163, opened 2020-10-20
Add Remaining Event Data to EDTR DOM Data Node WIP

tn3rb:
Currently the only Event data we are dumping into the DOM data is the Event ID.
This PR **attempts** to add the remaining event data so that we can begin to build out the rest of the Event Editor. I have no idea how to do so with our hybrid CPTs and tried to use the existing Attendee GQL Type as a guide, but get the following error when loading the Event Editor:

```
You cannot register duplicate fields on the same Type.
The field already exists on the type 'RootQuery'.
Make sure to give the field a unique name.
```


core/domain/entities/admin/GraphQLData/Events.php
```diff
@@ -0,0 +1,64 @@
...
+        }
+QUERY;
+        $this->setParams(
+            [
+                'operation_name' => 'GET_EVENTS',
+                'variables'      => [
+                    'first' => 500,
+                ],
+                'query'          => $query,
+            ]
```
manzoorwanijk, line 58: You are trying to add 500 events to DOM data? 😲

core/domain/services/graphql/connections/RootQueryEventsConnection.php
```diff
@@ -0,0 +1,92 @@
+<?php
+
+namespace EventEspresso\core\domain\services\graphql\connections;
+
+use EEM_Event;
+use EventEspresso\core\domain\services\graphql\abstracts\AbstractRootQueryConnection;
+use EventEspresso\core\domain\services\graphql\connection_resolvers\EventConnectionResolver;
+
+class RootQueryEventsConnection extends AbstractRootQueryConnection
```
manzoorwanijk, line 9: This is not needed, it's already provided by WP GQL. So, it won't allow you duplicate the field

core/domain/services/graphql/connection_resolvers/EventConnectionResolver.php
```diff
@@ -0,0 +1,191 @@
...
+use EventEspresso\core\exceptions\InvalidDataTypeException;
+use EventEspresso\core\exceptions\InvalidInterfaceException;
+use InvalidArgumentException;
+use WPGraphQL\Model\Post;
+
+/**
+ * Class EventConnectionResolver
+ *
+ */
+class EventConnectionResolver extends AbstractConnectionResolver
```
manzoorwanijk, line 19: This is also not needed, already in WP GQL

manzoorwanijk:
`Attendee` is a special CPT which has some special (private) fields like email, phone etc. Thus it has an explicit resolver to take care of those permission checks to read the data. `Event` and `Venue` are normal CPTs which are handled automatically by WPGQL.

PR #3173, opened 2020-10-21
Make Code Prettier

tn3rb:
WordPress GB is employing prettier now which is affecting our JS unit test linting so this PR installs Prettier and the corresponding `@wordpress/prettier-config`

it also auto-formats all of the `asset/src/` files as well as those in root, so please ignore those in your review


nicolad (approved):
LGTM :trollface:

manzoorwanijk (approved):
Lets hope all is well 😆

PR #3174, opened 2020-10-22
Add `$datetime` to Hook Args

tn3rb:
this PR simply adds another argument onto the existing `FHEE__espresso_list_of_event_dates__html` hook so that REM can gain access to some of the datetime data


manzoorwanijk (approved):
Hope that it doesn't break the Antarctic glaciers 😆

PR #3180, opened 2020-10-28
Ensure Default Price Order Is Set. closes #3080 

### LINKED ISSUE #3080: Default prices/taxes created from backend have order set to 0
When you create default taxes from back-end, their order should be non-zero to make sure they are placed after the base price when calculating the ticket price.

Here is the problem that arises because of that.

image

tn3rb:
plz see #3080 

Don't know how this was ever not noticed before.

This PR simply uses the base price type order as the order for any new prices that are created via the existing legacy pricing admin

### DIFF
--- a/caffeinated/admin/new/pricing/Pricing_Admin_Page.core.php
+++ b/caffeinated/admin/new/pricing/Pricing_Admin_Page.core.php
@@ -667,44 +667,43 @@ class Pricing_Admin_Page extends EE_Admin_Page
 
 
     /**
-     *        set_price_column_values
-     *
-     * @access protected
      * @return array
+     * @throws EE_Error
+     * @throws ReflectionException
      */
     protected function set_price_column_values()
     {
-
-        do_action('AHEE_log', __FILE__, __FUNCTION__, '');
-
-        $set_column_values = array(
-            'PRT_ID'         => absint($this->_req_data['PRT_ID']),
+        $PRC_order = 0;
+        $PRT_ID = absint($this->_req_data['PRT_ID']);
+        if ($PRT_ID) {
+            /** @var EE_Price_Type $price_type */
+            $price_type = EEM_Price_Type::instance()->get_one_by_ID($PRT_ID);
+            if ($price_type instanceof EE_Price_Type) {
+                $PRC_order = $price_type->order();
+            }
+        }
+        return array(
+            'PRT_ID'         => $PRT_ID,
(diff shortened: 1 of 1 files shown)


caffeinated/admin/new/pricing/Pricing_Admin_Page.core.php
```diff
@@ -667,44 +667,43 @@ public function _edit_price_details_meta_box()
...
+        }
+        return array(
+            'PRT_ID'         => $PRT_ID,
             'PRC_amount'     => $this->_req_data['PRC_amount'],
             'PRC_name'       => $this->_req_data['PRC_name'],
             'PRC_desc'       => $this->_req_data['PRC_desc'],
             'PRC_is_default' => 1,
             'PRC_overrides'  => null,
-            'PRC_order'      => 0,
+            'PRC_order'      => $PRC_order,
```
manzoorwanijk, line 692: Simple and elegant 👍

caffeinated/admin/new/pricing/Pricing_Admin_Page.core.php
```diff
@@ -738,8 +737,9 @@ protected function _insert_or_update_price($insert = false)
             }
 
             $PR = EEM_Price::instance()->get_one_by_ID($PRC_ID);
-            if ($PR->type_obj()->base_type() !== EEM_Price_Type::base_type_tax) {
-                // if this is $PRC_ID == 1, then we need to update the default ticket attached to this price so the TKT_price value is updated.
+            if ($PR instanceof EE_Price && $PR->type_obj()->base_type() !== EEM_Price_Type::base_type_tax) {
+                // if this is $PRC_ID == 1,
+                // then we need to update the default ticket attached to this price so the TKT_price value is updated.
```
manzoorwanijk, line 742: 👍

manzoorwanijk (approved):
Great! Thank you for the fix.

manzoorwanijk:
I think this fix should have been added to `master`.

tn3rb:
i debated about adding this fix to master instead, but chose this route because it could likely end up in master sooner going this route

By adding it to EDTR it can just get tested as part of the regular EDTR testing whereas merging the fix into master means it will need to be tested separately there... AS WELL AS still get tested within EDTR. Which will just delay things imho, and hopefully EDTR will be merged into master soon anyways. AS well, this error obviously existed in master for who knows how long without anyone noticing so it's not like a little bit of extra waiting will matter.

PR #3183, opened 2020-10-29
add EventManagers DOM data class to EDTR

tn3rb:
This PR:

- adds a new `\EventEspresso\core\domain\services\admin\events\editor\EventManagers` data class which queries for WP users with event edit caps

- converts that data to a format that is usable by GQL

### DIFF
--- a/core/domain/services/admin/events/editor/EventManagers.php
+++ b/core/domain/services/admin/events/editor/EventManagers.php
@@ -57,7 +57,7 @@ class EventManagers implements EventEditorDataInterface
 
 
     /**
-     * Returns a list of WP_Role names that have "event manager" capabilities
+     * Returns a list of WP_Role that have "event manager" capabilities
      * The list of "event manager" capabilities is filtered but defaults to:
      *      - 'ee_edit_events'
      *      - 'ee_edit_event'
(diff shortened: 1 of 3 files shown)


core/domain/services/admin/events/editor/EventManagers.php
```diff
@@ -0,0 +1,71 @@
...
+        $roles = $wp_roles->role_objects;
+        $capabilities = apply_filters(
+            'FHEE__EventEspresso_core_domain_services_admin_events_editor_EventManagers__getData__capability',
+            ['ee_edit_events', 'ee_edit_event'],
+            $roles
+        );
+        $SQL = "SELECT u1.ID, u1.display_name FROM {$wpdb->users} AS u1 "
+               . "INNER JOIN {$wpdb->usermeta} AS u2 ON u1.ID = u2.user_id "
+               . "AND u2.meta_key='{$wpdb->prefix}capabilities' "
+               . 'WHERE';
```
manzoorwanijk, line 50: This may not work because all the capabilities are not added to user meta

tn3rb: ahhhh... but I'm not getting the actual caps from the user_meta table... I'm getting them from the WP_Role objects, then checking for the roles that have the desired caps

tn3rb: I should add some comments to the code to make things clear

core/domain/services/admin/events/editor/EventManagers.php
```diff
@@ -0,0 +1,71 @@
...
+        $operator = '';
+        foreach ($roles as $role) {
+            foreach ($capabilities as $capability) {
+                if($role->has_cap($capability)) {
+                    $SQL .= $operator . ' u2.meta_value LIKE \'%"' . $role->name . '"%\' ';
```
manzoorwanijk, line 55: Shouldn't this be `$capability` instead of `$role->name`? 🤔

tn3rb: nope...
we are querying for roles not caps
if a role has any of the caps we are targeting, then we add a clause to check for that role

ooo... i didn't do anything to prevent duplicates, not that that would affect the queries, but would just be inefficient

manzoorwanijk (approved):
Looks good

PR #3184, opened 2020-10-29
Don't Update Legacy Registration Options Inputs

tn3rb:
this PR simply moves the capture of request inputs from the Legacy Event Registration Options meta box inside of a conditional that checks whether the EDTR is active or not so that they do not overwrite values set via GQL.

### ALSO...

made updates to the GQL schemas for events in:

- `core/domain/entities/admin/GraphQLData/Event.php`
- `core/domain/services/graphql/types/Event.php`

because there was an error when saving the EDTR

### DIFF
--- a/admin_pages/events/Events_Admin_Page.core.php
+++ b/admin_pages/events/Events_Admin_Page.core.php
@@ -1014,25 +1014,35 @@ class Events_Admin_Page extends EE_Admin_Page_CPT
             // get out we're not processing an event save.
             return;
         }
-        $event_values = array(
-            'EVT_display_desc'                => ! empty($this->_req_data['display_desc']) ? 1 : 0,
-            'EVT_display_ticket_selector'     => ! empty($this->_req_data['display_ticket_selector']) ? 1 : 0,
-            'EVT_additional_limit'            => min(
+        $event_values = [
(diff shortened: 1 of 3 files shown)


admin_pages/events/Events_Admin_Page.core.php
```diff
@@ -1014,25 +1014,35 @@ protected function _insert_update_cpt_item($post_id, $post)
...
-            'EVT_display_ticket_selector'     => ! empty($this->_req_data['display_ticket_selector']) ? 1 : 0,
-            'EVT_additional_limit'            => min(
+        $event_values = [
+            'EVT_member_only'     => ! empty($this->_req_data['member_only']) ? 1 : 0,
+            'EVT_allow_overflow'  => ! empty($this->_req_data['EVT_allow_overflow']) ? 1 : 0,
+            'EVT_timezone_string' => ! empty($this->_req_data['timezone_string'])
+                ? sanitize_text_field($this->_req_data['timezone_string'])
+                : null,
+        ];
+        if (! $this->admin_config->useAdvancedEditor()) {
```
manzoorwanijk, line 1024: 👍

manzoorwanijk (approved):
Nice 👍

PR #3205, opened 2020-11-24
new FeatureFlags class

tn3rb:
was having difficulties with conflicts and issues when cherry picking commits from the core barista branch into the `rc-barista-deployment` branch, which was necessary for removing features from the EDTR that were still in development. Since we have feature flag capabilities in barista, felt it was better to utilize that for controlling features instead of trying to do so via git, you know, like a fool!

this PR:

- adds a new class for controlling feature flags, which also allows feature flags to be filtered (like for example, by the barista plugin)

- uses that class in multiple places for controlling display and usage of features on an atomic level

- allow features to be controlled via user capabilities via the existing `CapCheck` system

- copies over barista assets from the `new-feature-flags` branch (see https://github.com/eventespresso/barista/pull/436) so that this branch can be tested if anyone so desires.


core/domain/services/capabilities/FeatureFlags.php
```diff
@@ -0,0 +1,82 @@
...
+     */
+    public function featureAllowed(string $feature)
+    {
+        $flag = isset($this->feature_flags[ $feature ]) ? $this->feature_flags[ $feature ] : false;
+        try {
+            return $flag instanceof CapCheck
+                ? $this->capabilities_checker->processCapCheck($flag)
+                : $flag;
+        } catch (InsufficientPermissionsException $e) {
+            // eat the exception
```
manzoorwanijk, line 68: 😄

PR #3207, opened 2020-11-24
Add support for Plain URLs

### LINKED ISSUE #3206: GraphQl Does Not Work With Plain Permalink Structure
### steps to reproduce

- go to `example.com/wp-admin/options-permalink.php` admin page

- set permalinks structure to `Plain` and click save

- go to `example.com/graphql` and/or attempt to use EDTR

I realize that anybody sane with at least half a brain will not be using plain permalink structures, but not everybody fits into that category unfortunately, so we need to at least investigate whether this issue can be mitigated

manzoorwanijk:
This PR adds support for plain permalinks for EDTR.

Note: It still needs some work to make sure WPGQL is loaded on such plain (ugly) URLs.

Closes #3206

### DIFF
--- a/core/domain/entities/routing/data_nodes/core/Api.php
+++ b/core/domain/entities/routing/data_nodes/core/Api.php
@@ -18,7 +18,6 @@ use WPGraphQL\Router;
  */
 class Api extends JsonDataNode
 {
-
     const NODE_NAME = 'api';
 
 
@@ -43,7 +42,13 @@ class Api extends JsonDataNode
         $this->addData('restApiRouteUrl', rest_url('ee/v4.8.36/'));
         $this->addData('restApiCollectionEndpoints', EED_Core_Rest_Api::getCollectionRoutesIndexedByModelName());
         $this->addData('restApiPrimaryKeys', EED_Core_Rest_Api::getPrimaryKeyNamesIndexedByModelName());
-        $graphqlEndpoint = class_exists('WPGraphQL') ? trailingslashit(site_url()) . Router::$route : '';
+
+        // route can be something like 'graphql'
+        $route = trim(Router::$route, '/');
+        // make sure we are dealing with sane folks
+        $has_pretty_permalinks = (bool) get_option('permalink_structure');
+        // if pretty permalinks, use '/graphql' otherwise '?graphql=1'
+        $graphqlEndpoint = $has_pretty_permalinks ? site_url($route) : add_query_arg($route, 1, site_url());
         $this->addData('graphqlEndpoint', esc_url($graphqlEndpoint));
         $this->setInitialized(true);
     }
--- a/core/domain/services/contexts/RequestTypeContextDetector.php
+++ b/core/domain/services/contexts/RequestTypeContextDetector.php
@@ -188,10 +188,10 @@ class RequestTypeContextDetector
     {
         // Check for URLs like http://mysite.com/?rest_route=/ee... and http://mysite.com/wp-json/ee/...
         return strpos(
-                   $this->request->getRequestParam('rest_route', false),
-                   '/' . Domain::API_NAMESPACE
-               ) === 0
-               || $this->uriPathMatches(trim(rest_get_url_prefix(), '/') . '/' . Domain::API_NAMESPACE);
+            $this->request->getRequestParam('rest_route', false),
+            '/' . Domain::API_NAMESPACE
+        ) === 0
+            || $this->uriPathMatches(trim(rest_get_url_prefix(), '/') . '/' . Domain::API_NAMESPACE);
     }
 
 
--- a/core/domain/services/contexts/RequestTypeContextDetector.php
+++ b/core/domain/services/contexts/RequestTypeContextDetector.php
@@ -3,10 +3,11 @@
 namespace EventEspresso\core\domain\services\contexts;
 
 use EventEspresso\core\domain\Domain;
+use EventEspresso\core\services\graphql\GraphQLEndpoint;
 use EventEspresso\core\services\request\RequestInterface;
 use EventEspresso\core\domain\entities\contexts\RequestTypeContext;
 use InvalidArgumentException;
-use Throwable;
+use WPGraphQL\Router;
 
 /**
  * Class RequestTypeContextDetector
@@ -18,6 +19,11 @@ use Throwable;
  */
 class RequestTypeContextDetector
 {
+
+    /**
+     * @var GraphQLEndpoint $gql_endpoint
+     */
+    private $gql_endpoint;
 
     /**
      * @var RequestTypeContextFactoryInterface $factory
@@ -38,6 +44,7 @@ class RequestTypeContextDetector
     /**
      * RequestTypeContextDetector constructor.
      *
+     * @param GraphQLEndpoint                    $gql_endpoint
      * @param RequestInterface                   $request
      * @param RequestTypeContextFactoryInterface $factory
      * @param array                              $globalRouteConditions an array for injecting values that would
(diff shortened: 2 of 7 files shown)


core/domain/entities/routing/data_nodes/core/Api.php
```diff
@@ -43,7 +42,13 @@ public function initialize()
         $this->addData('restApiRouteUrl', rest_url('ee/v4.8.36/'));
         $this->addData('restApiCollectionEndpoints', EED_Core_Rest_Api::getCollectionRoutesIndexedByModelName());
         $this->addData('restApiPrimaryKeys', EED_Core_Rest_Api::getPrimaryKeyNamesIndexedByModelName());
-        $graphqlEndpoint = class_exists('WPGraphQL') ? trailingslashit(site_url()) . Router::$route : '';
+
+        // route can be something like 'graphql'
+        $route = trim(Router::$route, '/');
+        // make sure we are dealing with sane folks
+        $has_pretty_permalinks = (bool) get_option('permalink_structure');
```
nicolad, line 49: `(bool)` here is doing typecasting?

manzoorwanijk: yes

tn3rb:
```
// make sure we are dealing with sane folks
```

LOLZ

tn3rb (approved):
### looks great! 💯 

didn't realize it would be this easy 😄

manzoorwanijk:
@tn3rb it doesn't work with plain URLs. When I try to update a datetime, it errs, which means WP GQL is not loaded.

tn3rb:
OK this ended up being NOT so easy 😆 so I take back my earlier comment 🤦‍♂️ 

It's a little bit hacky the way it works, but I really couldn't come up with a cleaner way to handle things without duplicating a bunch of code from WPGraphQL.

core/services/graphql/GraphQLEndpoint.php
```diff
@@ -0,0 +1,101 @@
...
+            // but they don't settle on their endpoint until `after_setup_theme` has run,
+            // which is too late for us, so we are just going to save this separately so that we have it.
+            // Oh, and this will essentially recheck this value on EVERY request
+            // but will only update our saved value if it has actually changed
+            $this->updateOption(
+                get_graphql_setting(
+                    'graphql_endpoint',
+                    apply_filters('graphql_endpoint', GraphQLEndpoint::DEFAULT_ENDPOINT)
+                )
+            );
```
manzoorwanijk, line 96: Not feeling confident about this logic here. It may result in unexpected conflicts with other plugins that may use WP GraphQL. For example some plugins may use `graphql_init` action to register their schema which means their callbacks will be called twice and it may result in duplication errors in schema registration.

tn3rb: ya that's a good point and I know better than to do stuff like this...
i guess sometimes i get so focused on trying to get a solution working that i just ignore the complications... thnx for keeping me honest 😆 

ok so then the only other option then is to scrap the logic inside that `class_exists('WPGraphQL')` consitional, remove the `!` from it, and then move all of that other logic into the conditional... meaning ONLY check for the gql endpoint if `WPGraphQL` is already fired up, ie:

```php
if (class_exists('WPGraphQL')) {
    $this->updateOption(
        get_graphql_setting(
            'graphql_endpoint',
            apply_filters('graphql_endpoint', GraphQLEndpoint::DEFAULT_ENDPOINT)
        )
    );
}
```

This means we won't always be able to obtain the correct endpoint that's actually in use.

Another option would be to find a hook that fires within the WPGraphQL logic sometime after they have decided upon their endpoint, and hook into that in order to save it for our purposes. But again, we won't always be able to obtain the correct endpoint immediately when we need it (which is before WGL does anything).

Going to look into trying to set up that second option. Should be pretty straightforward

manzoorwanijk: I have no idea about how to solve this puzzle to load WP GQL conditionally based on its own condition 😆

tn3rb:
ok so my latest changes hooked into the `graphql_register_settings` action which runs right at the end of their settings initialization, setup, etc, etc. So if someone is already running the WP GraphQL plugin and goes to change their endpoint, our logic should immediately grab a copy of the updated value and save it to our own WP option. When we retrieve our option we also run the same filter that WP GraphQL applies so if someone is changing the endpoint that way, we should still pick it up.

I've tested this using the following setups:

- `http://www.dev.test/graphql` default endpoint with pretty permalinks

- `http://www.dev.test/?graphql` default endpoint with fugly permalinks

- `http://www.dev.test/gql` filtered endpoint with pretty permalinks

- `http://www.dev.test/?gql` filtered endpoint with fugly permalinks

- `http://www.dev.test/mygraphql` custom endpoint set via WP GraphQL plugin with pretty permalinks

- `http://www.dev.test/?mygraphql` custom endpoint set via WP GraphQL plugin with fugly permalinks

core/services/graphql/GraphQLEndpoint.php
```diff
@@ -0,0 +1,99 @@
...
+        } catch (Throwable $t) {
+            // eat it
+        }
+    }
```
tn3rb, line 98: don't need to check if `class_exists('WPGraphQL')` anymore since this is hooking into `'graphql_register_settings'` action

core/domain/services/contexts/RequestTypeContextDetector.php
```diff
@@ -231,8 +246,8 @@ private function isFeedRequest()
     private function uriPathMatches($component)
     {
         $request_uri = $this->request->requestUri(true);
-        $parts       = explode('?', $request_uri);
-        $path        = trim(reset($parts), '/');
+        $parts = explode('?', $request_uri);
+        $path = trim(reset($parts), '/');
```
manzoorwanijk, line 250: This also matches `https://example.com/blog/why-you-should-use-graphql-for-wp` 😄

manzoorwanijk:
It looks great now, since it's my PR, I can't approve it 😄

core/domain/services/contexts/RequestTypeContextDetector.php
tn3rb: the code you are highlighting doesn't even look at the parameter you want to compare against, it is simply preparing the URL for comparison which happens on the last line

here's the entire method:

```php
    private function uriPathMatches($component)
    {
        $request_uri = $this->request->requestUri(true);
        $parts       = explode('?', $request_uri);
        $path        = trim(reset($parts), '/');
        return strpos($path, $component) === 0;
    }
```

- the `$component` parameter is the URL chunk you want to match, like `graphql`

- `$request_uri` is the request portion of the URI relative to the WP root.
  from the phpdocs of `Request::requestUri()`:
  > If `home_url()` is "http://mysite.com/wp/", and a request comes to "http://mysite.com/wp/wp-json", setting `$relativeToWpRoot=true` will return "/wp-json", whereas `$relativeToWpRoot=false` will return "/wp/wp-json/"
  For a URL like `example.com/blog/why-you-should-use-graphql-for-wp?some-param=123`
  `$request_uri` would be `/blog/why-you-should-use-graphql-for-wp?some-param=false`

- `$parts` simply splits the URI at the point where the parameters begin and dumps the two parts into an array

- `reset($parts)` grabs the first element from the above array and `trim( xxxxx, '/')` removes any backslashes. So in our example `$path` would now be `blog/why-you-should-use-graphql-for-wp`. At this point, no attempt to match this path with a string has occurred, that happens in the next step

- `strpos($path, $component) === 0` checks if the position of the `$component` string within the `$path` string is `0`, ie: it **only** returns `true` if the `$path` **begins** with `$component`. `$component` can be anywhere else in the string but if it's not at the very beginning, then it returns `false`

- passing `graphql` to this method would then be checking if the current request URI starts with `graphql`, and since `blog/why-you-should-use-graphql-for-wp` does not (`graphql` is at position 24), it would return `false`

- if the current URL was `example.com/graphql` then it would return `true`, OR if `$component` was "blog" and the URL was `example.com/blog/why-you-should-use-graphql-for-wp` then it would also return `true`

PR #3208, opened 2020-11-26
Better Event Manager Detection

tn3rb:
In https://github.com/eventespresso/barista/pull/436 I added the ability to toggle feature flags via user capabilities which led to the discovery that the logic in `EventEspresso\core\domain\services\admin\events\editor\EventManagers` was only determining event managers based on the "Event Manager" role.
This of course makes sense but disregards users with some other role that may have had the "ee_edit_event" capability added separately.

This PR adds detection for individual caps as well as any roles that possess them.

### DIFF
--- a/core/domain/services/admin/events/editor/EventManagers.php
+++ b/core/domain/services/admin/events/editor/EventManagers.php
@@ -17,6 +17,11 @@ class EventManagers implements EventEditorDataInterface
 {
 
     /**
+     * @var array
+     */
+    private $event_managers = [];
+
+    /**
      * @var Utilities
      */
     private $utilities;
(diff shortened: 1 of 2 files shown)


core/domain/services/admin/events/editor/EventManagers.php
```diff
@@ -39,49 +44,62 @@ public function __construct(Utilities $utilities)
...
-        // now convert to a format that's usable by GQL
-        $event_managers = [];
-        foreach ($event_manager_users as $user) {
-            $GUID             = $this->utilities->convertToGlobalId('user', $user->ID);
-            $event_managers[] = [
-                'id'   => $GUID,
-                'name' => $user->display_name,
-            ];
+        if (empty($this->event_managers)) {
+            [$roles, $capabilities] = $this->getRoleAndCapabilities();
```
manzoorwanijk, line 48: Are you trying to bring JS destructuring to PHP? 🤔 
No idea which PHP syntax is that. 🤷‍♂️

core/domain/services/admin/events/editor/EventManagers.php
```diff
@@ -116,6 +135,12 @@ private function getEventManagerUsers(array $event_manager_roles)
                 $operator = 'OR';
             }
         }
+        foreach ($capabilities as $capability) {
+            // for each capability, add a WHERE clause
+            $SQL     .= $operator . ' u2.meta_value LIKE \'%"' . $capability . '";b:1;%\' ';
```
manzoorwanijk, line 140: `$capability` does not need to be SQL escaped, right?

manzoorwanijk (approved):
Looks great! 🚀  just have a question

core/domain/services/admin/events/editor/EventManagers.php
tn3rb: ya that was added in PHP 7, so can only use it for EDTR related stuff for now:
https://www.php.net/manual/en/migration71.new-features.php#migration71.new-features.symmetric-array-destructuring

core/domain/services/admin/events/editor/EventManagers.php
tn3rb: no the capabilities are already coming from the db and can't have quotes or special characters in their names (afaik 🤔 )

core/domain/services/admin/events/editor/EventManagers.php
manzoorwanijk: Nice, thanks

PR #3209, opened 2020-11-27
Extra Txn Fees Distribution Strategy

tn3rb:
Tony informed me of a situation where additional fees applied to a transaction that exceed the final price for the primary registrant's registration, can result in the inability to make payments when returning to the checkout page.

### steps to reproduce the issue:

- add the following filter to your site:

	- [transaction surcharge gist](https://gist.githubusercontent.com/joshfeck/0e10424fff36bd62bcd72a4c76cd0687/raw/b849985524db28384718b855bb498adb02c6c035/cart_modifier.php)

	- but change Line 23 (unit price) to be $90

- create a ticket for $10

- switch to frontend then purchase and register for one or more tickets

- use an OFFLINE payment method (like Cheque) to finalize

- keep the Thank You page open

- in another tab or window switch back to the admin

- visit the TXN admin, find the transaction for the above and click to view its details

- apply a $50 payment with the following options:

	- for the "Registrations to Apply Payment to" option choose "ALL Registrations"

	- for the "Change Registration Status?" option choose "Leave the Same"

- switch back to your tab that's on the Thank You page

- click to view Payment Options (return to SPCO)

- prior to this branch you'll get 0 registrations and 0 amount left to pay

### This PR:

- adds the following new classes to `core/domain/services/checkout/fees`:

	- `ExtraTxnFeesHandler` coordinates the new logic added in this PR

	- `ApplyExtraFeesToAllRegistrationsEvenly` divides extra fees evenly between regs

	- `ApplyExtraFeesToPrimaryRegistrantOnly` applies extra fees to primary registrant only (duh)

	- `ExtraTxnFeesDistributionStrategyInterface` interface that the last two classes implement

- calls `ExtraTxnFeesHandler::applyExtraFeesToRegistrants()` from within the SPCO Payment Options reg step early on in its logic

- this will result in any extra fees applied to the TXN getting applied to the registrations based on the selected strategy

- there's currently the two strategy options as listed above with a filter allowing a user to change which class is used. If the user wants to, they could even add their own class as long as it implements `ExtraTxnFeesDistributionStrategyInterface`

- with this branch active, Tony's scenario described above will result in the extra $90 being split evenly between however many registrants you added tickets for. So if you purchased 3 tickets, then $30 would be added to each registration's `REG_final_price` field.


core/db_classes/EE_Registration.class.php
```diff
@@ -1115,28 +1115,42 @@ public function pretty_paid()
...
-        } else {
-            return false;
-        }
+        // so if this registration has a status that requires payment
+        return in_array($this->status_ID(), $requires_payment)
+               && ( // AND
+                   // ticket was not free and has not been fully paid for
+                   ($this->final_price() > 0 && $this->remaining() > 0)
+                   // OR this is the primary registrant and there are still monies owing on the TXN
+                   || ($this->is_primary_registrant() && $this->transaction()->remaining() > 0)
```
tn3rb, line 1152: this last line was Tony's fix for forcing the display of the SPCO payment options if the returning registrant is the primary and there are still monies owing on the transaction

core/domain/services/checkout/fees/ExtraTxnFeesHandler.php
```diff
@@ -0,0 +1,103 @@
...
+    {
+        // we're going to calculate if there are any extra fees on the TXN, so start with the full amount
+        $extra_fees = $transaction->total();
+        // then subtract the total for each registration from what's owing for the entire txn
+        foreach ($registrations as $registration) {
+            if ($registration instanceof EE_Registration) {
+                $extra_fees -= $registration->final_price();
+            }
+        }
+        return $extra_fees;
```
manzoorwanijk, line 58: Is this kind of business logic documented anywhere?

manzoorwanijk (approved):
LGTM, just have an unrelated question.

core/domain/services/checkout/fees/ExtraTxnFeesHandler.php
tn3rb: no, unfortunately we don't have any of our business logic documented anywhere (yet), but this is a bit of a new situation since it only arises if you add that extra functionality from the gist. There is some similar logic that distributes extra fees across the line items but that was never applied directly to the registration payments.

tn3rb:
@Pebblo can you plz test this with and without that gist active?

Pebblo:
Tested this with and without the snippet active on a single registration and multiple with multiple registrations.

Both worked as I would expect. Without the snippet payments work as they normally do as far as I can tell.

With the snippet active and testing the above scenario I can make additional payments on the transaction even if there is a payment (or multiple payments) that equal more than the registration values.

I'd say this is good to go.

@sethshoultes Brent asked if at least a couple of people can test this so would you like to give this a spin?

Pebblo:
Actually, I did a little more investigating and the `ApplyExtraFeesToPrimaryRegistrantOnly::class` strategy breaks with the setup from above.

@tn3rb 

The steps to reproduce, use a $10 ticket and the snippet.

I selected 3 tickets and then Invoice on the payment options, that'll load the thank you page.... keep that tab.

In another tab edit the transaction and apply a $80 payment (I left all options on the payment as default, except setting the reg status, I use 'leave the same').

Now back to the previous tab again and click to view your payment options. you'll have $40 to pay. I used PayPal Express to pay.

After that, remove the admin payment from the transaction, on the other tab load your payment options again.

Now you'll see something like this: https://monosnap.com/file/06Q6CN8yyPlnjmlqtmKwDMMnOOfbjl

$20 to pay should be $80, here is the transaction: https://monosnap.com/file/l58G0E0ZtNhMRH5Qrl5kNZ48wCKQGn

Pebblo:
Howdy @tn3rb 

Just checking this hasn't been missed?

kristinoc:
any updates?

PR #3221, opened 2020-12-08
Tweaks for WP User EDTR Integration

tn3rb:
This PR:

- adds a new method to `BaristaFactory` for generating a `Domain` class from an FQCN

-  adds a new method to `EventEspresso\core\services\routing\Route` for initializing Barista logic if Barista plugin is not active


manzoorwanijk (approved):
Looks great 🚀

PR #3226, opened 2020-12-11
Tweak Status Change Notices  closes #3018 closes #3199

### LINKED ISSUE #3018: Sold out status color change
Can we please change this to orange to keep the status colours consistent between the editors?

Current editor status colours: https://monosnap.com/file/amvMaDey0cFQ59SDAUHbxpJvwnGP9u

(Note they don't have to match exactly, 'orange' is close enough orange)

### LINKED ISSUE #3199: Add ability to hide note about status colors changing
Can you please add an option to hide this message?
"Click for an Important Notice regarding Status Color Codes"

Screen Shot 2020-11-05 at 4 53 31 PM

tn3rb:
plz see:

- #3018 
- #3199

this PR:

- allows notices to be dismissed permanently 

- tweaks layout as per Garth's request

- hides notices based on page as per Tony's request

- adds notes about additional message status changes

### DIFF
--- a/core/admin/StatusChangeNotice.php
+++ b/core/admin/StatusChangeNotice.php
@@ -55,7 +55,7 @@ class StatusChangeNotice extends WordPressOption
     private function getCurrentUser(): string
     {
         $user = wp_get_current_user();
-        if (! $user instanceof WP_User) {
+        if (! $user instanceof WP_User || ! $user->exists()) {
             throw new RuntimeException(
                 esc_html__('A valid WP User could not be retrieved.', 'event_espresso')
             );
--- a/core/admin/StatusChangeNotice.php
+++ b/core/admin/StatusChangeNotice.php
@@ -1,0 +1,65 @@
+<?php
+
+namespace EventEspresso\core\admin;
+
+use EventEspresso\core\services\database\WordPressOption;
+use RuntimeException;
+use WP_User;
+
+/**
+ * Class StatusChangeNotice
+ *
+ * @author  Brent Christensen
+ * @package EventEspresso\core\admin
+ * @since   $VID:$
+ */
+class StatusChangeNotice extends WordPressOption
+{
+    public function __construct()
+    {
+        parent::__construct('ee_hide_status_change_notices_for_users', [], false);
+    }
+
+
+    /**
+     * @return int
+     * @throws RuntimeException
+     */
+    public function dismiss(): int
+    {
+        $user      = $this->getCurrentUser();
+        $dismissed = (array) $this->loadOption();
+        if (! in_array($user, $dismissed)) {
+            $dismissed[] = $user;
+        }
+        return $this->updateOption($dismissed);
+    }
+
+
+    /**
+     * @return bool
+     * @throws RuntimeException
+     */
+    public function isDismissed(): bool
+    {
+        $user      = $this->getCurrentUser();
+        $dismissed = (array) $this->loadOption();
+        return in_array($user, $dismissed);
+    }
+
+
+    /**
+     * @return string
+     * @throws RuntimeException
+     */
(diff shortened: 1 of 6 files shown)


core/admin/StatusChangeNotice.php
```diff
@@ -0,0 +1,65 @@
...
+     * @throws RuntimeException
+     */
+    private function getCurrentUser(): string
+    {
+        $user = wp_get_current_user();
+        if (! $user instanceof WP_User) {
+            throw new RuntimeException(
+                esc_html__('A valid WP User could not be retrieved.', 'event_espresso')
+            );
+        }
```
manzoorwanijk, line 62: `wp_get_current_user()` always returns an instance of `WP_User` whether a user exists or not When the user does not exist, `$user->ID` is `0`. A safer approach would to to check for `$user->exists()`:

```suggestion
        if (! $user->exists()) {
            throw new RuntimeException(
                esc_html__('A valid WP User could not be retrieved.', 'event_espresso')
            );
        }
```

core/admin/templates/status_change_notice.template.php
```diff
@@ -1,32 +1,85 @@
...
+                    console.error(response.error);
+                } else if (typeof response.success !== undefined && response.success) {
+                    removeStatusNotice();
+                } else {
+                    console.error("<?php echo $unknown_error_msg; ?>");
+                }
+            }
+        };
+        xhr.send();
+    }
```
manzoorwanijk, line 67: Since `jQuery` is already loaded on admin pages, why not use `$.ajax()`?

core/admin/templates/status_change_notice.template.php
```diff
@@ -142,50 +203,145 @@ function closeStatusNotice() {
...
-        Click for an Important Notice regarding Status Color Codes
+        <?php esc_html_e('Click for an Important Notice regarding Status Color Codes', 'event_espresso'); ?>
     </a>
     <div id="ee-status-change-notice" class="ee-hide-container ee-close-notice">
         <div class="ee-status-change-notice-div">
 
-            <a href="javascript:void();" onclick="closeStatusNotice();" class="ee-close-notice-btn"></a>
-            <h3>Important Notice Regarding Status Color Codes</h3>
-            <p>In order to correct some inconsistencies in our event, datetime, and ticket status color codes, we have
-                made the following changes:</p>
```
manzoorwanijk, line 165: These strings have been there without i18n 😲

manzoorwanijk (approved):
Just have a suggestion and a question. Looks great otherwise. 🚀

core/admin/StatusChangeNotice.php
tn3rb: whale oil... never noticed that before... cool... thnx for that info... will change

core/admin/StatusChangeNotice.php
```diff
@@ -0,0 +1,65 @@
...
+
+
+    /**
+     * @return string
+     * @throws RuntimeException
+     */
+    private function getCurrentUser(): string
+    {
+        $user = wp_get_current_user();
+        if (! $user instanceof WP_User || ! $user->exists()) {
```
tn3rb, line 58: @manzoorwanijk 
I've kept the instanceof check because `wp_get_current_user()` is a pluggable function therefore there is no guarantee that it will always return a `WP_User` object, but I have added your suggestion afterwards to ensure that a valid user is always returned.

Thanks for pointing that function out as I have somehow overlooked that all of these years.

manzoorwanijk: Even better 👍

manzoorwanijk (approved):
🚀

manzoorwanijk (approved):

PR #3230, opened 2020-12-17
Fix Ticket Selector Table Display When # Tickets > Too Many! fixes #3228

### LINKED ISSUE #3228: Hide Extra Details Table Rows for Hidden Tickets
Using his incredibly impressive power to break things, Garth has discovered that creating 92 event dates using the new REM editor will result in a big solid bar appearing at the bottom of the ticket selector table on the frontend. 

The reason is because the bottom border for the "extra details" rows are still appearing for ticket rows that are hidden due to the date selector only showing a limited number of dates. We just never noticed before because the bar would normally have only been a few extra pixels.

tn3rb:
plz see #3228

this PR simply hides and displays the ticket details rows for each ticket when the corresponding ticket selector row is displayed or hidden.

### DIFF
--- a/modules/ticket_selector/assets/ticket_selector.js
+++ b/modules/ticket_selector/assets/ticket_selector.js
@@ -38,6 +38,7 @@ jQuery(document).ready(function ($) {
                 var $ticket_selector_rows = $ticket_selector.find('.tckt-slctr-tbl-tr');
                 $.each($ticket_selector_rows, function () {
                     var $ticket_selector_row = $(this);
+					var $ticket_details_row = $ticket_selector_row.next('.tckt-slctr-tkt-details-tr');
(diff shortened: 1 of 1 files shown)


modules/ticket_selector/assets/ticket_selector.js
```diff
@@ -38,6 +38,7 @@ jQuery(document).ready(function ($) {
                 var $ticket_selector_rows = $ticket_selector.find('.tckt-slctr-tbl-tr');
                 $.each($ticket_selector_rows, function () {
                     var $ticket_selector_row = $(this);
+					var $ticket_details_row = $ticket_selector_row.next('.tckt-slctr-tkt-details-tr');
```
nicolad, line 41: :+1:

nicolad (approved):
Looks good :rocket:

PR #3231, opened 2020-12-17
Hide Ticket Selector Ticket Rows Immediately

tn3rb:
If you create way too many dates then even on a decent server it takes a few seconds for an event listing to load, which results in every ticket row in the TS being displayed until the JS kicks in and hides them.

This PR simply hides ALL TS ticket rows by default, so now for the first few seconds the TS body will be empty until the JS kicks in and displays however many tickets have been configured for display.

### DIFF
--- a/modules/ticket_selector/TicketSelectorRowStandard.php
+++ b/modules/ticket_selector/TicketSelectorRowStandard.php
@@ -147,7 +147,7 @@ class TicketSelectorRowStandard extends TicketSelectorRow
         $ticket_selector_row_html = EEH_HTML::tr(
             '',
             '',
-            "tckt-slctr-tbl-tr {$this->status_class}{$this->ticket_datetime_classes} "
+            "tckt-slctr-tbl-tr ee-hidden-ticket-tr {$this->status_class}{$this->ticket_datetime_classes} "
             . espresso_get_object_css_class($this->ticket)
         );
         $filtered_row_content = $this->getFilteredRowContents();
--- a/modules/ticket_selector/assets/ticket_selector.css
+++ b/modules/ticket_selector/assets/ticket_selector.css
@@ -427,7 +427,14 @@
     top: -2px;
 }
 
-.ee-hidden-ticket-tr {
+
+.tckt-slctr-tbl-tr,
+.tckt-slctr-tkt-details-tr {
+    display: table-row;
+}
+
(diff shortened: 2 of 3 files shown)


modules/ticket_selector/TicketSelectorRowStandard.php
```diff
@@ -147,7 +147,7 @@ public function getHtml()
         $ticket_selector_row_html = EEH_HTML::tr(
             '',
             '',
-            "tckt-slctr-tbl-tr {$this->status_class}{$this->ticket_datetime_classes} "
+            "tckt-slctr-tbl-tr ee-hidden-ticket-tr {$this->status_class}{$this->ticket_datetime_classes} "
```
nicolad, line 150: would have been nice to apply BEM here as well
`tckt-slctr-tbl-tr ee-ticket-tr--hidden` or `tckt-slctr__tbl-tr tckt-slctr__tbl-tr--hidden`

modules/ticket_selector/assets/ticket_selector.css
```diff
@@ -427,7 +427,14 @@
...
 
-.ee-hidden-ticket-tr {
+
+.tckt-slctr-tbl-tr,
+.tckt-slctr-tkt-details-tr {
+    display: table-row;
+}
+
+.tckt-slctr-tbl-tr.ee-hidden-ticket-tr,
+.tckt-slctr-tkt-details-tr.ee-hidden-ticket-tr {
```
nicolad, line 437: is there any chance of using SASS here in the future?

nicolad (approved):
Left a BEM-related suggestion, other than that looks great :rocket: :satellite:

modules/ticket_selector/TicketSelectorRowStandard.php
tn3rb: oh i agree... SASS > CSS imho

modules/ticket_selector/assets/ticket_selector.css
tn3rb: afaik there's likely no easy way to convert all of these legacy assets to use a build system that can convert SASS to CSS, so it just isn't worth it imho

the current plan i'm considering is to convert the frontend presentation layer to utilize GB blocks, which we will be able to use SASS for since those will all be developed on the Barista repo.

PR #3233, opened 2020-12-18
Add Default Reg Status to Event GQL Data Type and Mutator

tn3rb:
Noticed while reviewing https://github.com/eventespresso/barista/pull/514 that out GQL Event object was missing a value for the default registration status.

This PR adds that in so that it can be used to wire up that input in the new Reg Options "meta box" in the EDTR

### DIFF
--- a/core/domain/services/graphql/data/mutations/EventMutation.php
+++ b/core/domain/services/graphql/data/mutations/EventMutation.php
@@ -23,7 +23,7 @@ class EventMutation
      * @return array
      * @throws Exception
      */
-    public static function prepareFields(array $input, $mutation_name)
+    public static function prepareFields(array $input, string $mutation_name): array
     {
         $args = [];
 
--- a/core/domain/services/graphql/data/mutations/EventMutation.php
+++ b/core/domain/services/graphql/data/mutations/EventMutation.php
@@ -28,15 +28,19 @@ class EventMutation
         $args = [];
 
         if (array_key_exists('allowDonations', $input)) {
-            $args['EVT_donations'] = (bool) ($input['allowDonations']);
+            $args['EVT_donations'] = filter_var($input['allowDonations'], FILTER_VALIDATE_BOOLEAN);
         }
 
         if (array_key_exists('allowOverflow', $input)) {
-            $args['EVT_allow_overflow'] = (bool) ($input['allowOverflow']);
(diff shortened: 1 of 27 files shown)


core/domain/services/graphql/data/mutations/EventMutation.php
```diff
@@ -28,35 +28,39 @@ public static function prepareFields(array $input, $mutation_name)
         $args = [];
 
         if (array_key_exists('allowDonations', $input)) {
-            $args['EVT_donations'] = (bool) ($input['allowDonations']);
+            $args['EVT_donations'] = filter_var($input['allowDonations'], FILTER_VALIDATE_BOOLEAN);
```
manzoorwanijk, line 31: Nice 👍

core/domain/services/graphql/types/Event.php
```diff
@@ -79,22 +73,22 @@ public function getFields()
...
+                'dbId',
+                ['non_null' => 'Int'],
+                'ID',
+                esc_html__('The event ID.', 'event_espresso')
             ),
-            new GraphQLInputField(
-                'manager',
+            new GraphQLField(
+                'defaultRegStatus',
                 'String',
```
manzoorwanijk, line 83: IMHO, we should allow a limited set of statuses by using enum instead of string.

manzoorwanijk (commented):
Just have that one suggestion.

core/domain/services/graphql/types/Event.php
tn3rb: great suggestion... will implement, thnx

core/domain/services/graphql/types/Event.php
```diff
@@ -79,22 +72,22 @@ public function getFields()
...
-            new GraphQLInputField(
-                'manager',
-                'String',
-                null,
-                esc_html__('Globally unique event ID for the event manager', 'event_espresso')
+            new GraphQLOutputField(
+                'defaultRegStatus',
+                $this->namespace . 'RegistrationStatusEnum',
+                'default_registration_status',
+                esc_html__('Default Event Registration Status', 'event_espresso')
```
tn3rb, line 84: this is the important change in this PR

manzoorwanijk (approved):
Needed a minor fix. Looks a great

PR #3245, opened 2021-01-05
Fix Hidden Ticket Selector Details Row 

### LINKED ISSUE #3244: Ticket Selector Ticket Rows Hidden if Event Has Only 1 Ticket
see: https://github.com/eventespresso/event-espresso-core/issues/3228

### steps to reproduce

- find event with only one ticket (or create one)

- visit event listing on frontend

- ticket is hidden because of code added in 3228

tn3rb:
This PR fixes an issue introduced by #3228

closes #3244

### DIFF
--- a/modules/ticket_selector/assets/ticket_selector.js
+++ b/modules/ticket_selector/assets/ticket_selector.js
@@ -128,4 +128,40 @@ jQuery(document).ready(function ($) {
 		$( this ).click();
 	} );
 
+
+	$('.display-tckt-slctr-tkt-details').on(
+		'click',
+		function () {
+			// get target element from "this" (the control element's) "rel" attribute
+			var target = $(this).attr("rel");
+			if (target.length) {
+				var $ticket_details_row = $('#' + target + '-dv').closest('.tckt-slctr-tkt-details-tr');
+				$ticket_details_row.removeClass('ee-hidden-ticket-tr');
+				if ($ticket_details_row.length) {
+					$ticket_details_row.removeClass('ee-hidden-ticket-tr');
+				}
+			}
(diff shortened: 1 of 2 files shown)


garthkoyle:
@tn3rb can this be closed?

tn3rb:
LOLZ

no

I just created this like 5 minutes ago

modules/ticket_selector/assets/ticket_selector.js
```diff
@@ -128,4 +128,40 @@ jQuery(document).ready(function ($) {
 		$( this ).click();
 	} );
 
+
+	$('.display-tckt-slctr-tkt-details').on(
+		'click',
+		function () {
+			// get target element from "this" (the control element's) "rel" attribute
+			var target = $(this).attr("rel");
+			if (target.length) {
```
nicolad, line 137:
```suggestion
			if (target?.length) {
```
maybe?

tn3rb: don't think that will work because this code doesn't run through babel/webpack or whatever does all of that magic of converting ES6 to vanilla JS

manzoorwanijk: `target.length` is always defined in that case and defaults to `0`

manzoorwanijk (approved):
Looks great 🚀

modules/ticket_selector/assets/ticket_selector.js
tn3rb: right.. jQuery always returns an object and checking its length is the easiest way to confirm that it is the expected DOM element

PR #3250, opened 2021-01-06
Fix event mutation for empty values

### LINKED ISSUE #3249: Can't Delete Alternative Registration Page URL Registration Option
### steps to reproduce

- find an event that has a URL set for the "Alternative Registration Page" reg option  (or add a URL for an existing event)

- refresh page just to make sure everything is persisted correctly

- click to edit the Alt Reg URL and delete the entire URL

- click away from the field to trigger saving

- refresh the page again

- alt reg URL will still be there

- the only way it appears you can delete the value is by deleting the value then hitting the spacebar before clicking out of the field

manzoorwanijk:
This PR fixes event mutation for empty values which were being ignored by `empty()` check.

Closes #3249


tn3rb (approved):
poifect 💯

PR #3254, opened 2021-01-07
Add datetime editor hooks

Pebblo:
This basically mimics the hooks we have for tickets that all us to hook in an add additional fields to the ticket editor and then save them in the admin.

When a new datetime is added to the editor, it won't have an ID before save() is called on the datetime object, which is why I used `AHEE__espresso_events_Pricing_Hooks___update_datetimes_after_save` right after that was called.

Unless I'm missing something that hook is required if you want to do anything related to the datetimes in the editor from outside of EE core.

---

`AHEE__event_tickets_datetime_attached_tickets_row_template__advanced_details_after_dtt_description` allows me to hook in and add some additional fields in the editor, I chose to do this after the assigned tickets section as its cleaner when you add fields into the editor, see rough example here:

https://monosnap.com/file/ZXYNGtQva24NkZFhrmXScX19EVh8yY

However, I'm not really fussed on where this goes.

The point of these is to allow me to hook in and add fields for zoom information onto the datetimes using extra_meta, those values can then be pulled into messages etc.

This isn't ideal but works out much better for the user than using a Venue object with a single value set, or event custom fields as each DateTime may have a different zoom URL, passcode, etc.

The latter here isn't essential for this but does make it easier. I could hook use `FHEE__EEH_Template__display_template__template_path` to hook into display_template and check the basename for template used here and override it with a custom version containing the hook if its not something you want included.

## Checklist

* [ ] My code is tested.

### DIFF
--- a/caffeinated/admin/new/pricing/espresso_events_Pricing_Hooks.class.php
+++ b/caffeinated/admin/new/pricing/espresso_events_Pricing_Hooks.class.php
@@ -374,6 +374,13 @@ class espresso_events_Pricing_Hooks extends EE_Admin_Hooks
                 }
             }
             $datetime->save();
+            do_action(
+                'AHEE__espresso_events_Pricing_Hooks___update_datetimes_after_save',
+                $datetime,
+                $row,
+                $datetime_data,
+                $data
+            );
             $datetime = $event->_add_relation_to($datetime, 'Datetime');
             // before going any further make sure our dates are setup correctly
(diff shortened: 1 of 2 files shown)


tn3rb (approved):
looks fine, but these won't be long lived unfortunately

Pebblo:
Thanks Brent.

I've tested this and a user has been using it with success for a little while now, so merging in.

PR #3271, opened 2021-01-19
Fix event mutation for event description

manzoorwanijk:
Inline styles by the new editor were being stripped out. So this PR fixes that issue.


tn3rb (approved):
# 👍🏻

tn3rb (approved):

PR #3274, opened 2021-01-19
Fix Exceptions on Post/Page Update

tn3rb:
plz see: https://github.com/eventespresso/eventsmart.com-website/issues/799#issuecomment-762390494

- error described above was because the block loading was disabled on Event Smart here: https://github.com/eventespresso/eventsmart.com-website/issues/486

- and after fixing that issue another error appeared on the new Post page on the frontend because the session appeared to be missing, although I didn't go looking to see if that was intended or not.

This PR:

- checks the `FHEE__EE_System__canLoadBlocks` filter before loading the GutenbergEditor admin route handler

- sets the `EE_Session` dependency on `InvisibleRecaptcha` to be optional and verifies its existence prior to access


knazart (approved):
🏆 Looking good to me.

PR #3277, opened 2021-01-20
Fix REST API Loading

### LINKED ISSUE #3276: EE REST API Broken on EDTR?
Found whilst testing #3261 

Attendee mover activated but testing its functionality failed, the REST request sent to pull in the events on step one showed 404.

So, then I started checking other endpoints locally and noticed that they no longer work.
 
Some quick examples:

https://rc.eventespresso.com/wp-json/ee/v4.8.36
https://rc.eventespresso.com/wp-json/ee/v4.8.36/site_info
https://rc.eventespresso.com/wp-json/ee/v4.8.36/events

I did checkout the commit right before the changes for above were made but that didn't seem to fix it so I don't think this is from the latest changes?

tn3rb:
closes #3276 

Fixes big oops I made with the setup for the `RestApiRequests` route handler class as well as a similar bit of code in the REST API itself.

Basically the EE request class has two functions for detecting whether the current request is for the REST API or not, `isApi()` and `isWordPressApi()`. The former detects requests that are specifically for the EE rest api and the other detects requests to the WP core api which i mistakenly thought would ALSO detect EE REST requests since we piggyback off of the core API, but I was wrong!

This PR: 

- adds the extra check for `isApi()` in `RestApiRequests::matchesCurrentRequest()`
- calls `EED_Core_Rest_Api::set_hooks_both()` from `RestApiRequests::requestHandler()` instead of doing so from the module since the route handler already has the `Request` class loaded
- does a little bit of formatting

### DIFF
--- a/core/libraries/rest_api/controllers/model/Meta.php
+++ b/core/libraries/rest_api/controllers/model/Meta.php
@@ -2,6 +2,7 @@
 
 namespace EventEspresso\core\libraries\rest_api\controllers\model;
 
+use EE_Belongs_To_Relation;
 use Exception;
 use EE_Boolean_Field;
 use EE_Maintenance_Mode;
@@ -10,6 +11,8 @@ use EE_Serialized_Text_Field;
 use EED_Core_Rest_Api;
 use EEM_System_Status;
 use EventEspresso\core\libraries\rest_api\ModelDataTranslator;
+use WP_REST_Request;
+use WP_REST_Response;
 
 /**
  * Controller for handling requests regarding meta info about the models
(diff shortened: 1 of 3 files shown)


core/libraries/rest_api/controllers/model/Meta.php
```diff
@@ -96,7 +99,7 @@ protected function getModelsMetadataEntity()
                 $relation_json = array(
                     'name'   => $relation_name,
                     'type'   => str_replace('EE_', '', get_class($relation_obj)),
-                    'single' => $relation_obj instanceof \EE_Belongs_To_Relation ? true : false,
+                    'single' => $relation_obj instanceof EE_Belongs_To_Relation,
```
knazart, line 102: 😁👍

knazart (approved):
Yes, looks good to me 👍

PR #3284, opened 2021-01-25
Fix Status Change Notice Display

### LINKED ISSUE #3199: Add ability to hide note about status colors changing

tn3rb:
closes #3199

In order to customise what status changes were displayed on various admin pages, I added a css class to the main wrapper div for the notices, then used that to selective hide certain notices. 

Here's the previous css:

```css 
    .espresso_events-page .ee-status-message,
    .espresso_messages-page .ee-status-event {
        display: none;
    }
```

The css class for the page was generated simply by appending "-page" to the admin page slug that is used throughout the admin for each admin page. This worked correctly on the messages admin page, but NOT on the event editor.

Why?

Because the messages admin page uses `espresso_messages` for its page slug but the event editor uses `espresso-events`.
Notice the difference? One is using underscores and the other is using dashes but the css class used underscores for both! 😕 

Took longer than I care to admit to figure this out.

### DIFF
--- a/core/admin/templates/status_change_notice.template.php
+++ b/core/admin/templates/status_change_notice.template.php
@@ -5,14 +5,14 @@
  */
 
 $page_slug          = ! empty($page_slug) ? "{$page_slug}-page" : '';
-$unknown_error_msg  = __(
+$unknown_error_msg  = esc_html__(
     'Oops... an unknown error has occurred on the server and this notice could not be dismissed.',
     'event_espresso'
 );
-$failed_request_msg = __('Request failed. The server returned status code: ', 'event_espresso');
+$failed_request_msg = esc_html__('Request failed. The server returned status code: ', 'event_espresso');
 ?>
 
-<script type="text/javascript">
+<script type='text/javascript'>
     function openStatusNotice() {
         let link = document.getElementById('ee-open-notice-link');
         link.classList.remove('ee-show-link');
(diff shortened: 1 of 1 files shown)


core/admin/templates/status_change_notice.template.php
```diff
@@ -148,7 +148,7 @@ function dismissStatusNotice() {
         background: #f1f1f1;
     }
 
-    .espresso_events-page .ee-status-message,
+    .espresso-events-page .ee-status-message,
```
nicolad, line 151: :+1:

nicolad (dismissed):
Looks great :zap:

nicolad (approved):

tn3rb:
OMG!!!
this is even sillier...

after pushing my fix, I got to wondering why I didn't notice the issue that Seth was seeing, because I **HAD** checked this. Realized at the beginning of my debugging that Seth was looking at the notice in the Event Editor, whereas I was looking at the notice on the Events List page.

But wait a minute... 
the problem was because the page slug for the event editor was "espresso-events" instead of "espresso_events"!?!?
So how come it worked on the Events List but not in the Event Editor, which are technically both using the same admin page controller and therefore **SHOULD** have the same page slug?!?!?

Went back to check the Events List and sure enough, my fix for the Event Editor, now breaks the Events List!

Turns out these pages are SOMEHOW USING DIFFERENT PAGE SLUGS!!!!

Events List uses: `espresso_events`
Event Editor uses: `espresso-events`

Those **SHOULD** be the same 🤦🏻‍♂️ 

Not even going to bother going down that rabbit hole to try and figure out where/how/why this is happening...

nicolad (approved):

PR #3285, opened 2021-01-26
Fix React asset loading

### LINKED ISSUE #3280: EDTR loads the WordPress React instead of the internal bundled React
In #2769, `react` and `react-dom` were bundled to load them externally. Both of those assets and their minified versions were added to `assets/vendor` directory and registered in WP as `react` and `react-dom` handles respectively.

But, due to the older version of deployment from barista, that `assets/vendor` directory got wiped out. But why wasn't there any error? 🤔

The reason for that being the name of the script handles - `react` and `react-dom`, which are already registered by WP with its internal URLs. So even thought the script was registered by EE, but it loaded and still loads the WP React and not the internal EE bundled react.

Two possible solutions:
1. Deregister `'react' ` and `'react-dom'` via `wp_deregister_script()` before registering them. But, this should not happen in blocks domain for obvious reasons.

2.
    - Rename the handles as something like `eventespresso-react` and `eventespresso-react-dom` to avoid the WP handle name.
    - Re-add those assets into `assets/vendor` directory.
    - Update dependency extraction script in barista to use the above handle names.

Notes:
- First solution is the easiest but we will need to be aware of the domain before-hand, which means addons will have to handle this logic separately, because core won't be aware of the addon domains or will it be? 🤔 
- Second solution is a bit complicated but the addons will have known and clear-cut dependencies defined (`eventespresso-react`, `eventespresso-react-dom`) and they won't need to re-run the deregister logic. Otherwise if we can expose that logic via core, it should work.

manzoorwanijk:
This PR fixes the loading of React assets (`react` and `react-dom`).

Fixes #3280


core/domain/services/assets/ReactAssetManager.php
```diff
@@ -35,6 +35,17 @@ class ReactAssetManager extends AssetManager
      */
     public function addAssets()
     {
+        // TODO get these values from router
+        $is_ee_domain = true;
+        $is_blocks_domain = false;
```
manzoorwanijk, line 40: @tn3rb this needs your expertise.

core/services/assets/AssetManager.php
```diff
@@ -156,7 +156,7 @@ public function addVendorJavascript(
         $load_in_footer = true,
         $version = ''
     ) {
-        $dev_suffix = wp_scripts_get_suffix('dev');
```
manzoorwanijk, line 159: `dev` works only for WP versions with `$wp_version` ending with `-src`. More [here](https://developer.wordpress.org/reference/functions/wp_scripts_get_suffix/)

core/domain/services/assets/ReactAssetManager.php
```diff
@@ -35,6 +35,17 @@ class ReactAssetManager extends AssetManager
...
+
+        if (!$is_ee_domain || $is_blocks_domain) {
```
tn3rb, line 42: This will almost certainly not work because it's nearly impossible to detect a "blocks" domain since users could put them nearly anywhere. We will most likely have to load blocks code for every frontend request.
 
 Other domains can be more targeted.

manzoorwanijk: Alright, then may be make it just `if (!$is_ee_domain) return`

tn3rb: well it won't matter now, because we are not implementing any frontend domains, but in the future it may be possible to have an overlap between blocks and other domains, although there are steps we could potentially take to avoid that. My point is that it may be tricky to try and implement a separate version of React than what core is using. Is there any critical reason we need to upgrade our version of React?

manzoorwanijk: > Is there any critical reason we need to upgrade our version of React

Yes, sometimes we rely on the latest React API and it won't be possible for us to be sure that the API will be available for a specific version of WP.

garthkoyle:
Are there any testing notes for this or is it obvious to someone with more technical experience than I have?

tn3rb (approved):

tn3rb:
> Are there any testing notes for this

no, things will either work or fail miserably

manzoorwanijk:
Lets stick to the WP react version for now.

Issue #3287, opened 2021-01-28
Prices not cloned from default tickets.


manzoorwanijk:
Steps:
- Create a default ticket or mark a ticket as default
- Add prices to the ticket
- Check the database for the correct relations being set between the ticket and the prices
- Add a new event
- Check the prices of the cloned default ticket

**Expected** The cloned ticket should have the prices that you added above

**Actual** The ticket has no prices.

tn3rb:
https://github.com/eventespresso/event-espresso-core/pull/3293 addresses some of this issue, but it appears there are still some client side changes required before the prices will show up correctly in the calculator

manzoorwanijk:
> but it appears there are still some client side changes required before the prices will show up correctly in the calculator

Everything looks fine to me. Closing this.

PR #3293, opened 2021-01-28
Moar Betterer Default Price and Tax Cloning

tn3rb:
plz see #3287

Previously the Default Prices DOM Data loader was only retrieving global default prices, which is the same behaviour as the existing legacy editor. The new Default Tickets UI however allows for much finer control over default tickets and pricing but the extra data needed to be added to the appropriate DOM data loaders.

This PR:

- temporarily adds relations between default ticket clones and their existing prices, which are later removed after the prices are cloned. This allows the newly created default tickets added to a new event to be freely mutated without affecting any existing entities

- improves the logic in `EventEspresso\core\domain\services\admin\entities\DefaultPrices` by:

    - cloning the default prices assigned to the ticket as mentioned above
    - assigning global default prices if the new ticket has no other prices
    - retrieves and assigns any global default taxes regardless of the above

- adds a new method to the Prices model for retrieving default taxes

- formats all touched files

### DIFF
--- a/core/db_models/EEM_Price.model.php
+++ b/core/db_models/EEM_Price.model.php
@@ -31,13 +31,13 @@ class EEM_Price extends EEM_Soft_Delete_Base
     {
         require_once(EE_MODELS . 'EEM_Price_Type.model.php');
         $this->singular_item = __('Price', 'event_espresso');
-        $this->plural_item = __('Prices', 'event_espresso');
+        $this->plural_item   = __('Prices', 'event_espresso');
 
-        $this->_tables = array(
+        $this->_tables          = [
             'Price' => new EE_Primary_Table('esp_price', 'PRC_ID'),
-        );
-        $this->_fields = array(
-            'Price' => array(
+        ];
+        $this->_fields          = [
+            'Price' => [
                 'PRC_ID'         => new EE_Primary_Key_Int_Field(
                     'PRC_ID',
                     'Price ID'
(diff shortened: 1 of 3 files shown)


core/db_models/EEM_Price.model.php
```diff
@@ -194,18 +196,43 @@ public function get_all_event_prices($EVT_ID = 0)
...
+                [
+                    'PRC_deleted'    => 0,
+                    'PRC_is_default' => 1,
+                    'Price_Type.PBT_ID' => EEM_Price_Type::base_type_tax
+                ],
+                'order_by' => [
+                    'Price_Type.PRT_order' => 'ASC',
+                    'PRC_order' => 'ASC'
+                ],
+            ]
```
manzoorwanijk, line 234: 👍

core/domain/services/admin/entities/DefaultPrices.php
```diff
@@ -55,66 +65,114 @@ public function __construct(EEM_Price $price_model, EEM_Price_Type $price_type_m
...
+            $default_prices,
+            $remove_existing_relations,
+            $has_base_price,
+            $is_free
+        );
+        if (! $has_base_price) {
+            $new_base_price                            = $this->createNewBasePrice($entity);
+            $this->new_prices[ $new_base_price->ID() ] = $new_base_price;
+        }
+        $this->applyTaxes($entity, $is_free);
```
manzoorwanijk, line 98: Does this modify `$entity` inn place? 🤔

core/domain/services/admin/entities/DefaultPrices.php
```diff
@@ -55,66 +65,114 @@ public function __construct(EEM_Price $price_model, EEM_Price_Type $price_type_m
...
                 $default_price_clone->set('PRC_is_default', false);
                 $default_price_clone->save();
-                $default_price_clone->_add_relation_to($entity, 'Ticket');
-                // verify that a base price has been set
-                $has_base_price = $default_price_clone->is_base_price() ? true : $has_base_price;
-                $new_prices[ $default_price_clone->ID() ] = $default_price_clone;
+                $default_price_clone->_add_relation_to($ticket, 'Ticket');
+                $this->new_prices[ $default_price_clone->ID() ] = $default_price_clone;
+                // then recheck that a base price has been set so that we don't add another one
+                $has_base_price = $is_base_price ? true : $has_base_price;
```
manzoorwanijk, line 172: Cool

manzoorwanijk (approved):
🚀 🥇  Comments are really great.

core/domain/services/admin/entities/DefaultPrices.php
tn3rb: yes but the entity (a ticket in this case) is a clone which was created for the sole purpose of mutating

manzoorwanijk (approved):

PR #3294, opened 2021-02-01
Prepare GQL for better local cache

### LINKED ISSUE #3248: Add Default Tickets to DOM Data
plz see https://github.com/eventespresso/barista/issues/590

- upon loading EDTR:
    - query for ALL default tickets in db
    - add to new field in DOM data

manzoorwanijk:
This PR prepares GQL endpoint for better querying to simplify local caching.

It also Closes #3248

### DIFF
--- a/core/domain/services/graphql/connection_resolvers/TicketConnectionResolver.php
+++ b/core/domain/services/graphql/connection_resolvers/TicketConnectionResolver.php
@@ -83,7 +83,10 @@ class TicketConnectionResolver extends AbstractConnectionResolver
 
             // Use the proper operator.
             if (! empty($input_fields['Datetime.DTT_ID']) && is_array($input_fields['Datetime.DTT_ID'])) {
-                $input_fields['Datetime.DTT_ID'] = ['in', $input_fields['Datetime.DTT_ID']];
+                $input_fields['Datetime.DTT_ID'] = ['IN', $input_fields['Datetime.DTT_ID']];
+            }
+            if (! empty($input_fields['Datetime.EVT_ID']) && is_array($input_fields['Datetime.EVT_ID'])) {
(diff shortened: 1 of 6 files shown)


core/domain/services/graphql/connection_resolvers/TicketConnectionResolver.php
```diff
@@ -113,6 +116,28 @@ public function get_query_args()
...
+                'OR' => [
+                    // use extra OR instead of AND to avoid it getting overridden
+                    'OR' => [
+                        'AND' => [
+                            'TKT_deleted'    => ['IN', [true, false]],
+                            'TKT_is_default' => 1,
+                        ]
+                    ],
+                    'AND' => $where_params,
+                ],
```
tn3rb, line 137: how long did it take you to wrap your head around this?
i find some of this logic in the models to be very UN-intuitive...
so good job figuring this out

tn3rb (approved):
### Code looks GREAT 💯 x ⭐⭐⭐⭐⭐

my only concern is with regards to the branch naming and the base branch...
is this going to break any of the existing query functionality if it is merged into `dev` prior to the corresponding branches in the barista repo?

If so, then we need to wait until those barista changes are ready to go. 

To help clarify that, or in case this new query approach will require additional PRs, 
we should use a `FET` branch to keep any breaking changes out of `dev` until we are fully ready to pull the trigger and add the new functionality.

Now all of that said, this is great and does not require any changes or fixes that I can see, so gladly approving this.

core/domain/services/graphql/connection_resolvers/TicketConnectionResolver.php
manzoorwanijk: Yeah, it definitely IS un-intuitive but we have to deal with it 😄

Issue #3308, opened 2021-01-11
Domain Exception when activating the WP User Add-on while testing EDTR /dev branch of EE4 core


joshfeck:
```
An DomainException was thrown!   code: RouteHandler - addRoute - 90
The following error occurred while trying to handle the
"EventEspresso\WpUser\domain\entities\routing\EspressoEventEditor" route:
The "/eea-wpuser-integration/assets/asset-manifest.json" file was not found or is not readable.
Please verify that the file exists and has appropriate permissions.
click to view backtrace and class/method details
/event-espresso-core/core/services/routing/RouteHandler.php   ( line no: 90 ) 
```

tn3rb:
i can add that

tn3rb:
plz see #3309

tn3rb:
fix merged and ready for testing

joshfeck:
I left a note in the PR https://github.com/eventespresso/event-espresso-core/pull/3309#issuecomment-772969397

tn3rb:
pushed another fix, i had accidentally pasted the wrong version string
ready for testing again

garthkoyle:
@Pebblo will you test this please?

garthkoyle:
I was able to activate `master` branch of WP Users Integration with `dev`.

What do we do with it now @tn3rb ?

PR #3309, opened 2021-02-02
Add WP Users to Incompatible Addons List

tn3rb:
plz see https://github.com/eventespresso/event-espresso-core/issues/3308#issuecomment-772003886

This PR simply makes any version of the WP Users add-on less than `2.1.0.rc.003` incompatible with EE Core


knazart (approved):
🏅

joshfeck:
@tn3rb I tested this and found that the exception still occurs. WP Users 2.1.0.rc.003 is not compatible with EE Core /dev branch. We could bump the version to a higher number, or maybe fix WP Users 2.1.0.rc.003 so it is compatible with EE Core /dev?

tn3rb:
### 🤦🏻‍♂️

I'm so sorry... 
I pasted the wrong version number and basically told the system that the version we wanted to exclude was the minimum required version. Fixed now.

Issue #3313, opened 2021-02-04
Cloned Default tickets have wrong prices.


manzoorwanijk:
Steps:
- Create a couple of default tickets or more
- Add prices to those tickets
- Save
- Add New event
- Check the prices of these cloned tickets

They are incorrect

tn3rb:
I think this may be some bad data in your db

When I first started looking into this, I was able to reproduce the issue!
I would create some default tickets, save them, then create a new event, and the prices would be different.

Initially I looked at the code in `\core\domain\services\admin\entities\DefaultPrices::cloneDefaultPrices()` which is responsible for creating the default prices for default tickets when you create a new event, but that code is not really doing anything except cloning all of the default tickets and prices and creating the appropriate relations.

I eventually discovered that the price order fields appeared to be wonky for all of my default prices.
I followed those rabbit holes and discovered that a bunch of my OLD default prices and tickets had wonky price orders set as well. WHY this affected NEW default tickets, I'm not sure, but here's what I did to get things working:

- deleted ALL of my old default tickets and prices (except for the one free ticket)
- also went directly into the db and deleted anything that looked like default data (again, except for the one free ticket and the default $0 base price)
- refreshed the EDTR
- created new default tickets with multiple price modifiers
- saved those new defaults 
- created a new event after all of updates finished
- the default tickets for this new event were all correct this time

Here's some screenies to show the results (don't have any of the initial wonky prices)
(veuillez excuser les français)

## my three default tickets just after initial creation

![ScreenShot_20210204170508](https://user-images.githubusercontent.com/1751030/106977150-9e6bcc80-670e-11eb-91d1-f89b2ebffa99.png)

## TPC for the "Default Prices in Order" ticket

![ScreenShot_20210204170534](https://user-images.githubusercontent.com/1751030/106977196-b7747d80-670e-11eb-8190-b0d02a9b11df.png)

## TPC for the "Default Prices in Reverse Order" ticket

![ScreenShot_20210204170557](https://user-images.githubusercontent.com/1751030/106977248-cd823e00-670e-11eb-81bf-2e9182f5c88f.png)

## the three new tickets created for my new event
plz note that the ticket prices are both correct and the same as they were when I initially created these defaults

![ScreenShot_20210204170703](https://user-images.githubusercontent.com/1751030/106977391-1d610500-670f-11eb-8ceb-e15ef0b9e6c7.png)

## TPC for the NEW "Default Prices in Reverse Order" ticket
[...]

PR #3314, opened 2021-02-05
check ticket::taxable() b4 taxing

tn3rb:
plz see #3313

this PR:

- checks that default tickets are taxable before applying any default taxes
- bumps order for a default tax to 60 so that it is added after other modifiers


manzoorwanijk (approved):
Looks great 🚀

PR #3316, opened 2021-02-05
Fix deletion of prices when permanently deleting a ticket

### LINKED ISSUE #3315: Deleting a ticket permanently, results in trashing of related default taxes.
Steps:
- Add a default tax from Event Espresso > Pricing
- Open EDTR and edit a ticket
- Add the default tax to the ticket and Save
- Now trash the ticket and then delete permanently
- Goto Event Espresso > Pricing again to see that default tax been trashed.

Problem:
In `EventEspresso\core\domain\services\graphql\mutators`, we have `deleteTicketAndRelations()` which is called for permanent deletions. It internally calls `EE_Ticket::delete_related_permanently('Price')`.

We need to get the related prices and delete only those that are not default ones.

manzoorwanijk:
This PR prevents trashing of default prices when a ticket is permanently deleted via GraphQL.

Fixes #3315

### DIFF
--- a/core/domain/services/graphql/mutators/TicketDelete.php
+++ b/core/domain/services/graphql/mutators/TicketDelete.php
@@ -78,8 +78,10 @@ class TicketDelete extends EntityMutator
      */
     public static function deleteTicketAndRelations($entity)
     {
-        // Remove related prices for the ticket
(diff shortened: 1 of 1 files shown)


tn3rb (approved):
looks good

PR #3318, opened 2021-02-15
Remove espresso-ui-theme as a script dependency and enqueue as a style.

Pebblo:
To reproduce, install the Query Monitor: https://wordpress.org/plugins/query-monitor/

Go to Event Espresso -> Payment methods and check the notices shown in the admin bar by the above plugin.

It'll show a missing script dependency - espresso-ui-theme

Removed 'espresso-ui-theme' from the script dependencies and enqueued the style on it's own.

This is currently only used on the payment logs datepicker.

Go to the logs tab and click on the date filter, in master, you don't get a datepicker at all, now you will.

As it's only actually used on the payment_log view this could be pulled out into a 'load_scripts_styles_payment_log()' method if needed.

* [x] My code is tested.

### DIFF
--- a/admin_pages/payments/Payments_Admin_Page.core.php
+++ b/admin_pages/payments/Payments_Admin_Page.core.php
@@ -234,12 +234,15 @@ class Payments_Admin_Page extends EE_Admin_Page
 
     public function load_scripts_styles()
     {
+        // styles
+        wp_enqueue_style('espresso-ui-theme');
+        // scripts
         wp_enqueue_script('ee_admin_js');
         wp_enqueue_script('ee-text-links');
         wp_enqueue_script(
             'espresso_payments',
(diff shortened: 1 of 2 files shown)


tn3rb (approved):

Pebblo:
During testing this worked fine one one site but not another, the datepicker worked fine but the becuase that site use good old English date format dd/mm/yy it broke the query.

So I've set the format of the datepicker to just always use mm/dd/yy now.

Pebblo:
Side note, the failed test don't appear to be from changes I've made so I'm staying out of those.

tn3rb (approved):

Pebblo:
Retested this and confirmed it worked with a couple of different date formats just to be sure.

Merging in.

PR #3336, opened 2021-03-04
Fix tax calculation with discounts

knazart:
This fixes the tax calculation when discount's are used.

For more info see: https://github.com/eventespresso/eea-square-gateway/pull/3#issuecomment-789178680

## How has this been tested
Requires quires the Promotions add-on.
* [ ] Purchase one or a few tickets that have applicable taxes. Apply any discount. The tax amounts should be counted on the discounted item price.
* [ ] Try mixing up the ticket prices and taxable options on them.
* [ ] Add more than one tax item to the mix.
* [ ] Try applying more than one type of a discount.
* [ ] Finally purchase taxable tickets without applying discounts.
* [ ] Test different payment methods we have. QuickBooks and PP Express for example.

## Checklist

* [x] My code is tested.

### DIFF
--- a/core/db_classes/EE_Line_Item.class.php
+++ b/core/db_classes/EE_Line_Item.class.php
@@ -1554,6 +1554,9 @@ class EE_Line_Item extends EE_Base_Class implements EEI_Line_Item
                     }
                 } elseif ($child_line_item->type() === EEM_Line_Item::type_sub_total) {
                     $total += $child_line_item->taxable_total();
+                } elseif ($child_line_item->is_percent()) {
+                    // Taxes should be charged on the discounted price.
+                    $total += $child_line_item->total();
                 }
             }
         }


core/db_classes/EE_Line_Item.class.php
```diff
@@ -1554,6 +1554,9 @@ public function taxable_total()
                     }
                 } elseif ($child_line_item->type() === EEM_Line_Item::type_sub_total) {
                     $total += $child_line_item->taxable_total();
+                } elseif ($child_line_item->is_percent()) {
+                    // Taxes should be charged on the discounted price.
+                    $total += $child_line_item->total();
```
tn3rb, line 1559: hmmm... this doesn't feel right to me...
why is the line item in question not being caught by the first clause above?

ie:
```php 
if ($child_line_item->type() === EEM_Line_Item::type_line_item && $child_line_item->is_taxable()) {
```

which already has some logic for handling precentages.

This new clause looks like it is going to add all percentage sub line items to the running total, which doesn't seem correct.

Can you plz show me an example of the kind of line item that would make it into this clause?

tn3rb (commented):
mmmmm... dunno... i would like to investigate this more and am not sure about this change.

tn3rb:
Found your problem...

in the promotions admin:

ScreenShot_20210304122112

core/db_classes/EE_Line_Item.class.php
knazart: >why is the line item in question not being caught by the first clause above?

Because a discounts is `! is_taxable()`. But now I think I know why, after you pointed out that promotions setting...

> This new clause looks like it is going to add all percentage sub line items to the running total

Aren't all the line items of type `"percentage"` considered to only be promotions ? Tax amounts in percent aren't counted as `true` in `is_percent()`. What else could be `true` in `is_percent()` beside a promotion ? Maybe I'm missing some. But a few PMs do use `is_percent()` to filter out only promos.

Anyways, this change is not suitable here, I totally agree. Your doubts did help figure this out. Thank you.

knazart:
This change is not suitable here. Closing.

PR #3341, opened 2021-03-22
Use model prop directly instead of get_model()

tn3rb:
Need to perform a bunch of code cleanup on the models before embarking on the next big project.

This PR changes references to `get_model()` within `EEM_Base_Class` to instead use the `_model` prop directly which has now been set immediately during construction. 

WHY?

because `get_model()` gets called constantly in nearly every major function in `EEM_Base_Class` and was unnecesarily perfomring some extra checks as well as repeatedly setting the timezone (which could be happening multiple times during a single function call).

The public `get_model()` method has been left since it is still used externally to retrieve the model for a model object. It has however also been slightly modified because every call to it was going through the process of setting the timezone on the model and model fields... using the SAME timezone property that was already set, which makes no sense whatsoever. Here's the line of code i removed:

```php
$this->_model->set_timezone($this->_timezone);
```

which would effectively be equivalent to:

```php
$this->_timezone = $this->_timezone;
```

but with some extra steps thrown in there for fun.

### DIFF
--- a/core/db_classes/EE_Base_Class.class.php
+++ b/core/db_classes/EE_Base_Class.class.php
@@ -140,8 +140,8 @@ abstract class EE_Base_Class
     {
         $className = get_class($this);
         do_action("AHEE__{$className}__construct", $this, $fieldValues);
-        $model = $this->get_model();
-        $model_fields = $model->field_settings(false);
+        $this->_model = $this->get_model();
+        $model_fields = $this->_model->field_settings(false);
         // ensure $fieldValues is an array
         $fieldValues = is_array($fieldValues) ? $fieldValues : array($fieldValues);
         // verify client code has not passed any invalid field names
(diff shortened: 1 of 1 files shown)


core/db_classes/EE_Base_Class.class.php
```diff
@@ -2027,8 +2017,6 @@ public function get_model()
         if (! $this->_model) {
             $modelName = self::_get_model_classname(get_class($this));
             $this->_model = self::_get_model_instance_with_name($modelName, $this->_timezone);
-        } else {
-            $this->_model->set_timezone($this->_timezone);
```
tn3rb, line 2031: this was nonsense

knazart: 🤨

knazart (approved):
Yes, a 🏅, why not.

manzoorwanijk (approved):
Nice

PR #3343, opened 2021-03-24
Allow payment methods to pass arrays as values in redirect_args

Pebblo:
Requested here: https://eventespresso.com/topic/generate-a-registration-form-from-2d-array/

Allows you to pass an array of values in redirect_args to be passed to the merchant.

Quick examples:

`$redirect_args['Products'] = array('Product1', 'Product2');`

Currently outputs:

`<input type="hidden" name="Products" value="array">`

With this change, outputs:

`<input type="hidden" name="Products[]" value="Product1">
<input type="hidden" name="Products[]" value="Product2">`

## How has this been tested
I used PayPal standard to test this.

In `\event-espresso-core-reg\payment_methods\Paypal_Standard\EEG_Paypal_Standard.gateway.php`

Add this:

`$redirect_args['Products'] = array('Product1', 'Product2');`

It can be anywhere before `$payment->set_redirect_args($redirect_args);`

Now select PayPal standard as a payment method.

When you click to proceed through to PayPal the page will show 'Forwarding to secure payment provider', at that point, hit the stop button on your browser.

Inspect the 'Click Here' button with dev tools and you'll now see the form data.

This is how it should show:

https://monosnap.com/file/tKxUM6pXdgAHHBf2SVhP7olcmpI7D0

----

This change leaves all other values working as they currently would, unless its an array (which wouldn't be working anyway).

## Checklist

* [x] My code is tested.

### DIFF
--- a/core/db_classes/EE_Payment.class.php
+++ b/core/db_classes/EE_Payment.class.php
@@ -689,7 +689,7 @@ class EE_Payment extends EE_Base_Class implements EEI_Payment
 
 
     /**
-     * Converts a 1d array of key-value pairs into html hidden inputs
+     * Converts a 2d array of key-value pairs into html hidden inputs
      * and returns the string of html
      *
      * @param array $args key-value pairs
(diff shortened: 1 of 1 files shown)


core/db_classes/EE_Payment.class.php
```diff
@@ -700,12 +700,23 @@ protected function _args_as_inputs($args)
...
+                             . '"/>';
+                    }
+                } else {
+                    $html .= EEH_HTML::nl(0)
                          . '<input type="hidden" name="'
                          . $name
                          . '" value="'
                          . esc_attr($value)
                          . '"/>';
+                }
```
tn3rb, line 719: how about extract this logic into a new method that can then recurse over values if passed an array:

```suggestion
                
				$html .= $this->generateInput($name, $value);
```

then add this below the current method:

```php
			private function generateInput($name, $value)
    		{
        		if (is_array($value)) {
            		$html = '';
            		$name = "{$name}[]";
            		foreach ($value as $array_value) {
                		$html .= $this->generateInput($name, $array_value);
            		}
            		return $html;
       			}
        		return EEH_HTML::nl()
               		. '<input type="hidden" name="' . $name . '"'
               		. ' value="' . esc_attr($value) . '"/>';
    		}
```

which has less duplication and means any future changes needed for that inputcan be made in one place

tn3rb (changes requested):
Looks good but can be better by reducing duplication through the power of [recursion](https://www.google.com/search?rlz=1C1ZKTG_enCA776CA776&sxsrf=ALeKk00FnLu481XnM2Ql5gcvZD4q6xaiWw:1616603029571&q=recursion&spell=1&sa=X&ved=2ahUKEwi4xMfKq8nvAhXFGjQIHQT1C_AQBSgAegQIARAw&biw=1024&bih=469)

Pebblo:
Thanks, Brent!

Applied and tested the above.

tn3rb (approved):
yup, thnx for making that change 👍🏻

PR #3345, opened 2021-03-25
EEM_Base Code CLeanup

tn3rb:
this PR:

- auto-formats to PSR12
- imports all dependencies via use statements
- declares all exceptions
- removes all unused vars or those matching defaults
- fixes formatting for @see links
- moves all constant declarations to top of class
- uses esc_html__()

because of the high number of changes made due to auto-formatting,
it may be beneficial to review this PR by looking at each commit individually, since they focus on doing one thing at a time

### DIFF
--- a/core/db_models/EEM_Base.model.php
+++ b/core/db_models/EEM_Base.model.php
@@ -570,7 +570,7 @@ abstract class EEM_Base extends EE_Base implements ResettableInterface
         if (! did_action('AHEE__EE_System__load_espresso_addons')) {
             throw new EE_Error(
                 sprintf(
-                    __(
+                    esc_html__(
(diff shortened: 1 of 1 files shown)


manzoorwanijk (approved):
Woah! That one file is 6500+ LoC 😱

manzoorwanijk:
I presume that the base branch will be tested when merged back to the actual base branch.

tn3rb:
> Woah! That one file is 6500+ LoC 😱

yup... Michael wrote amazing code except for majorly violating SRP.

that said, there's a LOT of code in EE that's like that to a degree because we had to support PHP 5.2 for so long (thanks WP) and therefore didn't have namespacing and easy class loading. 

> I presume that the base branch will be tested when merged back to the actual base branch.

yes I'll do a PR up for that after all of the smaller PRs are merged into `models-cleanup/main`

PR #3347, opened 2021-03-26
EE_Base_Class Code Cleanup

tn3rb:
This PR:

- auto-formats to PSR12
- removes all unnecessary vars and params
- renames vars & params
- removes unthrown exceptions
- fixes optional param before required

because of the high number of changes made due to auto-formatting,
it may be beneficial to review this PR by looking at each commit individually, since they focus on doing one thing at a time


knazart (approved):
Yup, for sure, 🏅

manzoorwanijk (approved):
I am glad you are doing it in small PR. Thanks

PR #3358, opened 2021-04-02
Fix Model Timezone Setting 

tn3rb:
This PR fixes a whole bunch of nonsense that I noticed happening in the models with respect to timezones. There appeared to be no sane / logical approach to ensuring that timezones were set properly and efficiently. To be honest, I'm not even sure how some of it wasn't completely crashing things because there appeared to be a lot recursion happening where setting a models' timezone would result in it setting the timezone for its related fields, which in turn would set their related models timezone, which included the original timezone that started the entire update stampede. This resulted in the timezone for a model getting updated hundreds upon hundreds (thousands maybe?) of times during a simple request.

With the changes in this PR< timezones should only get set once and only update related objects if the timezone has yet to be set or their is an obviously intentional change occurring.

My apologies for the big PR< but the timezone stuff need to all be changed in one fell swoop

### DIFF
--- a/admin_pages/events/Events_Admin_Page.core.php
+++ b/admin_pages/events/Events_Admin_Page.core.php
@@ -34,17 +34,25 @@ class Events_Admin_Page extends EE_Admin_Page_CPT
 
 
     /**
-     * This will hold the event model instance
-     *
      * @var EEM_Event $_event_model
      */
     protected $_event_model;
+
+    /**
+     * @var EEM_Datetime $datetime_model
+     */
+    protected $datetime_model;
+
+    /**
+     * @var EEM_Ticket $ticket_model
+     */
+    protected $ticket_model;
 
 
     /**
      * @var EE_Event
      */
-    protected $_cpt_model_obj = false;
+    protected $_cpt_model_obj;
 
 
     /**
@@ -768,11 +776,9 @@ class Events_Admin_Page extends EE_Admin_Page_CPT
      * When a user is creating a new event, notify them if they haven't set their timezone.
      * Otherwise, do the normal logic
      *
-     * @return string
      * @throws EE_Error
      * @throws InvalidArgumentException
      * @throws InvalidDataTypeException
-     * @throws InvalidInterfaceException
      */
     protected function _create_new_cpt_item()
     {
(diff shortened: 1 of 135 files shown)


admin_pages/events/Events_Admin_Page.core.php
```diff
@@ -1240,17 +1275,17 @@ protected function _default_tickets_update(EE_Event $evtobj, $data)
                 }
             }
             $DTM->save();
-            $DTT = $evtobj->_add_relation_to($DTM, 'Datetime');
+            $DTM = $evtobj->_add_relation_to($DTM, 'Datetime');
```
tn3rb, line 1278: `_add_relation_to()` returns the object you passed in, and since objects are always passed by reference in PHP... THERE WAS NO REASON OR LOGIC BEHIND CHANGING THE VAR NAME !!!

admin_pages/events/Events_Admin_Page.core.php
```diff
@@ -1240,17 +1275,17 @@ protected function _default_tickets_update(EE_Event $evtobj, $data)
...
+                $DTM->set('DTT_EVT_end', $DTM->get('DTT_EVT_start'));
+                $DTM = EEH_DTT_Helper::date_time_add($DTM, 'DTT_EVT_end', 'days');
+                $DTM->save();
             }
-            // now we got to make sure we add the new DTT_ID to the $saved_dtts array  because it is possible there was a new one created for the autosave.
-            $saved_dtt = $DTT;
-            $success = ! $success ? $success : $DTT;
+            // now we got to make sure we add the new DTT_ID to the $saved_dtts array
+            //  because it is possible there was a new one created for the autosave.
+            $saved_dtt = $DTM;
```
tn3rb, line 1288: THEN the same object as above is "copied" to the `$saved_dtt` var as well as to ...
 the`$success` var? which can also be a boolean ?!?!?

admin_pages/events/Events_Admin_Page.core.php
```diff
@@ -1265,8 +1300,6 @@ protected function _default_tickets_update(EE_Event $evtobj, $data)
             // trim inputs to ensure any excess whitespace is removed.
             $tkt = array_map('trim', $tkt);
             if (empty($tkt['TKT_start_date'])) {
-                // let's use now in the set timezone.
-                $now = new DateTime('now', new DateTimeZone($evtobj->get_timezone()));
```
tn3rb, line 1269: this was being done inside a loop !
moved it above

core/services/helpers/datetime/AbstractHelper.php
```diff
@@ -11,27 +11,84 @@
...
+     */
+    private $default_timezone_string = '';
+
+
+    /**
+     * AbstractHelper constructor.
+     */
+    public function __construct()
+    {
+        add_action('update_option_timezone_string', [$this, 'updateDefaultTimezoneString'], 10, 2);
```
tn3rb, line 25: IF the timezone is somehow changed during a request, then this resets the cached value here so that the new value gets picked up.

core/db_classes/EE_Base_Class.class.php
```diff
@@ -1506,16 +1512,37 @@ public function get_timezone()
...
+        // OR the existing timezone is already set and the validated value is the same as the existing timezone
+        if (
+            $set_in_progress
+            || (
+                ! empty($this->_timezone)
+                && (
+                    empty($timezone) || $valid_timezone === $this->_timezone
+                )
+            )
+        ) {
```
manzoorwanijk, line 1540: 👍

manzoorwanijk (approved):
Great effort 🥇

PR #3359, opened 2021-04-06
Request Routing Fixes

tn3rb:
While trying to figure out what was wrong in https://github.com/eventespresso/eea-automated-upcoming-event-notifications/pull/19 I discovered that the way a "previous dev" was doing things caused some issues with some of the request routing logic in core.

To protect against misuse I added some fixes to some of the request routing logic, but ended up settling on a better solution for the AUEN plugin. The changes I made here, although not necessary anymore, are still good so I would like to keep them.

I also discovered a couple other issues while going through error logs which I've also fixed here.

The changes include:

- loading the REST API module on all admin routes because the data nodes used for generating GQL data reference that class.

- fixed a minor data type error for the default value of `$route_request_type` in `core/services/routing/RouteHandler.php`

- added a couple of new methods to the Request class for resolving paths as well as a couple of other tweaks

### DIFF
--- a/core/services/request/Request.php
+++ b/core/services/request/Request.php
@@ -178,11 +178,7 @@ class Request implements InterminableInterface, RequestInterface, ReservedInstan
     public function setRequestParam($key, $value, $override_ee = false)
     {
         // don't allow "ee" to be overwritten unless explicitly instructed to do so
-        if (
-            $key !== 'ee'
-            || ($key === 'ee' && empty($this->request['ee']))
-            || ($key === 'ee' && ! empty($this->request['ee']) && $override_ee)
-        ) {
+        if ($key !== 'ee' || empty($this->request['ee']) || $override_ee) {
             $this->request[ $key ] = $value;
         }
     }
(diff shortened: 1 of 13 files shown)


core/services/request/Request.php
```diff
@@ -178,11 +178,7 @@ public function requestParams()
     public function setRequestParam($key, $value, $override_ee = false)
     {
         // don't allow "ee" to be overwritten unless explicitly instructed to do so
-        if (
-            $key !== 'ee'
-            || ($key === 'ee' && empty($this->request['ee']))
-            || ($key === 'ee' && ! empty($this->request['ee']) && $override_ee)
-        ) {
+        if ($key !== 'ee' || empty($this->request['ee']) || $override_ee) {
```
tn3rb, line 181: just simplified this conditional

core/services/request/Request.php
```diff
@@ -452,6 +450,10 @@ public function requestUri($relativeToWpRoot = false)
             // fallback sanitization if the above fails
             $request_uri = wp_sanitize_redirect($this->server['REQUEST_URI']);
         }
+        if($remove_query_params) {
+            $request_uri_parts = explode('?', $request_uri);
+            $request_uri = reset($request_uri_parts);
+        }
```
tn3rb, line 456: strips out the query params from the current request URI

core/services/request/Request.php
manzoorwanijk, line 181: Nice 👍

manzoorwanijk (dismissed):
Looks great 🚀

knazart (dismissed):
👍

Issue #3381, opened 2021-05-06
Add New `visibility` Field to Ticket Model


tn3rb:
plz see: https://github.com/eventespresso/barista/issues/473#issuecomment-833736960

- instead of a boolean field named adminOnly, we will use a string/enum field named visibility

- values for this new field could include things like:
    - admin UI only - only shows in Ticket Selector in the admin
    - admins only - only appears for logged in admin users
    - members only - only appears for logged in users
    - public - appears everywhere [default]

PR #3384, opened 2021-05-06
Add Ticket Visibility Field to Server Side GraphQL

tn3rb:
plz see #3381 

This PR adds the new ticket visibility field to Server Side GraphQL:
- query
- mutation handler
- type

### DIFF
--- a/core/db_models/EEM_Ticket.model.php
+++ b/core/db_models/EEM_Ticket.model.php
@@ -19,15 +19,20 @@ class EEM_Ticket extends EEM_Soft_Delete_Base
      *  TICKET_VISIBILITY_ADMIN_UI_ONLY - only displayed in the admin, never publicly
      *  TICKET_VISIBILITY_NONE          - will not be displayed anywhere
      */
-    public const TICKET_VISIBILITY_PUBLIC        = 100;
+    public const TICKET_VISIBILITY_PUBLIC_KEY          = 'PUBLIC';
+    public const TICKET_VISIBILITY_PUBLIC_VALUE        = 100;
 
-    public const TICKET_VISIBILITY_MEMBERS_ONLY  = 200;
+    public const TICKET_VISIBILITY_MEMBERS_ONLY_KEY    = 'MEMBERS_ONLY';
+    public const TICKET_VISIBILITY_MEMBERS_ONLY_VALUE  = 200;
 
-    public const TICKET_VISIBILITY_ADMINS_ONLY   = 300;
+    public const TICKET_VISIBILITY_ADMINS_ONLY_KEY     = 'ADMINS_ONLY';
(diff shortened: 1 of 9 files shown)


core/domain/services/graphql/types/Ticket.php
```diff
@@ -301,6 +301,12 @@ public function getFields()
                 null,
                 esc_html__('Ticket Creator ID', 'event_espresso')
             ),
+            new GraphQLField(
+                'visibility',
+                'Int',
+                'visibility',
+                esc_html__('Where the ticket can be viewed throughout the UI', 'event_espresso')
+            ),
```
manzoorwanijk, line 309: Since the client needs to be aware of server-side integer values, it's better to make it `Enum`, instead of `Int`

manzoorwanijk: Here is the difference between the two.
image


With enum, mutations are pretty nice
image

manzoorwanijk:
Here is the difference between `Int` and `Enum`.

image


With enum, mutations are pretty nice
image

core/domain/services/graphql/enums/TicketVisibilityEnum.php
```diff
@@ -0,0 +1,52 @@
...
+    {
+        $ticket_model = new ReflectionClass(EEM_Ticket::class);
+
+        $constants = $ticket_model->getConstants();
+
+        $enum = [];
+
+        foreach ($constants as $constant => $value) {
+            // if the constant name starts with 'TICKET_VISIBILITY_'
+            if (0 === strpos($constant, 'TICKET_VISIBILITY_')) {
```
tn3rb, line 44: Yoda conditionals use we do not!

core/db_models/EEM_Ticket.model.php
```diff
@@ -383,10 +379,73 @@ public function get_tickets_with_IDs(array $ticket_IDs)
...
+
+    /**
+     * @return array
+     */
+    public function getTicketVisibilityKeyValuePairs(): array
+    {
+        $ticket_visibility = [];
+        foreach ($this->ticket_visibility as $key => $ticket_visibility_option) {
+            if (isset($ticket_visibility_option['value'])) {
+                $ticket_visibility[ $key ] = $ticket_visibility_option['value'];
```
manzoorwanijk, line 438:
```suggestion
                $name = str_replace('TICKET_VISIBILITY_', '', $key);
                $ticket_visibility[ $name ] = ['value' => $ticket_visibility_option['value']];
```

tn3rb: i added new constants for those keys, like:

```php
public const TICKET_VISIBILITY_PUBLIC_KEY          = 'PUBLIC';
```

manzoorwanijk (approved):

PR #3390, opened 2021-05-08
Improve Event Manager Role Check in Ticket Selector

tn3rb:
This PR:

- extracts Event Manager logic from `core/domain/services/admin/events/editor/EventManagers` into separate class that should be more usable by others

- extracts Current User just added to `DisplayTicketSelector` into new class for same reasons as above BUT uses the new EventManagers class to determine if user is an Event Manager instead of simply checking for 'Administrator' role.

- updates references to the above logic and does some extra formatting and stuff
### testing

- LOG OUT

- log in again as an "Event Administrator" User

- visit registration page on frontend

- [ ] confirm that "Public", "Members Only", and "Admins Only" tickets are visible

- go to event lists admin page and click on the "Registrations" link under the name of the event from above

- click "Add New Registration" button at top of screen

- [ ] confirm that "Public", "Members Only", "Admins Only", and "Admin UI Only" tickets are visible


core/domain/entities/users/CurrentUser.php
```diff
@@ -0,0 +1,91 @@
...
+
+
+    public function setCurrentUser(): void
+    {
+        if (! $this->current_user instanceof WP_User) {
+            $this->current_user = wp_get_current_user();
+            $event_manager_roles    = array_keys($this->event_managers->roles());
+            $current_user_roles     = $this->current_user->roles;
+            $this->is_event_manager = ! empty(array_intersect($event_manager_roles, $current_user_roles));
+            $this->is_logged_in     = $this->current_user->ID !== 0;
```
manzoorwanijk, line 58: You can directly use this without the need for comparison.
```suggestion
            $this->is_logged_in     = $this->current_user->exists();
```

manzoorwanijk (approved):
Looks great 🚀

core/domain/entities/users/CurrentUser.php
tn3rb: cool thnx... forgot about that!

PR #3408, opened 2021-05-28
Extract Attendee Info Reg Step Form Generation into New Classes

tn3rb:
Before we can switch over to forms generated using the new reg form builder, we need to prepare SPCO to handle the transition. 

This PR extracts the attendee information reg step form generation into new classes that are then loaded as needed. When the new reg form builder is successfully generating data, then we can detect if an event has a form section UUID set and generate the form using a new set of classes based on the new data. The new system should be more more performant due to the dramatic reduction in queries needed to populate the reg form.


core/domain/services/registration/form/CountryOptions.php
```diff
@@ -0,0 +1,121 @@
...
+            : $this->country_model->get_all_active_countries();
+        // start with an empty option
+        $country_options = ['' => ''];
+        if (! empty($countries)) {
+            foreach ($countries as $country) {
+                if ($country instanceof EE_Country) {
+                    $country_options[ $country->ID() ] = $country->name();
+                }
+            }
+        }
```
manzoorwanijk, line 104: 👍

core/domain/services/registration/form/LegacyAutoCopyAttendeeInfoForm.php
```diff
@@ -0,0 +1,28 @@
...
+                apply_filters(
+                    'FHEE__EE_SPCO_Reg_Step_Attendee_Information__auto_copy_attendee_info__template_args',
+                    []
+                ),
+                true,
+                true
+            )
+        );
+    }
+}
```
manzoorwanijk, line 28: missing EOL

core/domain/services/registration/form/LegacyCopyAttendeeInfoForm.php
```diff
@@ -0,0 +1,87 @@
...
+                        'html_id'                 => 'spco-copy-attendee-chk-' . $registration->reg_url_link(),
+                        'html_class'              => 'spco-copy-attendee-chk ee-do-not-validate',
+                        'display_html_label_text' => false,
+                    ]
+                );
+            }
+        }
+        return $copy_attendee_info_inputs;
+    }
+}
```
manzoorwanijk, line 87: EOL

manzoorwanijk (approved):
Looks great 👍

core/domain/services/registration/form/LegacyAutoCopyAttendeeInfoForm.php
tn3rb: oops thnx, will fix

core/domain/services/registration/form/LegacyCopyAttendeeInfoForm.php
tn3rb: thnx, will fix

PR #3419, opened 2021-06-03
Extract Attendee Info Reg Form Processing into New Classes

tn3rb:
This PR:

- extracts logic from `EE_SPCO_Reg_Step_Attendee_Information::process_reg_step()` into the following new classes:

	- `RegFormHandler`
	- `RegFormInputHandler`
	- `RegFormAttendeeFactory`
	- `RegistrantData`

- extracts a bit more logic from the `RegForm` into `PrivacyConsentCheckboxForm`

- extracts most of the dependency registration from `EE_Dependency_Map` into `RegFormDependencyHandler`

- refactors a few other form related things after the above changes

- cleans up a few other things


core/EE_Dependency_Map.core.php
```diff
@@ -802,31 +802,8 @@ protected function _register_core_dependencies()
...
-            'EventEspresso\core\domain\services\registration\form\v1\CountryOptions' => [
-                null,
-                'EEM_Answer' => EE_Dependency_Map::load_from_cache,
-                'EEM_Country' => EE_Dependency_Map::load_from_cache,
-            ],
-            'EventEspresso\core\domain\services\registration\form\v1\StateOptions' => [
-                null,
-                'EEM_State' => EE_Dependency_Map::load_from_cache,
+            'EventEspresso\core\domain\services\registration\form\v1\RegFormDependencyHandler' => [
+                'EE_Dependency_Map' => EE_Dependency_Map::load_from_cache,
```
manzoorwanijk, line 806: 👍

core/domain/services/registration/form/v1/RegFormHandler.php
```diff
@@ -0,0 +1,327 @@
...
+        $this->initializeInputHandler();
+    }
+
+
+    private function initializeInputHandler()
+    {
+        $reg_form = $this->checkout->current_step->reg_form;
+        $required_questions = $reg_form instanceof RegForm ? $reg_form->requiredQuestions() : [];
+        $this->input_handler = LoaderFactory::getShared(
+            RegFormInputHandler::class,
```
manzoorwanijk, line 85: Now you like `::class` 😄

core/domain/services/registration/form/v1/RegFormHandler.php
```diff
@@ -0,0 +1,327 @@
...
+     * @return bool
+     * @throws EE_Error
+     * @throws ReflectionException
+     */
+    private function processRegFormData(EE_Registration $registration, string $reg_url_link, array $reg_form_data): bool
+    {
+        if (isset($reg_form_data[ $reg_url_link ])) {
+            // do we need to copy basic info from primary attendee ?
+            $copy_primary = isset($reg_form_data[ $reg_url_link ]['additional_attendee_reg_info'])
+                                  && absint($reg_form_data[ $reg_url_link ]['additional_attendee_reg_info']) === 0;
```
manzoorwanijk, line 278: Why not simply this?
```suggestion
            $copy_primary = empty($reg_form_data[ $reg_url_link ]['additional_attendee_reg_info']);
```

manzoorwanijk (approved):
Looks great 🚀

knazart (approved):
👍

core/domain/services/registration/form/v1/RegFormHandler.php
tn3rb: i'm guessing (since that code was probably originally written close to a decade ago) that `additional_attendee_reg_info` might hold value with different data types... like `bool`, `int`, `string` so running `absint()` on a string value (and then comparing against `0`) is very different than just using `empty()`.

what you've suggested might very well work fine, but i'm really hesitant to modify the logic in older code if I don't have to because there may be some weird edge case that some seemingly unnecessary line of code is accounting for. I've tried really hard to keep things operating exactly the same throughout this refactor because I don't want the existing code to get broken :\

tn3rb:
thnx guys

PR #3423, opened 2021-06-09
Add ramda globally

manzoorwanijk:
This PR adds/registers [`ramda`](https://ramdajs.com/) as a global dependency which can be used by all the client-side code.


tn3rb (dismissed):
Looks Great!

tn3rb (approved):

PR #3424, opened 2021-06-10
Set GRP_ID from the EE_Message in the message queue... on the EE_Message that is used.

Pebblo:
See this thread: https://eventespresso.com/topic/stylesheets-not-loaded-for-emailed-tickets-but-loaded-in-wp-backend/

If you have a custom message pack set on a custom message template, EE doesn't load that template pack when viewing the front end ticket links.

In the admin, when you view the tickets you'll get a link like this:

http://ee4.test/?ee=msg_url_trigger&snd_msgr=html&gen_msgr=html&message_type=ticketing&context=registrant&token=1-d463d9fb18eefea75cc536d2eb6f77c9&GRP_ID=27&id=0

The above works because it sets the GRP_ID, however, the user gets a link like this:

http://ee4.test/?ee=ee-txn-tickets-approved-url&token=1-d463d9fb18eefea75cc536d2eb6f77c9

That uses the registration to work out what is needed and for the most part, it works, in fact, it pulls all of the correct details and sets up the data in a 'message queue' in the background. However, the message queue system then basically reuses the details from that message queue (which is a bunch of EE_Message objects) to recreate EE_Message objects from and set values..... GRP_ID is NOT one of those values set and this fixes that.

@tn3rb when I mentioned this in slack you mentioned it didn't sound right and I agree, this could do with some refactoring as it just seems really inefficient currently. Maybe I'm missing something with it all but for now, this fixes the above.

## Checklist

* [x] My code is tested.

### DIFF
--- a/core/libraries/messages/EE_Message_To_Generate_From_Queue.php
+++ b/core/libraries/messages/EE_Message_To_Generate_From_Queue.php
@@ -32,7 +32,7 @@ class EE_Message_To_Generate_From_Queue extends EE_Message_To_Generate
         if ($this->valid()) {
             $this->_message->set_content($this->_get_content());
             $this->_message->set_subject($this->_get_subject($custom_subject));
-            $this->_message->set_GRP_ID($this->_get_GRP_ID());
+            $this->_message->set_GRP_ID($this->getGroupIdFromMessageRepo());
         }
     }
 
@@ -92,7 +92,7 @@ class EE_Message_To_Generate_From_Queue extends EE_Message_To_Generate
      * for the single EE_Message aggregate object returned by get_EE_Message
      * @return int;
      */
-    protected function _get_GRP_ID()
+    protected function getGroupIdFromMessageRepo()
     {
         $this->queue->get_message_repository()->rewind();
         if ($this->queue->get_message_repository()->valid()) {
--- a/core/libraries/messages/EE_Message_To_Generate_From_Queue.php
+++ b/core/libraries/messages/EE_Message_To_Generate_From_Queue.php
@@ -32,6 +32,7 @@ class EE_Message_To_Generate_From_Queue extends EE_Message_To_Generate
         if ($this->valid()) {
             $this->_message->set_content($this->_get_content());
             $this->_message->set_subject($this->_get_subject($custom_subject));
+            $this->_message->set_GRP_ID($this->_get_GRP_ID());
         }
     }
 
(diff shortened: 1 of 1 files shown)


core/libraries/messages/EE_Message_To_Generate_From_Queue.php
```diff
@@ -32,6 +32,7 @@ public function __construct($messenger_name, $message_type_name, EE_Messages_Que
         if ($this->valid()) {
             $this->_message->set_content($this->_get_content());
             $this->_message->set_subject($this->_get_subject($custom_subject));
+            $this->_message->set_GRP_ID($this->_get_GRP_ID());
```
tn3rb, line 35: this is technically fine but it _looks_ like you are passing the value returned from a getter to a setter, which would normally do absolutely nothing except eat up CPU cycles.

ie: getters and setters are _usually_named like:

 ```php
public function getProperty() {
	return $this->property;
}

public function setProperty( $value ) {
	$this->property = $value;
}
```

and your line of code here looks like:

```php
$obj->setProperty( $obj->getProperty() )
```

which would set the value of `property` using its current value (ie: no change).

Your `_get_GRP_ID()` function isn't a typical getter for the `GRP_ID` property, so it might be better to give it a name that clarifies that reality, such as:

```php
protected function getGroupIdFromMessageRepo()
```

tn3rb (commented):
Your changes are technically correct but this line here:

```php
 $this->_message->set_GRP_ID($this->_get_GRP_ID());
```

_looks_ like you are passing the value returned from a getter to a setter, which would normally do absolutely nothing except eat up CPU cycles.

ie: getters and setters are _usually_named like:

 ```php
public function getProperty() {
	return $this->property;
}

public function setProperty( $value ) {
	$this->property = $value;
}
```

and your line of code here looks like:

```php
$obj->setProperty( $obj->getProperty() )
```

which would set the value of `property` using its current value (ie: no change).

Your `_get_GRP_ID()` function isn't a typical getter for the `GRP_ID` property though, so it would be better to give it a name that clarifies that reality, such as:

```php
protected function getGroupIdFromMessageRepo()
```

Pebblo:
Yeah, I wasn't thrilled with the naming either tbh.

I just ended up following what is there as I'm pretty sure that whole `EE_Message_To_Generate_From_Queue` class could do with being refactored.

tn3rb (approved):
cool thnx

Pebblo:
Tested and confirmed this fixed the issue and messages seem to be working as expected.

Merging in.

PR #3427, opened 2021-06-11
Use getType() in place of getClass(), then use the getName() method.

Pebblo:
I was testing PHP8 and my logs start filling with tons of these, so fixed it and figured I might as push the change.

## How has this been tested
Use PHP8 and confirm no more of these notices are shown:

`PHP Deprecated:  Method ReflectionParameter::getClass() is deprecated in /wp-content/plugins/event-espresso-core-reg/core/services/container/Mirror.php on line 162`

If this broke something I think we'd know pretty quick, so that's it.

## Checklist

* [ ] My code is tested.


tn3rb (approved):
looks good... 
i pushed a potential fix for a failing unit test, although I'm not quite sure why it was failing because a proper default value **should** have been passed

PR #3432, opened 2021-06-16
Fix Webpack Build Errors

tn3rb:
When @Pebblo when to build a core release after merging #3431, he received the following error from the buildmachine:

 ```
Warning: Command failed: cd buildsrc/event-espresso-core-reg/./&&npm install&&npm run build
[webpack-cli] Failed to load '/event-espresso-core-reg/webpack.prod.js' config
[webpack-cli] TypeError: merge is not a function at /event-espresso-core-reg/webpack.prod.js:70:18
```

this was an EZ fix as it only required destructuring the `merge` function from the `webpack-merge` library as that was no longer exported as a default (see: https://github.com/eventespresso/event-espresso-core/commit/dffca7a0b51fc1c68e86e3591e86ac22f159818d). 

This however was followed by a new error:

```
(node:11283) [DEP_WEBPACK_EXTERNALS_FUNCTION_PARAMETERS] DeprecationWarning: 
The externals-function should be defined like ({context, request}, cb) => { ... }
/bin/i18n-map-extractor.js:76
                        Array.from( modules ),
                              ^

TypeError: Cannot convert undefined or null to object
    at Function.from (<anonymous>)
    at parseSourcesToMap (/bin/i18n-map-extractor.js:76:10)
    at /bin/i18n-map-extractor.js:140:5
    at /node_modules/lodash/lodash.js:4943:15
    at forEach (/node_modules/lodash/lodash.js:9410:14)
    at processChunks (/bin/i18n-map-extractor.js:131:3)
    at /bin/i18n-map-extractor.js:106:8
    at Hook.eval [as call] (
		eval at create(/node_modules/webpack/node_modules/tapable/lib/HookCodeFactory.js:19:10),
		<anonymous>:23:16
	)
    at Hook.CALL_DELEGATE [as _call] (/node_modules/webpack/node_modules/tapable/lib/Hook.js:14:14)
    at Compilation.seal (/node_modules/webpack/lib/Compilation.js:2393:36)
```

THIS was a far trickier issue to deal with. As far as I can tell it is partially because the [webpack-i18n-extractor-plugin](https://github.com/chuckdumont/webpack-i18n-extractor-plugin) does not support Webpack v5.

I was able to trace through the code and find where the issue was occurring and have added an utterly fugly fix (see: https://github.com/eventespresso/event-espresso-core/commit/8ee3906d5ed256a3ed82d650a084846d83f496b9#diff-67d01fa46b112a974132436d4f5a8a2282edd260f3c91e5ed49a3965aa8c5843R139). Basically it appears that the list of modules requiring i18n extraction was moved, but I have no confidence in my fix being correct. The build appears to run now, but again, I'm not sure if things are correct or not.

@manzoorwanijk can you plz investigate this issue and if need be, fix anything that is not up to your standards?


tn3rb:
this was fixed by Manzoor in #3433

PR #3433, opened 2021-06-17
Fix Webpack Build

manzoorwanijk:
This PR fixes Webpack build error by downgrading Webpack to v4 and the linked dependencies to the compatible versions.

Here are the dependencies that need to be frozen until we are able to upgrade to webpack 5

```json
"autoprefixer": "^9.8.6",
"cssnano": "^4.1.11",
"postcss-loader": "^4.2.0",
"webpack": "^4.44.2",
"webpack-assets-manifest": "^4.0.6",
```

See #3432


tn3rb (approved):
thank you so much Manzoor, I know you would have a better fix for this :)

PR #3434, opened 2021-06-22
Form Builder | Add GraphQL schema for Form Section queries

manzoorwanijk:
This PR adds the GQL schema for Form Builder section queries.

Next steps:
- Add the missing fields to `Form_Section` model - `adminLabel`, `description`, `showDescription`, `showName`
- Remove unnecessary fields - `FSC_relation`

image

### DIFF
--- a/core/db_classes/EE_Form_Section.class.php
+++ b/core/db_classes/EE_Form_Section.class.php
@@ -225,7 +225,7 @@ class EE_Form_Section extends EE_Base_Class
      * @throws EE_Error
      * @throws ReflectionException
      */
-    public function wpUser(): int
+    public function wp_user(): int
     {
         return $this->get('FSC_wpUser');
     }
--- a/core/db_models/EEM_Form_Section.model.php
+++ b/core/db_models/EEM_Form_Section.model.php
@@ -88,7 +88,7 @@ class EEM_Form_Section extends EEM_Form_Element
                 'FSC_appliesTo' => new EE_Enum_Text_Field(
                     'FSC_appliesTo',
                     esc_html__(
-                        'Form user types that this form section should be presented to. Values correspond to the EEM_Form_Section::APPLIES_TO_* constants.',
+                        'Form user type that this form section should be presented to. Values correspond to the EEM_Form_Section::APPLIES_TO_* constants.',
                         'event_espresso'
                     ),
                     false,
@@ -133,7 +133,7 @@ class EEM_Form_Section extends EEM_Form_Element
                 'FSC_status'    => new EE_Enum_Text_Field(
                     'FSC_status',
                     esc_html__(
-                        'Whether form section is active, archived, trashed, or used as a default on new forms. Values correspond to the EEM_Form_Section::STATUS_TO_* constants.',
+                        'Whether form section is active, archived, shared, trashed, or used as a default on new forms. Values correspond to the EEM_Form_Section::STATUS_TO_* constants.',
                         'event_espresso'
                     ),
                     false,
--- a/core/domain/services/graphql/connections/RootQueryFormSectionsConnection.php
+++ b/core/domain/services/graphql/connections/RootQueryFormSectionsConnection.php
@@ -1,0 +1,62 @@
+<?php
+
+namespace EventEspresso\core\domain\services\graphql\connections;
+
+use EEM_Form_Section;
+use EventEspresso\core\domain\services\graphql\connection_resolvers\FormSectionConnectionResolver;
+use EventEspresso\core\domain\services\graphql\abstracts\AbstractRootQueryConnection;
+use Exception;
+
+/**
+ * Class RootQueryFormSectionsConnection
+ * Description
+ *
+ * @package EventEspresso\core\domain\services\graphql\connections
+ * @author  Manzoor Wani
+ * @since   $VID:$
+ */
+class RootQueryFormSectionsConnection extends AbstractRootQueryConnection
+{
+
+
+    /**
+     * FormSectionConnection constructor.
+     *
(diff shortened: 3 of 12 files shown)


core/db_classes/EE_Form_Section.class.php
```diff
@@ -225,7 +225,7 @@ public function setStatus(string $status)
      * @throws EE_Error
      * @throws ReflectionException
      */
-    public function wpUser(): int
+    public function wp_user(): int
```
tn3rb, line 228: is this so things match up with WP GQL?

core/db_models/EEM_Form_Section.model.php
```diff
@@ -133,7 +133,7 @@ protected function __construct(Element $element, $timezone = null)
                 'FSC_status'    => new EE_Enum_Text_Field(
                     'FSC_status',
                     esc_html__(
-                        'Whether form section is active, archived, trashed, or used as a default on new forms. Values correspond to the EEM_Form_Section::STATUS_TO_* constants.',
+                        'Whether form section is active, archived, shared, trashed, or used as a default on new forms. Values correspond to the EEM_Form_Section::STATUS_TO_* constants.',
```
tn3rb, line 136: right! forgot about that 👍🏻 thnx

core/domain/services/graphql/connections/RootQueryFormSectionsConnection.php
```diff
@@ -0,0 +1,62 @@
...
+use EventEspresso\core\domain\services\graphql\abstracts\AbstractRootQueryConnection;
+use Exception;
+
+/**
+ * Class RootQueryFormSectionsConnection
+ * Description
+ *
+ * @package EventEspresso\core\domain\services\graphql\connections
+ * @author  Manzoor Wani
+ * @since   $VID:$
```
tn3rb, line 16: can remove `Description` if class is purpose is "self-explanatory" enough

core/domain/services/graphql/connections/RootQueryFormSectionsConnection.php
```diff
@@ -0,0 +1,62 @@
...
+     */
+    public function __construct(EEM_Form_Section $model)
+    {
+        $this->model = $model;
+    }
+
+
+    /**
+     * @return array
+     * @since $VID:$
```
tn3rb, line 35: can remove these on methods if the entire class is new and has an `@since` tag.

this is only needed on methods when a new method is added to an existing class

core/domain/services/graphql/connections/RootQueryFormSectionsConnection.php
```diff
@@ -0,0 +1,62 @@
...
+    public function config()
+    {
+        return [
+            'fromType'           => 'RootQuery',
+            'toType'             => $this->namespace . 'FormSection',
+            'fromFieldName'      => lcfirst($this->namespace) . 'FormSections',
+            'connectionTypeName' => "{$this->namespace}RootQueryFormSectionsConnection",
+            'resolve'            => [$this, 'resolveConnection'],
+        ];
+    }
```
tn3rb, line 46: we can also use PHP 7 return types in the function declarations, **IF** the method does not override a parent function that is not declared the same way.  ex:

```php
public function config(): array
```

is better, but NOT if this method is declared on a parent class (or in an interface) like:

```php
public function config()
```

because then the declarations don't match and hilarity will ensue in the form of exceptions :\

core/services/form/meta/Element.php
```diff
@@ -21,6 +21,11 @@ class Element
      */
     public const STATUS_DEFAULT = 'default';
 
+    /**
+     * indicates that a copy of the form section will be saved for use in other events but not loaded by default
+     */
+    public const STATUS_SHARED = 'shared';
```
tn3rb, line 27: thank you for adding this
# 🙇🏻‍♂️

tn3rb (approved):
### Great Job 💯 🥇 

I had a couple of super minor trivial comments pertaining to formatting, which you can feel free to leave to me to clean up as I see fit. Similarly you can also feel free to clean up any of my JS code I push to Barista.

tn3rb:
going to pull this branch and fix the conflicts after i rebased `FET/reg-form-builder/main`

core/db_classes/EE_Form_Section.class.php
manzoorwanijk: Yes, all other model objects use `wp_user`

core/domain/services/graphql/connections/RootQueryFormSectionsConnection.php
manzoorwanijk: OK, thanks

core/domain/services/graphql/connections/RootQueryFormSectionsConnection.php
manzoorwanijk: Yeah, I am not used to strict types in PHP yet. Need to start using those 😄

core/services/form/meta/Element.php
```diff
@@ -40,6 +45,7 @@ public function __construct()
                 Element::STATUS_ACTIVE   => esc_html__('Active', 'event_espresso'),
                 Element::STATUS_ARCHIVED => esc_html__('Archived', 'event_espresso'),
                 Element::STATUS_DEFAULT  => esc_html__('Default', 'event_espresso'),
+                Element::STATUS_DEFAULT  => esc_html__('Shared', 'event_espresso'),
```
tn3rb: keys were duplicated here by accident, but it's fixed

manzoorwanijk: Oops! Thank you

PR #3440, opened 2021-06-24
Form Builder | Add form section query args to GQL

manzoorwanijk:
This PR adds query args to filter form sections in GQL schema.

PR #3441, opened 2021-06-24
Form builder | Form Element GraphQL Schema

manzoorwanijk:
This PR adds GQL schema for Form Element queries.

### DIFF
--- a/core/db_models/EEM_Form_Input.model.php
+++ b/core/db_models/EEM_Form_Input.model.php
@@ -25,7 +25,7 @@ use EventEspresso\core\services\request\RequestInterface;
  * FIN_status    string    ex: 'archived'
  * FIN_type    string
  * FIN_wpUser    int
- * 
+ *
  * Model Fields:
  *
  * @author  Brent Christensen
@@ -141,7 +141,7 @@ class EEM_Form_Input extends EEM_Form_Element
                 'FIN_options'     => new EE_JSON_Field(
                     'FIN_options',
                     esc_html__(
-                        'Options for ENUM type inputs like checkboxes, radio buttons, select inputs, etc',
+                        'JSON string of options for ENUM type inputs like checkboxes, radio buttons, select inputs, etc.',
                         'event_espresso'
                     ),
                     true,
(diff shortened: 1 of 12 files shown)


core/domain/services/graphql/connection_resolvers/FormSectionConnectionResolver.php
```diff
@@ -128,14 +128,13 @@ public function get_query_args(): array
...
-        ];
         return $this->sanitizeWhereArgsForInputFields(
             $where_args,
-            $arg_mapping
+            [
+                'appliesTo' => 'FSC_appliesTo',
+                'belongsTo' => 'FSC_belongsTo',
+                'status'    => 'FSC_status',
+            ]
         );
```
manzoorwanijk, line 138: As per the suggestion [here](https://github.com/eventespresso/event-espresso-core/pull/3440#discussion_r658075179).

core/db_models/EEM_Form_Input.model.php
```diff
@@ -141,7 +141,7 @@ protected function __construct(Element $element, InputTypes $input_types, $timez
                 'FIN_options'     => new EE_JSON_Field(
                     'FIN_options',
                     esc_html__(
-                        'Options for ENUM type inputs like checkboxes, radio buttons, select inputs, etc',
+                        'JSON string of options for ENUM type inputs like checkboxes, radio buttons, select inputs, etc.',
```
tn3rb, line 144: 👍🏻

core/db_models/EEM_Form_Input.model.php
```diff
@@ -188,19 +188,31 @@ protected function __construct(Element $element, InputTypes $input_types, $timez
...
-                        'event_espresso'
+                    esc_html(
+                        sprintf(
+                            /* translators: 1 class name */
+                            __(
+                                'Whether form input is active, archived, trashed, or used as a default on new forms. Values correspond to the %s constants.',
+                                'event_espresso'
+                            ),
+                            'EEM_Form_Input::STATUS_*'
+                        )
```
tn3rb, line 199: why split out the escaping?

does this not do the same?

```suggestion
                        sprintf(
                            /* translators: 1 class name */
                            esc_html__(
                                'Whether form input is active, archived, trashed, or used as a default on new forms. Values correspond to the %s constants.',
                                'event_espresso'
                            ),
                            'EEM_Form_Input::STATUS_*'
                        )
```

manzoorwanijk: I thought it is better to use escaping at the end. If you prefer the other way, I am totally fine with that.

manzoorwanijk: Also, it's possible that the placeholder text may have some special characters sometimes, so escaping that will be skipped if we escape early.

tn3rb: ya I'm fine with either way too. I just wanted to know if there was some deeper reason for doing it this way that I was unaware of. 

i don't think it's necessary in this case since doing the escaping at the end would only additionally affect the variables injected into the placeholders. In this case `EEM_Form_Input::STATUS_*` will not require any escaping, but if it contained any HTML then we would likely want to do it the way you have here.

but again, I'm fine with either way, so no sense in changing it

tn3rb (approved):
# POIFECT 👌🏻 

great job

PR #3442, opened 2021-06-24
Make WP Users Add-on Work

tn3rb:
This PR simply reduces the min add-on version required set in the `IncompatibleAddonHandler`.
This allows older versions of WP User to continue working with the EDTR...

### **HOWEVER**...

PLEASE NOTE that NONE of the WP User Event Editor functionality (assigning caps to tickets) will work if the user has activated the Advanced Editor (EDTR). So in other words, they can use WP Users version `2.1.0.rc.003` with core dev branch (EDTR) and everything will work the same as long as they also continue to use the Legacy Event Editor. If the switch to use the new EDTR, then they just won't be able to set caps on tickets, but everything else _should_ work as SPCO remains unchanged.

### DIFF
--- a/admin_pages/events/Events_Admin_Page.core.php
+++ b/admin_pages/events/Events_Admin_Page.core.php
@@ -985,25 +985,24 @@ class Events_Admin_Page extends EE_Admin_Page_CPT
     protected function _events_overview_list_table()
     {
         do_action('AHEE_log', __FILE__, __FUNCTION__, '');
-        $this->_template_args['after_list_table']                           =
-            ! empty($this->_template_args['after_list_table'])
-                ? (array) $this->_template_args['after_list_table']
-                : [];
-        $this->_template_args['after_list_table']['view_event_list_button'] = EEH_HTML::br()
-                                                                              . EEH_Template::get_button_or_link(
-                get_post_type_archive_link('espresso_events'),
-                esc_html__('View Event Archive Page', 'event_espresso'),
-                'button'
-            );
-        $this->_template_args['after_list_table']['legend']                 =
-            $this->_display_legend($this->_event_legend_items());
-        $this->_admin_page_title                                            .= ' '
-                                                                               . $this->get_action_link_or_button(
+        $after_list_table                           = [];
+        $after_list_table['view_event_list_button'] = EEH_HTML::br();
+        $after_list_table['view_event_list_button'] .= EEH_Template::get_button_or_link(
+            get_post_type_archive_link('espresso_events'),
+            esc_html__('View Event Archive Page', 'event_espresso'),
+            'button'
+        );
+        $after_list_table['legend'] = $this->_display_legend($this->_event_legend_items());
+        $this->_admin_page_title    .= ' ' . $this->get_action_link_or_button(
                 'create_new',
                 'add',
                 [],
                 'add-new-h2'
             );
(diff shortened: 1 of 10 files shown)


core/services/addon/api/IncompatibleAddonHandler.php
```diff
@@ -28,7 +28,7 @@ public function deactivateIncompatibleAddons()
         $this->deactivateIncompatibleAddon(
             'WP Users Integration',
             'EE_WPUSERS_VERSION',
-            '2.1.1.rc.001',
+            '2.1.0.rc.003',
```
Pebblo, line 31: Can this not be set to the current release version?

And... having said that, if the current release version works with the legacy editor as is, do we need to set this as incompatible?

Pebblo:
Left a comment but will probably be easier to view this here.

So the idea behind this change is to prevent the requirement for the WP User Integration add-on need to have an update released on/around the same time as EDTR to keep things 'working'. With the current version number set in the above `2.1.0.rc.003` we would still need to push a required update before EDTR was released.

It's better but pretty much the same for the folks that don't pay attention to the updates, and yes there are a fair few of them.

I manually changed this to 2.0.17.p so that the current release version would remain enabled and it works as expected with the legacy editor (which is what I was expecting from the change) with a slight 'bug' on EDTR (it still shows the visibility field but then you can't really do anything with it).

So, can we change this to be the current release version of the WP User integration? I'm fine with anyone using anything older than that getting a deactivation and forced to update.... they should have updated AGES ago anyway.

If your not keen on doing it that way, can we change this deactivation to maybe just a notice that they need to update the WP user add-on to use it on the advanced editor if they are using 2.0.17.p still? (reading that back it sounds like the same thing tbh)

tn3rb:
> Can this not be set to the current release version?

ya that's fine with me, I thought I was setting it back to the previous version

> I manually changed this to 2.0.17.p

did you push that change?

tn3rb:
just looked at the git history and the very first "incompatible" version for WP Users was `2.1.0.rc.003`. So if you don't want that, then it makes more sense to simply remove this restriction and deal with whatever incompatibility it was protecting against.

admin_pages/events/Events_Admin_Page.core.php
```diff
@@ -985,25 +985,24 @@ public function extra_permalink_field_buttons($return, $id, $new_title, $new_slu
...
+        $this->_admin_page_title    .= ' ' . $this->get_action_link_or_button(
                 'create_new',
                 'add',
                 [],
                 'add-new-h2'
             );
+        $this->_template_args['after_list_table']   = array_merge(
+            $this->_template_args['after_list_table'],
+            $after_list_table
+        );
```
Pebblo, line 1005: `$this->_template_args['after_list_table']` is now an empty string, so array_merge() is throwing a warning on the event list table

Pebblo (changes requested):

admin_pages/events/Events_Admin_Page.core.php
tn3rb: kk we can just type cast that value, ie:

```php
        $this->_template_args['after_list_table']   = array_merge(
            (array) $this->_template_args['after_list_table'],
            $after_list_table
        );
```

Pebblo (approved):

Pebblo (approved):

PR #3445, opened 2021-06-25
Fix Delete Queries for Tables with String IDs

tn3rb:
@manzoorwanijk discovered that attempting to delete a model object from a table that uses strings for its PK column results in the following error:

```
WordPress database error Unknown column 'ckqcbw3j100iv3b9hvrtyfrl3' in 'where clause' for query 
DELETE Form_Input FROM  wp_esp_form_input AS Form_Input  WHERE Form_Input.FIN_UUID IN(ckqcbw3j100iv3b9hvrtyfrl3)
```

because the string IDs are not getting wrapped in quotes.

This was happening because `EEM_Base::_build_query_part_for_deleting_from_columns_and_values()` is constructing its own `IN` clause for the query with no regard for the PK data type.

This PR:

- uses `EEM_Base::_construct_in_value()` in `EEM_Base::_build_query_part_for_deleting_from_columns_and_values()` instead of constructing its own `IN` clause

- moves the use of `array_unique()` into `EEM_Base::_construct_in_value()` because that just makes sense to do that there (and likely should have been done there already as a good precaution)

### DIFF
--- a/core/db_models/EEM_Base.model.php
+++ b/core/db_models/EEM_Base.model.php
@@ -2169,9 +2169,7 @@ abstract class EEM_Base extends EE_Base implements ResettableInterface
         } elseif ($this->has_primary_key_field()) {
             $query = array();
             foreach ($ids_to_delete_indexed_by_column as $column => $ids) {
-                // make sure we have unique $ids
-                $ids = array_unique($ids);
-                $query[] = $column . ' IN(' . implode(',', $ids) . ')';
+                $query[] = $column . ' IN' . $this->_construct_in_value($ids, $this->_primary_key_field);
             }
             $query_part = ! empty($query) ? implode(' AND ', $query) : $query_part;
         } elseif (count($this->get_combined_primary_key_fields()) > 1) {
@@ -4475,7 +4473,6 @@ abstract class EEM_Base extends EE_Base implements ResettableInterface
     }
 
 
-
     /**
      * Takes an array or a comma-separated list of $values and cleans them
      * according to $data_type using $wpdb->prepare, and then makes the list a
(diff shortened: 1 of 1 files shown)


core/db_models/EEM_Base.model.php
```diff
@@ -2169,9 +2169,7 @@ protected function _build_query_part_for_deleting_from_columns_and_values(array
         } elseif ($this->has_primary_key_field()) {
             $query = array();
             foreach ($ids_to_delete_indexed_by_column as $column => $ids) {
-                // make sure we have unique $ids
-                $ids = array_unique($ids);
-                $query[] = $column . ' IN(' . implode(',', $ids) . ')';
+                $query[] = $column . ' IN' . $this->_construct_in_value($ids, $this->_primary_key_field);
```
manzoorwanijk, line 2172: 👍

core/db_models/EEM_Base.model.php
```diff
@@ -4490,29 +4487,27 @@ public function _construct_between_value($values, $field_obj)
...
-            $a_field = array_shift($all_fields);
+            $first_field    = reset($all_fields);
             $main_table = $this->_get_main_table();
-            $cleaned_values[] = "SELECT "
-                                . $a_field->get_table_column()
-                                . " FROM "
-                                . $main_table->get_table_name()
-                                . " WHERE FALSE";
+            $prepped[]  = "SELECT {$first_field->get_table_column()} FROM {$main_table->get_table_name()} WHERE FALSE";
         }
```
manzoorwanijk, line 4509: 👍

manzoorwanijk (approved):
Great, thank you for the fix

knazart (approved):

PR #3446, opened 2021-06-25
Form Builder | Fix Form Section delete mutation

manzoorwanijk:
This PR fixes the error when deleting a section before deleting its related elements/inputs. It sets the form section deletion to delete the related form elements. The client-side code does not need send the request to delete the related elements.


tn3rb (approved):
yup looks great

PR #3447, opened 2021-06-25
Consolidate Fields / Composite Form Models

tn3rb:
This PR:

 - updates DMS to add consolidated form fields
 - renames model to Form Element and updates usages
 - consolidates fields utilizing the following composites:
 	- Attributes
 	- FormLabel
 	- HelpText
 	- Required
 - renames and cleans up a few other things


core/services/form/meta/Attributes.php
```diff
@@ -0,0 +1,200 @@
...
+     */
+    private $attributes = [];
+
+    /**
+     * @var array
+     */
+    private $attribute_types = [
+        'accept'          => 'string',
+        'accesskey'       => 'string',
+        'alt'             => 'string',
```
tn3rb, line 41: i just copied all of these in here for now, but we can decide which to keep if we want to restrict things. I didn't add any of the fields that pertain to scripts, such as onChange, etc

core/db_classes/EE_Form_Section.class.php
```diff
@@ -161,6 +151,36 @@ public function setAppliesTo(string $user_type)
...
+     * @return Attributes
+     * @throws EE_Error
+     * @throws ReflectionException
+     */
+    public function attributes(): ?Attributes
+    {
+        if (! $this->attributes instanceof Attributes) {
+            $this->attributes = Attributes::fromJson($this->get('FSC_attributes'));
+        }
+        return $this->attributes;
```
manzoorwanijk, line 166: Super cool 👍

core/db_classes/EE_Form_Section.class.php
```diff
@@ -161,6 +151,36 @@ public function setAppliesTo(string $user_type)
...
+     * @param Attributes $attributes
+     * @throws EE_Error
+     * @throws ReflectionException
+     */
+    public function setAttributes(Attributes $attributes)
+    {
+        // set local object
+        $this->attributes = $attributes;
+        // then pass to model as an array which will get converted to JSON by the model field
+        $this->set('FSC_attributes', $attributes->toArray());
```
manzoorwanijk, line 180: 👍

core/db_classes/EE_Form_Section.class.php
```diff
@@ -186,98 +206,130 @@ public function setBelongsTo(string $parent_UUID)
...
 
 
     /**
-     * @param string $html_class
+     * @return EE_Form_Element[]
      * @throws EE_Error
      * @throws ReflectionException
      */
-    public function setHtmlClass(string $html_class)
+    private function getFormElements(): array
```
manzoorwanijk, line 224: Should this be `private`?

core/domain/services/graphql/connection_resolvers/FormElementConnectionResolver.php
```diff
@@ -21,17 +21,17 @@ public function get_loader_name(): string
...
      * @throws InvalidDataTypeException
      * @throws InvalidInterfaceException
      * @throws ReflectionException
      */
     // phpcs:ignore PSR1.Methods.CamelCapsMethodName.NotCamelCaps
-    public function get_query(): EEM_Form_Input
+    public function get_query(): EEM_Form_Element
     {
-        return EEM_Form_Input::instance();
+        return EEM_Form_Element::instance();
```
manzoorwanijk, line 34: 👍

core/services/form/meta/Attributes.php
manzoorwanijk: Nice

core/services/form/meta/Attributes.php
```diff
@@ -0,0 +1,198 @@
...
+                return filter_var($value, FILTER_SANITIZE_NUMBER_FLOAT, FILTER_FLAG_ALLOW_FRACTION);
+            case 'string':
+            default:
+                return filter_var(
+                    $value,
+                    FILTER_SANITIZE_STRING,
+                    FILTER_FLAG_ENCODE_LOW | FILTER_FLAG_ENCODE_HIGH | FILTER_FLAG_ENCODE_AMP
+                );
+        }
+    }
```
manzoorwanijk, line 153: 🚀

manzoorwanijk (approved):
Looks awesome

manzoorwanijk:
@tn3rb I think you forgot to rename the `form_input` table to `form_element`.

core/services/form/meta/Attributes.php
```diff
@@ -0,0 +1,198 @@
...
+
+    /**
+     * @param string                $attribute
+     * @param bool|float|int|string $value
+     */
+    public function addAttribute(string $attribute, $value): void
+    {
+        if (array_key_exists($attribute, $this->attributes)) {
+            $this->attributes[ $attribute ] = $this->sanitize($attribute, $value);
+        }
```
manzoorwanijk, line 176: Is this `array_key_exists` check correct?
Wouldn't it mean that an attribute will never be added if it's not already there?

core/db_models/EEM_Form_Element.model.php
```diff
@@ -195,28 +151,12 @@ protected function __construct(Element $element, InputTypes $input_types, $timez
...
-                'FIN_type'    => new EE_Enum_Text_Field(
-                    'FIN_type',
-                    esc_html(
-                        sprintf(
-                            /* translators: 1 class name */
-                            __(
-                                'Form input type. Values correspond to the %s constants.',
-                                'event_espresso'
-                            ),
-                            'EventEspresso\core\domain\entities\form\Input::TYPE_*'
```
manzoorwanijk, line 214: `FIN_type` should be a separate field because it is not actually a form attribute, rather it represents the type of element which allows us to decide how it should be rendered.

core/services/form/meta/Attributes.php
tn3rb: oops ya good catch... thnx for fixing 👍🏻

core/db_models/EEM_Form_Element.model.php
tn3rb: kk we can keep separate from the others
i was being absolutely literal with my interpretation and using the HTML definitions for what are and are not input attributes. 
It will be easier to handle if that is a property directly on the input as opposed to being encapsulated within the attributes JSON.

core/db_classes/EE_Form_Section.class.php
tn3rb: i guess it could be public... although currently form sections are either related to another form section or  to an event, which are both represented by:

- `getChildFormSections()`
- `getFormSectionsForEvent()`

which set the appropriate where params for you, but doesn't hurt to do this

core/services/form/meta/Attributes.php
tn3rb: oops... ya you are correct...
what i **_MEANT_** to do was check if the key exists in the `attribute_types` array:

```
if (array_key_exists($attribute, $this->attribute_types)) {
```

which then essentially operates as an `allow/deny` list, meaning an attribute can ONLY be added if it is hardcoded in the `attribute_types` array.

PR #3449, opened 2021-06-29
Form builder | Minor improvements

manzoorwanijk:
This PR:
- Adds `Block` to form input types 
- Adds Multi Checkbox to inputs
- Fixes delete section mutation

### DIFF
--- a/core/domain/services/graphql/mutators/FormSectionDelete.php
+++ b/core/domain/services/graphql/mutators/FormSectionDelete.php
@@ -77,7 +77,7 @@ class FormSectionDelete extends EntityMutator
     public static function deleteSectionAndRelations(EE_Form_Section $entity)
     {
         // Remove related non-default form elements
-        $entity->delete_related('Form_Input', [
+        $entity->delete_related('Form_Element', [
             [
                 'FIN_status' => ['NOT IN', [ FormStatus::SHARED, FormStatus::DEFAULT] ]
             ]
--- a/core/services/form/meta/InputTypes.php
+++ b/core/services/form/meta/InputTypes.php
@@ -2,6 +2,7 @@
 
 namespace EventEspresso\core\services\form\meta;
 
+use EventEspresso\core\services\form\meta\inputs\Block;
 use EventEspresso\core\services\form\meta\inputs\Button;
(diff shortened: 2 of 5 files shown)


core/services/form/meta/inputs/Input.php
```diff
@@ -10,6 +10,11 @@ class Input
      */
     public const TYPE_CHECKBOX = 'checkbox';
 
+    /**
+     * indicates that the HTML input type is 'multi checkbox'
+     */
+    public const TYPE_CHECKBOX_MULTI = 'checkbox-multi';
```
tn3rb, line 16: what's `checkbox-multi` and how is that different than using multiple `checkbox` inputs?

core/domain/services/graphql/mutators/FormSectionDelete.php
```diff
@@ -77,7 +77,7 @@ public static function mutateAndGetPayload(EEM_Form_Section $model)
     public static function deleteSectionAndRelations(EE_Form_Section $entity)
     {
         // Remove related non-default form elements
-        $entity->delete_related('Form_Input', [
+        $entity->delete_related('Form_Element', [
```
tn3rb, line 80: do you know how many times i searched for "Form_Input" ?!?!!? 😠

core/services/form/meta/InputTypes.php
```diff
@@ -2,6 +2,7 @@
 
 namespace EventEspresso\core\services\form\meta;
 
+use EventEspresso\core\services\form\meta\inputs\Block;
```
tn3rb, line 5: i might move this out of the `inputs` folder/namespace since it's not technically an input type...
will leave as is for now, until i think of how i want to approach this.
thnx for adding it though  as it's obviously something i missed 👍🏻

core/services/form/meta/inputs/Input.php
manzoorwanijk: That's actually a good question. Checkbox should ideally be multiple while as single checkbox should be replaced by `Switch`

tn3rb: 💯% agree

PR #3450, opened 2021-06-30
Add-on Registration Fixes

tn3rb:
While struggling through https://github.com/eventespresso/eea-automated-upcoming-event-notifications/pull/23 i discovered a couple of minor issues that this PR fixes.

### DIFF
--- a/admin_pages/messages/Messages_Admin_Page.core.php
+++ b/admin_pages/messages/Messages_Admin_Page.core.php
@@ -4018,7 +4018,7 @@ class Messages_Admin_Page extends EE_Admin_Page
             : array();
 
         // ensure is active
-        $this->_message_resource_manager->activate_messenger($messenger_name, $message_types_to_activate);
+        $this->_message_resource_manager->activate_messenger($active_messenger, $message_types_to_activate);
 
         // set response_data for reload
         foreach ($message_types_to_activate as $message_type_name) {
@@ -4085,7 +4085,7 @@ class Messages_Admin_Page extends EE_Admin_Page
         $message_type_to_activate = $this->_message_resource_manager->get_message_type($message_type_name);
 
         // ensure is active
-        $this->_message_resource_manager->activate_messenger($messenger_name, $message_type_name);
+        $this->_message_resource_manager->activate_messenger($active_messenger, $message_type_name);
 
         // set response for load
         if ($this->_message_resource_manager->is_message_type_active_for_messenger(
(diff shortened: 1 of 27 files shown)


admin_pages/messages/Messages_Admin_Page.core.php
```diff
@@ -4018,7 +4018,7 @@ protected function _activate_messenger($messenger_name)
             : array();
 
         // ensure is active
-        $this->_message_resource_manager->activate_messenger($messenger_name, $message_types_to_activate);
+        $this->_message_resource_manager->activate_messenger($active_messenger, $message_types_to_activate);
```
tn3rb, line 4021: first thing `activate_messenger()` does... is retrieve the messenger object given the supplied messenger name... or ya know... you could just pass the messenger existing object you already have!!!

which... also doesn't make sense... because how come the messenger object already exists if it hasn't been activated yet?!?!?

Anyways, in like 5 out of the 6 places that `activate_messenger()` is called... the messenger object was already instantiated.

core/EE_Addon.core.php
```diff
@@ -597,7 +597,7 @@ public function set_req_type($req_type)
      */
     public function detect_req_type()
     {
-        if (! $this->_req_type) {
+        if ($this->_req_type === null) {
```
tn3rb, line 600: the numeric value for a normal request is `0` which would result in calls to this function repeatedly running `detect_activation_or_upgrade()` because `! 0` == `true`

core/EE_System.core.php
```diff
@@ -419,7 +419,7 @@ public function detect_activations_or_upgrades()
         foreach ($this->registry->addons as $addon) {
             if ($addon instanceof EE_Addon) {
                 // detect teh request type for that addon
-                $addon->detect_activation_or_upgrade();
+                $addon->detect_req_type();
```
tn3rb, line 422: `detect_req_type()` only calls `detect_activation_or_upgrade()` if it hasn't been done already

core/helpers/EEH_Activation.helper.php
```diff
@@ -1349,7 +1349,7 @@ protected static function _activate_and_generate_default_messengers_and_message_
             // passing off to create templates (after the refactor is done).
             // @todo remove this comment when determined not necessary.
             $message_resource_manager->activate_messenger(
-                $messenger_to_generate->name,
+                $messenger_to_generate,
```
tn3rb, line 1352: lolz... ya let's use the object to get the name so that `activate_messenger()` can use the name to get the object!!! 

# 😆 🤔 😞

knazart (approved):
👍 Good stuff.

manzoorwanijk (approved):
LGTM

PR #3451, opened 2021-06-30
Form builder | EDTR Rehydration

manzoorwanijk:
This PR adds form builder rehydration data for EDTR:
- Fixes `EE_Event::registrationForm()`
- Fixes string foreign key UPPERCASE issue
- Fixes input options for form meta
- Adds Form Builder rehydration data for EDTR


core/db_classes/EE_Event.class.php
```diff
@@ -1427,20 +1427,20 @@ public function registrationFormUuid():? string
...
+                'OR' => [
+                    'FSC_UUID'      => $FSC_UUID, // top level form
+                    'FSC_belongsTo' => $FSC_UUID, // child form sections
                 ]
-            )
-            : [];
+                ],
+            'order_by' => ['FSC_order' => 'ASC'],
+        ]);
     }
```
manzoorwanijk, line 1444: @tn3rb the reason for this change is that `->get_many_related()` adds the relation field with `AND` clause to the query thus invalidating the query for sections, thus 
```php
$this->get_many_related(
    'Form_Section',
    [
        [
            'OR' => [
                'Form_Section.FSC_UUID'      => $FSC_UUID, // top level form
                'Form_Section.FSC_belongsTo' => $FSC_UUID, // child form sections
            ]
        ]
  ]
);
```
is equivalent to 
```php
$this->get_many_related(
    'Form_Section',
    [
        'AND' => [
            'OR' => [
                'Form_Section.FSC_UUID'      => $FSC_UUID, // top level form
                'Form_Section.FSC_belongsTo' => $FSC_UUID, // child form sections
            ]
            'Form_Section.FSC_UUID'      => $FSC_UUID, // top level form
        ]
    ]
);
```

core/db_models/fields/EE_Foreign_Key_String_Field.php
```diff
@@ -14,6 +36,7 @@ public function prepare_for_set($value_inputted_for_field_on_model_object)
         if ($this->is_model_obj_of_type_pointed_to($value_inputted_for_field_on_model_object)) {
             $value_inputted_for_field_on_model_object = $value_inputted_for_field_on_model_object->ID();
         }
-        return strtoupper(wp_strip_all_tags($value_inputted_for_field_on_model_object));
+        $clean_value = wp_strip_all_tags($value_inputted_for_field_on_model_object);
+        return $this->is_uppercase ? strtoupper($clean_value) : $clean_value;
```
manzoorwanijk, line 40: The reason for this change was that the `FIN_belongsTo` for `Form_Element` was converted to UPPERCASE, this making the relation invalid for queries.

core/db_classes/EE_Event.class.php
tn3rb: oh wow... hmmm.. ok we'll go with your fix for now, but I think i'll create an issue to fix that in the models

core/db_models/fields/EE_Foreign_Key_String_Field.php
tn3rb: oh yuk... i'm assuming Michael did that because the country table is the main table that uses string IDs, which are the ISO codes like `USA`, `CAD`, `IND`, etc and those are all capitalized. 

That's business logic that should be pushed into the Country model itself and not be part of EEM_Base

core/db_classes/EE_Event.class.php
```diff
@@ -1427,20 +1427,20 @@ public function registrationFormUuid():? string
...
+            [
+                'OR' => [
+                    'FSC_UUID'      => $FSC_UUID, // top level form
+                    'FSC_belongsTo' => $FSC_UUID, // child form sections
                 ]
-            )
-            : [];
+                ],
+            'order_by' => ['FSC_order' => 'ASC'],
+        ]);
```
tn3rb, line 1443: we'll roll with this fix for now, but I don't like having other models as dependencies within another model object as we have had issues with that before.

I'll create an issue for investigating the underlying problem and hopefully we can find a better solution.

core/db_classes/EE_Event.class.php
```diff
@@ -1427,20 +1427,20 @@ public function registrationFormUuid():? string
...
-                        'OR' => [
-                            'Form_Section.FSC_UUID'      => $FSC_UUID, // top level form
-                            'Form_Section.FSC_belongsTo' => $FSC_UUID, // child form sections
-                        ]
-                    ],
-                    'order_by' => ['FSC_order' => 'ASC'],
+
+        if (empty($FSC_UUID)) {
+            return [];
+        }
```
tn3rb, line 1433: this is cleaner and I should have done this regardless 👍🏻

core/db_classes/EE_Event.class.php
manzoorwanijk: I know there could be a better way that I didn't know 😆

tn3rb: impressive that you were able to find the underlying issues within `EEM_Base`'s 6500+ LoC

core/db_models/fields/EE_Foreign_Key_String_Field.php
```diff
@@ -2,6 +2,28 @@
...
+     * @param int|string      $default_value data type should match field type
+     * @param string|string[] $model_name    eg 'Event','Answer','Term', etc.
+     *                                       Basically its the model class's name without the "EEM_"
+     * @param boolean         $is_uppercase  Whether the value should be converted to uppercase on insertion.
+     */
+    public function __construct($table_column, $nicename, $nullable, $default_value, $model_name, $is_uppercase = true)
+    {
+        $this->is_uppercase = $is_uppercase;
+        parent::__construct($table_column, $nicename, $nullable, $default_value, $model_name);
+    }
```
tn3rb, line 26: hmmm... this is actually not bad at all in terms of a solution for this issue because it does tackle it at the root...

will think about this more because it still seems like this uppercase rule is specific to the Country model and not `EE_Foreign_Key_String_Field`

manzoorwanijk: I wanted the change to be fully backward compatible without the need to change any older code, you never know where things can explode 💣

tn3rb: yup... good thinking 💯

core/domain/services/admin/events/editor/FormBuilder.php
```diff
@@ -0,0 +1,85 @@
...
+                'status'     => GQLUtils::formatEnumKey($section->status()),
+            ];
+        }
+
+        /** @var EE_Form_Element[] */
+        $form_elements = EEM_Form_Element::instance()->get_all([
+            [
+                'FSC_UUID' => ['IN', array_keys($sections)]
+            ]
+        ]);
```
tn3rb, line 56: good stuff 👍🏻 
think I'll add this as a method to that model

core/services/form/meta/InputOptions.php
```diff
@@ -25,7 +25,7 @@ class InputOptions implements JsonableInterface
      *
      * @var array
      */
-    private $options;
+    private $options = [];
```
tn3rb, line 28: 👍🏻

core/services/form/meta/InputOptions.php
```diff
@@ -60,7 +59,7 @@ public static function fromJson(string $json): InputOptions
      */
     public function toArray(): array
     {
-        return ['options' => $this->options];
+        return array_values($this->options);
```
tn3rb, line 62: LOLZ 😆 

ya that makes more sense 👍🏻

core/services/form/meta/InputOptions.php
```diff
@@ -85,14 +84,16 @@ public function options(): array
 
 
     /**
-     * @param int|float|string $option_value
-     * @param int|float|string $display_text
+     * @param array $option
      */
-    public function addOption($option_value, $display_text): void
+    public function addOption($option): void
```
tn3rb, line 89: plz type hint for parameters

```suggestion
    public function addOption(array $option): void
```

manzoorwanijk: Done, thanks, need to make that a habit 😄

core/services/form/meta/InputOptions.php
```diff
@@ -85,14 +84,16 @@ public function options(): array
...
     {
-        $option_value = sanitize_key($option_value);
-        if (! isset($this->options[ $option_value ])) {
-            $this->options[ $option_value ] = sanitize_text_field($display_text);
+        if (isset($option['label'], $option['value'])) {
+            $label = sanitize_text_field($option['label']);
+            $value = sanitize_text_field($option['value']);
```
tn3rb, line 93:
```suggestion
            $value = sanitize_key($option['value']);
```

values can't have spaces or other oddities in them

manzoorwanijk: Done, thanks

core/domain/services/admin/events/editor/EventEditorGraphQLData.php
```diff
@@ -89,7 +95,8 @@ public function __construct(
         EventEntityRelations $relations,
         EventManagerData $managers,
         NewEventDefaultEntities $default_entities,
-        TicketMeta $ticket_meta
+        TicketMeta $ticket_meta,
+        FormBuilder $form_builder
```
tn3rb, line 99: not your problem or your fault, but this constructor is getting way too crowded 
10 parameters ?!?!?! yikes 😬

tn3rb (approved):
Great Stuff!!! 💯

core/db_models/EEM_Form_Element.model.php
```diff
@@ -163,7 +164,7 @@ protected function __construct(FormStatus $form_status, InputTypes $input_types,
                     'FIN_type',
                     esc_html__('Form element type.', 'event_espresso'),
                     false,
-                    false,
+                    null,
```
manzoorwanijk, line 167: This caused that `array_key_exists` error and thus the [test failures](https://github.com/eventespresso/event-espresso-core/runs/2955033784?check_suite_focus=true)

PR #3452, opened 2021-06-30
Fix Table Prefix Removal

### LINKED ISSUE #3436: Table ers doesn't exists
If table prefix is 'plus_', the class EE_Table_Base constructor method is trimming 'plus_users' to 'ers'.

https://github.com/eventespresso/event-espresso-core/blob/8a68549dbb6db29b0a1a7f87b8a8fbe8647b8d14/core/db_models/helpers/EE_Table_Base.php#L58

Should be something like: $table_name = preg_replace('/'.$prefix.'/', '', $table_name, 1);

tn3rb:
fixes #3436

`EE_Table_Base` was using `ltrim()` to remove the WP table prefix was is terrible because that will remove ALL of the supplied characters from the beginning of the string and NOT the exact prefix.

This PR uses `substr_replace()` because that will only remove the exact number of characters that the prefix is comprised of. 

In #3436 it was suggested to use `preg_replace()` but that has the potential to remove the prefix from other locations in the table name as well.

### DIFF
--- a/core/db_models/helpers/EE_Table_Base.php
+++ b/core/db_models/helpers/EE_Table_Base.php
@@ -1,7 +1,5 @@
 <?php
 
-
-
 /**
  * Base class for defining the tables that comprise models. This is used to store information
  * about the table\s alias, private key, etc.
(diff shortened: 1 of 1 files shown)


core/db_models/helpers/EE_Table_Base.php
```diff
@@ -41,47 +39,42 @@ abstract class EE_Table_Base
...
      */
     public function __construct($table_name, $pk_column, $global = false)
     {
         $this->_global = $global;
-        $prefix = $this->get_table_prefix();
+        $prefix        = $this->get_table_prefix();
         // if they added the prefix, let's remove it because we delay adding the prefix until right when its needed.
         if (strpos($table_name, $prefix) === 0) {
-            $table_name = ltrim($table_name, $prefix);
+            $table_name = substr_replace($table_name, '', 0, strlen($prefix));
```
manzoorwanijk, line 55: 👍

manzoorwanijk (approved):
Looks good

knazart (approved):
Looks good.

PR #3469, opened 2021-07-02
Create Default Form Sections and Form Elements

tn3rb:
This PR:

- detects whether an event:
	- has no registration form set
	- is a new event and has no registration form set... YET
- creates new form sections and elements (copied from defaults) if either of the above cases are true and saves them for the event
- does some other minor cleanup

### DIFF
--- a/core/EE_Psr4AutoloaderInit.core.php
+++ b/core/EE_Psr4AutoloaderInit.core.php
@@ -41,6 +41,7 @@ class EE_Psr4AutoloaderInit
             EE_Psr4AutoloaderInit::$psr4_loader->addNamespace('EventEspressoBatchRequest', EE_LIBRARIES . 'batch');
             EE_Psr4AutoloaderInit::$psr4_loader->addNamespace('EventEspressoVendor', EE_THIRD_PARTY);
             EE_Psr4AutoloaderInit::$psr4_loader->addNamespace('EETests', EE_PLUGIN_DIR_PATH . 'tests');
+            EE_Psr4AutoloaderInit::$psr4_loader->addNamespace('EndyJasmi', EE_THIRD_PARTY . 'cuid');
             $initialized = true;
         }
     }
--- a/core/db_models/EEM_Form_Section.model.php
+++ b/core/db_models/EEM_Form_Section.model.php
@@ -98,14 +98,14 @@ class EEM_Form_Section extends EEM_Base
                         'JSON string of HTML attributes, such as class, to be applied to this form section\'s container.',
                         'event_espresso'
                     ),
-                    true,
+                    false,
                     '{}'
                 ),
                 'FSC_belongsTo' => new EE_Plain_Text_Field(
                     'FSC_belongsTo',
                     esc_html__('UUID of parent form section that this one belongs to.', 'event_espresso'),
-                    true,
-                    null
+                    false,
+                    ''
                 ),
                 'FSC_label' => new EE_JSON_Field(
                     'FSC_label',
@@ -113,7 +113,7 @@ class EEM_Form_Section extends EEM_Base
                         'JSON string of properties pertaining to a form section\'s label.',
                         'event_espresso'
                     ),
-                    true,
+                    false,
                     '{}'
                 ),
                 'FSC_order'     => new EE_Integer_Field(
@@ -187,6 +187,18 @@ class EEM_Form_Section extends EEM_Base
     {
         $query_params['order_by'] = ['FSC_order' => 'ASC'];
         return $query_params;
+    }
+
+
+    /**
+     * returns an array of Form Sections that should be added by default to new Events
+     *
+     * @return EE_Form_Section[]
+     * @throws EE_Error
+     */
+    public function getDefaultFormSections(): array
+    {
+        return $this->getFormSections(['FSC_status' => FormStatus::DEFAULT]);
     }
 
 
--- a/core/domain/entities/routing/handlers/admin/EspressoEventEditor.php
+++ b/core/domain/entities/routing/handlers/admin/EspressoEventEditor.php
@@ -41,7 +41,7 @@ class EspressoEventEditor extends EspressoEventsAdmin
     protected function registerDependencies()
     {
         $editor_dependencies = [
-            'EventEspresso\core\domain\services\admin\events\editor\EventEditorGraphQLData' =>[
+            'EventEspresso\core\domain\services\admin\events\editor\EventEditorGraphQLData' => [
                 'EventEspresso\core\domain\entities\admin\GraphQLData\Datetimes'                 => EE_Dependency_Map::load_from_cache,
                 'EventEspresso\core\domain\entities\admin\GraphQLData\Event'                     => EE_Dependency_Map::load_from_cache,
                 'EventEspresso\core\domain\entities\admin\GraphQLData\Prices'                    => EE_Dependency_Map::load_from_cache,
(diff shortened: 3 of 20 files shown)


core/EE_Psr4AutoloaderInit.core.php
```diff
@@ -41,6 +41,7 @@ public function initializeAutoloader()
             EE_Psr4AutoloaderInit::$psr4_loader->addNamespace('EventEspressoBatchRequest', EE_LIBRARIES . 'batch');
             EE_Psr4AutoloaderInit::$psr4_loader->addNamespace('EventEspressoVendor', EE_THIRD_PARTY);
             EE_Psr4AutoloaderInit::$psr4_loader->addNamespace('EETests', EE_PLUGIN_DIR_PATH . 'tests');
+            EE_Psr4AutoloaderInit::$psr4_loader->addNamespace('EndyJasmi', EE_THIRD_PARTY . 'cuid');
```
tn3rb, line 44: setting this namespace early because this could become an integral part of EE

core/db_models/EEM_Form_Section.model.php
```diff
@@ -98,22 +98,22 @@ protected function __construct(FormStatus $form_status, ?string $timezone)
                         'JSON string of HTML attributes, such as class, to be applied to this form section\'s container.',
                         'event_espresso'
                     ),
-                    true,
+                    false,
                     '{}'
```
tn3rb, line 102: why deal with null cases when you can set an appropriate default?
(i mean other than the obvious answer that the db runs more efficiently with null values)

core/domain/entities/routing/handlers/admin/EspressoEventEditor.php
```diff
@@ -40,9 +40,8 @@ public function matchesCurrentRequest(): bool
      */
     protected function registerDependencies()
     {
-        $this->dependency_map->registerDependencies(
-            'EventEspresso\core\domain\services\admin\events\editor\EventEditorGraphQLData',
-            [
+        $editor_dependencies = [
+            'EventEspresso\core\domain\services\admin\events\editor\EventEditorGraphQLData' =>[
                 'EventEspresso\core\domain\entities\admin\GraphQLData\Datetimes'                 => EE_Dependency_Map::load_from_cache,
```
tn3rb, line 45: instead of calling `$this->dependency_map->registerDependencies()` repeatedly, 
i put everything in an array and then loop over it at the bottom of this method

core/domain/services/admin/events/editor/EventEditorGraphQLData.php
```diff
@@ -120,19 +120,11 @@ public function __construct(
...
         $eventManagers = $this->managers ->getData($eventId);
 
-        // Avoid undefined variable warning in PHP >= 7.3
-        $tickets = null;
-        $prices  = null;
-
-        if (empty($datetimes['nodes']) || (isset($_REQUEST['action']) && $_REQUEST['action'] === 'create_new')) {
-            $this->default_entities->getData($eventId);
-            $datetimes = $this->datetimes->getData(['eventId' => $eventId]);
-        }
```
tn3rb, line 134: creating default entities first which can then just be loaded by the other data generators

core/domain/services/admin/events/editor/FormBuilder.php
```diff
@@ -26,14 +28,16 @@ class FormBuilder implements EventEditorDataInterface
     /**
      * @param int $eventId
      * @return array
+     * @throws EE_Error
+     * @throws ReflectionException
      */
     public function getData(int $eventId): array
     {
-        /** @var EE_Event */
+        /** @var EE_Event $event */
```
tn3rb, line 36: inline phpdocs need to be the same format as those for a function,
ie: they need to state the type AND the variable name

core/domain/services/admin/events/editor/FormBuilder.php
```diff
@@ -47,26 +51,28 @@ public function getData(int $eventId): array
 
         $sections = [];
         foreach ($form_sections as $section) {
-            $UUID = $section->UUID();
+            if ($section instanceof EE_Form_Section) {
```
tn3rb, line 54: golden rule of programming: if it can be null, it will be null!

core/domain/services/admin/events/editor/NewEventDefaultEntities.php
```diff
@@ -90,6 +100,52 @@ public function getData(int $eventId)
...
+        $datetime_count = $this->datetime_model->count(
+            [
+                [
+                    'EVT_ID'      => $event->ID(),
+                    'DTT_deleted' => ['IN', [true, false]],
+                ],
+                'default_where_conditions' => EEM_Base::default_where_conditions_none,
+            ],
+            'EVT_ID'
+        );
```
tn3rb, line 129: counting dates is fastest way to determine if any are set

core/db_models/EEM_Form_Section.model.php
manzoorwanijk: Yeah, that's better. I wanted to the same the other day, but thought that there might be some underlying logic that uses those values.

core/domain/services/admin/entities/DefaultFormSections.php
```diff
@@ -0,0 +1,121 @@
...
+        if (is_array($default_form_sections)) {
+            foreach ($default_form_sections as $default_form_section) {
+                if (! $default_form_section instanceof EE_Form_Section) {
+                    throw new InvalidEntityException($default_form_section, 'EE_Form_Section');
+                }
+                // we're calling this inside the loop, because this might get set the first time around
+                // and it would always be blank for a new event if we were to call it before the loop,
+                // and then we couldn't set the "BelongsTo" for any child form sections
+                $top_level_form_section = $entity->registrationFormUuid();
+                $default_form_elements = $default_form_section->formElements();
```
manzoorwanijk, line 73: Perfect 👌

core/domain/services/admin/entities/DefaultFormSections.php
```diff
@@ -0,0 +1,121 @@
...
+                }
+                $new_form_section->save();
+                $new_form_sections[ $UUID ] = $new_form_section;
+                // now retrieve, clone, and save all of the form elements
+                $this->createDefaultFormElements($new_form_section, $default_form_elements);
+                // save form section UUID on event if it is the top-level form section
+                if ($new_form_section->isTopLevelFormSection()) {
+                    $entity->setRegistrationFormUuid($UUID);
+                    $entity->save();
+                }
```
manzoorwanijk, line 90: So it means that the first default section from the list will be set as the top level section for the event, right?

core/domain/services/admin/events/editor/FormBuilder.php
manzoorwanijk: Thanks

core/third_party_libs/cuid/Cuid.php
```diff
@@ -0,0 +1,258 @@
+<?php
+
+namespace EndyJasmi;
+
+/**
+ * Cuid is a library to create collision resistant ids optimized for horizontal scaling and performance.
+ *
+ * @package EndyJasmi
+ */
+class Cuid
```
manzoorwanijk, line 10: Should we create an EE alias of this namespace/class and use that instead of consuming it directly? Just like our adapters approach for client side third-party libs.

manzoorwanijk (approved):
Looks awesome

manzoorwanijk:
Why PR checks didn't run for this? 🤔

core/third_party_libs/cuid/Cuid.php
tn3rb: oh that's a great idea, going to create a PR for that, thnx

core/domain/services/admin/entities/DefaultFormSections.php
tn3rb: ya this is not ideal at all because that currently mostly happens by chance :\

as well, there is nothing in place to deal with nested form sections (can't remember atm what our final decision was on that... are we allowing those?? I mean, other than top-level + children?

tn3rb:
thanks for fixing the PR checks and the subsequent linting issues i missed

PR #3470, opened 2021-07-06
Fix dummy data related questions array

Pebblo:
This should be an array of question objects indexed by QST_ID, not an array of answers indexed by ANS_ID.

Generally, this won't cause issues unless you try to use answers which relate to questions within dummy data. 

We don't do this ourselves currently but the [EE4 Price Modifier Add-on](https://aparnascodex.com/downloads/price-modifier-for-event-espresso-4/) does so lets set the data up correctly so that (eventually) it can be used correctly.

## How has this been tested
Just switch to this branch and confirm that dummy data in the previews looks correct, there's not much you can test so as long as the preview loads, the questions look like questions and the answers look like the answers this should be good.

## Checklist

* [ ] My code is tested.


tn3rb (approved):
sure 👍🏻

tn3rb:
ignore test failures which won't be fixed until https://github.com/eventespresso/event-espresso-core/pull/3522 is merged

garthkoyle:
> dummy data in the previews looks correct,
Sorry it's not obvious to me, but what and where is this "dummy data"?

joaovictorangeline:
@Pebblo How can I test it? The "How this has been tested" is not clear to me.

Pebblo:
@garthkoyle @joaovictorangeline

The dummy data is the data used in the message system when generating a message preview, it doesn't matter which message template you view, they all use the dummy data generated within the messaging system.

So for example:

Event Espresso -> Messages -> Default Message Templates -> Registration Approve -> Edit Registrant.

When you click on the 'Preview' button for that message type the dummy data within EE is used: https://monosnap.com/file/O6dZ2wrSE3LsFVGLTJfycMdApkeV0U

---

However, I recently spent some time digging into a different issue with the messaging system which led me to something similar to the above change and there are a few places where the system is jumping through a bunch of hoops to do the same as the above already.

As this issue only happens with the Price Modifier add-on currently (because we don't relate questions in the same way within core) I'm putting this on the back burner until I can spend some time going back through the message system and fix this craziness in a single step to then be tested throughout.

tn3rb:
@Pebblo 
This change causes the following unit test to fail: `EEH_Parse_Shortcodes_Test::test_parsing_email_registration()`

I looked at the code again, and your initial comment on this PR:

> This should be an array of question objects indexed by QST_ID, not an array of answers indexed by ANS_ID.

is not correct.

The code in question was originally this:

```php
$q_n_as['questions'][ $ansobj->ID() ] = $qsts[ $ansobj->get('QST_ID') ];
```

which **IS** an array of question objects, but is indexed by ANS_ID. This makes sense as well, because we would most likely want the arrays of questions and answers to be indexed using the same values.

This code is not easy to read because of all of the unnecessary abbreviations used, and would be much easier to follow if rewritten:

```php
$questions = [];
// first the questions
foreach ($questions_data as $question_data) {
    $question_data_with_keys      = array_combine($question_columns, $question_data);
    $question                     = EE_Question::new_instance($question_data_with_keys);
    $questions[ $question->ID() ] = $question;
}

// now the answers (and we'll setup our arrays)
$questions_and_answers = [];
foreach ($answers_data as $answer_data) {
    $answer_data_with_keys                               = array_combine($answer_columns, $answer_data);
    $answer                                              = EE_Answer::new_instance($answer_data_with_keys);
    $questions_and_answers['answers'][ $answer->ID() ]   = $answer;
    $questions_and_answers['questions'][ $answer->ID() ] = $questions[ $answer->question_ID() ];
}
```

So basically the above code does the following:

- creates question objects from question data
- puts question objects into an array indexed by the question ID
- creates answer objects from answer data
- puts answer objects into a multidimensional array indexed by object type and answer ID
- uses the answer object's related question ID to retrieve that question from the questions array
- puts question objects into the multidimensional array indexed by object type and answer ID

So again, this is most likely correct, because then both an answer AND its corresponding question can be retrieved from that array using just the answer ID.

Looking at the logic in the failing unit test, it _appears_ to me that the above (previous) logic was correct and expected and the change in this PR is not correct.

Is it possible that the logic in the third party [EE4 Price Modifier Add-on](https://aparnascodex.com/downloads/price-modifier-for-event-espresso-4/) is incorrect?

PR #3483, opened 2021-07-28
Enhance Country and State GraphQL schema

manzoorwanijk:
This PR adds support for querying the country and state lists in GQL schema in order to facilitate better querying needed for venue data.


tn3rb (approved):
# POIFECT 💯

PR #3499, opened 2021-08-16
Fix data node order

tn3rb:
see https://github.com/eventespresso/barista/issues/980 && https://github.com/eventespresso/eea-wpuser-integration/issues/40

It appeared that the WP User ticket meta was not getting loaded because of an issue with filters, but it was the order of the data nodes that was causing the filters to be triggered out of order.

This PR adds some logic for ensuring that data nodes:

- are given a predictable and repeatable default ordering 
- have a mechanism in place for changing that order if need be

This was achieved by:

- adding an order property plus a getter and setter to the `JsonDataNode` class
- sorting the data node data array when needed using a custom sorting algorithm

### DIFF
--- a/core/helpers/EEH_Array.helper.php
+++ b/core/helpers/EEH_Array.helper.php
@@ -34,9 +34,9 @@ class EEH_Array extends EEH_Base
      * @param array $array
      * @return boolean
      */
-    public static function is_associative_array(array $array)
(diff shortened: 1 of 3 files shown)


manzoorwanijk (approved):
Looks great. Thank you for the fix

PR #3502, opened 2021-08-23
GitHub Actions | Use composite action for LAMP set up

manzoorwanijk:
This PR uses the extracted composite action to checkout the commit and set up LAMP stack to avoid unnecessary repetition.


tn3rb (approved):
NOICE 👌🏻 love it

PR #3435, opened 2021-06-22
Cleanup Form Builder Models and GQL Classes

tn3rb:
This PR:

- adds DB model & model object changes from another WIP branch to bring `FET/reg-form-builder/main` up to latest schema

- adds PHP 7 typing wherever possible

- cleans up phpdocs

PR #3518, opened 2021-09-21
:bug: Fix bulk delete entities error

manzoorwanijk:
This PR fixes the issue mentioned [here](https://github.com/eventespresso/barista/issues/1012#issuecomment-922768933) which was apparently caused because of [this change](https://github.com/eventespresso/event-espresso-core/pull/3435/files#r712992126).

### DIFF
--- a/core/domain/services/graphql/mutators/BulkEntityDelete.php
+++ b/core/domain/services/graphql/mutators/BulkEntityDelete.php
@@ -72,7 +72,7 @@ class BulkEntityDelete extends EntityMutator
             $deleted = [];
             $failed  = [];
 
-            foreach ($details['entityDbids'] as $key => $entityDbid) {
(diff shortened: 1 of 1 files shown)


core/domain/services/graphql/mutators/BulkEntityDelete.php
```diff
@@ -72,7 +72,7 @@ public static function mutateAndGetPayload()
             $deleted = [];
             $failed  = [];
 
-            foreach ($details['entityDbids'] as $key => $entityDbid) {
+            foreach ($details['entity_db_ids'] as $key => $entityDbid) {
```
tn3rb, line 75: oops... find and replace error followed by lack of oversight... sry about that :(

tn3rb (approved):
# 💯

PR #3529, opened 2021-10-01
Add a deploy workflow for demoee

knazart:
This adds a new GH workflow that updates EE core on demoee.org whenever a master branch is pushed to.


knazart:
Although, now thinking about this... We want the `dev` branch there for now. On the other side I can change the branch in the scripts on the server.

.github/workflows/deploy-to-demoee.yml
```diff
@@ -0,0 +1,19 @@
+name: Deploy to demoee.org
+
+on:
+    push:
+        branches: [master]
```
manzoorwanijk, line 5: To test the things, you can temporarily do this.
```suggestion
        branches: [master, deploy-to-demoee-org]
```

knazart:
Permissions updated. Works now. 

Note that now there are two copies of EE Core on demoee. The active one is on the dev branch. Left it be in case we want to switch to the other version, that now has auto deployments through GH actions.

.github/workflows/deploy-to-demoee.yml
```diff
@@ -0,0 +1,19 @@
+name: Deploy to demoee.org
+
+on:
+    push:
+        branches: [master, deploy-to-demoee-org]
```
manzoorwanijk, line 5:
```suggestion
        branches: [master]
```

manzoorwanijk (approved):
🚢 it

PR #3546, opened 2021-10-06
Fix Default Tickets Editor Screen

### LINKED ISSUE #3526: WSOD on Legacy Default Tickets List Table
https://garth.eventespresso.com/wp-admin/admin.php?page=espresso_events&action=ticket_list_table

list table needs to be replaced with notice about location of new default tickets list

tn3rb:
Closes #3526 

this PR:

- loads FeatureFlags manager for all admin pages
- uses shared FeatureFlags property throughout admin
- restores the Default Tickets route in the admin menu 
- loads purdy notice on default tickets screen if EDTR is used and `ee_edtr_use_default_ticket_manager` cap is enabled
- renames `FeatureFlags::featureAllowed()` to `allowed()` cuz `$this->feature->allowed('flag-name')` looks nice
- adds `is_caffeinated` prop to admin config and caches value
- fixes warning that was popping up in admin caused by old jQuery

so IF:

- EDTR is not enabled
- OR Barista tester plugin is deactivated
- OR current user does not have the `ee_edtr_use_default_ticket_manager` cap 

then the legacy default tickets list table is displayed

but IF:

- EDTR IS enabled
- AND
	- Barista tester plugin IS enabled
	- OR current user HAS the `ee_edtr_use_default_ticket_manager` cap 

then the following notice appears (screenshot cut off - sry)

Screenshot 2021-10-06 154618

### DIFF
--- a/caffeinated/admin/extend/events/templates/default_tickets_moved_notice.template.php
+++ b/caffeinated/admin/extend/events/templates/default_tickets_moved_notice.template.php
@@ -42,7 +42,7 @@
                 </p>
                 <img alt='event editor tickets list menu button'
                      class='ee-new-edtr-notice__img'
-                     src='<?php echo EE_IMAGES_URL . 'default-tickets-1.png'; ?>'
+                     src='<?php echo esc_url_raw(EE_IMAGES_URL . 'default-tickets-1.png'); ?>'
                 />
             </div>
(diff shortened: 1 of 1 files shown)


garthkoyle:
Does this still need merged?

PR #3554, opened 2021-10-07
Fix Event Active Status Tag

### LINKED ISSUE #3553: Event Status Tag Style Issues
In the Event Registration Options, the 'upcoming' highlight is not formatted correctly. Broken on Chrome and Safari: https://www.screencast.com/t/jcoRFNI51a

It seems fine on Event Smart.

tn3rb:
Fixes #3553 

This PR sets the css width for the event active status tag in the reg options meta box to `fit-content`.
It was previously set to `52px` ?!?!?

The only other place this class is used is for a sold out status tag in SPCO and `fit-content` should work fine for that.

### DIFF
--- a/core/admin/assets/ee-admin-page.css
+++ b/core/admin/assets/ee-admin-page.css
@@ -1133,7 +1133,7 @@ label.error {
     border-radius: 3px;
     color: white;
     font-size: 10px;
-    width: 52px;
+    width: fit-content;
     text-decoration: none;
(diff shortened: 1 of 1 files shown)


garthkoyle:
Does this still need merged?

garthkoyle:
Does this still need to be merged?

tn3rb:
just needs re-approval after being rebased

PR #3557, opened 2021-10-11
Improve Request Param Sanitization for Array Values

tn3rb:
from @Pebblo in chat:

> Hey Brent, adding this now so I don't forget to chat about it later.
> Is it possible for getRequestParam()  to be made backwards compatible with the 'old' way of getting those values (using get())? For example can it default to return an array of values if one has been passed, but no $type has been passed?
> Asking as I think its related to [THIS MER issue](https://github.com/eventespresso/ee4-mer/issues/29).

This PR:

- adds an explicit `$is_array` parameter to `Request::getRequestParam()` interface and all implementing usages

- doubles checks if incoming values to `RequestSanitizer::clean()` are an array, and adjusts other parameters accordingly, meaning legacy request usages involving array values are now detected and handled properly. 

- updates all request usages to utilize `$is_array` parameter instead of having to define type as `arrayOf|string` or `arrayOf|int`, etc

### DIFF
--- a/core/services/request/sanitizers/RequestSanitizer.php
+++ b/core/services/request/sanitizers/RequestSanitizer.php
@@ -23,7 +23,7 @@ class RequestSanitizer
         // check if we are getting an improperly typed array and correct
         $is_array = $is_array || is_array($param);
         if ($is_array) {
-            $values        = [];
+            $values = [];
             foreach ($param as $key => $value) {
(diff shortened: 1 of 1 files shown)


Pebblo:
Thanks, Brent. 

This fixed the issues I was having with MER and I also the previous issue related to this (selecting registrations and triggering messages) works as expected too.

Merging in.

PR #3570, opened 2021-10-21
Handle Legacy Editor Tax Changes

tn3rb:
from chat:

So this whole line of dominoes started going down because Garth was mixing legacy editor events with EDTR created events (which is good, and something I hadn't done in a while) which caused chaos with the taxes because we now apply them directly to tickets as opposed to keeping them as "global" price modifiers.
Anyways one of the issues that happens is if you create events in the EDTR and then revert to the legacy editor, the taxes are loaded as price modifiers, but we currently also mark those tickets with the `isTaxable` flag. I'm thinking we will need to change the use of that flag and keep it solely for legacy tickets that require global taxes applied. There's no need to add that for EDTR tickets that add taxes directly as ticket price modifiers.

I've had to add a way for the legacy editor to switch back n forth between global taxes and the EDTR price modifier taxes, but the legacy editor uses jQuery and all of the data is stored in the DOM and to make matters worse...
this particular code was written by "a previous dev" so it's somewhat convoluted (not as bad as some of his other code though)
For example, data required for calculating ticket prices and/or applying taxes is naturally stored... on the trash can icon!!!
so ya, when recalculating ticket prices or taxes... it retrieves data from... the trash can icon ¯\\_(ツ)_/¯

anyways... I have some things working, but if you monkey around with the UI,
ie: click things randomly and change things around and basically keep clicking on things...
it will get to a point where something gets out of sync and the ticket price is wrong.
It can be reset by manually re-selecting options, but i don't like that it can get out of sync in the first place

My guess is that it is just a side-effect of using jQuery and having to manipulate the DOM when executing logic, so there might not be any way to 100% prevent issues from happening, other than using React + EDTR

From the following screenie:

1. click the gear to view ticket details including price modifiers (base price is set in that initial row)
2. click global taxes to switch between legacy tax mode (checked: use global taxes) or EDTR method (unchecked: add taxes as price modifiers)
3. select modifier types, if "Use Global Taxes" is checked, then tax modifiers are disabled and any existing tax modifiers are changed to % surcharge (effectively the same as a tax) The previous price type is stored in a hidden input and is restored is "Use Global Taxes" is unchecked

Screenshot 2021-10-21 105236

the code in question is in `caffeinated/admin/new/pricing/assets/ee-datetime-ticket-metabox.js`
please note that I have left a bunch of debug code in there for the time being but will remove that later


manzoorwanijk (commented):
Reviewing for the sake of cleaning up my list of PRs with requested reviews.

tn3rb:
# MANZOOR!!!  How are you? 😃 

How's A8c?

manzoorwanijk:
Hey Brentt 👋 

I am good. How about you and EE?

All good so far :)

tn3rb:
This week has been especially crazy, but we have EDTR on Event Smart now, so that's good :)

When are you coming back to EE? (I can dream 😉) he he he

How's Shagufta? You two still in Hyderabad?

manzoorwanijk:
> we have EDTR on Event Smart now

That's Awesome 🎉 

> When are you coming back to EE? (I can dream 😉) he he he

😄 



> How's Shagufta? You two still in Hyderabad?

All good. Will email you where I am :trollface:

PR #3595, opened 2021-11-17
Fix Unit Tests

tn3rb:
Currently all unit tests are failing because of incompatibilities between WP core and PHPUnit.

This PR:

- adds the [yoast/phpunit-polyfills](https://github.com/Yoast/PHPUnit-Polyfills) via composer and updates all deps
- finds and loads the above polyfills accordingly depending on where tests are run (local vs github)
- removes speed trap listeners

these fixes were copied over from the dev branch where they were first implemented


knazart (approved):
Yup, looks god to me 👍

PR #3599, opened 2021-11-18
Fix schema eventStatus escaping.

Pebblo:
Create an event and check the schema output, the 'eventStatus' parameter has escaped double quotes `"`

`"eventStatus": [ &quot;https://schema.org/EventScheduled&quot; ]`

This change swaps out how that value is output to match the others used in that file, the quotes are hardcoded and no longer escaped.

## Checklist

* [ ] My code is tested.

### DIFF
--- a/core/templates/json_linked_data_for_event.template.php
+++ b/core/templates/json_linked_data_for_event.template.php
@@ -26,7 +26,7 @@ defined('EVENT_ESPRESSO_VERSION') || exit;
   "description": <?php echo wp_json_encode($event_description); ?>,
   "url": "<?php echo esc_url_raw($event_permalink); ?>",
   "eventAttendanceMode": "<?php echo esc_url_raw('https://schema.org/' . $event_attendance_mode); ?>",
-  "eventStatus": [ <?php echo esc_html($event_status); ?> ],
+  "eventStatus": [ "<?php echo esc_html($event_status); ?>" ],
   "offers": [
     <?php
     $i = 0;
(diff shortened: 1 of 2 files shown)


core/templates/json_linked_data_for_event.template.php
```diff
@@ -26,7 +26,7 @@
   "description": <?php echo wp_json_encode($event_description); ?>,
   "url": "<?php echo esc_url_raw($event_permalink); ?>",
   "eventAttendanceMode": "<?php echo esc_url_raw('https://schema.org/' . $event_attendance_mode); ?>",
-  "eventStatus": [ <?php echo esc_html($event_status); ?> ],
+  "eventStatus": [ "<?php echo esc_html($event_status); ?>" ],
```
tn3rb, line 29: since the content of that variable is a URL, this should be

```suggestion
  "eventStatus": [ "<?php echo esc_url_raw($event_status); ?>" ],
```

tn3rb (changes requested):
looks good, but had one important suggestion

core/templates/json_linked_data_for_event.template.php
Pebblo: Thanks, Brent.

I've no problem committing this but just checking you are aware that for [esc_url_raw()](https://developer.wordpress.org/reference/functions/esc_url_raw/) the codex has:

> This function is not safe to use for displaying the URL, use esc_url() instead.

tn3rb: Yes, but this file is not used for display. It supplies event data for Google and other bots to read, therefore the URL needs to be functional. 

`esc_url()` will make the URL safe for display, but can potentially render it unusable depending on how it encodes things.

tn3rb: If you wanted to both display a link AND keep it functional, you would use both `esc_url_raw()` and `esc_url()` like:

```php
<a href="<?php echo esc_url_raw( $url );?>"><?php echo esc_url( $url );?></a>
```

Pebblo: Coolio ta, just double checkin.

tn3rb:
just a heads up that this needs to be rebased onto the latest master

Issue #3596, opened 2021-11-17
Fix 'Resend Registration Details' button 


Pebblo:
Go to Event Espresso -> Registrations.

Click on the 'Resent Registration Details' button on a registration you'll see:

> Sorry, you are not allowed to access this page.

Screenshot: https://monosnap.com/file/2ULnbwJYTUgJZIoCzA7WPiITpB5RPg

The link works in 4.10.14.p but not in 4.10.15.p+

The broken link looks like this:

`http://ee4.test/wp-admin/admin.php?page=http%3A%2F%2Fespresso_registrations&action=default&return=resend_registration&_wpnonce=56888b6958&default_nonce=56888b6958`

Not sure where that rouge HTTP:// is coming from.

Pebblo:
Howdy Brent, just letting you know what I found on this:

Afaict comes from: https://github.com/eventespresso/event-espresso-core/blob/master/core/admin/EE_Admin_Page.core.php#L577-L581

The 'url' type being passed means that the requestSaniziter runs it through esc_url_raw():
https://github.com/eventespresso/event-espresso-core/blob/master/core/services/request/sanitizers/RequestSanitizer.php#L44

Anyway, just adding here in case it saves you having to dive around a little.

Pebblo:
Threw together a quick snippet I can give to users that need a fix for this:

```
add_filter('FHEE_redirect_Extend_Registrations_Admin_Pageresend_registration', 'tw_ee_fix_resend_registrations_redirect_url', 20, 2 );
function tw_ee_fix_resend_registrations_redirect_url($redirect_url, $query_args) {
    return str_replace('http%3A%2F%2Fespresso_registrations', 'espresso_registrations', $redirect_url);
}
```

That works, however, I think I found another bug with that... I get another error:

> Something went wrong because we're missing the registration ID

PR #3600, opened 2021-11-18
Fix  'something went wrong because we're missing the registration ID'

Pebblo:
_REG_ID will either be an array of IDs or a string with a single ID depending on how the message was triggered so need to remove the is_array flag as it's not always an array.

Testing this is a little tricky right now due to another bug, see #3596 

When the above is fixed, you can test in:

Event Espresso -> Registrations.

Click the 'Resend Registration Details' icon in actions.

A notice should show on the page to show it triggered the message correctly.

---

Next, use the checkboxes next to the registration and select 1 registration. 

In bulk actions, select 'Set {status} and Notify' (status being the same status the registration currently is)

That should trigger another message for that registration.

Finally, select 2 or more registrations with the checkboxes, do the same test about, it should trigger messages for both.

## Checklist

* [x] My code is tested.


Pebblo:
This is no longer needed.

#3601 fixes this.

PR #3601, opened 2021-11-18
Typecast Array Values Passed to RequestSanitizer

### LINKED ISSUE #3597: Fix 'An error has occurred: You need to select a ticket quantity...' when Max Tickets is set to 1
Create an event with just one paid ticket type.

Set 'Maximum number of tickets allowed per order for this event:' to 1.

Trying to register onto the ticket and you'll see an error message like this:

> An error has occurred:
> You need to select a ticket quantity before you can proceed.

Dude ticket selector works as expected (same set up just with a free ticket rather than paid).

tn3rb:
This PR:

- closes #3597
- ensures that parameters supplied to `RequestSanitizer` are always an array if they are expected to be

### DIFF
--- a/core/services/request/sanitizers/RequestSanitizer.php
+++ b/core/services/request/sanitizers/RequestSanitizer.php
@@ -24,7 +24,7 @@ class RequestSanitizer
         $is_array = $is_array || is_array($param);
         if ($is_array) {
             $values = [];
-            foreach ($param as $key => $value) {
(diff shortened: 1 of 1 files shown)


Pebblo (approved):

Pebblo:
Noice!

This fixes a couple of others I'd found as well.

#3600 for example, I don't think that's needed now but will confirm.

PR #3609, opened 2021-11-19
Fix upcoming events widget echo date range.

Pebblo:
This just sets the echo flag on espresso_event_date_range() used within the widget.

Before this change: 
image

After (and how release shows the dates): 
image

Use the upcoming events widget and set the 'Show date range' option to be true.


tn3rb (approved):
sure

PR #3477, opened 2021-07-17
Replace Legacy Request Handler

tn3rb:
This PR:

- replaces all uses of the legacy `EE_Request_Handler` class with the newer `Request` class as well as a totally new `CurrentPage` class

- adds automatic sanitization to Request class getters

- LOTS of auto PSR12 formatting and code cleanup

sorry for the massive PR

tests won't pass on this brach because otherwise PR would have been waaaaay bigger.
tests pass locally for the `decaf-fixes/sanitize-all-request-params` branch though

PR #3613, opened 2021-11-24
Lazy Load CurrentPage in Legacy Request Handler

### LINKED ISSUE #3612: Promotions - 'Promotion applies to...' paging not working
Adding this to core because it's an issue from 4.10.14+

Create/edit a promotion and on the 'Promotion applies to' meta box try to use paging.
(To view the issue you'll need more than 10 events available to select a promotion on)

The 'spinner' will stay on the page and the admin-ajax.php request returns status code 400 with a response of just `0`

Checking into this it looks like the [EE_Promotion_Scope](https://github.com/eventespresso/EE4-Promotions/blob/master/lib/scopes/EE_Promotion_Scope.lib.php#L109) isn't being called at all on those requests.

I added some output into `init()` there and it runs within the admin itself, it doesn't seem to be on the ajax request when using master (4.10.13 shows the output on the ajax request).

tn3rb:
This PR:

- closes #3612 
- was crazy fun to diagnose! Here's a brief outline of what went wrong:
    - in #3477 the legacy `EE_Request_Handler` was deprecated 
    - its remaining useful logic was extracted into `CurrentPage`
    - `CurrentPage` was injected into the constructor of `EE_Request_Handler` as a dependency
    - any calls to deprecated methods were then forwarded to the corresponding method in `CurrentPage` to maintain back compatability
    - `CurrentPage` injected `EE_CPT_Strategy` as a dependency, which has `CustomPostTypeDefinitions` as a dependency, which has  `EE_Core_Config` as a dependency
    - this caused `EE_Config` to initialize a tiny tiny bit earlier than it did previously
    - this caused the Promotions add-on to miss its opportunity to hook into  `EE_Config`  so that `EE_Promotions_Config` could be loaded
    - `EE_Promotions_Config` loads all of the promotion scope classes 
    - `EE_Promotion_Event_Scope` was therefore not loaded when the `promotion_scope_items` AJAX action was triggered from the  promotions admin page by the event list pagination
- this domino effect was fixed simply by NOT injecting  `CurrentPage`  into the `EE_Request_Handler` constructor and instead lazy loading the class ONLY if one of the deprecated methods that require it was called
- whew 😓

### DIFF
--- a/core/EE_Dependency_Map.core.php
+++ b/core/EE_Dependency_Map.core.php
@@ -450,7 +450,6 @@ class EE_Dependency_Map
                 'EventEspresso\core\services\request\Request'     => EE_Dependency_Map::load_from_cache,
             ],
             'EE_Request_Handler'                                                                                          => [
-                'EventEspresso\core\services\request\CurrentPage' => EE_Dependency_Map::load_from_cache,
                 'EventEspresso\core\services\request\Request'     => EE_Dependency_Map::load_from_cache,
                 'EventEspresso\core\services\request\Response'    => EE_Dependency_Map::load_from_cache,
             ],
--- a/core/EE_Request_Handler.core.php
+++ b/core/EE_Request_Handler.core.php
@@ -1,6 +1,7 @@
 <?php
 
 use EventEspresso\core\interfaces\InterminableInterface;
+use EventEspresso\core\services\loaders\LoaderFactory;
 use EventEspresso\core\services\request\CurrentPage;
 use EventEspresso\core\services\request\RequestInterface;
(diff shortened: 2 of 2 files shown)


knazart (approved):
Looks good 🤌

Pebblo:
Wow..... **hands @tn3rb some rope** ..... let me help pull you out of that rabbit hole! Nice find!

This may also help a couple of issues posted by other users (not related to the promotions add-on).

Pebblo:
Tested and confirmed it fixes promotions.

Also tested adding a registration and viewing/saving various pages in the admin.

Issue #3608, opened 2021-11-19
Fix Fatal error: Uncaught TypeError: Argument 1 passed to EventEspresso\core\services\request\CurrentPage::__construct() must be an instance of EE_CPT_Strategy, null given


Pebblo:
Howdy Brent, I don't know much about this one yet but adding the stack trace in case something sticks out for you:

```
Fatal error: Uncaught TypeError: 
Argument 1 passed to EventEspresso\core\services\request\CurrentPage::__construct() must be an instance of EE_CPT_Strategy, 
null given in /core/services/request/CurrentPage.php:68
Stack trace:
#0 [internal function]: EventEspresso\core\services\request\CurrentPage->__construct(NULL, Object(EventEspresso\core\services\request\Request))
#1 /core/EE_Registry.core.php(1275): ReflectionClass->newInstanceArgs(Array)
#2 /core/EE_Registry.core.php(784): EE_Registry->_create_object('EventEspresso\\c...', Array, '', false)
#3 /core/services/loaders/CoreLoader.php(116): EE_Registry->create('EventEspresso\\c...', Array, true, false, false, '')
#4 /core/services/loaders/CachingL in /core/services/request/CurrentPage.php on line 68
```

A user gets this error as soon as they update past 4.10.13 to current.

Any idea what it could be?

tn3rb:
my guess is they have some custom code or a gist or something running on their site.
Can you try to get access and then look through their plugins and function.php file for any custom code plz?

I've looked over ALL of the usages of that class and everything looks good.

Pebblo:
I tried copying all plugins and the theme locally, no issues with those.

I did some digging into this and hacked up core to get a full stack trace (they have 4.10.13.p and v4.10.17.p installed so I'm hacking up 4.10.17 and then switching back to keep them up and running).

I changed up CurrentPage.php so the constructor become this:

```
  public function __construct($cpt_strategy, RequestInterface $request)
  {
      if(! $cpt_strategy instanceof EE_CPT_Strategy ){
          $e = new \Exception;
          echo $e->getTraceAsString();
          exit('$cpt_strategy is not an instance of EE_CPT_Strategy');
      }
      $this->cpt_strategy = $cpt_strategy;
      $this->request      = $request;
      $this->initialized  = is_admin();
      // analyse the incoming WP request
      add_action('parse_request', [$this, 'parseQueryVars'], 2, 1);
  }
```

From that, when I activate EE on their site I get this:

```
[...]

PR #3619, opened 2021-11-26
Remove return type declarations from EEM_Country

Pebblo:
These make EE4's min PHP version number requirement jump to PHP7

See #3608 for related fun times.

## Checklist

* [ ] My code is tested.

### DIFF
--- a/core/db_models/EEM_Country.model.php
+++ b/core/db_models/EEM_Country.model.php
@@ -37,7 +37,7 @@ class EEM_Country extends EEM_Base
     }
 
 
-    protected function __construct(?string $timezone = '')
+    protected function __construct($timezone = '')
     {
         $this->singular_item = esc_html__('Country', 'event_espresso');
         $this->plural_item   = esc_html__('Countries', 'event_espresso');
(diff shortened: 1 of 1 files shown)


core/db_models/EEM_Country.model.php
```diff
@@ -184,7 +184,7 @@ public function get_all_active_countries(): array
      * @return string
      * @throws EE_Error
      */
-    public function get_country_name_by_ISO(string $country_ISO): string
+    public function get_country_name_by_ISO(string $country_ISO)
```
tn3rb, line 187: think you'll need to remove the `string` type hinting too, as I believe that was only added in PHP 7.0

```suggestion
    public function get_country_name_by_ISO($country_ISO)
```

core/db_models/EEM_Country.model.php
```diff
@@ -206,7 +206,7 @@ public function get_country_name_by_ISO(string $country_ISO): string
      * @return EE_Country|null
      * @throws EE_Error
      */
-    public function getCountryByName(string $country_name): ?EE_Country
+    public function getCountryByName(string $country_name)
```
tn3rb, line 209: same here

```suggestion
    public function getCountryByName($country_name)
```

tn3rb (approved):
looks good,  but didn't quite go far enough if you are trying to achieve compatibility with PHP < 7

PR #3617, opened 2021-11-25
Refactor ProcessTicketSelector & Add Unit Tests

tn3rb:
This PR:

- extracts logic from `ProcessTicketSelector::validatePostData()` into new `ProcessTicketSelectorPostData` class
- adds additional sanitization and security checks
- adds unit tests and supporting utility classes required to add tests

PR #3621, opened 2021-11-26
Remove Ticket Selector Typing

tn3rb:
Seeing #3619 made me realize that #3617 also used PHP 7+ type hinting so best to remove that before it becomes an issue.
Also noticed some variables that could potentially cause undefined issues, so fixed that too.

### DIFF
--- a/core/services/request/sanitizers/RequestSanitizer.php
+++ b/core/services/request/sanitizers/RequestSanitizer.php
@@ -15,7 +15,7 @@ class RequestSanitizer
      * @return array|bool|float|int|string
      * @since 4.10.14.p
      */
-    public function clean($param, string $type = 'string', bool $is_array = false, string $delimiter = '')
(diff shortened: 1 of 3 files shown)


knazart:
Huh, ok. This now raises a question - do we stick to the same min PHP version with add-ons as well ?

knazart (approved):
Looks good.

Pebblo:
I got a fatal on this branch on PHP5.6.40 so went back through the changes and spotted type hinting in RequestSanitizer so I've also removed those.

Merging this in as no more fatal.

tn3rb:
re: 
>  do we stick to the same min PHP version with add-ons as well ?

no i don't think so. This was only done because the code in question could be going out the door real soon.

Once EDTR is released, our min PHP version needs to be bumped to 7.1 and I don't think there is any reason not to do so with any add-ons as well.

PR #3622, opened 2021-11-26
Fix Event CPT Status Dropdown

### LINKED ISSUE #3614: Event editor no longer lists custom event statuses (sold out etc)
Since 4.10.14 the custom EE event statuses don't show up in the event editor.

The only options I have are 'Published', 'Draft' and 'Pending Review'

I found this but setting up and event to test something on the waitlist add-on. Setting up a ticket to be sold out, the updating the ticket so the ticket is no longer sold out.... then back again, leaves the event as upcoming and I'm guessing thats related to this?

For example set up a ticket with a qty of 1.

Add an Approved registration to that ticket... its now sold out.

Edit the event and set Qty to 2, update.

Switch the Qty back to 1 and update again.

The event remains upcoming.

tn3rb:
This PR:

- closes #3614 
- uncomments html in `status_dropdown.template.php` and re-adds `display: none;` as a style, because the best way to make things appear is to set their display property to "none" 😕 

I'm guessing that the dropdown selector in `status_dropdown.template.php`  was hidden, not because it was not being used (as I had assumed) but because it was being used to replace the existing WordPress post status dropdown via javascript.

### DIFF
--- a/core/admin/templates/status_dropdown.template.php
+++ b/core/admin/templates/status_dropdown.template.php
@@ -8,9 +8,7 @@
  * @var string[] $statuses
  */
 ?>
-
-<?php /* ?>
-<div id="ee-status-container">
+<div id="ee-status-container" style="display: none;">
     <span id="cur_status"><?php echo esc_html($cur_status_label); ?></span>
     <span id="localized_status_save"><?php echo esc_html($localized_status_save); ?></span>
     <span id="cur_stat_id"><?php echo esc_html($cur_status); ?></span>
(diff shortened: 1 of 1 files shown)


tn3rb:
also confirmed that this fixes the issue with events switching to sold out status if ticket quantities are changed via the admin

knazart (approved):
👍

Pebblo:
There's a change in behaviour between 4.10.13.p and current but it's not directly related to this change so I'm merging this in, will do another ticket for the change.

PR #3624, opened 2021-11-30
Don't escape EE3 migration table output

Pebblo:
Migrations showed up as escaped html, so looked like this:

https://monosnap.com/file/SKfRMhn0gaJBnigbKAY577PReaLowY

This change fixes the table and one of the table's contents as that was also escaped but contains HTML:

https://monosnap.com/file/WFOF9NugQexx3ANonBjQAfIVnkVt6O
(The option 1 section uses HTML tags)

Side note... I'm aware 'escapation' isn't a word, but migration escapation kinda rolls off the tongue right?

I'll leave you with this: 

escape

## Checklist

* [ ] My code is tested.


Pebblo:
Rolling with this as it's better than what w have now.

PR #3606, opened 2021-11-18
Fix Message Preview

tn3rb:
This PR:

- closes #3604 
- fixes parameter prefixes

PR #3625, opened 2021-11-30
Fix creating custom messages

Pebblo:
In #3606 I missed that the parameter prefixes had also been added to the _add_message_template() method, that change has broke the 'Create custom' button within message, it now throws a fatal with:

```
2021-11-29 23:42:33]  Exception Details
Message: Sorry, but we can&#039;t create new templates because we&#039;re missing the messenger or message type
Code: Messages_Admin_Page - _add_message_template - 1251
File: C:\laragon\www\ee4\wp-content\plugins\event-espresso-core-reg\admin_pages\messages\Messages_Admin_Page.core.php
Line No: 1251
```

This change just removes the prefix again.

---

To test, go to Event Espresso -> Messages -> Default message templates.

Click 'create custom' on any message type.


Pebblo:
Tested this fix on a users site (manually applied) and it fixed the issue for them.

Rolling with it for now.

PR #3626, opened 2021-12-01
Fix question editor stripping values

Pebblo:
Currently, if you create a question, the values will save as expected when saved.

If you then edit the question, none of the values show on the form and if you save, the empty values are saved to the DB.

I switched these fields back to using `f()` as these are form inputs and that appears to be what that function is intended for.

## How has this been tested
Create or edit a question, I used both text and checkbox question types to test.

Add some values for each of the question types and also add some HTML (where applicable).

On master when you edit the question now of the values will show, on this branch they will.

Make a change to the values and save, make sure that change shows when saved and that everything within th question shows as expected.

A couple of examples, first question text HTML:
https://monosnap.com/file/UVLsBJgSLcmbfwUZ03thq64wNCs142

Checkbox HTML:
https://monosnap.com/file/PXZhtB57NHEyePvdpSBxuhwuoJlGAt

## Checklist

* [ ] My code is tested.

### DIFF
--- a/admin_pages/registration_form/templates/questions_main_meta_box.template.php
+++ b/admin_pages/registration_form/templates/questions_main_meta_box.template.php
@@ -322,7 +322,7 @@ if ($QST_system === 'country') {
                                                 <?php if ($has_answers) : ?>
                                                     <input type="hidden"
                                                            name="question_options[<?php echo $count; ?>][QSO_value]"
-                                                           value="<?php echo esc_attr($option->f('QSO_value')); ?>"
+                                                           value="<?php esc_attr($option->f('QSO_value')); ?>"
                                                     />
                                                 <?php endif; ?>
                                             </td>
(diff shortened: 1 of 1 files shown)


admin_pages/registration_form/templates/questions_main_meta_box.template.php
```diff
@@ -316,21 +316,21 @@ class="QSO_order"
                                                 <input type="text"
                                                        class="option-value regular-text"
                                                        name="question_options[<?php echo $count ?>][QSO_value]"
-                                                       value="<?php echo esc_attr($option->get_pretty('QSO_value')) ?>"
+                                                       value="<?php echo esc_attr($option->f('QSO_value')) ?>"
```
knazart, line 319: Are you sure this will work ok ? Because `f()` usually echoes out.

knazart (commented):
Have a question there.

admin_pages/registration_form/templates/questions_main_meta_box.template.php
Pebblo: Yeah it works, but will remove the additional echo. Nice catch.

knazart (approved):

PR #3627, opened 2021-12-01
Beaver Builder Upcoming events widget fix

Pebblo:
I don't really agree with having to do this, but Beaver Builder now fatals with Event Espresso because BB is accessing a `category` property on every widget and we now have a private category property.

I switched out 'category' property for 'events_category' as it's a quick an easy fix for us to work around.

## Checklist

* [ ] My code is tested.

### DIFF
--- a/widgets/upcoming_events/EEW_Upcoming_Events.widget.php
+++ b/widgets/upcoming_events/EEW_Upcoming_Events.widget.php
@@ -19,7 +19,7 @@ class EEW_Upcoming_Events extends EspressoWidget
     /**
      * @var string
      */
-    private $category;
+    private $events_category;
 
     /**
      * @var bool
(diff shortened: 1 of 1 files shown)


tn3rb (approved):
yup, that totally works and I think it's a good solution

widgets/upcoming_events/EEW_Upcoming_Events.widget.php
```diff
@@ -66,7 +66,7 @@ class EEW_Upcoming_Events extends EspressoWidget
      * Register widget with WordPress.
      */
     public function __construct()
-    {
+    {
```
tn3rb, line 69: oops... this caused the linting to fail

Pebblo:
Tested and confirmed this works with Beaver Builder.

PR #3628, opened 2021-12-01
Fix thank you page resend email link

Pebblo:
reg_url_link isn't a URL here, it's used as a 'token' to know which registration to resend the email for.

If you complete a registration andget to the thank you page, then click on the resen email link it'll show an error:

> The Registration Confirmation email could not be sent because a valid Registration could not be retrieved from the database.

That's happening because the reg_url_link value is being passed but has `http://` at the beginning:

https://monosnap.com/file/4br6OywykafrrH0dAX3seFO8ocv3kr

This change fixes that value:

https://monosnap.com/file/SJ03voenecnMsh0pInTQzdy694KCzA

## Checklist

* [ ] My code is tested.

### DIFF
--- a/modules/thank_you_page/templates/thank-you-page-registration-details.template.php
+++ b/modules/thank_you_page/templates/thank-you-page-registration-details.template.php
@@ -87,7 +87,7 @@
                                'Click here to resend the Registration Confirmation email',
                                'event_espresso'
                            ); ?>"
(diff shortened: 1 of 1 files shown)


tn3rb (approved):
oops... yup... looks good

Pebblo:
Tested and confirmed it fixes the link.

PR #3644, opened 2021-12-04
Calculate Ticket Selector Rows Dynamically

tn3rb:
Had a customer that was using a filter to remove tickets from the ticket selector. This unfortunately throws off the value set in a hidden input that is used for tracking the number of ticket rows in the TS. This was because that value was being set simply by counting the number of elements in the initial tickets array. That doesn't work however if tickets are being removed afterwards via a filter. This ultimately results in an error being thrown because the ticket selections can not be processed properly.

This PR:

- uses a separate counter for tracking the number of ticket rows 
- increments / decrements the above counter based on whether HTML is actually returned from the ticket row filters
- fixes some formatting and phpdoc issues

### DIFF
--- a/modules/ticket_selector/TicketSelector.php
+++ b/modules/ticket_selector/TicketSelector.php
@@ -2,6 +2,14 @@
 
 namespace EventEspresso\modules\ticket_selector;
 
+use EE_Error;
+use EE_Event;
+use EE_Ticket;
+use EEH_Template;
+use EEH_URL;
+use Exception;
+use ReflectionException;
+
 /**
  * Class TicketSelector
  * Description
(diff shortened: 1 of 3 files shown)


modules/ticket_selector/TicketSelector.php
```diff
@@ -13,42 +21,45 @@ abstract class TicketSelector
...
-     * @throws \EE_Error
      */
-    public function __construct(\EE_Event $event, array $tickets, $max_attendees, array $template_args)
+    public function __construct(EE_Event $event, array $tickets, $max_attendees, array $template_args)
     {
         $this->event = $event;
         $this->tickets = $tickets;
         $this->max_attendees = $max_attendees;
         $this->template_args = $template_args;
-        $this->template_args['hidden_inputs'] = $this->getHiddenInputs();
```
tn3rb, line 51: moved this to just before we load the final template and after ticket rows have been generated (or removed via a filter) so that the new ticket rows counter is accurate

modules/ticket_selector/TicketSelectorSimple.php
```diff
@@ -56,6 +56,7 @@ public function __construct(EE_Event $event, EE_Ticket $ticket, $max_attendees,
      */
     protected function addTemplateArgs()
     {
+        $this->ticket_rows = 1;
```
tn3rb, line 59: the simple TS only ever has 1 ticket

modules/ticket_selector/TicketSelectorStandard.php
```diff
@@ -80,8 +80,8 @@ public function __construct(
      */
     protected function addTemplateArgs()
     {
-        $row                      = 1;
-        $ticket_row_html          = '';
+        $this->ticket_rows        = 0;
+        $all_ticket_rows_html     = '';
```
tn3rb, line 84: $row started at 1 before because it was only used in the individual ticket row html and then incremented AFTER each row was generated. This approach doesn't work if we want to use that number afterwards for the total count of tickets because it always ends up one higher than the actual number of of tickets.

Pebblo:
Tested this fixed the reported issue and also confirmed DWMTS still works

PR #3648, opened 2021-12-09
Fix Ticket Selector Row Count

### LINKED ISSUE #3647: Return of dynamic ticket selector rows error
Sorry, Brent! I missed an output to test with #3644 and it's causing the same error.

If you have the WP Users integration add-on installed and have a ticket(s) that the current user doesn't have the capability the access the ticket row is replaced with:

> The {ticket_name} is available to members only. On Sale

Meaning it has a row but no Qty field to submit and throws off the count.

So this works: https://monosnap.com/file/mosxsORIRLTr1oNkv54hS7GxHePY6i

But if I add another ticket and set the min cap to be `something` (which is obviously a cap I don't have)
Then this fails: https://monosnap.com/file/fRgAB2PdVhQsiTErqCbIBtcGzPuY6G

tn3rb:
this PR:

- closes #3647 
- adds a prop to `TicketSelectorRow` for tracking if ticket is on sale or not
- decrements ticket row count if ticket is not on sale
- does a bunch of formatting

### DIFF
--- a/modules/ticket_selector/TicketSelectorRow.php
+++ b/modules/ticket_selector/TicketSelectorRow.php
@@ -95,6 +95,8 @@ abstract class TicketSelectorRow
     protected $status_class = 'ticket-sales-sold-out lt-grey-text';
 
     /**
+     * used as an alternate way to indicate that a ticket is not available for sale
+     *
      * @var bool
      */
     protected $is_on_sale = true;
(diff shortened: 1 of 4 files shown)


knazart (approved):
Looks good.

Pebblo:
This fixed the issue with the WP User tickets showing on the TS with no qty.

However, I reviewed th changes and noticed it involved sold out, expired etc so I test them.

The count for those tickets is correct within the TS `-rows` field but the submissions don't work.

To reproduce just add a ticket to an otherwise 'working' event and set the Qty field to 0 so that specific tickets shows as sold out.... or set the ticket to be expired.

You can't select tickets on that event anymore with an expired and/or sold out ticket.

https://monosnap.com/file/bIsrqH2ccXMgj831SdPaYRD4dkdP8e

knazart (approved):

Pebblo:
Ok, so I tested:

DWMTS
'Normal' ticket (One ticket type with a qty dropdown)
'Single' tickets (Different ticket types with a max of 1 on each)
'Max Atndz 1' ticket, 3 different ticket types but only 1 max attendee.
WP User restricted ticket with a normal ticket.
'Mixed tickets' (Normal, sold out, expired, wp user restricted)

Pretty sure that should cover everything.... right?

tn3rb:
@Pebblo 
ya that seems like a comprehensive list of ticket selector functionality, in fact, think i'm going to copy that for future use.

Aslo, i looked into adding a unit test for this latest issue, but the current TS unit tests don't operate at the UI level (which is where that filter runs) so we'll likely have to wait for an E2E test to be added.

PR #3649, opened 2021-12-09
Fix Missing RegIDs After Reg Status Update

### LINKED ISSUE #3646: Fix 'Something went wrong because we're missing the registration ID' error
If you manually update a registration and trigger messages, you'll see:

'Something went wrong because we're missing the registration ID'

---

To reproduce go to Event Espresso -> Registrations -> Edit any registration.

Set 'Send related messages' to Yes and click update. 

(You don't need to change the reg status to reproduce this, doing the above should resend whatever the current reg status email is)

Tracked this down to [THIS LINE](https://github.com/eventespresso/event-espresso-core/blob/master/core/libraries/messages/EE_Messages_Processor.lib.php#L572) which returns an empty array.

tn3rb:
So this is what was really happening here:

- while replacing our usage of the legacy `EE_Request_Handler` I swapped out a local copy of the request data on `EE_Admin_Page` with the request data from the new `Request` class
- `Registrations_Admin_Page::_set_registration_status_from_request()` was mutating the request data (BIG no-no) and copying the REG IDs from one location within the request data to another location
- the message trigger was looking in this new location but the REG IDs were no longer there

this PR:

- closes #3646 
- returns to passing the mutated `EE_Admin_Page->_req_data` array to `EED_Messages::process_resend()`
- also adds a conditional so that request data is not overwritten if it already exists in ANOTHER spot in the code that mutates request data >:(

### DIFF
--- a/core/admin/EE_Admin_Page.core.php
+++ b/core/admin/EE_Admin_Page.core.php
@@ -4174,7 +4174,7 @@ abstract class EE_Admin_Page extends EE_Base implements InterminableInterface
      */
     protected function _process_resend_registration()
     {
-        $this->_template_args['success'] = EED_Messages::process_resend($this->request->requestParams());
+        $this->_template_args['success'] = EED_Messages::process_resend($this->_req_data);
         do_action(
             'AHEE__EE_Admin_Page___process_resend_registration',
             $this->_template_args['success'],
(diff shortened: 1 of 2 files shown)


Pebblo (approved):

Pebblo:
Tested and confirmed

Thanks, Brent!

PR #3651, opened 2021-12-15
Don't urlencode the RSS feed url

Pebblo:
The 'New @ Event Espresso' RSS feed has been broken for a little while now so decided to fix it.

This change just removes the call to URL encode the URL passed.

Tested with:

```
//Test URL
$url = 'https://eventespresso.com/feed/';
echo wp_widget_rss_output($url, ['show_date' => 0, 'items' => 5]);

//Make wayyyyy.
echo '<br><br>';

// Test encoded URL
$encoded_url = urlencode($url);
echo wp_widget_rss_output($encoded_url, ['show_date' => 0, 'items' => 5]);
```

Output: 
image

Go to an EE admin page, Event Espresso -> General Settings is fine.

In master the 'New @ EE' metabox will show nothing or an error.

In this branch, it loads correctly.

### DIFF
--- a/core/admin/EE_Admin_Page.core.php
+++ b/core/admin/EE_Admin_Page.core.php
@@ -2503,9 +2503,11 @@ abstract class EE_Admin_Page extends EE_Base implements InterminableInterface
                 // Get RSS Feed(s)
                 self::cached_rss_display(
                     'espresso_news_post_box_content',
-                    apply_filters(
-                        'FHEE__EE_Admin_Page__espresso_news_post_box__feed_url',
-                        'https://eventespresso.com/feed/'
+                    esc_url_raw(
+                        apply_filters(
+                            'FHEE__EE_Admin_Page__espresso_news_post_box__feed_url',
+                            'https://eventespresso.com/feed/'
+                        )
                     )
                 );
                 ?>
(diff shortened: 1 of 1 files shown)


core/admin/EE_Admin_Page.core.php
```diff
@@ -2503,11 +2503,9 @@ public function espresso_news_post_box()
...
                 self::cached_rss_display(
                     'espresso_news_post_box_content',
-                    urlencode(
-                        apply_filters(
-                            'FHEE__EE_Admin_Page__espresso_news_post_box__feed_url',
-                            'https://eventespresso.com/feed/'
-                        )
+                    apply_filters(
+                        'FHEE__EE_Admin_Page__espresso_news_post_box__feed_url',
+                        'https://eventespresso.com/feed/'
```
tn3rb, line 2508: does it still work if you use `esc_url_raw()` ? 
considering it is a filtered url, might be good to have some kind of protection on it.

tn3rb (approved):
LGTM

had one comment but this is fine regardless

tn3rb:
WTF is up with the PHP linting?!?!? 😲 ... I'll look into that right now

core/admin/EE_Admin_Page.core.php
Pebblo: Yep, switched it out to use it _raw

PR #3654, opened 2021-12-17
Fix PHP Linting

tn3rb:
we're seeing tons of linting issues in recent PRs like:
https://github.com/eventespresso/event-espresso-core/runs/4538556394?check_suite_focus=true

This PR: 

- sets composer dependencies & options in GHA workflow


knazart (approved):
👍

PR #3655, opened 2021-12-22
Fix registration messages not triggered when applying/deleting a payment

Pebblo:
Reported here: https://eventespresso.com/topic/transaction-screen-send-related-messages-registration-messages-not-working/

Registration related messages are not being triggered when a payment is applied/deleted from a transaction.

In 4.10.13 just an is_admin() check was enough to confirm if we are working within the admin (including ajax requests sent via the admin) however, in the new Request class we have both isAdmin() and isAdminAjax().

isAdmin() on its own returns false for admin-ajax requests so added this additional check.

## How has this been tested
Use an incomplete transaction (set it up however you prefer, create a new one and use an offline payment method or delete a payment from an existing transaction).

Manually apply a payment to the transaction (full amount is fine).

Make sure under 'Send Related Messages?' the 'Registration Messages' option is checked 
(Payment related messages are still working in master anyway, they trigger differently)

If applying a full payment then a 'Payment Received' and a 'Registration Approved' message should be triggered, in master only the payment received messages triggers.

Now delete the payment from the transaction, if you select 'Leave the same' for the reg status then even if you select to send related messages, nothing should send. If you change the registration status option to anything else, then the messages should trigger (if send related messages is checked).

## Checklist

* [ ] My code is tested.

### DIFF
--- a/modules/messages/EED_Messages.module.php
+++ b/modules/messages/EED_Messages.module.php
@@ -630,7 +630,10 @@ class EED_Messages extends EED_Module
         }
         $request = self::getRequest();
         // first we check if we're in admin and not doing front ajax
-        if ($request->isAdmin() && ! $request->isFrontAjax()) {
+        if (
+            ($request->isAdmin() || $request->isAdminAjax())
+            && ! $request->isFrontAjax()
+        ) {
             $status_change = $request->getRequestParam('txn_reg_status_change', [], 'int', true);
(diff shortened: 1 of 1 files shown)


tn3rb (approved):
yup 👍🏻 💯

Pebblo:
Thanks Brent, merging this in.

PR #3668, opened 2022-01-05
fix placeholder case

tn3rb:
see: https://eventespresso.com/topic/transactions-tab-gives-error/

this PR fixes a weird format/typo error where some placeholders in some templates got pasted as `%1$S` instead of `%1$s`


tn3rb:
plz update that forum issue when this is released

knazart (approved):
👍

Pebblo:
I was confused as to why this didn't show up during testing but turns out this is a PHP8 issue.

Advised the user to switch back to PHP7 for the time being but will include this fix anyway.

Confirmed the fatal is gone with PHP8 and that PHP7 also works as expected on this branch.

PR #3669, opened 2022-01-05
Replace Legacy Request Data in Registrations Admin

tn3rb:
This PR simply updates the Registrations Admin by replacing all uses of the legacy `_req_data` array with functionality on the new Request class

### DIFF
--- a/core/exceptions/ExceptionStackTraceDisplay.php
+++ b/core/exceptions/ExceptionStackTraceDisplay.php
@@ -100,14 +100,14 @@ class ExceptionStackTraceDisplay
 	<div class="ee-error-dev-msg-dv">
 		<p class="ee-error-dev-msg-pg">
 		    '
-               . sprintf(
-                   esc_html__('%1$sAn %2$s was thrown!%3$s code: %4$s', 'event_espresso'),
-                   '<strong class="ee-error-dev-msg-str">',
-                   get_class($exception),
-                   '</strong>  &nbsp; <span>',
-                   $code . '</span>'
-                )
-               . '<br />
+            . sprintf(
+               esc_html__('%1$sAn %2$s was thrown!%3$s code: %4$s', 'event_espresso'),
+               '<strong class="ee-error-dev-msg-str">',
+               get_class($exception),
+               '</strong>  &nbsp; <span>',
+               $code . '</span>'
+            )
+            . '<br />
             <span class="big-text">"' . trim($msg) . '"</span><br/>
             <a id="display-ee-error-trace-1'
                . $time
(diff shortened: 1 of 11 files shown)


Pebblo (approved):
Noice 👍

knazart (approved):
Like the changes. Looks good to me.

Pebblo:
I spotted something whilst testing this although so far it hasn't caused issues as such.

The `_wp_http_referer` query string value has the slashes removed:

`&_wp_http_referer=wp-adminadmin.phppageespresso_registrations`

Prior to this branch it shows:

`&_wp_http_referer=%2Fwp-admin%2Fadmin.php%3Fpage%3Despresso_registrations`

To reproduce use the above branch.

Go to Event Espresso -> Registrations.

Select a month to filter down to (any will do, just select a month and click filter).

Now click on the the 'All Registrations CSV Report'.

Once you are redirected back to the registration list table look in the address bar.

tn3rb:
good catch @Pebblo I'll look into that right now

tn3rb:
the exception stack trace display class wasn't working so I fixed that too because i wanted to use that while debugging

core/exceptions/ExceptionStackTraceDisplay.php
```diff
@@ -41,191 +53,185 @@ public function __construct(Exception $exception)
...
-        if (! WP_DEBUG) {
-            $output .= '
-	<p>';
+        $output = '
+<div id="ee-error-message" class="error">
+    ' . $error_message . '
+</div>';
+        $styles = $this->exceptionStyles();
+        $scripts = $this->printScripts(true);
+        if (defined('DOING_AJAX')) {
```
knazart, line 76: Should we also check for the value of this constant ?

core/exceptions/ExceptionStackTraceDisplay.php
```diff
@@ -41,191 +53,185 @@ public function __construct(Exception $exception)
...
+        echo $styles, $output, $scripts; // already escaped
+    }
+
+
+    private function genericError($msg, $code)
+    {
+        return '
+    <p>
+        <span class="ee-error-user-msg-spn">' . trim($msg) . '</span> &nbsp; <sup>' . $code . '</sup>
+	</p>';
```
knazart, line 89: Kinda messed up spacing.

knazart (approved):
👍

This new `getRequestParam()` is a really nice addition.

core/exceptions/ExceptionStackTraceDisplay.php
tn3rb: it is only ever set to `true` by WP core, so unless a third party is setting that constant to `false`,  it's value is kinda moot... 
ie: it's either `true` or undefined

core/exceptions/ExceptionStackTraceDisplay.php
tn3rb: good eye, will fix, thnx

core/exceptions/ExceptionStackTraceDisplay.php
knazart: Right, ok.

Pebblo:
Tested various sections on the Event Espresso -> Registrations section.

Actions links.
Table views.
Table filters.
Exporting all registrations.
Filtered exports.
Event Check-in tab (selecting an event and checking in/out a registration)
Trashing and Perm Deleting registrations.

Then tested changing the registration status for a single registration and also retriggering messages for it.

Also tested the original reason for this change and that seems fixed.

@tn3rb I think the above covers this, right?

PR #3671, opened 2022-01-07
Replace Legacy Request Data in Transactions Admin

tn3rb:
Just like in #3669, this PR updates the Transactions Admin by replacing all uses of the legacy `_req_data` array with functionality on the new Request class

### DIFF
--- a/admin_pages/transactions/Transactions_Admin_Page.core.php
+++ b/admin_pages/transactions/Transactions_Admin_Page.core.php
@@ -98,11 +98,7 @@ class Transactions_Admin_Page extends EE_Admin_Page
     {
 
         $this->_set_transaction_status_array();
-
-        $txn_id = ! empty($this->_req_data['TXN_ID'])
-                  && ! is_array($this->_req_data['TXN_ID'])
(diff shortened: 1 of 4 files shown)


knazart (approved):

Pebblo:
Garth found this, but something is up with the changes here.

If edit a transaction and click the 'Recalculate Taxes and Total' button you'll get a fatal.

The URL for that button on my test site is:

`/wp-admin/admin.php?page=espresso_transactions&action=espresso_recalculate_line_items&TXN_ID=103&redirect_to=view_transaction&espresso_recalculate_line_items_nonce=9515ddf15d&return=view_transaction`

However, when I click it and get the fatal, the address bar shows:

`/wp-admin/admin.php?page=espresso_transactions&action=http://view_transaction&TXN_ID=103&_wpnonce=59011f916c&http://view_transaction_nonce=59011f916c&return=espresso_recalculate_line_items`

tn3rb:
@Pebblo good catch! pushed a fix

garthkoyle:
That issue seems to be fixed now.

PR #3677, opened 2022-01-17
Set 'Send test email to' field to not be required as it prevents message template changes when empty

Pebblo:
If you edit a message template and don't have a value set for the 'Send test message to' field, when you hit save you get a 'required field' error, like this:

https://share.getcloudapp.com/GGuErw0d

Setting this to not be required means if someone doesn't set the email field and clicks the test send button they see:

https://monosnap.com/file/jz68ys5UJUH3PNnEAPMrsh0pmgfcin

----

2 issues with this:

1. It prevents message template changes if you don't set a value, people get confused by this when they don't want to send a test email.

2. Because you have to set a value, you can't then set an empty value to remove your email. I remember this coming up a while back where a developer tested emails and their email saved within the templates. They couldn't figure out how to remove it again as they needed a value (the solution was to save the admins email there but this removes that issue).


tn3rb (approved):
makes sense 👍🏻

Issue #3686, opened 2022-01-24
Fix htmlspecialchars() expects parameter 1 to be string, array given


Pebblo:
Reported here: https://eventespresso.com/topic/last-two-updates-had-errors/

I can't reproduce this but I've asked for more details on when they are getting it.

The user is reporting this warning:

```
Warning: htmlspecialchars() expects parameter 1 to be string, array given in /home/patroller/www/www/events/wp-content/plugins/event-espresso-core-reg/core/admin/EE_Admin_Page.core.php on line 1164
```

Any idea on this or anything specific I should ask for?

Pebblo:
@hoseinrafiei could you take a look at this?

hoseinrafiei:
@Pebblo I fixed it and submitted a PR: https://github.com/eventespresso/event-espresso-core/pull/3687

PR #3687, opened 2022-01-25
Fixed an issue on calling htmlspecialchars only on strings.

hoseinrafiei:
Fixed an issue which described here: https://github.com/eventespresso/event-espresso-core/issues/3686

### DIFF
--- a/core/admin/EE_Admin_Page.core.php
+++ b/core/admin/EE_Admin_Page.core.php
@@ -1161,7 +1161,7 @@ abstract class EE_Admin_Page extends EE_Base implements InterminableInterface
                 if (strpos($key, 'nonce') !== false) {
                     continue;
                 }
-                $args[ 'wp_referer[' . $key . ']' ] = (is_string($value) ? htmlspecialchars($value) : $value);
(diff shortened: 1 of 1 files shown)


core/admin/EE_Admin_Page.core.php
```diff
@@ -1161,7 +1161,7 @@ public static function add_query_args_and_nonce(
                 if (strpos($key, 'nonce') !== false) {
                     continue;
                 }
-                $args[ 'wp_referer[' . $key . ']' ] = htmlspecialchars($value);
+                $args[ 'wp_referer[' . $key . ']' ] = (is_string($value) ? htmlspecialchars($value) : $value);
```
tn3rb, line 1164: fix is absolutely correct, but the extra brackets are unnecessary:

```suggestion
                $args[ 'wp_referer[' . $key . ']' ] = is_string($value) ? htmlspecialchars($value) : $value;
```

hoseinrafiei:
I applied the changes. Please check and merge.

tn3rb (approved):

PR #3680, opened 2022-01-20
Replace Legacy Request Data in Venues Admin

tn3rb:
like some previous PRs, this PR updates the Venues Admin by replacing all uses of the legacy _req_data array with functionality on the new Request class

PR #3688, opened 2022-01-25
Changed G-Maps Zoom Level fields to number instead of text

hoseinrafiei:
I was checking the https://github.com/eventespresso/event-espresso-core/pull/3680/ and I found that Google Maps Zoom Level fields on Venues -> Google Maps had some issues. Therefore I fixed the issues by changing the field type to number instead of text and also by adding some HTML validation options.

### DIFF
--- a/admin_pages/venues/templates/google_map.template.php
+++ b/admin_pages/venues/templates/google_map.template.php
@@ -106,10 +106,13 @@
                 </th>
                 <td>
                     <input id="event_details_map_zoom"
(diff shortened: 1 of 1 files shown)


tn3rb (approved):
### Nice Find!!! 💯 

thnx for taking initiative like that!

hoseinrafiei:
Great.

tn3rb:
tested this and fix works great! 
rebased onto latest master branch

Issue #3674, opened 2022-01-13
Permanently Deleting Auto Draft Events


garthkoyle:
Checkout this video: https://www.screencast.com/t/cOw5y4szO

Deleting trashed Auto Draft events: 

-  Branch: `FET/update-event-admin-request-params` , bulk deleting events donesn’t work. Individually deleting events creates a loop.

- Branch: `master` , bulk deleting events creates a loop, individually deleting individual works.

garthkoyle:
https://user-images.githubusercontent.com/3094547/149416484-1e0c553f-793a-4d54-a384-179bdfdcfc28.mp4

hoseinrafiei:
The issue is fixed in #3689 PR.

Issue #3652, opened 2021-12-15
Infinite Loop When Bulk Deleting Zero Events


tn3rb:
@marlonhario found the following bug while working on E2E tests for the Events admin:

- go to the Events admin overview tab on trash link 
- select "Delete Permanently" bulk action 
- click apply without selecting any event
- infinite loop of "Identified 0 items for deletion." error messages appears

![bug_1_at_overview_trash](https://user-images.githubusercontent.com/1751030/146241832-4e04c366-4934-48da-a218-39bd0dfd7710.gif)

garthkoyle:
See also #3674

tn3rb:
@hoseinrafiei 

for this issue you will need to poke around in one or more of the following:

- `admin_pages/events/Events_Admin_Page.core.php`
- `core/domain/services/admin/events/data/ConfirmDeletion.php`
- `core/libraries/batch/JobHandlers/PreviewEventDeletion.php`
- `core/domain/services/admin/events/data/ConfirmDeletion.php`
- and/or any of the classes/files related to the above

hoseinrafiei:
This issue is fixed in https://github.com/eventespresso/event-espresso-core/pull/3689 PR.

PR #3689, opened 2022-01-26
Fix/events deletion

hoseinrafiei:
Fixed following two issues.

https://github.com/eventespresso/event-espresso-core/issues/3652
https://github.com/eventespresso/event-espresso-core/issues/3674

I tested the followings:

1- Bulk delete of multiple events using the bulk edit form.
2- Single delete of one event using the link.
3- Bulk delete when there are no events in trash.

### DIFF
--- a/admin_pages/events/Events_Admin_Page.core.php
+++ b/admin_pages/events/Events_Admin_Page.core.php
@@ -2211,8 +2211,7 @@ class Events_Admin_Page extends EE_Admin_Page_CPT
     {
         if ($this->request->requestParamIsSet('EVT_IDs')) {
             return $this->request->getRequestParam('EVT_IDs', [], 'int', true);
-        }
-        else {
+        } else {
             return $this->request->getRequestParam('EVT_ID', [], 'int', true);
         }
     }
--- a/admin_pages/events/Events_Admin_Page.core.php
+++ b/admin_pages/events/Events_Admin_Page.core.php
@@ -2209,7 +2209,12 @@ class Events_Admin_Page extends EE_Admin_Page_CPT
      */
     private function getEventIdsFromRequest()
     {
-        return $this->request->getRequestParam('EVT_IDs', [], 'int', true, ',');
+        if ($this->request->requestParamIsSet('EVT_IDs')) {
+            return $this->request->getRequestParam('EVT_IDs', [], 'int', true);
+        }
+        else {
+            return $this->request->getRequestParam('EVT_ID', [], 'int', true);
+        }
     }
 
 
--- a/core/libraries/batch/JobHandlers/PreviewEventDeletion.php
+++ b/core/libraries/batch/JobHandlers/PreviewEventDeletion.php
@@ -226,7 +226,7 @@ class PreviewEventDeletion extends JobHandler
         $job_parameters->mark_processed($units_processed);
         // If the most-recently processed root node is complete, we must be all done because we're doing them
         // sequentially.
-        if (isset($root_node) && $root_node instanceof ModelObjNode && $root_node->isComplete()) {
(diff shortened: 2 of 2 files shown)


admin_pages/events/Events_Admin_Page.core.php
```diff
@@ -2209,7 +2209,12 @@ private function cleanEventIds(array $event_ids)
      */
     private function getEventIdsFromRequest()
     {
-        return $this->request->getRequestParam('EVT_IDs', [], 'int', true, ',');
+        if ($this->request->requestParamIsSet('EVT_IDs')) {
+            return $this->request->getRequestParam('EVT_IDs', [], 'int', true);
+        }
+        else {
+            return $this->request->getRequestParam('EVT_ID', [], 'int', true);
+        }
```
tn3rb, line 2217: looks good but needs a formatting fix
(see: https://github.com/eventespresso/event-espresso-core/runs/4951951150?check_suite_focus=true)

core/libraries/batch/JobHandlers/PreviewEventDeletion.php
```diff
@@ -226,7 +226,7 @@ public function continue_job(JobParameters $job_parameters, $batch_size = 50)
         $job_parameters->mark_processed($units_processed);
         // If the most-recently processed root node is complete, we must be all done because we're doing them
         // sequentially.
-        if (isset($root_node) && $root_node instanceof ModelObjNode && $root_node->isComplete()) {
+        if (! isset($root_node) || (isset($root_node) && $root_node instanceof ModelObjNode && $root_node->isComplete())) {
```
tn3rb, line 229: hmmmm....

would be better if this displayed a warning/error notice that nothing was selected for deletion but that's not a quick fix, so we'll ride with this for the time being, but I'm going to create an issue for improving this UX

hoseinrafiei:
@tn3rb Should be fine.

tn3rb (approved):
poifect 👌🏻

tn3rb:
hold on Hossein, don't merge this yet, I need to fix something with the branch.

Because we use rebasing, you should normally NEVER need to merge master into your work branch. So plz don't do that in the future (and don't worry, everything's ok, rebasing is tricky but you'll get used to it)

tn3rb (changes requested):
can't merge because base is messed up.. i will fix

tn3rb (approved):
OK Hossein, 

I've fixed this branch by resetting it back to the commit before you merged master into it, and force pushed it again. Please DO NOT pull the changes for this branch. If for whatever reason you want to work on this branch some more, you will need to delete your local copy and then re-checkout the branch again from this repo so that you get the correct updated files.

In case you didn't see the notices i posted elsewhere, here's a link to some documentation i created re: rebasing: https://github.com/eventespresso/ee-and-es-docs/discussions/81

If all of the tests on this PR have passed by the time you read this, and nobody else has merged anything to master, then you can go ahead and merge this (errr... IF you have permissions.. not sure atm) If, not then I can do so later.

hoseinrafiei:
@tn3rb

I checked the commits logs of master branch and I didn't find any commit so I assumed that nobody merged anything into the master branch. Then I merged this one and everything works well so I had permission to do it.

BTW, Please check it and let me know if there is any issue.

tn3rb:
hmmm... something doesn't look right with the merge...
did you select the "Squash and merge" option?

Screenshot 2022-01-26 100110

It's ok if you didn't and I'll likely leave it as is since it is only 2 commits

hoseinrafiei:
@tn3rb 

No, unfortunately I selected the first option. I just read the rules and I saw that as the fifth rule.
I will follow from now on.

Issue #3683, opened 2022-01-20
Reset event list settings is not working when selecting yes


marlonhario:
To replicate: Event Templates Admin - Reset Event List Settings

![image](https://user-images.githubusercontent.com/37426301/150369718-f0b1431a-f1f9-4ece-8ca5-5def9e9340dc.png)

PR #3693, opened 2022-01-27
Fix Events Archive Config Reset

hoseinrafiei:
Fixed the issue that is explained in https://github.com/eventespresso/event-espresso-core/issues/3683 issue.

### DIFF
--- a/caffeinated/admin/extend/events/Extend_Events_Admin_Page.core.php
+++ b/caffeinated/admin/extend/events/Extend_Events_Admin_Page.core.php
@@ -1,7 +1,5 @@
 <?php
 
-use EventEspresso\core\exceptions\InvalidDataTypeException;
-use EventEspresso\core\exceptions\InvalidInterfaceException;
 use EventEspresso\core\services\loaders\LoaderFactory;
 
 /**
@@ -20,6 +18,8 @@ class Extend_Events_Admin_Page extends Events_Admin_Page
      * Extend_Events_Admin_Page constructor.
      *
      * @param bool $routing
+     * @throws EE_Error
+     * @throws ReflectionException
      */
     public function __construct($routing = true)
     {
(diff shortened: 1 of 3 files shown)


tn3rb:
hmmmm....
looks like this branch might have been cut from a branch other than master.

I think you missed this comment on Tuesday: https://eventespresso.slack.com/archives/C080DDVLM/p1643132147616559

That would have messed up your local copy of master and you likely saw a notice saying that your local copy was # commits ahead of origin, however merging was not a good solution.

So this is what you need to do to fix things:

- rename this branch to `fix/events-archive-config-reset-BAK`
- rename your local `master` branch to `master-BAK`
- checkout a fresh copy of `master` from the repo
- create a new branch off of your new `master` branch named  `fix/events-archive-config-reset`
- cherry pick your last commit ([a8fff21](https://github.com/eventespresso/event-espresso-core/pull/3693/commits/a8fff2168fdbffa59cd728591893c2f14ca14b6e)) from `fix/events-archive-config-reset-BAK` into  `fix/events-archive-config-reset`
- FORCE push `fix/events-archive-config-reset` to origin to overwrite the messed up version
- verify that everything looks ok
- delete your local `fix/events-archive-config-reset-BAK` and `master-BAK` branches
- don't worry about this or think that you messed anything up, it's ALLLL GOOOOD 😄 

then in the future:

- NEVER merge master into anything (there are a few rare exceptions)
- ALWAYS rebase your work branches onto master then force push them back up to the repo
- if a work branch  you are working on has been rebased and force pushed to origin, then you need to delete your local copy and re-checkout a fresh copy. You can NOT merge in those changes because rebasing essentially changes the commit history for the branch and that can not be merged back into your local branch.

I'm going to create some docs in our Wiki right now explaining how we use rebasing to keep our git log clean.

EDIT
Looking at your actual fix in this branch, I think the real underlying issue is with that incoming `REQ` variable and not those conditionals. That will need a different more complex fix (which i might have to do). That said, I would still like to see you perform the steps I outlined above for fixing this branch, just as an exercise / learning experience.

tn3rb:
Here's a link to the discussion I created re: rebasing:
https://github.com/eventespresso/ee-and-es-docs/discussions/81

hoseinrafiei:
@tn3rb 

Sorry for messing up and thanks for your detailed guide.
I walked through the steps and I think I fixed everything. I just did not used cherry pick and applied the changes again.

I will try use the instruction that you sent from now on.

Regarding the fix, Since I saw an if condition for the times that user is not reseting the config, I thought that we should have an else for when the user is trying to reset the config.
But as you said, if we change the $REQ variable before this function, we don't need any conditional codes here and we will just update the new configuration based on the request that we received. Please let me know some guides if you prefer to proceed with that solution.

tn3rb:
> Sorry for messing up

that's totally ok. Just so you are aware, i have absolutely no interest in playing the blame game as i don't believe that it helps anything. Mistakes can happen and the best solution IMHO (in my humble opinion) is to just move forwards. It is of course worthwhile to examine a situation where things went wrong and identify _why_ it occurred so that everybody can learn from it. In this scenario, you are a new employee and we have a different and complex approach to repo branch management, AND... we were lacking documentation regarding our approach to branch management, so you are **expected** to mess things up 😆 

> Since I saw an if condition for the times that user is not reseting the config, I thought that we should have an else for when the user is trying to reset the config

hmmmm... ya that's  a good point and makes 100% sense.
the reason i assumed it was a problem with the incoming `$REQ` var is because that has had some recent changes, but more importantly... this setting **used to work**!!! 

So if the issue was a lack of an else clause... then how come it worked before?!?! 🤔 

I kinda want to take a look at this and see if the issue is still orginiating somewhere else.

tn3rb:
@hoseinrafiei your fix was good and is still needed, so good job with that.

I went through and replaced the legacy request data that was being used because I didn't trust that that was not somehow contributing to this issue showing up when it did.

knazart (approved):
Looking good to me 👍

tn3rb:
ok @hoseinrafiei this is ready for merging

hoseinrafiei:
@tn3rb 

I rebased the branch and force pushed it and then I merged it into the master using Squash method.
Please check it to make sure everything works well.

tn3rb:
yup everything looks great, thnx 💯

PR #3694, opened 2022-02-01
Fixed Event Status Banner Not Showing in Block Themes.

### LINKED ISSUE #3676: Event Status Banner Not Showing in Default Twenty Twenty-Two Theme
The event Status banner does show in the default Twenty Twenty-One theme, but not the Twenty Twenty-Two theme:

Twenty Twenty-One: 
image

Twenty Twenty-Two:
image

hoseinrafiei:
Fixes #3676 issue.

To test please do the followings:

1- Activate Twenty Twenty-Two theme on your local website.
2- Make sure that Display Status Banner is set to Yes for both of single and archive pages.
3- Check both of pages to see the Status Banner.

### DIFF
--- a/modules/event_single/EED_Event_Single.module.php
+++ b/modules/event_single/EED_Event_Single.module.php
@@ -224,7 +224,7 @@ class EED_Event_Single extends EED_Module
     public static function the_title($title = '', $id = 0)
     {
         global $post;
-        return in_the_loop() && $post->ID === (int) $id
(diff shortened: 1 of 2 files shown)


tn3rb (approved):
Looks great!

lorenzocaum:
I tested here (https://enzo12dev.wpengine.com/) and this is ready to go.

- I set up two events and set different statuses.
- Then I turned on the display status banner setting for the archives and the single event pages.

I see the correct display status for each event on both the event archives and the single event pages. Then I switched to Twenty Twenty and confirmed the statuses are also shown correctly there.

PR #3705, opened 2022-02-02
Fix JS Build

tn3rb:
This PR simply adds a property check to a script that parses i18n data, then rebuilds all production assests


bin/i18n-map-extractor.js
```diff
@@ -138,7 +138,9 @@ class wpi18nExtractor {
 				chunkName = options.aliases.hasOwnProperty( chunk.name ) ?
 					options.aliases[ chunk.name ] :
 					chunk.name;
-				parseSourcesToMap( chunk._modules, chunkName, extractor );
+				if (chunk._modules) {
+					parseSourcesToMap( chunk._modules, chunkName, extractor );
+				}
```
tn3rb, line 143: this was the actual fix

bin/i18n-map-extractor.js
```diff
@@ -138,7 +138,9 @@ class wpi18nExtractor {
 				chunkName = options.aliases.hasOwnProperty( chunk.name ) ?
 					options.aliases[ chunk.name ] :
 					chunk.name;
-				parseSourcesToMap( chunk._modules, chunkName, extractor );
+				if (chunk._modules) {
```
hoseinrafiei, line 141: I normally use following to check existence of a key in JS.

    if(typeof chunk._modules !== 'undefined')

Or since it's an object, you can use [hasOwnProperty](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwnProperty) method like the following:

    if(chunk.hasOwnProperty('_modules'))

hoseinrafiei (approved):
It should be fine but I suggested the two other solutions that we can use.

knazart (approved):
Looks ok, not including Hossein's suggestion.

bin/i18n-map-extractor.js
tn3rb: oh good call, ya a `hasOwnProperty()` check would be wise here since it's lower level JS

thnx, will fix

Issue #3656, opened 2021-12-29
Attendee Mover - no messages triggered when moving attendee and send messages set to yes


Pebblo:
Reported here: https://eventespresso.com/topic/did-not-trigger-a-registration-message-sent-from-event-espresso-attendee-mover/

**Describe the bug**
If you use the attendee mover to move a registration to a new ticket and set the 'Trigger Notifications?' option to Yes on the confirmation page, no messages trigger.

**To Reproduce**
Event Espresso -> Registrations.

Move a registration to a new ticket and make sure trigger notifications is checked:
https://monosnap.com/file/Nj6oRWvlzvVgMZfPsI53UcRwwKBH1r

Check messages after that step, none will e generated.

**Additional context**
Tracked this down to this:
https://github.com/eventespresso/eea-attendee-mover/blob/master/services/commands/MoveAttendeeCommandHandler.php#L131

That value isn't on the Request class when core does its thing here:
https://github.com/eventespresso/event-espresso-core/blob/master/modules/messages/EED_Messages.module.php#L636

It throws:
`PHP Notice:  Undefined index: send_notifications in \event-espresso-core-reg\modules\messages\EED_Messages.module.php on line 639`

And returns false.

---

I'm not sure if this needs a fix in core or an update the attendee mover. 

I added an isset() check to core for the above notice and noticed messages sent. I then tested completely removing the above line (setting $_REQUEST) from the attendee mover add-on and the messages still trigger.

Note I tested moving a $0 ticket to another $0 ticket so the status didn't change and also changing from a $10 ticket (paid and approved) to a $20 ticket to confirm that the pending payment emails triggered.

So I'm wondering if that hack for msgs is still needed.

PR #3706, opened 2022-02-03
Fix PHP Notice When "send notification" Checkbox is Not Checked.

hoseinrafiei:
Fixes https://github.com/eventespresso/event-espresso-core/issues/3656 issue.

The PHP notice issue is fixed by applying two patches on PHP form handler and template parts.

The main issue described in the issue thread is happening only for non primary attendees because we're sending the notifications only to main attendees.
If it's not an issue and the feature is designed to work that way, then we can close the issue and merge this PR otherwise we can fix the issue by removing the following if condition from the **/event-espresso-core/core/business/EE_Registration_Processor.class.php** and **/event-espresso-core/modules/messages/EED_Messages.module.php** files.

    if (! $registration->is_primary_registrant()) {
        return false;
    }

### DIFF
--- a/admin_pages/registrations/templates/reg_admin_register_new_attendee_step_content.template.php
+++ b/admin_pages/registrations/templates/reg_admin_register_new_attendee_step_content.template.php
@@ -23,6 +23,10 @@
                 <label for="txn_reg_status_change" class="last">
                     <?php esc_html_e('Send Related Messages?', 'event_espresso'); ?>
                 </label>
+                <input name="txn_reg_status_change[send_notifications]"
+                       type="hidden"
+                       value="0"
+                />
                 <input checked=checked
                        id="txn_reg_status_change"
                        name="txn_reg_status_change[send_notifications]"
--- a/admin_pages/registrations/templates/reg_admin_register_new_attendee_step_content.template.php
+++ b/admin_pages/registrations/templates/reg_admin_register_new_attendee_step_content.template.php
@@ -23,10 +23,6 @@
                 <label for="txn_reg_status_change" class="last">
                     <?php esc_html_e('Send Related Messages?', 'event_espresso'); ?>
                 </label>
-                <input name="txn_reg_status_change[send_notifications]"
-                       type="hidden"
-                       value="0"
-                />
                 <input checked=checked
                        id="txn_reg_status_change"
                        name="txn_reg_status_change[send_notifications]"
--- a/core/third_party_libs/pue/pue-client.php
+++ b/core/third_party_libs/pue/pue-client.php
@@ -1373,7 +1373,7 @@ if (! class_exists('PluginUpdateEngineChecker')):
             //oh let's generate the download_url otherwise it will be old news...
 
             if (! empty($this->download_query)) {
-                $d_install_key = $this->install_key_arr['key'];
+                $d_install_key = isset($this->install_key_arr['key']) ? $this->install_key_arr['key'] : NULL;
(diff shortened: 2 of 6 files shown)


admin_pages/registrations/templates/reg_admin_register_new_attendee_step_content.template.php
```diff
@@ -23,6 +23,10 @@
                 <label for="txn_reg_status_change" class="last">
                     <?php esc_html_e('Send Related Messages?', 'event_espresso'); ?>
                 </label>
+                <input name="txn_reg_status_change[send_notifications]"
+                       type="hidden"
+                       value="0"
+                />
```
tn3rb, line 29: this is a great approach and i wish we had used this from the start, but adding this in now could easily cause issues.

How, you say?

Well if there is any code that assumes that the mere presence of the `send_notifications` is enough to trigger that logic, without checking the value, then this would throw off that logic. 

In other words, currently we either:
- HAVE a `send_notifications` request param, meaning "trigger notices"
 or
- NO `send_notifications` request param, meaning "do NOT trigger notices"

It might look ok to do that within the core plugin (and it most certainly is) but we also have a ton of other addons that would need to be checked, AND you also have to remember that EE is a distributed product so there could potentially be third party or user customizations out there that are relying on the above logic and this change could break that.

core/third_party_libs/pue/pue-client.php
```diff
@@ -1373,7 +1373,7 @@ private function requestUpdate()
             //oh let's generate the download_url otherwise it will be old news...
 
             if (! empty($this->download_query)) {
-                $d_install_key = $this->install_key_arr['key'];
+                $d_install_key = isset($this->install_key_arr['key']) ? $this->install_key_arr['key'] : NULL;
```
tn3rb, line 1376: lolz... started to panic when I saw a change in `pue-client.php`
but this is a good change!  thnx  👍🏻

modules/messages/EED_Messages.module.php
```diff
@@ -636,7 +636,10 @@ protected static function _verify_registration_notification_send(
         ) {
             $status_change = $request->getRequestParam('txn_reg_status_change', [], 'int', true);
             // make sure appropriate admin params are set for sending messages
-            if (! $status_change['send_notifications']) {
+            if (
+                ! isset($status_change['send_notifications'])
+                || (isset($status_change['send_notifications']) && ! $status_change['send_notifications'])
+            ) {
```
tn3rb, line 642: also a good change and won't cause issues because the value of request param is checked.

tn3rb (changes requested):
I like the changes that you made, but I think we need to remove those hidden inputs because they have the potential to break user customizations (highly unlikely but still possible)

hoseinrafiei:
@tn3rb Please check the PR again and let me know your thoughts.
A part of fix is applied in the Attendee Mover add-on: https://github.com/eventespresso/eea-attendee-mover/pull/18

tn3rb (approved):
Looks great now thnx 💯

hoseinrafiei:
@tn3rb I merged this one into the master after rebasing. Should I merge the https://github.com/eventespresso/eea-attendee-mover/pull/18 PR as well. If any E2E test is required just let me know if there is anything I can do on that.

tn3rb:
yes that one is good to go as well, assuming it fixes the issue 😆 

in terms of testing, if you follow the steps outlined in the original issue to reproduce the problem, and it no longer occurs, then it's all good.

PR #3707, opened 2022-02-04
Remove "demoee" From "remoteNamesToPushTo"

tn3rb:
from chat:

> tony
> So it looks like the build completed as exepcted, until it got to Running "shell:remoteSync" (shell) task, then it errored out:
    ```Warning: Command failed: cd buildsrc/event-espresso-core-reg&&unset GIT_DIR&&git push demoee master 
    fatal: 'demoee' does not appear to be a git repository
    fatal: Could not read from remote repository.
    Please make sure you have the correct access rights and the repository exists.
    Use --force to continue.```
> 
> nazar
> we need to remove demoee from the "remoteNamesToPushTo" list in EE core. That's in info.json  file.

this PR does exactly what Nazar recommended doing


knazart (approved):
👌  🙇

Issue #2572, opened 2020-03-19
Deactivating Receipts message type also deactivates Invoice payment method


joshfeck:
**Describe the bug**

If you deactivate the receipts message type, the Invoice payment method will also deactivate
**To Reproduce**

1. Go to Event Espresso > Payment Methods and activate Invoice payment method if not already activated
2. Go to Event Espresso > Messages > Settings, click Html, then click and drag "Receipt" over to "Inactivate Message Types"
3. Note the message: "Deactivating the html messenger also automatically deactivates the invoice payment method. In order for invoices to be generated the html messenger must be be active. If you completed this action by mistake, simply reactivate the html messenger, then visit the payment methods admin page to reactivate the invoice payment method."
4. Go to Event Espresso > Payment Methods and behold! A deactivated Invoice payment method

The notification in the admin is wrong: The html messenger **has not** been deactivated. Only the Receipt Message Type has been deactivated.

**Expected behavior**

You should be able to deactivate the Receipt Message type without the automatic deactivation of the Invoice Payment Method.

**Screenshots**

<img width="1083" alt="Screen Shot 2020-03-19 at 11 44 00 AM" src="https://user-images.githubusercontent.com/891156/77085652-03271700-69d7-11ea-914f-f526fa7d48d4.png">

PR #3708, opened 2022-02-08
Fix Invoice Payment Deactivation Issue After Disabling Receipt and Ticket Message Types.

hoseinrafiei:
Fixes https://github.com/eventespresso/event-espresso-core/issues/2572

The issue happened due to a wrong logic for an if condition. I fixed it by correcting the condition.
I tested many possible configuration and everything looks good from my side.

Also I tested booking and payment process when receipt and ticket message types are disabled and Invoice Payment is enabled and everything went well.

### DIFF
--- a/admin_pages/messages/Messages_Admin_Page.core.php
+++ b/admin_pages/messages/Messages_Admin_Page.core.php
@@ -4325,7 +4325,13 @@ class Messages_Admin_Page extends EE_Admin_Page
         }
 
         // if messenger was html or message type was invoice then let's make sure we deactivate invoice payment method.
-        if ($messenger->name === 'html' || $message_type_name === 'invoice') {
+        if (
+            $messenger->name === 'html'
+            && (
+                is_null($message_type)
+                || $message_type_name === 'invoice'
+            )
+        ) {
             EE_Registry::instance()->load_lib('Payment_Method_Manager');
             $count_updated = EE_Payment_Method_Manager::instance()->deactivate_payment_method('invoice');
             if ($count_updated > 0) {


admin_pages/messages/Messages_Admin_Page.core.php
```diff
@@ -4325,7 +4325,13 @@ protected function _setup_response_message_for_deactivating_messenger_with_messa
         }
 
         // if messenger was html or message type was invoice then let's make sure we deactivate invoice payment method.
-        if ($messenger->name === 'html' || $message_type_name === 'invoice') {
+        if (
+            $messenger->name === 'html'
+            && (
+                is_null($message_type)
```
tn3rb, line 4331: hmmmm...
can't fathom why this would fix the issue...
that said the messages system is a bit tricky to grok, so... 🤷🏻 
I want to have a bit more of a look into this, so just going to hold off on this for now.

hoseinrafiei: First part of the if condition was always true if you disable one of the HTML message types. It's why the Invoice payment method got disabled after disabling receipt and ticket message types.

I changed the condition to disable the Invoice only when HTML messenger got disabled ( `$messenger->name === 'html' && is_null($message_type)` ) OR invoice message type got disabled ( `$messenger->name === 'html' && $message_type === 'invoice'` ).

It fixed those issues and also based on my E2E tests, everything works fine after disabling the receipt or ticket message types.

tn3rb: ug... i read the previous code wrong when I first looked at this

ya sry, this is fine, thnx for the explanation 👍🏻

Issue #3715, opened 2022-02-11
Fatal error: Uncaught RuntimeException: Something went wrong! A valid Datetime could not be retrieved from the database using the supplied ID: 2


lorenzocaum:
Seen on master branch with **decaf** mode.

We saw this on Event Smart when the basic event editor is running. That happens on a free site which has the basic plan.

Garth verified the issue: https://github.com/eventespresso/eventsmart.com-website/issues/897#issuecomment-1036631044

**Steps to reproduce**

- Switch to master branch on a WordPress site.
- Enable **decaf** mode by using the constant.
- Create an event that has basic details like a name and description. Then publish it.
- Next, create and publish a venue.
- Go back to the event and select the venue. 
- Then click the update button. (The fatal happens here)

**Reported or seen on**

- https://lorenzoj13square.makeeventsnotwar.com/wp-admin/admin.php?page=espresso_events&action=edit&post=11&edit_nonce=330db7a176
- https://lorenzofirefox21.makeeventsnotwar.com/wp-admin/admin.php?page=espresso_events&action=edit&post=11&edit_nonce=a0cdca716b

We can find more sites here: https://github.com/eventespresso/eventsmart.com-website/issues/897#issuecomment-1036617109

**Other notes**

PR #3717, opened 2022-02-12
Fix an invalid instance check.

hoseinrafiei:
Fixes https://github.com/eventespresso/event-espresso-core/issues/3715

The error was related to a wrong `instanceof` condition. The buggy code was checking if `$datetime` is an instance of `EE_Ticket` class.


admin_pages/events/Events_Admin_Page.core.php
```diff
@@ -1211,7 +1211,7 @@ protected function _default_tickets_update(EE_Event $event, $data)
             //  Otherwise we instantiate a new object for save.
             if (! empty($datetime_data['DTT_ID'])) {
                 $datetime = EEM_Datetime::instance($event_timezone)->get_one_by_ID($datetime_data['DTT_ID']);
-                if (! $datetime instanceof EE_Ticket) {
+                if (! $datetime instanceof EE_Datetime) {
```
tn3rb, line 1214: OOF! bad copy n paste?

tn3rb (approved):
Looks Good

Issue #3728, opened 2022-02-16
Can't Change Event Slug


tn3rb:
### steps to reproduce

- log into admin

- go to **Events** > **Templates** tab

- scroll down to the **Event List Pages** section

- type a new value for the "Event Slug" setting

    <img width="571" alt="Screenshot 2022-02-16 103832" src="https://user-images.githubusercontent.com/1751030/154336913-f9fdb1f4-04bd-4d2c-ad9c-a07e4d6928ce.png">

- click "Save" 

- setting remains unchanged

PR #3734, opened 2022-02-17
Fix Event Slug Issue

hoseinrafiei:
Fixes https://github.com/eventespresso/event-espresso-core/issues/3728 issue.
Tested and works fine.


tn3rb (changes requested):
needs another change to fix a mistake I made previously

tn3rb (approved):
# 💯

tn3rb:
> sorry for my mistake on not applying slugify function to the value and also wrong code for the falsey condition.

that was my fault not yours dude 😆 

I'm the one who broke it when I converted things to use the new request class because I read that ternary incorrectly due to the way it was structured. Not your fault at all.

Issue #3667, opened 2022-01-04
Remove Help Tours 


tn3rb:
plz see: https://github.com/eventespresso/eventsmart.com-website/issues/448

the help tours were initially meant as a way to aid users when migrating from EE3 to EE4 but have now become an unnecessary distraction and can be removed

tn3rb:
@hoseinrafiei 
Here's a low priority issue that you can pick away at if you need a task.

If you look at an admin page like: `admin_pages/events/Events_Admin_Page.core.php`
you will see references throughout to "help tours". 

There's also a corresponding folder for that page for help tour classes: `admin_pages/events/help_tours`
that is _probably_ safe to remove... hopefully... 🤞🏻 

Please note that there are the base admin pages like the one above, but then there are also class extensions found in the caffeinated folder like: `caffeinated/admin/extend/events/Extend_Events_Admin_Page.core.php`
that may also have help tour logic and more importantly, may make reference to some of those help tour classes, so it will be important to remove any references first and/or at the same time those classes are removed.

Please create a new PR for each admin page you work on.

PR #3736, opened 2022-02-19
Removed rest of help tour codes

hoseinrafiei:
Fixes https://github.com/eventespresso/event-espresso-core/issues/3667 issue.
Rest of tour codes are removed from the repository. Only following two files are remained which I kept to make sure that customized websites continue to run without any problem.

core/admin/EE_Help_Tour.core.php
core/admin/EE_Help_Tour_final_stop.class.php


tn3rb (approved):
# 💯

Issue #3726, opened 2022-02-16
Personal information question group can be removed from Primary Registrant questions


Pebblo:
Found by @mohdshakeel 

Create an event and give it a title.

Check the 'Question Groups for Primary Registrant'.

Notice the 'Personal Information' question group is no longer disabled and CAN be unchecked.

It's checked by default but if you uncheck it and update the event it's removed from the reg form:

https://monosnap.com/file/8a0MhXJ6iS6IOhyelQh2wJRAgGxFly

When the event reloads after saving, Personal Information gets automatically checked again, but it has been unselected from the event. 

Saving again updates the question groups to include it on the event.

**Expected behavior**

Personal info should not be allowed to be unchecked from the Primary Registrant group.

Pebblo:
First report of this on the forum: https://eventespresso.com/topic/first-attendee-information-form-not-being-displayed/?view=all

garthkoyle:
@mohdshakeel Will you double check this? Test to be sure you can not remove the Primary Registrant Personal Information assignment.

mohdshakeel:
I had many run to test and found that issue has been fixed completely. I attached the screenshot to check.
![image](https://user-images.githubusercontent.com/1772444/155279129-d6c56676-a084-48bb-b12a-63c5524fe262.png)

garthkoyle:
Thank you @mohdshakeel

PR #3737, opened 2022-02-21
Fix Escaping Issues.

hoseinrafiei:
Fixes https://github.com/eventespresso/event-espresso-core/issues/3726 issue.

Some other instances are removed as well. We should escape only the attribute value and not attribute name and double quotes.

### DIFF
--- a/admin_pages/messages/Messages_Admin_Page.core.php
+++ b/admin_pages/messages/Messages_Admin_Page.core.php
@@ -2682,7 +2682,7 @@ class Messages_Admin_Page extends EE_Admin_Page
                         foreach ($context_templates as $context => $template_fields) :
                             $checked = ($context === $args['context']) ? 'selected="selected"' : '';
                             ?>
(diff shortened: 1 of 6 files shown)


admin_pages/messages/Messages_Admin_Page.core.php
```diff
@@ -2682,7 +2682,7 @@ protected function _set_context_switcher(EE_Message_Template_Group $template_gro
                         foreach ($context_templates as $context => $template_fields) :
                             $checked = ($context === $args['context']) ? 'selected="selected"' : '';
                             ?>
-                            <option value="<?php echo esc_attr($context); ?>" <?php echo esc_attr($checked); ?>>
+                            <option value="<?php echo esc_attr($context); ?>" <?php echo $checked; ?>>
```
tn3rb, line 2685: sigh... ya of course that would convert the quotes to entities

tn3rb (approved):
yup, looks good

Issue #3738, opened 2022-02-22
Apply WP_KSES() to ALL Unescaped Output


tn3rb:
According to the ppl at wp.org, the following template (example) is a horrible security issue:

```php
<div class="container">
	<h4><?php echo $title;?></h4>
	<p><?php echo $text ;?></p>
	<p><?php echo $button;?></p>
</div>
```

because those variables are NOT being escaped upon echoing, despite them being sanitized before being passed to the template, like:

```php
$title = esc_html__('some title', 'ee');
$text = esc_html__('some paragraph text', 'ee');
$button = '<a class="button" href="' . esc_url_raw($url) . '">' . esc_html__('click me', 'ee') . '</button>';
```

No amount of reasoning is acceptable to them.

So in order to satisfy the ppl at wp.org, we need to apply `wp_kses()` to ALL echoed content that is not already being escaped, regardless of whether the content has been _previously_ escaped prior to appearing in the template. It's possible that we may have to undo some of the other escaping as well.

For the example above, we would have to change the template to be something like:

```php
<div class="container">
	<h4><?php echo wp_kses($title);?></h4>
	<p><?php echo wp_kses($text) ;?></p>
	<p><?php echo wp_kses($button);?></p>
</div>
```

(just to be 100% clear, if an echoed variable is already being escaped via another function like one of the  `esc_attr()` or `esc_html()` variations, then that does not need to have `wp_kses()` applied.)

In cases, where more complex tags and attributes are required, we can grab one of the global `$allowed_tags` variables (or use `wp_kses_allowed_html()`) and pass that as the second parameter to `wp_kses()`.

For cases where we need an even _MORE_ complex set of allowed tags and attributes, I have created the `EventEspresso\core\services\request\sanitizers\AllowedTags` class that can be used for building a hardcoded list of tags and attributes that can be easily referenced like so:

```php
// at top of file:
use EventEspresso\core\services\request\sanitizers\AllowedTags;

// in code
$allowed_tags = AllowedTags::getAllowedTags();
wp_kses($html, $allowed_tags );
```

but I'm open to variations on that particular usage.  For example, maybe we don't want to have one accessor for ALL tags and instead want to break things up like:

```php
AllowedTags::getAll();

AllowedTags::getButtonTags();

AllowedTags::getMessageContainerTags();
```

i dunno, whatever... just saying that i'm not opposed to getting granular with the hard coded tags if we need to.

### ANYWAYS...

[...]

PR #3739, opened 2022-02-23
Escaping - Part 1

hoseinrafiei:
It's a starter for https://github.com/eventespresso/event-espresso-core/issues/3738 issue.

## How has this been tested
I did full E2E tests for the upcoming event widget in both of classic editor and block editor.

Also I changed some codes on a shortcode for thank you page which I think is deprecated and have no use at the moment.

## Screenshots for Tests

Screen Shot 2022-02-23 at 18 01 25

Screen Shot 2022-02-23 at 18 00 52

### DIFF
--- a/shortcodes/espresso_thank_you/templates/thank-you-page-transaction-details.template.php
+++ b/shortcodes/espresso_thank_you/templates/thank-you-page-transaction-details.template.php
@@ -5,8 +5,6 @@
  * @var string         $SPCO_payment_options_url
  * @var string         $primary_registrant_name
  */
-
-$SPCO_payment_options_url = esc_url_raw($SPCO_payment_options_url);
 ?>
 
 <h3><?php esc_html_e('Transaction Details', 'event_espresso'); ?></h3>
@@ -44,7 +42,7 @@ $SPCO_payment_options_url = esc_url_raw($SPCO_payment_options_url);
                     <?php $transaction->e_pretty_status(true);
                     if ($show_try_pay_again_link && ! $transaction->is_completed()) { ?>
                         <span class="small-text">
-                        <a href='<?php echo $SPCO_payment_options_url ?>'>
+                        <a href='<?php echo esc_url($SPCO_payment_options_url); ?>'>
                             <?php esc_html_e('View Payment Options', 'event_espresso'); ?>
                         </a>
                     </span>
@@ -68,7 +66,7 @@ $SPCO_payment_options_url = esc_url_raw($SPCO_payment_options_url);
 
     <?php if ($show_try_pay_again_link && ! $transaction->is_completed()) { ?>
         <p class="small-text jst-rght">
-            <a href='<?php echo $SPCO_payment_options_url ?>'><?php
+            <a href='<?php echo esc_url($SPCO_payment_options_url); ?>'><?php
                 esc_html_e("Click here to view Payment Options", 'event_espresso'); ?></a>
         </p>
         <br />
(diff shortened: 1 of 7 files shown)


hoseinrafiei:
@tn3rb 

I investigated the ticket and there are lots of jobs to be done for this so I split the task to following 2 phases.

### Escaping Phase
I found about 800 variables that should be checked for escaping in following directories. I will check and fix them part by part and will create separate PRs for them.

admin_pages (188)
caffeinated (330)
core (120)
modules (110)
public (19)
shortcodes (5)
widgets (21)

Other directories are ignored either due to development purposes or zero variables to get escaped.

### wp_kses Phase
After the first phase, I will start checking the remained variables to apply `wp_kses` function. Right now there are lots of variables that should be checked but after doing the first phase, there will be much less to for investigation.

I will improve the `core/services/request/sanitizers/AllowedTags.php` since there are many more tags and attributes that should be added. I may add some functions to get the tags based on context as you suggested.

Please let me know your thoughts for this PR and also for the whole plan. Your feedback will be a great help.

shortcodes/espresso_thank_you/templates/thank-you-page-transaction-details.template.php
```diff
@@ -44,7 +42,7 @@
                     <?php $transaction->e_pretty_status(true);
                     if ($show_try_pay_again_link && ! $transaction->is_completed()) { ?>
                         <span class="small-text">
-                        <a href='<?php echo $SPCO_payment_options_url ?>'>
+                        <a href='<?php echo esc_url($SPCO_payment_options_url); ?>'>
```
tn3rb, line 45: `esc_url()` is fine for display, but for URLs that need to work, it's better to use `esc_url_raw()`

shortcodes/espresso_thank_you/templates/thank-you-page-transaction-details.template.php
```diff
@@ -68,7 +66,7 @@
 
     <?php if ($show_try_pay_again_link && ! $transaction->is_completed()) { ?>
         <p class="small-text jst-rght">
-            <a href='<?php echo $SPCO_payment_options_url ?>'><?php
+            <a href='<?php echo esc_url($SPCO_payment_options_url); ?>'><?php
```
tn3rb, line 69:
```suggestion
            <a href='<?php echo esc_url_raw($SPCO_payment_options_url); ?>'><?php
```

widgets/upcoming_events/EEW_Upcoming_Events.widget.php
```diff
@@ -357,7 +357,7 @@ public function make_the_title_a_link($title)
      */
     public function fieldID($field_name)
     {
-        return esc_attr(parent::get_field_id($field_name));
+        return parent::get_field_id($field_name);
```
tn3rb, line 360: 👍🏻

tn3rb (changes requested):
looks great but need to swap out of of the escape functions being used

hoseinrafiei:
> looks great but need to swap out of of the escape functions being used

Thanks @brent. I fixed those items, Please check it again.

tn3rb (approved):
yup looks great now, thnx 💯

hoseinrafiei:
@mohdshakeel Please check this branch and do an E2E test.
After that please inform me to merge the branch.

mohdshakeel:
I have tested the widget **Event Espresso Upcoming Events** . I picked each settings options one by one and nothing found any issue with its back-end settings and front-end display.
Front end is displaying perfect with each setting of widget.

image

mohdshakeel (commented):
Can We use the **wp_get_attachment_image** function at **public/Espresso_Arabica_2014/content-espresso_events-thumbnail.php** to display the feature image? It may better in terms of speed optimization.

tn3rb:
> Can We use the wp_get_attachment_image function at public/Espresso_Arabica_2014/content-espresso_events-thumbnail.php to display the feature image? It may better in terms of speed optimization.

that would be addressed in a separate issue, but is not important atm as we will be migrating everything over to blocks at some point

PR #3757, opened 2022-03-07
Move EDTR Admin Option

### LINKED ISSUE #3756: Move EDTR Admin Option Loading
Basic Plan users on MENW don't have access to the option for enabling the EDTR:

Screenshot 2022-03-07 132446

because (assumption) `AdvancedEditorAdminFormSection` is being loaded via `Extend_Events_Admin_Page`

need to switch things around so that it is loaded via `Events_Admin_Page`

tn3rb:
closes #3756

this PR moves where the EDTR admin option is loaded


tn3rb:
need #3759 for tests to pass 😖

knazart (dismissed):
👍

PR #3762, opened 2022-03-09
Fix "Invalid arguments supplied for the SessionStartHandler" error

tn3rb:
plz see: https://github.com/eventespresso/eventsmart.com-website/issues/923

This PR, ensures session handler is loaded on iFrame requests.
Oddly, this error does not appear to happen on single sites, and only seems to affect multisite.


knazart (approved):
👍

PR #3766, opened 2022-03-14
Set Feature Flags by Domain

tn3rb:
Ran into an issue where the "Advanced Event Editor" feature flag needs to be toggled based on both caps AND the current domain configuration, as per the following:

<table>
<tr>
	<td>Condition
	<td>Use Advanced Event Editor
<tr>
	<td>User is on Decaf
	<td>No
<tr>
	<td>User is Caffeinated
	<td>Yes
<tr>
	<td>User is on Multisite
	<td>defer to Capabilities
<tr>
	<td>User missing cap
	<td>No
<tr>
	<td>User has required cap
	<td>Yes
</table>

This PR:

- updates the Domain class to track whether site is part of a multisite install
- injects the Domain class into the FeatureFlags class 
- uses Domain class to check if site is decaf or multisite 
- sets feature flags accordingly


core/domain/Domain.php
```diff
@@ -81,4 +88,14 @@ public static function brandName()
     {
         return (string) apply_filters('FHEE__EventEspresso_core_domain_Domain__brandName', 'Event Espresso');
     }
+
+
+    /**
+     * @return bool
+     * @since $VID:$
+     */
+    public function isMultiSite(): bool
```
hoseinrafiei, line 97: Return type declaration only works in PHP 7+. Is that fine?

hoseinrafiei (approved):
Perfect. Only one concern about return type declaration.

core/domain/Domain.php
tn3rb: yes, this is based off of the dev branch which bumps the min PHP version requirement to 7.1.
great observation, thnx.

PR #3771, opened 2022-03-21
Kses - Part 1

hoseinrafiei:
This is related to the https://github.com/eventespresso/event-espresso-core/issues/3738 issue.
Escaped some PHP variables and used EE class instead of global WP variable in most of sections.

Use this as the testing checklist: https://trello.com/c/Kxpt7F6K/1147-escaping-ee4-decaf-testing-checklist-pr-3771

* [ ] After this is merged to Main branch, be sure to update https://demoee.org to use the latest version of master (instead of the static version we are using to avoid a bug).


admin_pages/general_settings/templates/country_details_settings.template.php
```diff
@@ -26,7 +26,7 @@ class="button-primary save"
             <input
                 type="hidden"
                 id="country-currency-setting-disabled-input"
-                value="<?php echo $CNT_cur_disabled ? 'true' : 'false'; ?>"
+                value="<?php echo ($CNT_cur_disabled ? 'true' : 'false'); ?>"
```
hoseinrafiei, line 29: @tn3rb What is your opinion about using this code style. I know the WP guys searches for `echo $` to find non escaped variables. Based on my previous experience they do not go into the details and as soon as they see that something is matched with their search they think that we did not escaped all the variables.

Using this syntax we can remove these stuff from their search.

tn3rb: lolz... good demonstration of how useless their security reviews are 😆 

i personally don't like the extra brackets and they may conflict with our linting rules, which are set to try and follow the [PHP PSR 12](https://www.php-fig.org/psr/psr-12/) style guidelines.

That said, it's fine in this case as a quick fix. Would be better imho to refactor things so that we don't have to convert that boolean to string `"true"` or `"false"` but that could end up being a rabbit hole we don't want to go down, so best to go with the quick fix for now.

So ya, that's fine with me

hoseinrafiei:
@tn3rb, Please review this branch and let me know your thoughts.

admin_pages/events/templates/event_tickets_metabox_main.template.php
```diff
@@ -116,7 +119,7 @@
                 </tr>
             </thead>
             <tbody id="all-ticket-rows">
-                <?php echo $ticket_rows; // already escaped ?>
+                <?php echo wp_kses($ticket_rows, AllowedTags::getWithFormTags()); ?>
```
tn3rb, line 122: this looks great, but because there is the _potential_ to unintentionally strip out content, i would like you to do the following for at least the first few PRs like this:

- switch to master branch

- find exactly where a variable like `$ticket_rows` is being echoed out in the DOM by doing something like:
    ```
    echo '<h1>START</h1>' . $ticket_rows . '<h1>END</h1>';
    ```

- using your browser's dev tools, copy all of the HTML for that part of the DOM into a file named `master.html`

- discard your changes

- switch to this branch

- find that same code in DOM again (add your START & END markers if need be)

- copy all of the HTML for that part of the DOM into a file named `kses.html`

- use your IDE or Git Client to perform a file compare or diff on those two files to ensure that nothing is being unintentionally stripped out 

i know that's a huge PITA (pain in the 4$$) but it's the only way for us to know for sure if something is being messed up.

An alternate approach is to backtrace all of the variables and find every bit of code that is feeding content to the template and check each and every bit of HTML to see what attributes are being used, but that seems like way more work to me

tn3rb (approved):
### this all looks great 💯 

however... because there is the _potential_ to unintentionally strip out content, i would like you to do the following for at least the first few PRs like this:

- switch to master branch

- find exactly where a variable like `$ticket_rows` is being echoed out in the DOM by doing something like (or whatever method you prefer):
    ```
    echo '<h1>START</h1>' . $ticket_rows . '<h1>END</h1>';
    ```

- using your browser's dev tools, copy all of the HTML for that part of the DOM into a file named `master.html`

- discard your changes

- switch to this branch

- find that same code in DOM again (add your START & END markers if need be)

- copy all of the HTML for that part of the DOM into a file named `kses.html`

- use your IDE or Git Client to perform a file compare or diff on those two files to ensure that nothing is being unintentionally stripped out 

i know that's a huge PITA (pain in the 4$$) but it's the only way for us to know for sure if something is being messed up.

An alternate approach is to backtrace all of the variables and find every bit of code that is feeding content to the template and check each and every bit of HTML to see what attributes are being used, but that seems like way more work to me.

I'm going to approve this because everything you have done is awesome, but I would like you to confirm everything via the above process (or something like it). Thanks

hoseinrafiei:
@mohdshakeel and @Saam01 

Please start the test process based on trello checklist and report all the issues here.

Saam01:
@hoseinrafiei Sure, we'll check that. Thanks.

Pebblo:
Side note for this issue, master has been merged into Dev recently so this issue also affects dev branch.

Please merge master into dev again once this has been fixed, thanks :)

hoseinrafiei:
@mohdshakeel and @Saam01 any update?
I'm waiting for your report.

hoseinrafiei:
> Biography text (EE > Registration > Contact List > Edit Contact) is showing in the middle in the section:

Editor style depends to the theme that you're using. For me with theme Twenty Twenty-One the editor looks good.

Screen Shot 2022-03-29 at 15 01 46
Screen Shot 2022-03-29 at 15 02 01

hoseinrafiei:
> Queue message is not showing in the Queue for resend tab:

It depends to your message configuration. Please put the **Generate and send all messages** option in Messages -> Settings menu to **One a separate request** and then test.

For me it is working fine.

hoseinrafiei:
> Permalink structure not working:

It's not an issue. It's how WordPress works by default for custom post types such as EE event.

hoseinrafiei:
> We cannot add more dates in decaf mode.
>   But when we disable the decaf and add new date we are getting this error:

Thanks for your report. I fixed this issue.

Saam01:
> > Biography text (EE > Registration > Contact List > Edit Contact) is showing in the middle in the section:
> 
> Editor style depends to the theme that you're using. For me with theme Twenty Twenty-One the editor looks good.
> 
> Screen Shot 2022-03-29 at 15 01 46
> 
> Screen Shot 2022-03-29 at 15 02 01

@hoseinrafiei It seems to be happening in Twenty Twenty only.

Saam01:
> > We cannot add more dates in decaf mode.
> > But when we disable the decaf and add new date we are getting this error:
> 
> Thanks for your report. I fixed this issue.

Confirmed. Working fine now.

Saam01:
> > Queue message is not showing in the Queue for resend tab:
> 
> @Saam01 It depends to your message configuration. Please put the **Generate and send all messages** option in Messages -> Settings menu to **One a separate request** and then test.
> 
> For me it is working fine.

Yes, It's working fine today for me.

PR #3774, opened 2022-03-25
Pass $relativeToWpRoot to server_params->requestUri() call

Pebblo:
We've had a couple of reports of issues with the app when EE is installed in a subdirectory:
https://eventespresso.com/topic/mobile-app-connectivity-issue-logged-in-but-missing-permissions/
https://eventespresso.com/topic/ee-app-you-are-not-allowed-list-events-missing-permissions/

It looks like the issue fixed here is causing it, I applied this fix to the first thread and the app started working for me.

Unfortautnely the user then updated EE to then test the app again (meaning they tested within my 'patch') so haven't confirmed just yet but think this needs fixing regardless.

## Checklist

* [ ] My code is tested.

### DIFF
--- a/core/services/request/Request.php
+++ b/core/services/request/Request.php
@@ -298,7 +298,7 @@ class Request implements InterminableInterface, RequestInterface, ReservedInstan
      */
     public function requestUri($relativeToWpRoot = false)
     {
-        return $this->server_params->requestUri();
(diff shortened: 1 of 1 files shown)


Pebblo:
Confirmed this fixed the issue for the 2 users reporting problems (both using subdirectories) and also works on my test site that is NOT a subdirectory.

Merging in

Issue #3770, opened 2022-03-18
Modify CSV Report to Clarify Check-in Data (see Gist)


garthkoyle:
Can we add this to core so that people can get the per-datetime check in/out data?

https://gist.github.com/joshfeck/4a8ea5b02855ebd4d7cc720dd6112d5f

hoseinrafiei:
@garthkoyle 
I added these columns in the https://github.com/eventespresso/event-espresso-core/pull/3808 PR.

PR #3808, opened 2022-04-18
Added Check-Ins DateTime data to the CSV Registration Report.

hoseinrafiei:
Fixes https://github.com/eventespresso/event-espresso-core/issues/3770 issue.

Added following two new columns next to the `Check-Ins` column in the Registration Report.

- `Check-In for Datetime(s)`
- `Check-In Time(s)`


hoseinrafiei:
@tn3rb Please check this and let me know your feedback.

garthkoyle:
This is what I see for a registration that is: 

- check-in
- check-out
- check-in
image

hoseinrafiei:
@garthkoyle 

It's exactly what the code designed to do. I used the the code snippet that you sent on the issue topic.
Do you believe that we should modify it? If yes, please let me know your feedback otherwise if everything is fine, I will merge it.

garthkoyle:
That's a good question. I was a little perplexed at the output and wondered if now is the time to make a change rather than later. 

I've looked at this data again this morning. I think all the data is there for people to know whether people checked-in or out, but it's kind of hard to use or work the data to get the true meaning out of the data.

For instance, this registration has 3 check-ins; two for Datetime B and 1 for Datetime A. So, that means the person checked in to both Datetime A and B, but checked-out of datetime B. That's understandable if you're looking at individual rows, but if you want to do this with hundreds or thousands of registrations, then it's hard to use. 

image

It is possible, but difficult to make the data more usable by using Excel tools to work the data. If I had to do this with lots of registrations, I'd create a column for each datetime and count out the number of check-ins for each datetime by examining the data in each cell and looking for a specific pattern or by breaking the data in the cell apart into a column for each checkin, then I'd convert the even quantities to check-in and the odd quantities to check-outs. I haven't tried it, my wife is better at working with Excel data than I am, but I'm a lot better than average at using excel. I think 98% of our users wouldn't know how to do that if they needed to.

I wonder if it's possible to keep what you have but more user friendly if we added one columns and represented the data a little different in the export.  

What do you think of just adding a column for each datetime name with a status inside the cell, for example:
image
This would allow people to easily see and sort the registration records for people who of each datetime that: 1) did not check-in/out, 2) checked in, 3) checked-out

The trouble with building that export is that we have to dynamically build csv output check-in/out report with more columns depending on the number of datetimes.

What do you guys think? cc @tn3rb

tn3rb:
oh yikes... 
ya that output looks horrible (not your fault @hoseinrafiei) and yes I think it would be much better to collate the datetimes with the check-in times. That said, this might be one of those situations where there is simply no good solution that will work well for all use cases.

> The trouble with building that export is that we have to dynamically build csv output check-in/out report with more columns depending on the number of datetimes.

right, this is one of the reasons that this will still be unwieldly to manage for some users (think REM with 52 weekly datetimes). This approach would most likely be slower because we will have to loop through the data twice; once to build the column list and then again to populate the rows (or once to analyse and sort data and a second time to build the table from the sorted data).

My local system doesn't have any registrations because I'm currently working on an ES migration issue, so do you have an example CSV file you can post in Slack that I could have a quick look at?

garthkoyle:
See attached
[event-espresso-registrations-2022-04-20-14-13-41.csv](https://github.com/eventespresso/event-espresso-core/files/8523500/event-espresso-registrations-2022-04-20-14-13-41.csv)
.

tn3rb:
Here's a mockup of the decided output:

Screenshot 2022-04-20 112843

- create 1 column for each datetime
- column headers contain the following info:
   - Datetime Name
   - Datetime ID
- cells display the following:
   - empty cell - no check-in record
   - IN - if the most recent check-in record has a value of 1 (CHK_in = 1)
   - OUT - if the most recent check-in record has a value of 0 (CHK_in = 0)


As well, to reduce complexity, we need to hide the "Filtered CSV" button until the registrations list table has been filtered for a single event, ie: do not display the "Filtered CSV" button if registrations for more than one event appear in the list table.

hoseinrafiei:
@tn3rb

I did the task based on the output that you sent. Please check following image.
Screen Shot 2022-04-22 at 18 10 48

Also instead of hiding the **Filtered CSV Report**, I implemented a new parameter to include check-in and check-out data only when the registrations are sorted by event. Therefore the button works as before when the results are not filtered. If you believe that hiding the button is better just let me know to hide the **Filtered CSV Report** button.

Also in addition to the Filtered CSV Report button there are two more buttons included in the table footer. One named **All Registrations CSV Report** and another named **Event Registrations CSV Report**. The second button shows only when the results are filtered by event so I applied same feature to this button and it will include the check-in and check-out data too.

Screen Shot 2022-04-22 at 18 16 18

Also I removed some useless loops and indents from the function the make it a little more read-able.

garthkoyle:
Thank you for doing this, it's a good improvement. Here is my feedback.

* [ ] Did the order of the columns change? It'd be good if we had the first name, last name, and email address at the beginning of the csv. This may not have changed, but only more noticeable? What do you think?
* [ ] The availability of the export buttons, the labels, and the csv output is inconsistent or unchanged. I'll explain more.

So, there are three levels to the UI:

1. Check-in (no filtering)
2. Event Filtering
3. Datetime Filtering

1. Check-in (no filtering) - Number of Check-ins added to the csv.
image

2. Event Filtering - Datetime columns with check in/out status added.
image

3. Datetime Filtering - The csv report looks the same as Event filtering
image

What I'm questioning is:

- Are the queries the same for Event and Datetime filtering?
- Do we have too many buttons that do the same thing?
- Can the button labels be improved?
- The `Check-Ins` column reports the quantity of check-ins, but what number does that represent? Is that a representation of all check-ins (in/out) for that registration or is it a quantity of check-in/out for a particular datetime? No matter which level of filtering and which button I use, the `Check-Ins` column count is the same.

garthkoyle:
What if we did this: 

- Just have one button for each filtering level
- Change the label of the button to indicate the filtering label e.g. `All Check-in/out CSV Report`, `Event Check-in/out CSV Report`, `Datetime Check-in/out CSV Report`
- Change the `Check-Ins` column name to be dynamic to represent the level of filtering, e.g.: `All Check-Ins`, `Event Check-ins`, "Datetime Check-ins`.
- When we're at the top level of filtering (no filtering) The `Check-ins` column will be the sum of all check-ins for that registration across all the associated events or datetimes; would this be the same as the event filtering?
- When we're filtering at the event level, only produce the sum of the `Event Check-Ins` for that event.
- When we're filtering at the datetime level, only produce the sum of all the `Datetime Check-Ins` for that datetime.

tn3rb:
## my feedback

> I did the task based on the output that you sent. Please check following image.

that looks great, but I only realized now that my mock-up did not include anything to indicate what those columns actually are! So instead of just displaying `{ datetime name } - ID: { DTM_ID }` let's add "Check-ins for" to the beginning of those column headers. 

example:

<table>
<tr>
	<td>Check-ins for Date A - ID: 1
	<td>Check-ins for Date B - ID: 2
<tr>
	<td>IN: 2022-04-22  8:02 am
	<td>OUT: 2022-04-22  10:27 am
<tr>
	<td>
	<td>IN: 2022-04-22   3:05 pm
</table>

> instead of hiding the Filtered CSV Report, I implemented a new parameter to include check-in and check-out data only when the registrations are sorted by event. Therefore the button works as before when the results are not filtered. If you believe that hiding the button is better just let me know to hide the Filtered CSV Report button.

Yes I like that approach and think it makes sense. Thnx

> Also in addition to the Filtered CSV Report button there are two more buttons included in the table footer. One named All Registrations CSV Report and another named Event Registrations CSV Report. The second button shows only when the results are filtered by event so I applied same feature to this button and it will include the check-in and check-out data too.

I think what you did makes sense. Might be better to just have one button and then change the output based on the view. More about this below.

>  I removed some useless loops and indents from the function the make it a little more read-able.

AWESOME 💯 

---

## re: Garth's comments, 

> Are the queries the same for Event and Datetime filtering?

not sure which queries you are referring to specifically, but I'm certain that some of them would need to be different.

> Do we have too many buttons that do the same thing? 
&& 
Can the button labels be improved?

YES && YES

i  think it would be better to just have one button for the Registration CSV reports. Then we can:

- change the button text and action (targeted report) based on the view. ie:
   - No filtering : **All Registrations CSV Report**
   - Event filtering : **Event Registrations CSV Report**
   - Datetime filtering : **Datetime Registrations CSV Report**

> The Check-Ins column reports the quantity of check-ins, but what number does that represent? Is that a representation of all check-ins (in/out) for that registration or is it a quantity of check-in/out for a particular datetime? No matter which level of filtering and which button I use, the Check-Ins column count is the same.

It's likely counting all check-in records for each registration and therefore should be changed to take filtering into consideration.

> Do we want to add the other time stamps to the Datetime cells?

no that's not what we decided on yesterday, which was to only display the status and timestamp for the LAST check-in record for each attendee. Let's be agile and leave that off for the time being. IF customers ask for it, we can add it later.

> Just have one button for each filtering level 
&& 
Change the label of the button to indicate the filtering label e.g. All Check-in/out CSV Report, Event Check-in/out CSV Report, Datetime Check-in/out CSV Report

Yup, I've included this is my instructions below.

> Change the Check-Ins column name to be dynamic to represent the level of filtering, e.g.: All Check-Ins, Event Check-ins, "Datetime Check-ins`.

agreed, included in instructions 

> When we're at the top level of filtering (no filtering) The Check-ins column will be the sum of all check-ins for that registration across all the associated events or datetimes; would this be the same as the event filtering?

Since registrations have a 1:1 relation with Events, you are correct that this would essentially be the same for the ALL registrations CSV report with NO filtering. We should NOT include check-in data in the ALL events report since that would then add a column for each datetime for EVERY event. That will be overkill IMHO. As well, that's going to slow things down dramatically as it massively increases the amount of querying required for each row.

So imho:

- ALL Registrations CSV (no filtering) = NO check-in details
- Event CSV = check-in details for all datetimes for the selected event showing last status only
- Datetime CSV = check-in details for just the selected datetime showing ALL check-ion records and timestamps

> When we're filtering at the event level, only produce the sum of the Event Check-Ins for that event. && When we're filtering at the datetime level, only produce the sum of all the Datetime Check-Ins for that datetime.

agreed && agreed, included in instructions 

---

# ok @hoseinrafiei 

**Based on Garth's other comments, let's do the following:**

#### Registrations list table with no filtering
- change CSV report button text to "All Registrations CSV Report"
- have it run whatever report function it is running now `*`
- remove the Check-ins count column from the csv
- do NOT display any other Check-in columns

#### Registrations list table with Event filtering
- change CSV report button text to "Event Registrations CSV Report"
- have it run the report function that you were working on `*`
- **move** the Check-ins count column to just before the other Check-in columns at the far end of the csv
- change the Check-ins count column text to "Event Check-ins #"
- change the value displayed for each registration in the Check-ins count column text to the sum of ALL Check-Ins for that registrant for that event (ie: across all datetimes)
- display Check-in columns for ALL datetimes for the selected event
- the value for each Check-in column shows:
   - BLANK cell if no check-in record exists, or:
   - check-in status as either `IN` or `OUT`
   - the timestamp for the check-in using format like `2022-04-22 11:45 am`

example:

<table>
<tr>
	<td>Event Check-ins #
	<td>Check-ins for Date A - ID: 1
	<td>Check-ins for Date B - ID: 2
<tr>
	<td>3
	<td>IN: 2022-04-22 8:02 am
	<td>OUT: 2022-04-22 10:27 am
<tr>
	<td>1
	<td>
	<td>IN: 2022-04-22 3:05 pm
</table>


#### Registrations list table with Event filtering
- change CSV report button text to "Datetime Registrations CSV Report"
- have it run the report function that you were working on `*`
- **move** the Check-ins count column to just before the other Check-in columns at the far end of the csv
- change the Check-ins count column text to "Datetime Check-ins #"
- change the value displayed for each registration in the Check-ins count column text to the sum of the Check-Ins for that registrant for just the selected datetime
- display Check-in columns for only the ONE datetime that the list table was filtered for
- the value for each Check-in column shows:
   - BLANK cell if no check-in record exists, or:
   - the following status - timestamp pattern for EACH check-in record
   - check-in status as either `IN` or `OUT`
   - the timestamp for the check-in using format like `2022-04-22 11:45 am`

example:

<table>
<tr>
	<td>Check-ins for Date A - ID: 1
<tr>
	<td>IN: 2022-04-22 8:02 am -- OUT: 2022-04-22 10:27 am
<tr>
	<td>IN: 2022-04-22 3:05 pm
</table>



As well, please put the attendee personal info in the far left columns at the very start of the csv if they are not there already (with Registration ID as the very first column)

 `*` I'm not sure at the moment how those reports are set up, ie: whether they are completely different classes/functions for handling the different reports or whether they share the same logic and then respond dynamically depending on the current request. 

If the current code is a huge mess (it's pretty old so it might be really nasty) then you can extract and refactor logic if needed. 
IF you do decide to extract and refactor everything into new classes, here are a few guidelines you need to follow:

- any public API (public functions, etc) need to remain in place as is. They can be deprecated but will need to try and call any replacement functions. If need be, use a façade or adapter to translate between the old API and the new API. SO in other words, you need to maintain backwards compatibility for any existing publicly accessible code.

- any new classes created should utilize namespacing and can be located in:
    ```
	/event-espresso-core/core/domain/services/admin/registrations/list_table/csv_reports
	```
   and be namespaced like:
    ```php
	namespace EventEspresso\core\domain\services\admin\registrations\list_table\csv_reports;
	```

- IF you extract any generic common code that could be utilized by any CSV report or other logic, then that should be moved to an appropriate folder under the `EventEspresso\core\services\` namespace. 
Most new code is being moved into the following two primary namespaces:
	-  `EventEspresso\core\domain` for logic that is specific to particular business domain logic, which can be though of as a specific "use case". So a "Registrations CSV Report" is a specific business use case and therefore belongs within that "domain" namespace/folder. 
	-  `EventEspresso\core\services` for generic logic that is not for a specific use case and could be utilized by any class. This is essentially the "framework" classes that are used to build logic for the specific use cases in the domain folder above. so for example, some generic utilities to help build a CSV file, might go somewhere like: 
    ```php
	namespace `EventEspresso\core\services\csv\` 
    ```
    If you unsure of where to put something, ask in chat or in whatever issue/PR you are working on, and I will help guide you. If I'm not around, then just do what you think is best and we can change things around later.

As always, I'm here to help and expect you to ask lots of questions to help you deal with all of this chaos 😆

admin_pages/registrations/EE_Registrations_List_Table.class.php
```diff
@@ -137,6 +138,7 @@ protected function _set_properties()
             'route'         => 'registrations_report',
             'extra_request' => [
                 'use_filters' => true,
+                'incude_checkins' => isset($req_data['event_id']) ? true : false,
```
tn3rb, line 141: great approach 💯

hoseinrafiei:
@tn3rb 

Thanks for writing all of these details. I have one question regarding removing one of the buttons and keep only 1 button. I can remove one and apply the changes on the remained button but the actual question is Which button should I remove? The **All / Event Registrations** button or **Filtered** button?

The first one includes all registrations but the second one applies current filter options (like status, date) to the export. Based on your feedback, I assume that I should remove the Filtered button but I want to make sure that we're on the same page.

tn3rb:
> Which button should I remove?

well it doesn't really matter because we will be folding the functionality for all buttons into one.

> The first one includes all registrations but the second one applies current filter options (like status, date) to the export. 

right, so what we want to do now is have one button that changes dynamically based on the current filtering.

this should result in three different options/states:

- ### No filtering
   - this should essentially be the same as the current all registrations button since the reg list has not been reduced in size via filtering and should contain all registrations
   - button text: "All Registrations CSV Report"

- ### Event filtering
   - reduces the size of the previous list by only containing data for the selected event
   - button text: "Event Registrations CSV Report"

- ### Datetime filtering
   - reduces the size of the previous list even further by only containing data for a single datetime for the selected event
   - button text: "Datetime Registrations CSV Report"

My apologies for not being more clear about that. 
I knew in my head what I wanted... so like, come on... you should be able to read my mind by now!!! 😆

hoseinrafiei:
@tn3rb 

I removed one of the buttons and implemented the feature based on your guide. Also I checked all the items one by one to make sure they're developed.

Please check the 3 buttons and their CSV output and let me know if you have any feedback regarding the functionality.
For the extraction, I didn't do anything yet. I just wanted to do the functionality first and after that go for extraction.

admin_pages/registrations/Registrations_Admin_Page.core.php
```diff
@@ -227,6 +227,7 @@ protected function _define_page_props()
                 'add-attendee'        => esc_html__('Add Contact', 'event_espresso'),
                 'edit'                => esc_html__('Edit Contact', 'event_espresso'),
                 'report'              => esc_html__('Event Registrations CSV Report', 'event_espresso'),
+                'report_datetime'     => esc_html__('Datetime Registrations CSV Report', 'event_espresso'),
                 'report_all'          => esc_html__('All Registrations CSV Report', 'event_espresso'),
```
tn3rb, line 231: poifect 👌🏻

core/libraries/batch/JobHandlers/RegistrationsReport.php
```diff
@@ -55,34 +55,30 @@ public function create_job(JobParameters $job_parameters)
...
-            if ($event_id) {
-                $query_params[0]['EVT_ID'] = $event_id;
-            } else {
-                $query_params['force_join'][] = 'Event';
-            }
+                'Ticket.TKT_deleted' => array('IN', array(true, false)),
+            ),
+            'order_by'   => array('Transaction.TXN_ID' => 'asc', 'REG_count' => 'asc'),
+            'force_join' => array('Transaction', 'Ticket', 'Attendee'),
+            'caps'       => \EEM_Base::caps_read_admin,
```
tn3rb, line 77: same here plz

caffeinated/admin/extend/registrations/EE_Event_Registrations_List_Table.class.php
```diff
@@ -100,17 +101,30 @@ protected function _set_properties()
...
+                    'report' => array(
+                        'route'         => 'registrations_report',
+                        'extra_request' =>
+                            array(
+                                'EVT_ID'     => $evt_id,
+                                'return_url' => $return_url,
+                            ),
+                    ),
+                );
+            }
```
tn3rb, line 127: another potentially better way to handle this is to "isolate the thing that changes", which is a common quote from functional programming which could have also been written as "extract shared logic".

since there is a lot of duplication in the above code, we can separate things so that the shared logic is on its own, and then apply the changes in isolation. 

for example:

```php
$route_details = [
	'route'         => 'registrations_report',
	'extra_request' => [
		'EVT_ID'     => $evt_id,
		'return_url' => $return_url,
	]
];
if ($DTT_id) {
	$route_details['extra_request']['DTT_ID'] = $DTT_id;
	$this->_bottom_buttons['report_datetime'] = $route_details;
} else {
	$this->_bottom_buttons['report'] = $route_details;
}
```

Now I'm guessing that the `'report'` and `'report_datetime'` routes are very very similar, so you may want to think about consolidating those into a single route and then handle the differences based on the presence or absence of a datetime ID in the request parameters. But I'm totally guessing, and there may be bigger functional differences that require a separate route for those two reports. I'll totally leave it up to you to decide which path to take.

core/libraries/batch/JobHandlers/RegistrationsReport.php
```diff
@@ -47,6 +47,7 @@ class RegistrationsReport extends JobHandlerFile
     public function create_job(JobParameters $job_parameters)
     {
         $event_id = intval($job_parameters->request_datum('EVT_ID', '0'));
+        $DTT_id = intval($job_parameters->request_datum('DTT_ID', '0'));
```
tn3rb, line 50: he he.. Michael loved to use `intval()`  but it's a function from PHP 4 and it's advised to not use it since it may get deprecated at some point. It's better to simply type hint values. 

```php
        $event_id = (int) $job_parameters->request_datum('EVT_ID', '0');
        $DTT_id   = (int) $job_parameters->request_datum('DTT_ID', '0');
```

core/libraries/batch/JobHandlers/RegistrationsReport.php
```diff
@@ -244,17 +241,17 @@ public function continue_job(JobParameters $job_parameters, $batch_size = 50)
...
      * @param array $question_labels the IDs for all the questions which were answered by someone in this selection
      * @param array $query_params for using where querying the model
-     * @param bool $incude_checkins
+     * @param int|null $DTT_id
      * @return array top-level keys are numeric, next-level keys are column headers
      * @throws \EE_Error
      * @throws \ReflectionException
      */
-    public function get_csv_data_for($event_id, $offset, $limit, $question_labels, $query_params, $incude_checkins = false)
+    public function get_csv_data_for($event_id, $offset, $limit, $question_labels, $query_params, $DTT_id = null)
```
tn3rb, line 249: plz try to maintain type consistency whenever possible. So since `$DTT_id` is an integer, it would be better to use an integer for the default value. Now that said, this doesn't always work and there are some cases where you need to be able to determine when a value has simply not been set, and need to use a value that can not be misinterpreted as a valid value. 

For example, using `0` as a default value for a quantity may not be good in cases where you only want to apply a change when a valid new value is provided. Since zero is a potentially valid value for a quantity, it's impossible to know whether that value was simply not set, or whether the user really wants to set the quantity to zero. In that case, `null` may work better. 

But for the most part, plz try to avoid type switching as much as possible as it greatly increases the chances of introducing issues that are tough to troubleshoot.

Oh and can you plz change `$DTT_id` to `$DTT_ID`?

core/libraries/batch/JobHandlers/RegistrationsReport.php
```diff
@@ -506,12 +498,74 @@ public function get_csv_data_for($event_id, $offset, $limit, $question_labels, $
...
+                        esc_html($datetime->get('DTT_ID'))
+                    );
+                }
+                $checkins = [];
+                foreach ($checkin_rows as $checkin_row) {
+                    if ((int) $checkin_row['Checkin.CHK_in'] === 1) {
+                        /* translators: 1: check-in status, 2: check-in timestamp */
+                        $checkin_value = sprintf(
+                            esc_html__('%1$s: %2$s', 'event_espresso'),
+                            esc_html__('IN', 'event_espresso'),
```
tn3rb, line 547: don't need/want nested escaping, so this would be better as:

```php
                            esc_html__('%1$s: %2$s', 'event_espresso'),
                            __('IN', 'event_espresso'),
```

I'm also on the fence as to whether the following would be better:

```php
                        $checkin_value = sprintf(
                        	/* translators: 1: check-in timestamp */
                            esc_html__('IN: %1$s', 'event_espresso'),
                            date('Y-m-d g:i a', strtotime($checkin_row['Checkin.CHK_timestamp']))
                        );
```

core/libraries/batch/JobHandlers/RegistrationsReport.php
```diff
@@ -506,12 +498,74 @@ public function get_csv_data_for($event_id, $offset, $limit, $question_labels, $
...
+                            date('Y-m-d g:i a', strtotime($checkin_row['Checkin.CHK_timestamp']))
+                        );
+                        $checkins[] = $checkin_value;
+                    } elseif ((int) $checkin_row['Checkin.CHK_in'] === 0) {
+                        /* translators: 1: check-in status, 2: check-in timestamp */
+                        $checkin_value = sprintf(
+                            esc_html__('%1$s: %2$s', 'event_espresso'),
+                            esc_html__('OUT', 'event_espresso'),
+                            date('Y-m-d g:i a', strtotime($checkin_row['Checkin.CHK_timestamp']))
+                        );
```
tn3rb, line 557: plz change this (or not) to match whatever you end up doing above

core/libraries/batch/JobHandlers/RegistrationsReport.php
```diff
@@ -506,12 +498,74 @@ public function get_csv_data_for($event_id, $offset, $limit, $question_labels, $
...
+                // get whether or not the user has checked in
+                $reg_csv_array[ (string) esc_html__('Event Check-ins #', 'event_espresso') ] = $reg_model->count_related(
+                    $reg_row['Registration.REG_ID'],
+                    'Checkin'
+                );
                 $datetimes = EEM_Datetime::instance()->get_all_wpdb_results(
                     [
                         [
-                            'Ticket.TKT_ID' => $reg_row['Ticket.TKT_ID']
+                            'Ticket.TKT_ID' => $reg_row['Ticket.TKT_ID'],
```
tn3rb, line 571: 👍🏻

core/libraries/batch/JobHandlers/RegistrationsReport.php
```diff
@@ -548,16 +602,16 @@ public function get_csv_data_for($event_id, $offset, $limit, $question_labels, $
                     if ($checkin_row instanceof EE_Checkin && $checkin_row->get('CHK_in') === true) {
                         /* translators: 1: check-in status, 2: check-in timestamp */
                         $checkin_value = sprintf(
-                            esc_html__('%1$s - %2$s', 'event_espresso'),
+                            esc_html__('%1$s: %2$s', 'event_espresso'),
                             esc_html__('IN', 'event_espresso'),
-                            date('Y-m-d H:i:s', $checkin_row->get_raw('CHK_timestamp'))
+                            $checkin_row->get_datetime('CHK_timestamp', 'Y-m-d', 'g:i a')
```
tn3rb, line 610: LOLZ!  ya that's much better! 💯 good job

core/libraries/batch/JobHandlers/RegistrationsReport.php
```diff
@@ -554,9 +554,7 @@ public function get_csv_data_for($event_id, $offset, $limit, $question_labels, $
                     }
                 }
                 $reg_csv_array[ (string) $datetime_name ] = implode(' --- ', $checkins);
-            }
-            // Include check-in data for event
-            elseif ($event_id) {
+            } elseif ($event_id) {
```
tn3rb, line 562: thank you 🙇🏻‍♂️

core/libraries/batch/JobHandlers/RegistrationsReport.php
```diff
@@ -291,6 +297,36 @@ public function get_csv_data_for($event_id, $offset, $limit, $question_labels, $
                     $reg_row['Event_CPT.ID']
                 );
             }
+            // add attendee columns
+            foreach ($att_fields_to_include as $att_field_name) {
+                $field_obj = EEM_Attendee::instance()->field_settings_for($att_field_name);
+                if ($reg_row['Attendee_CPT.ID']) {
+                    if ($att_field_name == 'STA_ID') {
+                        $value = EEM_State::instance()->get_var(
```
tn3rb, line 305: good stuff, looks like you are having no troubles adapting to use the model system 👍🏻

core/libraries/batch/JobHandlers/RegistrationsReport.php
```diff
@@ -55,34 +55,30 @@ public function create_job(JobParameters $job_parameters)
...
+                        array(
+                            EEM_Transaction::failed_status_code,
+                            EEM_Transaction::abandoned_status_code,
                         ),
-                        // unless the registration is approved, in which case include it regardless of transaction status
-                        'STS_ID'             => \EEM_Registration::status_id_approved,
                     ),
-                    'Ticket.TKT_deleted' => array('IN', array(true, false)),
+                    // unless the registration is approved, in which case include it regardless of transaction status
+                    'STS_ID'             => \EEM_Registration::status_id_approved,
```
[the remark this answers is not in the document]
hoseinrafiei: Good point. I always use the same structure. Removed the \ characters the modified files.

caffeinated/admin/extend/registrations/EE_Event_Registrations_List_Table.class.php
hoseinrafiei: You're right. I did it.

core/libraries/batch/JobHandlers/RegistrationsReport.php
hoseinrafiei: Nice!

core/libraries/batch/JobHandlers/RegistrationsReport.php
hoseinrafiei: Yes, another good point. Thanks for sharing it.

core/libraries/batch/JobHandlers/RegistrationsReport.php
hoseinrafiei: I did the first approach since the translated IN and OR texts were totally unnecessary and they could be merged into the first `esc_html__(` function.

    esc_html__('IN: %1$s', 'event_espresso'),

I believe the translators are still able to change `: ` characters to something else.

hoseinrafiei:
@tn3rb

I applied the changes based on your feedback.
Thanks again for sharing good points.

I also have one question.

    $datetime = EEM_Datetime::instance()->get_one_by_ID($DTT_ID);

Above function returns an object but following function returns an array of arrays.

    $datetimes = EEM_Datetime::instance()->get_all_wpdb_results(...);

Is there any function that works like above code with query options but returns array of objects. An object like the output of `get_one_by_ID` function?

core/libraries/batch/JobHandlers/RegistrationsReport.php
tn3rb: ya as long as that is part of the string passed to one of the i18n functions then they can change it. 
on a tangentially related point, it's also super important to always use numbered placeholders like `%1$s` & `%2$s` as opposed to just `%s`, because it gives translators the freedom to rearrange the order of the placeholders, but you likely already knew that 😉

tn3rb:
> Is there any function that works like above code with query options but returns array of objects? An object like the output of `get_one_by_ID()` function.

If you want full control over the query parameters but only receive a single object as opposed to an array, then you can use `get_one()` which is used by `get_one_by_ID()` after adding props for the model primary key (PK)

tn3rb (approved):
This all looks GREAT now ! 💯 x 🌮

hoseinrafiei:
@tn3rb 

I extracted some codes into separate classes and removed some duplicate codes. I added following classes.

    EventEspresso\core\domain\services\admin\registrations\list_table\csv_reports\PaymentsInfo;
    EventEspresso\core\domain\services\admin\registrations\list_table\csv_reports\Attendee;
    EventEspresso\core\domain\services\admin\registrations\list_table\csv_reports\Checkins;
    EventEspresso\core\domain\services\admin\registrations\list_table\csv_reports\Registration;
    EventEspresso\core\domain\services\admin\registrations\list_table\csv_reports\Answers;

Please check the new and modified files and let me know your thoughts.
Since it was my first extraction there might be something that should be fixed based on EE standards.

core/domain/services/admin/registrations/list_table/csv_reports/Checkins.php
```diff
@@ -0,0 +1,76 @@
...
+            );
+        } else {
+            /* translators: %s: datetime ID */
+            $name = sprintf(
+                esc_html__('ID: %1$s', 'event_espresso'),
+                esc_html($datetime->get('DTT_ID'))
+            );
+        }
+
+        return $name;
```
tn3rb, line 45: this is great, but it's even cleaner and easier to read if you remove the `$name` variable and just return immediately. 
this also allows you to get rid of the else clause which reduces complexity:

```php
        if (trim($datetime->get('DTT_name'))) {
            /* translators: 1: datetime name, 2: datetime ID */
            return sprintf(
                esc_html__('Check-ins for %1$s - ID: %2$s', 'event_espresso'),
                esc_html($datetime->get('DTT_name')),
                esc_html($datetime->get('DTT_ID'))
            );
        } 
        /* translators: %s: datetime ID */
        return sprintf(
            esc_html__('ID: %1$s', 'event_espresso'),
            esc_html($datetime->get('DTT_ID'))
        );
```

core/domain/services/admin/registrations/list_table/csv_reports/Checkins.php
```diff
@@ -0,0 +1,76 @@
...
+            );
+        } elseif ($checkin instanceof EE_Checkin && $checkin->get('CHK_in') === false) {
+            /* translators: 1: check-in timestamp */
+            $value = sprintf(
+                esc_html__('OUT: %1$s', 'event_espresso'),
+                $checkin->get_datetime('CHK_timestamp', 'Y-m-d', 'g:i a')
+            );
+        }
+
+        return $value;
```
tn3rb, line 74: same here, just return immediately

core/domain/services/admin/registrations/list_table/csv_reports/PaymentsInfo.php
```diff
@@ -0,0 +1,36 @@
...
+        $payment_methods = [];
+        $gateway_txn_ids_etc = [];
+        $payment_times = [];
+        foreach ($payments_info as $payment_method_and_gateway_txn_id) {
+            $payment_methods[] = isset($payment_method_and_gateway_txn_id['name'])
+                ? $payment_method_and_gateway_txn_id['name'] : esc_html__('Unknown', 'event_espresso');
+            $gateway_txn_ids_etc[] = isset($payment_method_and_gateway_txn_id['gateway_txn_id'])
+                ? $payment_method_and_gateway_txn_id['gateway_txn_id'] : '';
+            $payment_times[] = isset($payment_method_and_gateway_txn_id['payment_time'])
+                ? $payment_method_and_gateway_txn_id['payment_time'] : '';
```
tn3rb, line 32: plz format ternary options on separate lines (if the entire thing doesn't fit on one line)

```php
            $payment_methods[] = isset($payment_method_and_gateway_txn_id['name'])
                ? $payment_method_and_gateway_txn_id['name'] 
                : esc_html__('Unknown', 'event_espresso');
            $gateway_txn_ids_etc[] = isset($payment_method_and_gateway_txn_id['gateway_txn_id'])
                ? $payment_method_and_gateway_txn_id['gateway_txn_id'] 
                : '';
            $payment_times[] = isset($payment_method_and_gateway_txn_id['payment_time'])
                ? $payment_method_and_gateway_txn_id['payment_time'] 
                : '';
```

core/domain/services/admin/registrations/list_table/csv_reports/Registration.php
```diff
@@ -0,0 +1,111 @@
...
+                    $this->reg_row['Registration.REG_date'],
+                    'no_html'
+                );
+            } else {
+                $value = EEH_Export::prepare_value_from_db_for_display(
+                    $this->reg_model,
+                    $field_name,
+                    $this->reg_row[ $field->get_qualified_column() ]
+                );
+            }
```
tn3rb, line 102: maybe a switch instead of multiple `if` `else` clauses

```php
switch ($field_name) {
	case 'REG_final_price':
		// etc
```

core/libraries/batch/JobHandlers/RegistrationsReport.php
```diff
@@ -96,16 +97,7 @@ public function create_job(JobParameters $job_parameters)
...
-            EEM_Registration::instance()->count(
-                array_diff_key(
-                    $query_params,
-                    array_flip(
-                        array('limit')
-                    )
-                )
-            )
-        );
+        $job_parameters->set_job_size($this->count_units_to_process($query_params));
```
tn3rb, line 100: 👍🏻  much better

core/libraries/batch/JobHandlers/RegistrationsReport.php
```diff
@@ -308,81 +302,11 @@ public function get_csv_data_for($event_id, $offset, $limit, $question_labels, $
...
-                if ($field_name == 'REG_final_price') {
-                    // add a column named Currency after the final price
-                    $reg_csv_array[ (string) esc_html__("Currency", "event_espresso") ] = EE_Config::instance()->currency->code;
-                }
-            }
+            $reg_csv_array = (new Attendee($att_fields_to_include, $reg_row))
+                ->addAttendeeColumns($reg_csv_array);
+            // add registration columns
+            $reg_csv_array = (new Registration($reg_fields_to_include, $reg_row))
+                ->addRegistrationColumns($reg_csv_array);
```
tn3rb, line 309: let's split these up and verify the instance so that we don't violate the "golden rule" of programming

> "If it can be `null`, it will be `null`

As well, I'm thinking we should maybe clarify the names of these new classes a bit to differentiate them from regular entities, ie: `AttendeeCSV` and `RegistrationCSV`

```php
            $attendee_csv = new AttendeeCSV($att_fields_to_include, $reg_row);
            $reg_csv_array = $attendee_csv instanceof AttendeeCSV 
		? $attendee_csv->addAttendeeColumns($reg_csv_array)
		: $reg_csv_array;
            // add registration columns
	    $registration_csv = new RegistrationCSV($reg_fields_to_include, $reg_row);
            $reg_csv_array = $registration_csv instanceof RegistrationCSV
		? $registration_csv->addRegistrationColumns($reg_csv_array)
		: $reg_csv_array;
```

core/libraries/batch/JobHandlers/RegistrationsReport.php
```diff
@@ -469,50 +387,9 @@ public function get_csv_data_for($event_id, $offset, $limit, $question_labels, $
...
-                            EEM_Answer::instance(),
-                            'ANS_value',
-                            $answer_row['Answer.ANS_value']
-                        )
-                    );
-                }
-            }
+            // add answer columns
+            $reg_csv_array = (new Answers($reg_row))
+                ->addAnswerColumns($reg_csv_array, $question_labels);
```
tn3rb, line 392: same deal here as above

core/libraries/batch/JobHandlers/RegistrationsReport.php
```diff
@@ -525,55 +402,33 @@ public function get_csv_data_for($event_id, $offset, $limit, $question_labels, $
                         ]
                     ]
                 );
+                /** @var EE_Datetime $datetime */
                 $datetime = EEM_Datetime::instance()->get_one_by_ID($DTT_ID);
-                $checkin_rows = EEM_Checkin::instance()->get_all_wpdb_results(
+                $checkin_rows = EEM_Checkin::instance()->get_all(
```
tn3rb, line 407: 👍🏻

core/libraries/batch/JobHandlers/RegistrationsReport.php
```diff
@@ -662,19 +492,20 @@ public function get_csv_data_for($event_id, $offset, $limit, $question_labels, $
...
-        return EEM_Registration::instance()->count($query_params);
+        return EEM_Registration::instance()->count(
+            array_diff_key(
+                $query_params,
+                array_flip(
+                    ['limit']
+                )
+            )
+        );
     }
```
tn3rb, line 509: hmmm... this is a big change... 🤔 
was this method not being used before? IF you are 1000% certain this was not being used anywhere else (ie: search for the method name directly as opposed to relying on your IDE to know) then this is great!

tn3rb (commented):
### LOVE IT 💯 

Absolutely great changes that make the code so much easier to read and maintain.

I did have a few suggestions for improving things a little bit more, but can't stress enough how much I appreciate good work like this 👍🏻

core/domain/services/admin/registrations/list_table/csv_reports/Answers.php
```diff
@@ -0,0 +1,93 @@
...
+use EEM_State;
+
+/**
+ * Class Answers
+ * Add question / answer columns to the CSV row
+ *
+ * @author  Hossein Rafiei
+ * @package EventEspresso\core\domain\services\admin\registrations\list_table\csv_reports
+ */
+class Answers
```
tn3rb, line 18: maybe rename to `AnswersCSV` to help differentiate this from any other similarly named entities

core/domain/services/admin/registrations/list_table/csv_reports/Attendee.php
```diff
@@ -0,0 +1,86 @@
...
+use EEM_State;
+
+/**
+ * Class Attendee
+ * Adds attendee columns to the CSV row
+ *
+ * @author  Hossein Rafiei
+ * @package EventEspresso\core\domain\services\admin\registrations\list_table\csv_reports
+ */
+class Attendee
```
tn3rb, line 18: maybe `AttendeeCSV`

core/domain/services/admin/registrations/list_table/csv_reports/Checkins.php
```diff
@@ -0,0 +1,76 @@
...
+use ReflectionException;
+
+/**
+ * Class Checkins
+ * Renders the Checkin data
+ *
+ * @author  Hossein Rafiei
+ * @package EventEspresso\core\domain\services\admin\registrations\list_table\csv_reports
+ */
+class Checkins
```
tn3rb, line 17: `CheckinsCSV`

core/domain/services/admin/registrations/list_table/csv_reports/PaymentsInfo.php
```diff
@@ -0,0 +1,36 @@
...
+namespace EventEspresso\core\domain\services\admin\registrations\list_table\csv_reports;
+
+/**
+ * Class PaymentsInfo
+ * Extracts payment information using the payment records
+ *
+ * @author  Hossein Rafiei
+ * @package EventEspresso\core\domain\services\admin\registrations\list_table\csv_reports
+ */
+class PaymentsInfo
```
tn3rb, line 12: `PaymentsInfoCSV`

core/domain/services/admin/registrations/list_table/csv_reports/Registration.php
```diff
@@ -0,0 +1,111 @@
...
+     * @param array $fields
+     * @param array $reg_row
+     * @throws EE_Error
+     * @throws ReflectionException
+     */
+    public function __construct(array $fields, array $reg_row)
+    {
+        $this->fields = $fields;
+        $this->reg_row = $reg_row;
+        $this->reg_model = EE_Registry::instance()->load_model('Registration');
```
tn3rb, line 52: the model singletons can be accessed directly like: `EEM_Registration::instance()`
and are always loaded on every request, so no need to call this loader method (it's super old)

tn3rb (commented):
Had a few other afterthoughts regarding the new classes you created in `core/domain/services/admin/registrations/list_table/csv_reports/` 

- these could potentially be refactored into static services, meaning they would solely operate on data while maintaining no internal state. So instead of creating an instance and injecting _some_ of your data into the constructor where it gets set as a class property, you would simply pass **everything** to the appropriate method. This would be a more pure functional approach with no unexpected side effects: data in, data out, nothing else.

- wouldn't be possible for ALL of these classes since a couple of them operate quite differently, but it would be real nice if the API could be made more consistent where possible, even to the point of utilizing a common interface (as long as that happened easily/naturally and didn't require any hacky weirdness). So for example, the `AttendeeCSV`, & `RegistrationCSV` classes could both potentially share a common interface like:

	```php
	interface EntityCSV
	{
		public static function generateCsvColumns(array $reg_row, array $fields, array $data): array;
	}
	```

these comments are just food for thought and not anything that you need to act upon. I'm simply trying to give you further ideas about how things can be approached. Over my years in programming I've been gravitating more and more towards trying to make clear boundaries between types of classes, such as:

 - Entities: objects that represent a **unique** real world thing that most likely requires persistence to the database, and therefore some kind of unique identifier.

 - Value Objects: also represent real world things, but generally do not require a unique instance, for example, an object representing a currency. A USD currency object should be interchangeable with any other USD currency object.

 - Data Transfer Objects (DTOs): super simple objects used for gathering data and passing along to another entity or service. 

 - Services: classes that operate on data objects like those above in order to "do something". Ideally these classes should consist of "**pure**" functions that maintain zero internal state and cause zero side effects. This means that service class functions should always produce the exact same output when given the same input.

obviously that is not a complete list of class types, but aiming for those types can seriously improve the quality of one's coding.

Anyways, again, this is just food for thought intended as some general guidance and not anything you need to act upon for this PR which is already a great improvement over what was there previously.

core/libraries/batch/JobHandlers/RegistrationsReport.php
hoseinrafiei: Yes, I already confirmed it by searching the function name on whole `plugins` directory.

tn3rb: awesome thnx 👍🏻

core/domain/services/admin/registrations/list_table/csv_reports/Checkins.php
hoseinrafiei: Yes, It's better. Done.

core/domain/services/admin/registrations/list_table/csv_reports/Checkins.php
hoseinrafiei: Done.

core/domain/services/admin/registrations/list_table/csv_reports/PaymentsInfo.php
hoseinrafiei: Done.

core/domain/services/admin/registrations/list_table/csv_reports/Registration.php
hoseinrafiei: Done.

core/domain/services/admin/registrations/list_table/csv_reports/Registration.php
hoseinrafiei: Thanks for sharing. With the new method (static functions), I receive it as parameter.

hoseinrafiei:
@tn3rb 

All of the new classes renamed to what you suggested and the codes are reformatted based on your advise.
Please check them and let me know if there is anything else that should be done.

core/domain/services/admin/registrations/list_table/csv_reports/CheckinsCSV.php
```diff
@@ -56,21 +54,20 @@ public static function getDatetineLabel(EE_Datetime $datetime)
...
             /* translators: 1: check-in timestamp */
-            $value = sprintf(
+            return sprintf(
                 esc_html__('OUT: %1$s', 'event_espresso'),
                 $checkin->get_datetime('CHK_timestamp', 'Y-m-d', 'g:i a')
             );
         }
 
-        return $value;
+        return null;
```
tn3rb, line 71: plz return an empty string in order to maintain data type consistency

core/libraries/batch/JobHandlers/RegistrationsReport.php
```diff
@@ -302,11 +302,9 @@ public function get_csv_data_for($event_id, $offset, $limit, $question_labels, $
                 );
             }
             // add attendee columns
-            $reg_csv_array = (new Attendee($att_fields_to_include, $reg_row))
-                ->addAttendeeColumns($reg_csv_array);
+            $reg_csv_array = AttendeeCSV::addAttendeeColumns($att_fields_to_include, $reg_row, $reg_csv_array);
             // add registration columns
-            $reg_csv_array = (new Registration($reg_fields_to_include, $reg_row))
-                ->addRegistrationColumns($reg_csv_array);
+            $reg_csv_array = RegistrationCSV::addRegistrationColumns($reg_fields_to_include, $reg_row, $reg_model, $reg_csv_array);
```
tn3rb, line 307: this is both easier to read and looks more reliable, good job 👍🏻

tn3rb (approved):
### Great Job 💯 🎆

core/domain/services/admin/registrations/list_table/csv_reports/CheckinsCSV.php
hoseinrafiei: Fixed.

hoseinrafiei:
@tn3rb

The last item is done.
Do you think that we should ask Sam or Moh to test it before merge or we're already good to go for merge?

garthkoyle:
I would like to test this before it's merged. Is it ready for testing? I think @joaovictorangeline and @Saam01 could help test this.

garthkoyle:
I tested:
- single registration
- multiple registrations
- check-in
- check-out
- check-in again
- filtering to the event and datetime level
- exporting data

This is an example of:
- the first row checking in and out multiple times, in to Datetime B, and the in and out to Datetime A
- the second row checking in to Datetime B
- the thrid row checking in and out to datetime b
 
image

I think this looks really good.

hoseinrafiei:
> I think this looks really good.

Great. Therefore, I will merge this into `master`.

PR #3812, opened 2022-04-27
Fix StatusChangeNotice Code Vomit

tn3rb:
Due to the escaping work being done in master, the status change notice in dev branch vomited code all over the screen because some JS and CSS that was lazily dumped into the template by myself was now being escaped and rendered as HTML instead of code:

Screenshot 2022-04-27 103524

this PR fixes that by extracting the assets from that template and loading them properly.


knazart (dismissed):
Looks good to me 👍

PR #3817, opened 2022-04-28
Handle Event Query `orderby` Params as Arrays

### LINKED ISSUE #3795: Fix EEH_Event_Query orderby parameters returning empty string
This has been reported in relation to the table view template add-on but I tracked it down to core, so adding this to core for now.

Easiest way to reproduce is using the table view template and setting the order_by parameter to use `event_name`:

`[ESPRESSO_EVENTS_TABLE_TEMPLATE order_by=event_name]`

The output will be the default order by which uses start_date.

Tracked this down to [HERE](https://github.com/eventespresso/event-espresso-core/blob/master/core/helpers/EEH_Event_Query.helper.php#L183)

It returns an empty string and I think it is because `event_query_orderby` is an array with a single string element.

If I switch line 185 to be:

`$event_query_orderby = self::getRequest()->getRequestParam('event_query_orderby', $orderby, 'string');`

It works but wasn't sure if the other calls to getRequestParam needed the same change and/or if it would cause issues with other code using that class?

tn3rb:
fixes #3795

this PR ensures that the Event Query `orderby` param is handled as an array

### DIFF
--- a/core/helpers/EEH_Event_Query.helper.php
+++ b/core/helpers/EEH_Event_Query.helper.php
@@ -3,6 +3,7 @@
 use EventEspresso\core\exceptions\InvalidDataTypeException;
 use EventEspresso\core\exceptions\InvalidInterfaceException;
 use EventEspresso\core\services\loaders\LoaderFactory;
+use EventEspresso\core\services\request\DataType;
 use EventEspresso\core\services\request\RequestInterface;
 
 /**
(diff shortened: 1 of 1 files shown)


core/helpers/EEH_Event_Query.helper.php
```diff
@@ -172,15 +173,20 @@ private static function _show_expired($show_expired = false)
...
      */
     private static function _orderby($orderby = 'start_date')
     {
-        $event_query_orderby = self::getRequest()->getRequestParam('event_query_orderby', $orderby);
+        $event_query_orderby = self::getRequest()->getRequestParam(
+            'event_query_orderby',
+            (array) $orderby,
+            DataType::STRING,
+            true
+        );
```
hoseinrafiei, line 189: Nice 👍

PR #3831, opened 2022-05-04
Kses - Part 9

hoseinrafiei:

Applied `wp_kses` function to some more variables related to admin panel.

Checklist: https://trello.com/c/WWMDoGAH/1160-ee4-core-testing-checklist-pr-3831

### DIFF
--- a/core/admin/EE_Admin_List_Table.core.php
+++ b/core/admin/EE_Admin_List_Table.core.php
@@ -1,5 +1,7 @@
 <?php
 
+use EventEspresso\core\services\request\sanitizers\AllowedTags;
+
 if (! class_exists('WP_List_Table')) {
     require_once ABSPATH . 'wp-admin/includes/class-wp-list-table.php';
 }
@@ -468,7 +470,7 @@ abstract class EE_Admin_List_Table extends WP_List_Table
             return;
         }
         foreach ($filters as $filter) {
-            echo $filter; // already escaped
+            echo wp_kses($filter, AllowedTags::getWithFormTags());
         }
         // add filter button at end
         echo '<input type="submit" class="button-secondary" value="'
@@ -749,7 +751,7 @@ abstract class EE_Admin_List_Table extends WP_List_Table
                     $item,
                     $this
                 );
-                echo $this->handle_row_actions($item, $column_name, $primary);
+                echo wp_kses($this->handle_row_actions($item, $column_name, $primary), AllowedTags::getWithFormTags());
                 echo "</td>";
             } else {
                 echo "<td $attributes>"; // already escaped
@@ -760,7 +762,7 @@ abstract class EE_Admin_List_Table extends WP_List_Table
                     $column_name,
                     $this
                 );
-                echo $this->handle_row_actions($item, $column_name, $primary);
+                echo wp_kses($this->handle_row_actions($item, $column_name, $primary), AllowedTags::getWithFormTags());
                 echo "</td>";
             }
         }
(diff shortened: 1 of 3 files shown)


hoseinrafiei:
@tn3rb 

Please check this PR and let me know your thoughts.
Also I believe we're on final stages of escaping printed variables. What do you suggest to do after this?

tn3rb (approved):
looks good, this will require a lot of testing since it affects the base admin classes

tn3rb:
re:
> Also I believe we're on final stages of escaping printed variables. What do you suggest to do after this?

hmmm... have we done the modules yet? don't think we've done the ticket selector or SPCO (single page checkout) yet 🤔

hoseinrafiei:
> hmmm... have we done the modules yet? don't think we've done the ticket selector or SPCO (single page checkout) yet 🤔

Yes, already done. Only 3 items are remained in the `/modules` directory which do not need any escape.

Screen Shot 2022-05-05 at 18 32 29

hoseinrafiei:
@mohdshakeel @Saam01 

Please check this PR using the https://trello.com/c/WWMDoGAH/1160-ee4-core-testing-checklist-pr-3831 checklist and report any bug you found. You should focus on admin menus.

Saam01:
@hoseinrafiei Start testing the above.

hoseinrafiei:
@Saam01 @mohdshakeel 
Please update us about the process.

Saam01:
@hoseinrafiei I am testing the list.

For now, I found a minor issue, so not creating any separate ticket for it: 
image

This issue is not available in the Master branch.

hoseinrafiei:
> image

@tn3rb the reported problem is due to following code in line 1355 of `/core/admin/EE_Admin_Page_CPT.core.php` file.

Screen Shot 2022-05-10 at 15 39 41

Curious to know why it implemented this way and is there any reason that we didn't fix it before?

tn3rb:
well...  it looks like `inject_nav_tabs()` is a callback for this hook:

```php
add_action('post_edit_form_tag', [$this, 'inject_nav_tabs']);
```

which fires INSIDE of the `<form` tag in `wp-admin/edit-form-advanced.php`

So looks like the dev that did that was trying to figure out a way to make those nav tabs appear above everything else.
Why they did not just use the `edit_form_top` hook point a little bit below that is a good question 🤔 
It _appears_ that there are only hidden input fields rendered at that point, which should not interfere with the display of the nav tabs.

So can you plz try this:

- change the `add_action()` on line 354 of `EE_Admin_Page_CPT` to:

    ```php
		add_action('edit_form_top', [$this, 'inject_nav_tabs']);
	```

- then change `inject_nav_tabs()` to simply be:

    ```php
    	public function inject_nav_tabs()
	    {
	        echo $this->_get_main_nav_tabs();
	    }
	```

   but add the appropriate escaping to that 

 - then confirm that the nav tabs at the top of the admin pages work as expected

Saam01:
The "Payment Received" message not queued (same issue in the masters):

https://user-images.githubusercontent.com/99984614/168121519-65254eaa-eb49-435e-9e0e-cfa93162f8fc.mp4

tn3rb (approved):
👍🏻  poifect

tn3rb:
@Saam01 

re:
> The "Payment Received" message not queued (same issue in the masters):

if you mean that the same issue exists in `master` branch then it is not a side effect of the work done in this branch.
can you look if there is already an existing ticket for the issue you found, and if not, then create one plz?

hoseinrafiei:
@Saam01 

Please finish since I need to merge it to master if everything was fine.

For the issue that you found as I said, if you can replicate the same issue in `master` branch, then you should create a separate issue topic since it's not related to this specific branch.

Saam01:
@raphaelkross I've finished testing the PR. Also, I'm adding a new issue.

PR #3836, opened 2022-05-11
Fix Entity Sorting

tn3rb:
This PR:

- fixes https://github.com/eventespresso/eventsmart.com-website/issues/965
- simply fixes the name for an array key that was preventing updates from happening

### DIFF
--- a/core/domain/services/graphql/mutators/EntityReorder.php
+++ b/core/domain/services/graphql/mutators/EntityReorder.php
@@ -49,23 +49,18 @@ class EntityReorder
                     esc_html__('Sorry, you do not have the required permissions to reorder entities', 'event_espresso')
                 );
             }
-
+            $ok = true;
+            // 'entities', 'entityGuids', 'entity_db_ids', 'entityType', 'keyPrefix'
             $details = EntityReorder::prepareEntityDetailsFromInput($input);
-
             $orderKey  = $details['keyPrefix'] . '_order'; // e.g. "TKT_order"
-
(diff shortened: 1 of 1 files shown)


core/domain/services/graphql/mutators/EntityReorder.php
```diff
@@ -49,23 +49,18 @@ public static function mutateAndGetPayload()
...
-
             // We do not want to continue reorder if one fails.
             // Thus wrap whole loop in try-catch
             try {
-                foreach ($details['entityDbids'] as $order => $entityDbid) {
-                    $args = [
-                        $orderKey => $order + 1,
-                    ];
-                    $details['entities'][ $entityDbid ]->save($args);
+                foreach ($details['entity_db_ids'] as $order => $entityDbid) {
```
tn3rb, line 60: this was the actual bug... it was previously using `'entityDbids'` as the key and therefore not finding any data to update

core/domain/services/graphql/mutators/EntityReorder.php
```diff
@@ -49,23 +49,18 @@ public static function mutateAndGetPayload()
...
+                    $args = [ $orderKey => $order + 1 ];
+                    $ok = $details['entities'][ $entityDbid ]->save($args) ? $ok : false;
```
tn3rb, line 62: this was done to improve feedback, since the previous way would return true even if updates failed

knazart (approved):
👍

PR #3840, opened 2022-05-16
Kses - Part 11

hoseinrafiei:

Applied `esc_attr` function to some HTML attributes in pricing and ticket management system. Whole event creation and pricing menu should be checked.

Testing checklist: https://trello.com/c/uQ44e6bH/1162-ee4-core-testing-checklist-pr-3840

### DIFF
--- a/caffeinated/admin/new/pricing/espresso_events_Pricing_Hooks.class.php
+++ b/caffeinated/admin/new/pricing/espresso_events_Pricing_Hooks.class.php
@@ -1246,7 +1246,7 @@ class espresso_events_Pricing_Hooks extends EE_Admin_Hooks
                 ),
         );
         $template_args['show_trash'] = count($all_datetimes) === 1 && $template_args['trash_icon'] !== 'ee-lock-icon'
-            ? ' style="display:none"'
+            ? 'display:none'
             : '';
         // allow filtering of template args at this point.
         $template_args = apply_filters(
@@ -1288,7 +1288,7 @@ class espresso_events_Pricing_Hooks extends EE_Admin_Hooks
             'event_datetimes_name'              => $default ? 'DTTNAMEATTR' : 'edit_event_datetimes',
             'DTT_description'                   => $default ? '' : $datetime->get_f('DTT_description'),
             'datetime_tickets_list'             => $default ? '<li class="hidden"></li>' : '',
-            'show_tickets_row'                  => ' style="display:none;"',
+            'show_tickets_row'                  => 'display:none;',
             'add_new_datetime_ticket_help_link' => EEH_Template::get_help_tab_link(
                 'add_new_ticket_via_datetime',
                 $this->_adminpage_obj->page_slug,
@@ -1465,7 +1465,7 @@ class espresso_events_Pricing_Hooks extends EE_Admin_Hooks
             $TKT_taxable = '';
         } else {
             $TKT_taxable = $ticket->taxable()
-                ? ' checked'
+                ? 'checked'
                 : '';
         }
         if ($default) {
@@ -1488,7 +1488,7 @@ class espresso_events_Pricing_Hooks extends EE_Admin_Hooks
             'TKT_order'                     => $default ? 'TICKETNUM' : $ticket_row,
             // on initial page load this will always be the correct order.
             'tkt_status_class'              => $ticket_status_class,
-            'display_edit_tkt_row'          => ' style="display:none;"',
+            'display_edit_tkt_row'          => 'display:none;',
             'edit_tkt_expanded'             => '',
             'edit_tickets_name'             => $default ? 'TICKETNAMEATTR' : 'edit_tickets',
             'TKT_name'                      => $default ? '' : $ticket->get_f('TKT_name'),
@@ -1547,11 +1547,11 @@ class espresso_events_Pricing_Hooks extends EE_Admin_Hooks
             'TKT_base_price_ID'             => $default || ! $base_price instanceof EE_Price ? 0 : $base_price->ID(),
             'show_price_modifier'           => count($prices) > 1 || ($default && $count_price_mods > 0)
                 ? ''
-                : ' style="display:none;"',
+                : 'display:none;',
             'show_price_mod_button'         => count($prices) > 1
                                                || ($default && $count_price_mods > 0)
                                                || (! $default && $ticket->deleted())
-                ? ' style="display:none;"'
+                ? 'display:none;'
                 : '',
             'total_price_rows'              => count($prices) > 1 ? count($prices) : 1,
             'ticket_datetimes_list'         => $default ? '<li class="hidden"></li>' : '',
@@ -1562,7 +1562,7 @@ class espresso_events_Pricing_Hooks extends EE_Admin_Hooks
             'TKT_taxable'                   => $TKT_taxable,
             'display_subtotal'              => $ticket instanceof EE_Ticket && $ticket->taxable()
                 ? ''
-                : ' style="display:none"',
+                : 'display:none;',
             'price_currency_symbol'         => EE_Registry::instance()->CFG->currency->sign,
             'TKT_subtotal_amount_display'   => EEH_Template::format_currency(
                 $ticket_subtotal,
@@ -1585,7 +1585,7 @@ class espresso_events_Pricing_Hooks extends EE_Admin_Hooks
                 : 'clone-icon ee-icon ee-icon-clone clickable',
         );
         $template_args['trash_hidden'] = count($all_tickets) === 1 && $template_args['trash_icon'] !== 'ee-lock-icon'
-            ? ' style="display:none"'
+            ? 'display:none'
             : '';
         // handle rows that should NOT be empty
         if (empty($template_args['TKT_start_date'])) {
@@ -1699,7 +1699,7 @@ class espresso_events_Pricing_Hooks extends EE_Admin_Hooks
             $template_args = array(
                 'display_tax'       => ! empty($ticket) && $ticket->get('TKT_taxable')
                     ? ''
-                    : ' style="display:none;"',
+                    : 'display:none;',
                 'tax_id'            => $tax->ID(),
                 'tkt_row'           => $ticket_row,
                 'tax_label'         => $tax->get('PRC_name'),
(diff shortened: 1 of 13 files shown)


caffeinated/admin/new/pricing/espresso_events_Pricing_Hooks.class.php
```diff
@@ -1794,21 +1794,21 @@ protected function _get_ticket_price_row(
...
                 : '',
             'show_currency_symbol'  => ($default && empty($price)) || $price->is_percent()
-                ? ' style="display:none"'
+                ? 'display:none'
                 : '',
             'PRC_amount'            => $default && empty($price)
                 ? 0
                 : $price->get_pretty('PRC_amount', 'localized_float'),
             'show_percentage'       => ($default && empty($price)) || ! $price->is_percent()
-                ? ' style="display:none;"'
```
tn3rb, line 1811: this was a silly way to do this wasn't it?
should have been done via a css class, oh well 🤷🏻

caffeinated/admin/new/pricing/templates/event_tickets_datetime_attached_tickets_row.template.php
```diff
@@ -18,7 +18,7 @@
 <tr id="advanced-dtt-edit-row-<?php echo esc_attr($dtt_row); ?>" class="advanced-dtt-edit-row">
     <td colspan="7">
         <section id="edit-event-datetime-tickets-<?php echo esc_attr($dtt_row); ?>"
-                 class="datetime-tickets-edit"<?php echo $show_tickets_row; ?>>
+                 class="datetime-tickets-edit" style="<?php echo esc_attr($show_tickets_row); ?>">
```
tn3rb, line 21: 👍🏻

caffeinated/admin/new/pricing/templates/event_tickets_metabox_main.template.php
```diff
@@ -140,15 +140,15 @@ class="ee-icon ee-icon-tickets ee-icon-size-20"></span><?php
                         'Available Tickets',
                         'event_espresso'
                     ); ?></span></h3>
-        <div class="event-tickets-container ee-create-ticket-button"<?php echo $show_tickets_container; ?>>
+        <div class="event-tickets-container ee-create-ticket-button">
```
tn3rb, line 143: i'm assuming (and hoping) this template variable doesn't exist anywhere?

tn3rb (approved):
Looks Great 👍🏻 🦖 🌴

caffeinated/admin/new/pricing/templates/event_tickets_metabox_main.template.php
hoseinrafiei: It was always empty so I deleted it instead of escaping an empty string 😉

caffeinated/admin/new/pricing/espresso_events_Pricing_Hooks.class.php
hoseinrafiei: Yes, a lot better!

caffeinated/admin/new/pricing/templates/event_tickets_metabox_main.template.php
tn3rb: was the variable being passed though?
it may have been run through a filter so that users could change it.
so it may have always been empty using the plugin as is, but may have a value when filtered on a user site.
can you plz confirm?

hoseinrafiei: It defined in `espresso_events_Pricing_Hooks::pricing_metabox` function located in `caffeinated/admin/new/pricing/espresso_events_Pricing_Hooks.class.php` file.

In that function there is no filter to modify the values of template args but there is a general hook in `EEH_Template::display_template` function named `FHEE__EEH_Template__display_template__template_args` which is used for all templates and not specifically this one.

tn3rb: hmmm... ya... we should keep that in just in case. 
Can you plz add the `$show_tickets_container` variable back and escape it?
thnx

tn3rb: you'll have to use `wp_kses()` with your custom form attributes

tn3rb: > After applying wp_kses function we will have exact same string.

??? 

How do you figure that? 
You're saying it wouldn't strip out anything that isn't an allowed tag or attribute?
I would expect the resulting string to only be `value="foobar"` since `value` is likely the only attribute that is currently allowed. The others would be removed, no?

IIRC we filtered the allowed tags, so if a user wants to inject `never-exists-such-attribute` they will need to first add that to the allowed attributes. If we didn't filter the allowed tags then we should.

Am i missing something here?

hoseinrafiei: $show_tickets_container = 'value="foobar" notallowed="value" never-exists-such-attribute="yes" onerror="alert(\'XSS\');"';
    <div class="event-tickets-container ee-create-ticket-button" <?php echo wp_kses($show_tickets_container, AllowedTags::getAllowedTags()); ?>">

Above code will return the exact string becuase `wp_kses` function won't consider the string as HTML (which is not) and returns it without any change. If we put a HTML code then it works as expected and will remove the unwanted attributes and tags.

Another option can be to convert whole `div` (from opening tag to the closing one) to a PHP string and then apply `wp_kses` to the string and print it but it will very messy.

tn3rb: didn't know `wp_kses()` had to be "primed" with a tag to make it work! TIL

hmm ok... so what about just slapping `sanitize_text_field()` on that then?
That won't work on full HTML but might be ok in this case

core/services/request/sanitizers/MultipleAttributes.php
```diff
@@ -0,0 +1,35 @@
+<?php
+
+namespace EventEspresso\core\services\request\sanitizers;
+
+/**
+ * Class MultipleAttributes
+ *
+ * @author  Hossein Rafiei
+ * @package EventEspresso\core\services\request\sanitizers
+ * @since   4.10.30
```
tn3rb, line 10: using the following will swap the placeholder for whatever version this actually gets released under

```php
 * @since   $VID:$
```

core/services/request/sanitizers/MultipleAttributes.php
```diff
@@ -0,0 +1,35 @@
...
+ */
+class MultipleAttributes
```
tn3rb, line 12: can you plz rename this class to `AttributesSanitizer` and update the file name accordingly?

core/services/request/sanitizers/MultipleAttributes.php
```diff
@@ -0,0 +1,35 @@
...
+{
+    /**
+     * @param string $attributes
+     * @param array $allowed_tags
+     * @param string $tag
+     * @return mixed|string
+     */
+    public static function clean(string $attributes, array $allowed_tags, string $tag = 'div')
```
tn3rb, line 20: is there any reason you added the `$tag` parameter?
seems it would make zero difference what the tag was since we remove it anyways.
just hardcoding that to `div` would also mean we don't have to calculate the `strpos()` for the start.

core/services/request/sanitizers/MultipleAttributes.php
```diff
@@ -0,0 +1,35 @@
...
+    {
+        if (trim($attributes) === '') {
+            return '';
+        }
+        $html = '<' . $tag . ' ' . $attributes . '>';
```
tn3rb, line 25: maybe just:

```php
        $html = "<div $attributes>";
```

core/services/request/sanitizers/MultipleAttributes.php
```diff
@@ -0,0 +1,35 @@
...
+        $escaped = wp_kses($html, $allowed_tags);
+        $start = strpos($escaped, ' ');
```
tn3rb, line 27: unnecessary if we drop `$tag` and hardcode its value as `div`

tn3rb (changes requested):
### LOVE IT! 

Super clever solution to this problem, absolutely great stuff 💯 x 💯 X 🍰 

I did have a couple of requests and questions

core/services/request/sanitizers/MultipleAttributes.php
hoseinrafiei: Since it's already part of Sanitizers namespace, I thought that `MultipleAttributes` is cleaner and present the function of this class better but will do that.

hoseinrafiei: Yes, great points. Renamed the class 👍

core/services/request/sanitizers/MultipleAttributes.php
hoseinrafiei: Nice.
What will change that to the latest version? Our own script or Github itself?

tn3rb: currently it's a script on the `buildmachine` server that packages things up for release.
there has been some work done to migrate that to GitHub actions, but it's not fully completed yet.
eventually everything will be done via GitHub actions as part of our CI/CD pipeline.

tn3rb: OH and FYI, that tag originally came from phpStorm and you can have that added automatically by editing your live templates:

Screenshot 2022-05-26 102917

core/services/request/sanitizers/MultipleAttributes.php
tn3rb: OHHHHHHH   yes of course!

great awesome, thnx for the explanation!

Great job then and plz disregard my comments about changing that 😆

core/services/request/sanitizers/MultipleAttributes.php
tn3rb: pffft disregard!

core/services/request/sanitizers/MultipleAttributes.php
tn3rb: pffft disregard!

tn3rb (approved):
# GREAT STUFF!!! 💯 🌮

hoseinrafiei:
@Saam01 Please start working on the checklist of this PR: https://trello.com/c/uQ44e6bH/1162-ee4-core-testing-checklist-pr-3840

Saam01:
I've finished testing the PR.

PR #3841, opened 2022-05-19
Remove fallback reCAPTCHA area ?

knazart:
Resolves: https://github.com/eventespresso/eventsmart.com-website/issues/973

As noted in the the above issue, when the recaptcha theme is black or white, there's two `g-recaptcha-response` text areas being rendered on the registration form page (at least on ES). Those were conflicting so I removed that extra area from the EE recaptcha template.

As of today, [gracaptcha mostly only needs](https://developers.google.com/recaptcha/docs/display#auto_render) the `g-recaptcha` div with the key and a callback.

Maybe I'm missing something, but I was not able to find the cause of why we needed that extra `<noscript>` area. So please note here if someone remembers the history of this template, because maybe this is not the best way to fix this issue. Looks like Brent was the one who created it.

Black white and invisible reCAPTCHA themes still seem to be working fine.

### Testing:
- Try if all the reCAPTCHA option variations still work in the Reg Form Settings, checking if the recaptcha passes on form submission etc. The setup that produced the issue was: https://monosnap.com/file/6cJ7c0I79pTdR04tOWTMSlURKeeyof

### DIFF
--- a/caffeinated/modules/recaptcha/templates/recaptcha.template.php
+++ b/caffeinated/modules/recaptcha/templates/recaptcha.template.php
@@ -12,14 +12,12 @@ if (! defined('EVENT_ESPRESSO_VERSION')) {
              data-theme="<?php echo esc_attr($recaptcha_theme); ?>" data-type="<?php echo esc_attr($recaptcha_type); ?>"
              data-callback="espresso_recaptcha_verification"></div>
         <noscript>
-            <div style="width: 302px; height: 352px;">
-                <div style="width: 302px; height: 352px; position: relative;">
-                    <div style="width: 302px; height: 352px; position: absolute;">
-                        <iframe
-                            src="https://www.google.com/recaptcha/api/fallback?k=<?php echo esc_attr($recaptcha_publickey); ?>"
-                            frameborder="0" scrolling="no" style="width: 302px; height:352px; border-style: none;">
-                        </iframe>
-                    </div>
+            <div style="position: relative;">
+                <div style="position: absolute;">
+                    <iframe
+                        src="https://www.google.com/recaptcha/api/fallback?k=<?php echo esc_attr($recaptcha_publickey); ?>"
+                        style="border-style: none;">
+                    </iframe>
                 </div>
             </div>
         </noscript>


tn3rb:
My memory isn't the greatest these days but i'm thinking it had to do with iframe support and/or shortcodes.
Plz try embedding the ticket selector in another site and see if the recaptcha still works, if it doesn't, then try it again without this change to verify things.

It's also possible that this was just how Google did things in their example code. I would have just followed any of their documentation as closely as possible. But again, iirc it was to facilitate iframes.

tn3rb (approved):
technically nothing wrong with this change as long as things still works everywhere

hoseinrafiei (approved):
Looks fine 💯

knazart:
I'd rather leave the fallback iframe area, just in case.

From my testing I was not able to spot any difference besides that it fixes the issue with submitting the registration form when recaptcha is enabled.

I did test Brent's suggestion, but I don't see any difference there as well. If you embed the ticket selector on a post or event of a totally different site it still works ok. But with the invisible recaptcha enabled for the ticket selector, it doesn't seem to load the captcha on the ticket selector page nor with the changes in this PR nor without.

knazart:
@tn3rb should we make this change on ES and see if this does fix the recaptcha issue and at the same time give it time to be tested ? Or just merge it ?

garthkoyle:
Can we test this on MENW first?

knazart:
Yes, we can test MENW, but I'm not sure how to re-enable recaptcha there after Brent's changes.

I'll create an issue.

garthkoyle:
@knazart is this how we re-enable recaptcha? https://makeeventsnotwar.com/wp-admin/network/admin.php?page=ee_saas_solution&action=default

image

knazart:
Rebased just now.

tn3rb (approved):

garthkoyle:
What are we testing specifically and does this incorporate what's happening here? https://github.com/eventespresso/event-espresso-core/issues/3879

knazart:
I don't think we need this fix anymore. I tested the recent changes made by Tony and Hossein, and those seem to be working just as well, the https://github.com/eventespresso/event-espresso-core/pull/3890 in particular.

If those were already deployed we can test on MENW.

Issue #3839, opened 2022-05-16
CSS and JavaScript Extraction


hoseinrafiei:
**Describe the bug**
We should avoid css and JavaScript codes printed among HTML codes since we cannot escape them using WP escape functions.

**Instances**

1. /core/exceptions/ExceptionStackTraceDisplay.php #81

PR #3848, opened 2022-05-25
Kses - Part 12

hoseinrafiei:
Also fixes https://github.com/eventespresso/event-espresso-core/issues/3839 issue.

Removed some inline styles from the HTML codes to escape the HTML codes properly.

### DIFF
--- a/core/EE_Addon.core.php
+++ b/core/EE_Addon.core.php
@@ -203,7 +203,7 @@ abstract class EE_Addon extends EE_Configurable implements RequiresDependencyMap
      * Sets addon_name
      *
      * @param string $addon_name
-     * @return boolean
+     * @return bool
      */
     public function set_name($addon_name)
     {
@@ -313,6 +313,7 @@ abstract class EE_Addon extends EE_Configurable implements RequiresDependencyMap
      * If the site isn't in maintenance mode, should setup the addon's database
      *
      * @return void
+     * @throws EE_Error
      */
     public function new_install()
     {
@@ -332,6 +333,7 @@ abstract class EE_Addon extends EE_Configurable implements RequiresDependencyMap
      * it's good to just check that your data is still intact
      *
      * @return void
+     * @throws EE_Error
      */
     public function reactivation()
     {
@@ -367,13 +369,14 @@ abstract class EE_Addon extends EE_Configurable implements RequiresDependencyMap
      * initializing this addon's necessary initial data. This is called by default on new activations
      * and reactivations.
      *
-     * @param boolean $verify_schema whether to verify the database's schema for this addon, or just its data.
+     * @param bool $verify_schema whether to verify the database's schema for this addon, or just its data.
      *                               This is a resource-intensive job so we prefer to only do it when necessary
      * @return void
-     * @throws \EE_Error
+     * @throws EE_Error
      * @throws InvalidInterfaceException
      * @throws InvalidDataTypeException
      * @throws InvalidArgumentException
+     * @throws ReflectionException
      */
     public function initialize_db_if_no_migrations_required($verify_schema = true)
     {
@@ -482,6 +485,7 @@ abstract class EE_Addon extends EE_Configurable implements RequiresDependencyMap
      * they're ran
      *
      * @return void
+     * @throws EE_Error
      */
     public function upgrade()
     {
@@ -610,6 +614,7 @@ abstract class EE_Addon extends EE_Configurable implements RequiresDependencyMap
      * Should only be called once per request
      *
      * @return void
+     * @throws EE_Error
      */
     public function detect_activation_or_upgrade()
     {
@@ -659,7 +664,7 @@ abstract class EE_Addon extends EE_Configurable implements RequiresDependencyMap
      *
      * @param array  $version_history
      * @param string $current_version_to_add
-     * @return boolean success
+     * @return bool success
      */
     public function update_list_of_installed_versions($version_history = null, $current_version_to_add = null)
     {
@@ -810,50 +815,12 @@ abstract class EE_Addon extends EE_Configurable implements RequiresDependencyMap
                 $after_plugin_row .= '<tr id="' . sanitize_title($plugin_file) . '-ee-addon" class="' . $class . '">';
                 $after_plugin_row .= '<th class="check-column" scope="row"></th>';
                 $after_plugin_row .= '<td class="ee-addon-upsell-info-title-td plugin-title column-primary">';
-                $after_plugin_row .= '<style>
-.ee-button,
-.ee-button:active,
-.ee-button:visited {
-	box-sizing: border-box;
-	display: inline-block;
-	position: relative;
-	top: -1px;
-	padding:.5em 1em;
-	margin: 0;
-	background: #00B1CA -webkit-linear-gradient( #4EBFDE, #00B1CA ); /* For Safari 5.1 to 6.0 */
-	background: #00B1CA -o-linear-gradient( #4EBFDE, #00B1CA ); /* For Opera 11.1 to 12.0 */
-	background: #00B1CA -moz-linear-gradient( #4EBFDE, #00B1CA ); /* For Firefox 3.6 to 15 */
-	background: #00B1CA linear-gradient( #4EBFDE, #00B1CA ); /* Standard syntax */
-	border: 1px solid rgba(0,0,0,0.1) !important;
-	border-top: 1px solid rgba(255,255,255,0.5) !important;
-	border-bottom: 2px solid rgba(0,0,0,0.25) !important;
-	font-weight: normal;
-	cursor: pointer;
-	color: #fff !important;
-	text-decoration: none !important;
-	text-align: center;
-	line-height: 1em;
-/*	line-height: 1;*/
-	-moz-border-radius: 3px;
-	-webkit-border-radius: 3px;
-	border-radius: 3px;
-	-moz-box-shadow: none;
-	-webkit-box-shadow: none;
-	box-shadow: none;
-}
-.ee-button:hover {
-	color: #fff !important;
-	background: #4EBFDE;
-}
-.ee-button:active { top:0; }
-</style>';
-                $after_plugin_row .= '
-<p class="ee-addon-upsell-info-dv">
-	<a class="ee-button" href="' . $link_url . '">'
-                                     . $link_text
-                                     . ' &nbsp;<span class="dashicons dashicons-arrow-right-alt2" style="margin:0;"></span>'
-                                     . '</a>
-</p>';
+                $after_plugin_row .= '<p class="ee-addon-upsell-info-dv">
+	                <a class="ee-button" href="' . esc_url_raw($link_url) . '">'
+                    . $link_text
+                    . ' &nbsp;<span class="dashicons dashicons-arrow-right-alt2" style="margin:0;"></span>'
+                    . '</a>
+                </p>';
                 $after_plugin_row .= '</td>';
                 $after_plugin_row .= '<td class="ee-addon-upsell-info-desc-td column-description desc">';
                 $after_plugin_row .= $description;
--- a/core/admin/assets/admin-menu-styles.css
+++ b/core/admin/assets/admin-menu-styles.css
@@ -624,3 +624,41 @@ input.ee-toggle-round-flat:checked + label:after {
     top: -1px;
     margin-right: .25em;
 }
+
+/** ee buttons **/
+.ee-button,
+.ee-button:active,
+.ee-button:visited {
+    box-sizing: border-box;
+    display: inline-block;
+    position: relative;
+    top: -1px;
+    padding:.5em 1em;
+    margin: 0;
+    background: #00B1CA -webkit-linear-gradient( #4EBFDE, #00B1CA ); /* For Safari 5.1 to 6.0 */
+    background: #00B1CA -o-linear-gradient( #4EBFDE, #00B1CA ); /* For Opera 11.1 to 12.0 */
+    background: #00B1CA -moz-linear-gradient( #4EBFDE, #00B1CA ); /* For Firefox 3.6 to 15 */
+    background: #00B1CA linear-gradient( #4EBFDE, #00B1CA ); /* Standard syntax */
+    border: 1px solid rgba(0,0,0,0.1) !important;
+    border-top: 1px solid rgba(255,255,255,0.5) !important;
+    border-bottom: 2px solid rgba(0,0,0,0.25) !important;
+    font-weight: normal;
+    cursor: pointer;
+    color: #fff !important;
+    text-decoration: none !important;
+    text-align: center;
+    line-height: 1em;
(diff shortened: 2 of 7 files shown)


core/EE_Addon.core.php
```diff
@@ -810,50 +815,12 @@ public function after_plugin_row($plugin_file, $plugin_data, $status)
                 $after_plugin_row .= '<tr id="' . sanitize_title($plugin_file) . '-ee-addon" class="' . $class . '">';
                 $after_plugin_row .= '<th class="check-column" scope="row"></th>';
                 $after_plugin_row .= '<td class="ee-addon-upsell-info-title-td plugin-title column-primary">';
-                $after_plugin_row .= '<style>
-.ee-button,
-.ee-button:active,
-.ee-button:visited {
-	box-sizing: border-box;
-	display: inline-block;
```
hoseinrafiei, line 818: Moved these styles to `core/admin/assets/admin-menu-styles.css` file. I don't know if it is the right file or not but based on my checks it was loading on the plugins page.

core/exceptions/ExceptionStackTraceDisplay.php
```diff
@@ -68,73 +68,75 @@ protected function displayException(Exception $exception)
...
+            ' . $error_message . '
+        </div>';
         $scripts = $this->printScripts(true);
         if (defined('DOING_AJAX')) {
-            echo wp_json_encode(array('error' => $styles . $output . $scripts));
+            echo wp_json_encode(array('error' => $output . $scripts));
             exit();
         }
-        echo $styles, $output, $scripts;
+        echo wp_kses($output . $scripts, AllowedTags::getWithScriptAndStyleTags());
```
hoseinrafiei, line 79: Removed the styles from here and added them as a new file located in `core/templates/global_assets/css/ee-exception-stack-display.css` path. Added the link tag to the HTML codes. Also modified the `AllowedTags` class to support link tag.

Since this class can be called from both of frontend and backend, I was going to add the CSS codes to a file that is loading in both of them but didn't find anything so I added the new file as mentioned earlier. I'm not sure if adding the file, path of it and naming is correct so if there is anything just let know to correct it.

core/exceptions/ExceptionStackTraceDisplay.php
```diff
@@ -416,18 +361,18 @@ private function printScripts($force_print = false)
...
-        $jquery = esc_url_raw(includes_url() . 'js/jquery/jquery.js');
-        $core = esc_url_raw(EE_GLOBAL_ASSETS_URL . 'scripts/espresso_core.js?ver=' . espresso_version());
-        $ee_error = esc_url_raw(EE_GLOBAL_ASSETS_URL . 'scripts/EE_Error.js?ver=' . espresso_version());
+        $jquery = includes_url() . 'js/jquery/jquery.js';
+        $core = EE_GLOBAL_ASSETS_URL . 'scripts/espresso_core.js?ver=' . espresso_version();
+        $ee_error = EE_GLOBAL_ASSETS_URL . 'scripts/EE_Error.js?ver=' . espresso_version();
+        $style = EE_GLOBAL_ASSETS_URL . 'css/ee-exception-stack-display.css';
         return '
-<script>
-/* <![CDATA[ */
```
hoseinrafiei, line 424: Removed the CDATA because `kses` considered it as an unknown HTML tag and removed the tag and its contents.

core/admin/assets/admin-menu-styles.css
```diff
@@ -624,3 +624,41 @@ input.ee-toggle-round-flat:checked + label:after {
     top: -1px;
     margin-right: .25em;
 }
+
+/** ee buttons **/
+.ee-button,
+.ee-button:active,
+.ee-button:visited {
```
tn3rb, line 631: to avoid any conflicts with other styles, can you plz increase the specificity of these particular rules by prepending them all with `.ee-addon-upsell-info-dv` 

ie:

```css
.ee-addon-upsell-info-dv .ee-button,
.ee-addon-upsell-info-dv .ee-button:active,
.ee-addon-upsell-info-dv .ee-button:visited {
```

core/admin/assets/admin-menu-styles.css
```diff
@@ -624,3 +624,41 @@ input.ee-toggle-round-flat:checked + label:after {
...
+    -moz-box-shadow: none;
+    -webkit-box-shadow: none;
+    box-shadow: none;
+}
+.ee-button:hover {
+    color: #fff !important;
+    background: #4EBFDE;
+}
+.ee-button:active {
+    top:0;
```
tn3rb, line 663: same here plz

core/bootstrap_espresso.php
```diff
@@ -1,5 +1,13 @@
 <?php
 
+use EventEspresso\core\exceptions\InvalidClassException;
+use EventEspresso\core\exceptions\InvalidDataTypeException;
+use EventEspresso\core\exceptions\InvalidEntityException;
+use EventEspresso\core\exceptions\InvalidIdentifierException;
+use EventEspresso\core\exceptions\InvalidInterfaceException;
+use EventEspresso\core\services\container\exceptions\ServiceExistsException;
+use EventEspresso\core\services\container\exceptions\ServiceNotFoundException;
```
tn3rb, line 9: 👍🏻 

this should be fine... 
we couldn't do this before here because there were still a few users on servers running PHP 5.2 😕 
and this originally was not behind a php version check

core/exceptions/ExceptionStackTraceDisplay.php
```diff
@@ -68,73 +68,75 @@ protected function displayException(Exception $exception)
...
             : $this->genericError($msg, $code);
         // start gathering output
-        $output = '
-<div id="ee-error-message" class="error">
-    ' . $error_message . '
-</div>';
-        $styles = $this->exceptionStyles();
+        $output = '<div id="ee-error-message" class="error">
+            ' . $error_message . '
+        </div>';
```
tn3rb, line 73: can you keep these line breaks as part of the formatting plz?

the only reason they are there and why i would like to keep them is because it makes it easier to find this content in the DOM if you are looking for it. The way you have it, `<div id="ee-error-message" class="error">` will get tacked onto the end of whatever other line of HTML is there like this:

```html
<div id="post-body-content">
	<div class="ee-admin-container">
		<div class="padding">
			<form action="https://makeeventsnotwar.com/wp-admin/admin.php?page=espresso_events&amp;route=default" id="Extend_Events_Admin_List_Table-table-frm" method="get"><div id="ee-error-message" class="error">
            Unable to find markup
        </div>
    <p class="search-box">
	<label class="screen-reader-text" for="toplevel_page_espresso_events-search-input">Search Events:</label>
	<input type="search" id="toplevel_page_espresso_events-search-input" name="s" value="">
		<input type="submit" id="search-submit" class="button" value="Search Events"></p>
		        <ul class="subsubsub">
```

which makes it hard to find if you are looking for `id="ee-error-message"`

Another solution would be to remove that markup from this file and place it in a template file and then load that as needed.

core/exceptions/ExceptionStackTraceDisplay.php
tn3rb: the `core/templates/global_assets/css/ee-exception-stack-display.css ` file name and location is PERFECT 💯

core/exceptions/ExceptionStackTraceDisplay.php
```diff
@@ -68,73 +68,75 @@ protected function displayException(Exception $exception)
...
         $scripts = $this->printScripts(true);
         if (defined('DOING_AJAX')) {
-            echo wp_json_encode(array('error' => $styles . $output . $scripts));
+            echo wp_json_encode(array('error' => $output . $scripts));
```
tn3rb, line 76: this might be an issue.

you'll need to try and chase down a spot in the code where exceptions are being caught and returned via AJAX and see what things look like.

I can't remember where that might be happening (EE core alone is over 620k lines of code now 😬) but I suspect it is during the registration process since that relies heavily on AJAX. You should also checkout and activate [MER (Multi Event Registration)](https://github.com/eventespresso/ee4-mer) as it also relies heavily on AJAX.

You'll then need to find a spot in the code where `ExceptionStackTraceDisplay` is being used and try to simulate an error state that will trigger this code during an AJAX response. Fun times 😆 

I'm totally open to suggestions as to how to solve this problem, but first you will need to understand the full scope of it.

core/exceptions/ExceptionStackTraceDisplay.php
```diff
@@ -68,73 +68,75 @@ protected function displayException(Exception $exception)
...
 
     private function genericError($msg, $code)
     {
-        return '
-    <p>
-        <span class="ee-error-user-msg-spn">' . trim($msg) . '</span> &nbsp; <sup>' . $code . '</sup>
-    </p>';
+        return '<p>
+            <span class="ee-error-user-msg-spn">' . trim($msg) . '</span> &nbsp; <sup>' . $code . '</sup>
+        </p>';
```
tn3rb, line 87: same thing here, plz keep newlines in any formatted markup like this...

OR

extract to a separate template file (but maintain formatting there as well!)

core/exceptions/ExceptionStackTraceDisplay.php
```diff
@@ -68,73 +68,75 @@ protected function displayException(Exception $exception)
...
+                    esc_html__('%1$s( line no: %2$s )%3$s', 'event_espresso'),
+                    ' &nbsp; <span class="small-text lt-grey-text">',
+                    $exception->getLine(),
+                    '</span>'
+                )
+                . '
+            </p>
+            <div id="ee-error-trace-1'
+                   . $time
+                   . '-dv" class="ee-error-trace-dv hidden">
```
tn3rb, line 134: + 💯  for wanting to replace `style="display: none;"` with a `hidden` class 

**However**, I can almost guarantee that JS is being used to show and hide that container by adding and removing `style="display: none;"`, most likely via `jQuery.show()` and  `jQuery.hide()` or one of its variants (`.toggleSlide()` etc). 
So clicking on whatever listener is currently set up will most likely attempt to display that container by setting `style="display: block;"`.

That will work fine if the `hidden` class is merely:

```css
.hidden {
	display: none;
}
```

however...  `hidden` is a **VERY** common class name used in plugins and templates so there is an extremely high chance of conflicting with one of those. IF the conflicting `hidden` class is doing something other than the above, then things will not work.

possible solutions:

- namespace classname, ex: `ee-hidden` this is something that should be practiced 100% of the time, but even this could still conflict with another usage of `ee-hidden` somewhere else in the plugin (so search first)

- make classname more specific ex: `ee-error-trace-dv--hidden` BEM is great for this (search "css BEM")

- depending on how the JS works, it may also be necessary to modify that logic to toggle classnames instead of using other jQuery visibility handlers like `.show()`,  `.hide()`, `.toggleSlide()` etc. Using css classnames can also give more control over display.

- some other possible solution that you can think of?

I'm always open to new ideas for solving problems, but just be aware that a seemingly simple change can have a domino affect and cause all kinds of unanticipated problems.

core/EE_Addon.core.php
tn3rb: ya that will be fine for now... I think in the dev branch there is a better place, but it's not a  big deal at all

tn3rb (changes requested):
This is all looking good, but there are a few issues that need to be addressed.

My concern is that there will be some unintended side effects from a few of these seemingly insignificant changes, so you'll need to do a bunch of detective work to ensure that things don't break somewhere.

core/exceptions/ExceptionStackTraceDisplay.php
hoseinrafiei: I simulated the AJAX error on normal registration and also MER using following code and the error with scripts and link tag (style) returns as response of AJAX request but it didn't show to the end user.

```
add_action('init', function()
{
    new \EventEspresso\core\exceptions\ExceptionStackTraceDisplay(new Exception('Simple Error'));
});
```

Please check this image.
Screen Shot 2022-05-26 at 20 56 57

core/bootstrap_espresso.php
tn3rb: i call those days the "dark ages"

core/exceptions/ExceptionStackTraceDisplay.php
tn3rb: 🙇🏻‍♂️

core/exceptions/ExceptionStackTraceDisplay.php
tn3rb: ya I've been trying to implement it whenever I can, it's a little bit verbose, but works REALLY well with Sass css

core/exceptions/ExceptionStackTraceDisplay.php
```diff
@@ -334,64 +336,7 @@ protected function generate_error_code($file = '', $func = '', $line = '')
      */
     private function exceptionStyles()
     {
-        return '
```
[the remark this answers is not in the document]
tn3rb: 🙇🏻‍♂️

tn3rb (approved):
cool, thnx for making those changes

core/exceptions/ExceptionStackTraceDisplay.php
tn3rb: can you plz switch to master branch, try your snippet there and see what happens?

if it doesn't display the error, then you will need to back trace the code, or find some other way to determine where those AJAX errors were being used, and find a way to reliably reproduce them. THEN switch back to this branch and test out your changes.

hoseinrafiei: I switched to **master** branch and added the previous code snippet in espresso.php file (main plugin file) and check the MER AJAX but it didn't show to user in frontend.

Then I added following codes in **EED_Multi_Event_Registration::remove_ticket** and some other functions. The errors was reflected in the AJAX response but they didn't show to user in frontend.

```
new \EventEspresso\core\exceptions\ExceptionStackTraceDisplay(new Exception('Simple Error'));
```

```
EE_Error::add_error(
    esc_html__( 'An invalid event object was received.', 'event_espresso' ),
    __FILE__,
    __FUNCTION__,
    __LINE__
);
```
Screen Shot 2022-05-30 at 18 08 29

Also added the above codes into **EED_Single_Page_Checkout::process_ajax_request** function without MER and same happened.

Screen Shot 2022-05-30 at 18 23 15
Screen Shot 2022-05-30 at 18 24 18

If the error display was working fine before then perhaps a recent change (most probably on JS codes) destroyed the functionality of error display. I can check it if I'm in the right path of investigation.

hoseinrafiei: Finally found fixed it. Please check the latest commit.

Screen Shot 2022-05-30 at 20 28 20

tn3rb: AWESOME! 
good job working through that.
sometimes it can take a long time to wade through all of the mountains of code to find the source of the issue, so thanks for sticking with it 💯

tn3rb (approved):
good stuff 👍🏻

PR #3850, opened 2022-05-27
Fix Missing "CNT_ISO" Parameter in AJAX Request

tn3rb:
from chat:

>Does anyone know why this happens with the Countries admin?
> ```An error has occurred:
> No Country ISO code or an invalid Country ISO code was received.

This PR fixes the above error


garthkoyle:
Reported here:

https://secure.helpscout.net/conversation/1844100581/0/?folderId=284418

I was able to reproduce this issue on ES and Master. I tested this on garth.eventespresso.com and it works now.

PR #3522, opened 2021-09-24
Fix PHP 8.0 Issues

tn3rb:
plz see: #3515 && #3521 & #3851

This PR:

- fixes some more named parameter mismatches caused by using `call_user_func_array()`

- fixes a bunch of required parameters positioned after optional parameters, sometimes by also making them optional (if not necessary to function or appropriate checks in place) or by making the optional parameters before them required parameters (if appropriate checks already in place)

- ensures that some classes are setup properly prior to loading, now that optional parameters are required

- changes how some dependency resolution occurs in `EE_Registry` due to changes in how PHP 8's Reflection classes report information about code.

- closes #3210  closes #3521

PR #3851, opened 2022-05-30
Update Unit Tests

tn3rb:
WordPress 5.9 introduced a significant number of changes with regards to unit testing after updating to allow for compatibility with PHP 8  and PHPUnit 9+

This PR updates our unit tests and codebase to work with the above updates.


hoseinrafiei (approved):
Looks great. I just checked the files under `tests` since the rest seems to be a linter fix.

tn3rb:
changes in this branch got ported to #3522 since the two issues go hand in hand

PR #3866, opened 2022-06-03
More EDTR UI Fixes & Updates

tn3rb:
These are some other UI updates for the EDTR and admin refresh that have been sitting in a branch on my local system that I wanted to move so that I don't have to keep rebasing it.

Most changes are fixes for responsive layouts, but the biggest visible change is a redesign to the editor tabs, ex:

Screenshot 2022-06-03 105305

Screenshot 2022-06-03 105347

works on older add-ons too:

Screenshot 2022-06-03 105416

although they have not been updated to show icons


hoseinrafiei (dismissed):
Great

hoseinrafiei (dismissed):
💯👍

PR #3868, opened 2022-06-04
Datetime Registrations Button Issue

hoseinrafiei:
Fixed an issue on not displaying datetime registrations button on Registrations -> Overview tab.
Related PR: https://github.com/eventespresso/event-espresso-core/pull/3808

PR #3875, opened 2022-06-07
Fix Registrations CSV Report Filters

tn3rb:
In https://github.com/eventespresso/event-espresso-core/pull/3868 a bunch of work was done to reconfigure how the Registration CSV Reports worked, and amidst the flurry of changes, the ability to apply the reg list filters was lost. 
I missed that in my code reviews and it was also overlooked during testing.

This PR:

- extracts the logic for setting the CSV filter params for the csv reg report into a new class
- removes ALL of the various button labels for different CSV report buttons in favour of one single button
- adds an info notice after the reg CSV button so that people are aware that the functionality for all buttons has been combined
- reinstates the application of the filter query params to the CSV report if present

Screenshot 2022-06-07 134355

### DIFF
--- a/admin_pages/registrations/EE_Registrations_List_Table.class.php
+++ b/admin_pages/registrations/EE_Registrations_List_Table.class.php
@@ -1,5 +1,6 @@
 <?php
 
+use EventEspresso\core\domain\services\admin\registrations\list_table\csv_reports\RegistrationsCsvReportParams;
 use EventEspresso\core\exceptions\EntityNotFoundException;
 use EventEspresso\core\exceptions\InvalidDataTypeException;
 use EventEspresso\core\exceptions\InvalidInterfaceException;
@@ -87,8 +88,11 @@ class EE_Registrations_List_Table extends EE_Admin_List_Table
         $ID_column_name      .= ' : <span class="show-on-mobile-view-only" style="float:none">';
         $ID_column_name      .= esc_html__('Registrant Name', 'event_espresso');
         $ID_column_name      .= '</span> ';
-        $req_data            = $this->_admin_page->get_request_data();
-        if (isset($req_data['event_id'])) {
+
+        $EVT_ID = $this->_req_data['event_id'] ?? 0;
+        $DTT_ID = $this->_req_data['DTT_ID'] ?? 0;
+
+        if ($EVT_ID) {
             $this->_columns        = [
                 'cb'               => '<input type="checkbox" />', // Render a checkbox instead of text
                 '_REG_ID'          => $ID_column_name,
@@ -101,19 +105,6 @@ class EE_Registrations_List_Table extends EE_Admin_List_Table
                 'TXN_paid'         => esc_html__('Paid', 'event_espresso'),
                 'actions'          => esc_html__('Actions', 'event_espresso'),
             ];
-            $route_details = [
-                'route'         => 'registrations_report',
-                'extra_request' => [
-                    'EVT_ID'     => $this->_req_data['event_id'],
-                    'return_url' => $return_url,
-                ]
-            ];
-            if (isset($req_data['datetime_id']) && $req_data['datetime_id']) {
-                $route_details['extra_request']['DTT_ID'] = $req_data['datetime_id'];
-                $this->_bottom_buttons['report_datetime'] = $route_details;
-            } else {
-                $this->_bottom_buttons['report'] = $route_details;
-            }
         } else {
             $this->_columns        = [
                 'cb'               => '<input type="checkbox" />', // Render a checkbox instead of text
@@ -126,15 +117,13 @@ class EE_Registrations_List_Table extends EE_Admin_List_Table
                 '_REG_paid'        => esc_html__('Paid', 'event_espresso'),
                 'actions'          => esc_html__('Actions', 'event_espresso'),
             ];
-            $this->_bottom_buttons = [
-                'report_all' => [
-                    'route'         => 'registrations_report',
-                    'extra_request' => [
-                        'return_url' => $return_url,
-                    ],
-                ],
-            ];
         }
+
+        $csv_report = RegistrationsCsvReportParams::getRequestParams($return_url, $this->_req_data, $EVT_ID, $DTT_ID);
+        if (! empty($csv_report)) {
+            $this->_bottom_buttons['csv_reg_report'] = $csv_report;
+        }
+
         $this->_primary_column   = '_REG_ID';
         $this->_sortable_columns = [
             '_REG_date'     => ['_REG_date' => true],   // true means its already sorted
--- a/admin_pages/registrations/EE_Registrations_List_Table.class.php
+++ b/admin_pages/registrations/EE_Registrations_List_Table.class.php
@@ -89,8 +89,8 @@ class EE_Registrations_List_Table extends EE_Admin_List_Table
         $ID_column_name      .= esc_html__('Registrant Name', 'event_espresso');
         $ID_column_name      .= '</span> ';
 
-        $EVT_ID = $this->_req_data['event_id'] ?? 0;
-        $DTT_ID = $this->_req_data['DTT_ID'] ?? 0;
+        $EVT_ID = isset($this->_req_data['event_id']) ? $this->_req_data['event_id'] : 0;
+        $DTT_ID = isset($this->_req_data['DTT_ID']) ? $this->_req_data['DTT_ID'] : 0;
 
         if ($EVT_ID) {
             $this->_columns        = [
@@ -219,6 +219,8 @@ class EE_Registrations_List_Table extends EE_Admin_List_Table
      *    _get_table_filters
      *
      * @return array
+     * @throws EE_Error
+     * @throws ReflectionException
      */
     protected function _get_table_filters()
     {
--- a/admin_pages/registrations/Registrations_Admin_Page.core.php
+++ b/admin_pages/registrations/Registrations_Admin_Page.core.php
@@ -8,6 +8,7 @@ use EventEspresso\core\exceptions\InvalidFormSubmissionException;
 use EventEspresso\core\exceptions\InvalidInterfaceException;
 use EventEspresso\core\domain\services\attendee\forms\AttendeeContactDetailsMetaboxFormHandler;
 use EventEspresso\core\services\request\CurrentPage;
+use EventEspresso\core\services\request\DataType;
 use EventEspresso\core\services\request\sanitizers\AllowedTags;
 
 /**
@@ -256,7 +257,7 @@ class Registrations_Admin_Page extends EE_Admin_Page_CPT
         $ATT_ID             = $this->request->getRequestParam('post', $ATT_ID, 'int');
         $this->_page_routes = [
             'default'                             => [
-                'func'       => '_registrations_overview_list_table',
+                'func'       => [$this, '_registrations_overview_list_table'],
                 'capability' => 'ee_read_registrations',
             ],
             'view_registration'                   => [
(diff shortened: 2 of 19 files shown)


joaovictorangeline (changes requested):
@tn3rb I have tested the CSV Report and found issues.

**1st scenario:**
Event Espresso > Registrations > Registrations CSV Report is coming with the Event ID and empty fields
CleanShot 2022-06-07 at 11 03 27@2x
CleanShot 2022-06-07 at 11 03 48@2x

**2sd scenario**
Event Espresso > Registrations > Select one attendee 
CleanShot 2022-06-07 at 11 04 36@2x
It's giving the same CSV report as the first test.

Both scenarios are reproducible only through Event Espresso > Registrations (Not from Events > EVENT X > Registrations).

**3rd scenario**
Event Espresso > Check-in tab > Select ONE user > CSV Report
CleanShot 2022-06-07 at 11 06 53@2x
It should include just one user that I filtered (Joao and not both).
The same behavior when we export the report from Events > EVENT X > Registrations.

Let me know if a loom is necessary.

tn3rb:
### Issue 1
I'll have to look into that. I thought my tests were returning attendee data, but that may have broken and I didn't notice

### Issue 2
The checkboxes are for the bulk actions and do not affect the CSV report

### Issue 3
not sure what you mean by "Select ONE user" but if that means using the check boxes in the first column, then that's a non-issue as those are for the bulk actions and not the CSV report

joaovictorangeline:
> ### Issue 1
> I'll have to look into that. I thought my tests were returning attendee data, but that may have broken and I didn't notice
> 
> ### Issue 2
> The checkboxes are for the bulk actions and do not affect the CSV report
> 
> ### Issue 3
> not sure what you mean by "Select ONE user" but if that means using the check boxes in the first column, then that's a non-issue as those are for the bulk actions and not the CSV report

I thought we were able to select each user and create a report including the selected ones. And Regarding the filters, it's not working, I filtered the Event Date by Today
CleanShot 2022-06-07 at 11 19 31@2x
And got all registrations in the report
CleanShot 2022-06-07 at 11 19 55@2x

About the "Loom", I was referring to a screen record showing the problems using [Loom](https://loom.com/)

tn3rb:
unless I'm mistaken, these are the filters:

Screenshot 2022-06-07 202447

i can easily see how selecting those other views would make it look like they would affect the filters, but afaik, they don't.

IF you get a chance, can you plz checkout an earlier version of master from before all of these CSV changes were made (`4.10.30.rc.022` or less) and confirm how things worked.

core/domain/services/admin/registrations/list_table/csv_reports/RegistrationsCsvReportParams.php
```diff
@@ -0,0 +1,82 @@
...
+            $route_details['extra_request']['EVT_ID'] = $EVT_ID;
+        }
+        if ($DTT_ID) {
+            $route_details['extra_request']['DTT_ID'] = $DTT_ID;
+        }
+        if (
+            isset($request_params['month_range'])
+            || isset($request_params['EVT_CAT'])
+            || isset($request_params['_reg_status'])
+        ) {
```
Pebblo, line 58: The filters on the registration report are filtered so we can't use a check like this, right?

Custom filter example:
https://gist.github.com/Pebblo/22995de5472c46a361ecc206bc8e9f9b

core/libraries/batch/JobHandlers/RegistrationsReport.php
```diff
@@ -347,7 +363,7 @@ public function get_csv_data_for($event_id, $offset, $limit, $question_labels, $
                     ARRAY_A,
                     'Payment_Method.PMD_admin_name as name, Payment.PAY_txn_id_chq_nmbr as gateway_txn_id, Payment.PAY_timestamp as payment_time'
                 );
-                list($payment_methods, $gateway_txn_ids_etc, $payment_times) = PaymentsInfoCSV::extractPaymentInfo($payments_info);
+                [$payment_methods, $gateway_txn_ids_etc, $payment_times] = PaymentsInfoCSV::extractPaymentInfo($payments_info);
```
Pebblo, line 366: Is this correct? 

Now we have an array of arrays but we need each individual variable to be used later.

For example, Line 368 below uses the $payment_times var, but that's still set to an empty array from a little above this line.

Pebblo (changes requested):
I'm reproducing the issue Joao mentioned and noticed a couple of issues during the review

Pebblo:
Hey @joaovictorangeline 

I just want to check I follow what you are trying to do here as you are mentioned 'users', I'm assuming you mean registrations?

So you're selecting specific registrations using the checkboxes and clicking to export, but getting all of the registrations?

If so, Brent is correct, the checkboxes aren't a way to export individual registrations like that so the above is expected.

----

With regards to this:

> And Regarding the filters, it's not working, I filtered the Event Date by Today

Those links you are clicking at the top are called 'views', they are a WP List Table feature which we use for various outputs. They switch the 'view' for a specific output.

'Filters' are the sections Brent highlighted, they allow you to 'filter' registrations based on selections made in the dropdowns.

This breaks them down a little so you know what you are testing: https://monosnap.com/file/7cVOjqZqgQyK2gWKO1z4gxM5uyS8tV

So originally the 'Filtered CSV Registrations' button allowed you to filter say registrations with a status of 'Abandoned' and then click that button to apply the filters to the export. You could combine those to view say Abandoned registration in January. The button has been merged but the functionality should remain the same here, so setting filters to the registration list should export the same registrations you see when filtered.

core/libraries/batch/JobHandlers/RegistrationsReport.php
```diff
@@ -57,43 +57,59 @@ class RegistrationsReport extends JobHandlerFile
...
+                ],
+                'order_by'   => ['Transaction.TXN_ID' => 'asc', 'REG_count' => 'asc'],
+                'force_join' => ['Transaction', 'Ticket', 'Attendee'],
+                'caps'       => EEM_Base::caps_read_admin,
+            ];
+
         if ($event_id) {
             $query_params[0]['EVT_ID'] = $event_id;
         } else {
             $query_params['force_join'][] = 'Event';
```
Pebblo, line 101: Previously, this was only set on the 'default' query params, now it's set on both the default and filtered query params.

This means line 104 no longer jumps in and sets the 'correct' force_join for the filtered registrations, this is why the attendee fields no longer show. The query no longer returns attendee data from the DB.

Pebblo:
@tn3rb I added some comments in the review, normally I'd make some changes and pass them over for you to pick apart/make better :) but I think there are a few changes needed here so I've left alone and just added what I found to help narrow it down some.

core/domain/services/admin/registrations/list_table/csv_reports/RegistrationsCsvReportParams.php
tn3rb: ahh... missed that filter since it was in the list table parent class...
hmmm... ya that's going to be tricky, because the filter params were previously being toggled by a specific request parameter added when you clicked the "Filtered CSV report" button.
going to need to figure out a way to detect when filters have been applied when we don't know what they actually are 🤔

core/libraries/batch/JobHandlers/RegistrationsReport.php
tn3rb: those are functionally equivalent.
The array on the assignment side of things is called "destructuring" and assigns each array element to a variable:

```php
 [ $a, $b, $c ] = [ 'A',  'B',  'C' ];
```

which works with the above code because `PaymentsInfoCSV::extractPaymentInfo()` returns an array.
 
My IDE does that automatically and I did not intend to make that change.

Pebblo: >those are functionally equivalent.

Oh... cool! I did not know that. Thanks :)

admin_pages/registrations/EE_Registrations_List_Table.class.php
```diff
@@ -126,15 +117,13 @@ protected function _set_properties()
...
-                    ],
-                ],
-            ];
         }
+
+        $csv_report = RegistrationsCsvReportParams::getRequestParams($return_url, $this->_req_data, $EVT_ID, $DTT_ID);
+        if (! empty($csv_report)) {
+            $this->_bottom_buttons['csv_reg_report'] = $csv_report;
+        }
+
```
hoseinrafiei, line 126: 👏

admin_pages/registrations/Registrations_Admin_Page.core.php
```diff
@@ -226,10 +227,7 @@ protected function _define_page_props()
                 'add-registrant'      => esc_html__('Add New Registration', 'event_espresso'),
                 'add-attendee'        => esc_html__('Add Contact', 'event_espresso'),
                 'edit'                => esc_html__('Edit Contact', 'event_espresso'),
-                'report'              => esc_html__('Event Registrations CSV Report', 'event_espresso'),
-                'report_datetime'     => esc_html__('Datetime Registrations CSV Report', 'event_espresso'),
-                'report_all'          => esc_html__('All Registrations CSV Report', 'event_espresso'),
-                'report_filtered'     => esc_html__('Filtered CSV Report', 'event_espresso'),
+                'csv_reg_report'      => esc_html__('Registrations CSV Report', 'event_espresso'),
```
hoseinrafiei, line 230: Perfect

hoseinrafiei (approved):
Looks great.

Pebblo:
I'm not sure if this is actually ready for testing but saw the above review so ran through it anyway.

Setting the Reg Status filter to Approved and clicking to export, exports all registrations again so filters don't seem to be applied currently.

tn3rb:
> Setting the Reg Status filter to Approved and clicking to export, exports all registrations again so filters don't seem to be applied currently.

hmmm... not getting that at all locally...
this branch has been rebased so you need to delete your local branch and then re-checkout  from origin. 
Can you plz do that and then test it again?
If you still get the same results, can you then make a screen cast  and post it plz?

joaovictorangeline:
@tn3rb A couple of details to be fixed.

1. The "Registration CSV Report" needs a margin-top and the Legend is not aligned (maybe too big):
CleanShot 2022-06-11 at 5 29 04@2x

2. I need to filter twice when I want to get a different filter. Watch my screen record:
https://www.loom.com/share/81a70cecc7314d94b2a8ea4cb1a775fc

tn3rb:
@joaovictorangeline 

re:
> The "Registration CSV Report" needs a margin-top and the Legend is not aligned 

what OS, browser, screen size is your system?

>  I need to filter twice when I want to get a different filter

well that's bonkers... can confirm locally though!  thnx for the screenie 👍🏻

tn3rb:
plz ignore JS test failures, it's just dependency issues for the legacy JS config

joaovictorangeline:
@tn3rb It's working fine now! Great job 💯

Pebblo:
Thanks @tn3rb 

The `_wp_http_referrer` cleanup is awesome, that fixes an issue people run into when they run multiple searches for registrations and the referrer gets so long it triggers a [414 URI Too Long](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/414) error. Awesome stuff.

The filtering is now working great!

However, the new styles added in this branch breaks the layout as you reduce screen size:

See top of list table: https://monosnap.com/file/xzIvsLutoX77niFYSM4y23Orex12Y8

Bottom: https://monosnap.com/file/DFcmCVeaFM8Z2GYEgtStWlauCd1Ah4

Pebblo:
Tested the above 2 additional commits and they fix the styling issues found.

If no additional changes needed I think this can be merged. 

Nice work @tn3rb

PR #3877, opened 2022-06-08
Add Cap Checks to Routing

tn3rb:
This PR:

- adds `RequiresCapCheckInterface` to the base `Route` class
- injects the `CapChecker` class into `RouteHandler`
- retrieves each route's `CapCheck` object and processes it before handling route
- adds `$suppress_exceptions` parameter to `CapChecker` and toggles to `true` during routing
- adds default caps to public and admin routes
- updates typing, dependencies, and phpdocs

### DIFF
--- a/core/domain/entities/routing/handlers/admin/AdminRoute.php
+++ b/core/domain/entities/routing/handlers/admin/AdminRoute.php
@@ -4,6 +4,8 @@ namespace EventEspresso\core\domain\entities\routing\handlers\admin;
 
 use EE_Admin_Config;
 use EE_Dependency_Map;
+use EventEspresso\core\domain\services\capabilities\CapCheckInterface;
+use EventEspresso\core\domain\services\capabilities\CapCheck;
 use EventEspresso\core\services\routing\Route;
 use EventEspresso\core\domain\entities\routing\specifications\RouteMatchSpecificationInterface;
 use EventEspresso\core\services\json\JsonDataNode;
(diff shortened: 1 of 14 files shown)


core/domain/services/capabilities/CapabilitiesChecker.php
```diff
@@ -48,20 +48,25 @@ protected function capabilities()
...
+    public function processCapCheck($cap_check, bool $suppress_exceptions = false): bool
     {
         if (is_array($cap_check)) {
+            $passed = true;
             foreach ($cap_check as $check) {
-                $this->processCapCheck($check);
+                $passed = $this->processCapCheck($check) ? $passed : false;
             }
-            return true;
+            return $passed;
```
tn3rb, line 63: we initially set `$passed` to `true`, 
then if any cap check fails, it will get toggled to `false`

core/domain/entities/routing/handlers/admin/AdminRoute.php
```diff
@@ -53,6 +65,21 @@ public function __construct(
     }
 
 
+    public function getCapCheck()
+    {
+        return new CapCheck('edit_posts', 'access Event Espresso admin route');
```
tn3rb, line 70: default cap check for admin routes

core/services/routing/Route.php
```diff
@@ -157,6 +159,12 @@ protected function dataNodeClass(): string
     }
 
 
+    public function getCapCheck()
+    {
+        return new PublicCapabilities('', 'access Event Espresso route');
```
tn3rb, line 164: default cap check for public routes (which is no cap)

hoseinrafiei (dismissed):
👍

core/domain/entities/routing/handlers/admin/AdminRoute.php
tn3rb: right, i basically wanted a cap that would be typical of any role higher than subscriber.
the idea is that this cap check would be a fallback if nothing else is set for a route, so it's intentionally trying to allow nearly everyone through.

PR #3880, opened 2022-06-13
Improved "Free Admission" Label and Used "Admission" Instead

hoseinrafiei:
Fixes: https://github.com/eventespresso/event-espresso-core/issues/3883

This will fix the problem for new installations.


hoseinrafiei:
@tn3rb

Based on my understanding this text is not a hardcoded text in the codes nor in **barista** neither is **event-espresso-core** plugin. It comes from the table `esp_price` of database.

There are some instances of hardcoded `Free Admission` text in the following file of **event-espresso-core** plugin but it seems we never used this template.

    admin_pages/events/templates/new_price_layout.template.php
    
Therefore I found that we can change the text for new installations by modification on line 542 of following file.

    core/data_migration_scripts/EE_DMS_Core_4_5_0.dms.php
    
Please check it and let me know if there is anything else on this that can be done.

PR #3887, opened 2022-06-15
Fix CSV custom question output when questions have been removed from event(s)

Pebblo:
@tn3rb Apologies for the wall of text.... I'm explaining my logic for adding this and my understanding of the problem, feel free to point out something I'm missing/have completely wrong.

I know you've gone through this previously which is why I've added all the details to explain my thinking here.

----

When a registration CSV report is generated EE first pulls in the questions assigned to the events and eventually generates an array of question labels, see:

https://github.com/eventespresso/event-espresso-core/blob/master/core/libraries/batch/JobHandlers/RegistrationsReport.php#L162-L187

(It's important to note that what EE is doing above is pulling in all of the questions based on the groups they are related to being linked to events, that could be all events or an individual event)

That array is then passed around and used within the report to set which question labels are included in the CSV.
However, to pull the **values**  (the answers) for those questions for each registration, what we do right now is pull **all** answers related to the current registration we are processing and work with those answers to find the related question, then from that use that question label to include it in the CSV (and append... hint to the issue there) and include it in the CSV, you can see this here:

https://github.com/eventespresso/event-espresso-core/blob/master/core/domain/services/admin/registrations/list_table/csv_reports/AnswersCSV.php#L38-L41

In English, that code says 'Gimme all of the answers related to REG_ID X, oh and join the related question so we can use it'

So at that point, `$question_lables` has been used to add all of the custom questions labels we are looking for into `$data` as array keys.

`$answers` is an array of answers (and each answer's related question) relating to the current REG_ID and their related questions.

What we do now is loop over `$answers`, work out the question label for each individual answer, and then add the answer to the `$data` array.... the problem is we are **always** assuming that all answers related to REG_ID actually apply to the current dataset.

You can see that [HERE](https://github.com/eventespresso/event-espresso-core/blob/master/core/domain/services/admin/registrations/list_table/csv_reports/AnswersCSV.php#L43-L75)

If we remove the need for the additional processing required for various question/answer types (yes it's actually here needed but to give a basic example) that code is doing something like this:

```
foreach ($answers as $answer_row) {
    $data[$answer_row['Question.QST_admin_label']] = $answer_row['Answer.ANS_value];
}
```

In English..... for **every** answer row in `$answers`, add the value to `$data` using the question label as the key. 

If we already have the question label in `$data`, it simply adds the value to it... if we don't, its adding an additional element to the array using the current answers question label as the key and then adds the value to that. The problem with doing that is we are looking for a specific set of questions now, the CSV headers have been create for those specific questions so adding additional keys to `$data` throws the row out of sync when it's written to the CSV.

This is what happens with the export....

Master: https://monosnap.com/file/VlDzGylysUU6oBcoSXhMy23s7821fQ

This branch: https://monosnap.com/file/CAQSScl43dMzCsAXMs9oY26BT3dxRG

To highlight this more, here is what I see when I edit a single registration from that event: https://monosnap.com/file/zkqKr6FlsTSbvSeb8E66zvgT16lRsW

See how we have a bunch of questions/answers that apply the registration, but NOT the current event I'm exporting anymore.

---

In short, we include ALL registration answers when what we should be including in these reports is whatever has been set within `$question_lables` (which is then added to `$data`) because that is what EE is expecting values for, not everything.

Reproducing this is actually pretty easy now we know the above.

Add questions to either a new or current group.... add some registrations with answers to those questions, now remove the questions from the group, or remove the question group from the event itself (you don't need to delete the group or the questions, just unassign them so they no longer apply to the event) and run an export.

The export will look 'off' like values are missing headers etc.... but its not the headers that are missing, its additional values that have thrown out the CSV data row then pushing data around where it shouldn't be.

### DIFF
--- a/core/domain/services/admin/registrations/list_table/csv_reports/AnswersCSV.php
+++ b/core/domain/services/admin/registrations/list_table/csv_reports/AnswersCSV.php
@@ -55,6 +55,11 @@ class AnswersCSV
             } else {
                 $question_label = sprintf(esc_html__('Question $s', 'event_espresso'), $answer_row['Answer.QST_ID']);
             }
+            if(! array_key_exists($question_label, $data)) {
+                // We don't need an answer for this specific question in the current dataset
+                // so skip adding this value to $data.
+                continue;
+            }
             if (
                 isset($answer_row['Question.QST_type'])
                 && $answer_row['Question.QST_type'] == EEM_Question::QST_type_state


tn3rb (approved):
# AWESOME 💯 x 🍩

garthkoyle:
So @Pebblo what's the expected output now after a question/group are removed from an event?

Pebblo:
@garthkoyle currently EE expects the CSV report to only contain answers that are assigned to the events in question (be that all events, single events or multiple), so the CSV should only show the answers for questions assigned to the event(s).

Meaning if you have custom questions 1-5 assigned to an event and have registrations on the event that answer those questions, but you remove say questions 2 and 3 from the group (meaning they are no longer assigned to the event) then the CSV for that event should only show questions 1, 4 and 5 in the export (along with the correct answers for those questions).

That same logic applies throughout, it's not about what answers the registration(s) has, but the questions on the event(s) at the time of export.

Pebblo:
To test... first use **MASTER** branch, you need to see the issue yourself first to follow this one.

Create some new custom questions in Event Espresso -> Registration Form -> Questions.

I created 5 with the question text set to 'Custom Question X' (where X was 1 through 5)

Mine - https://monosnap.com/file/kZwywUElL4gVdqguwNJ192uqg4Ic7G

Assign those questions to a custom question group, like so: https://monosnap.com/file/Ag1BCAAVlrpgGMS98ztTk8inhvQchh

Create a new event and assigned that group on the primary reg questions: https://monosnap.com/file/mvIBGLUeZSude7iRATF2XpsEU1VP0X

Register onto that event a couple of times, I used 'Some Question X' for each questions answer (again X being the number for the question itself)

Run the events registration report.

This is what I see: https://monosnap.com/file/0OVX4UsveWDYLWa1nXgOzp50vl4k8d

You can see we have all 5 questions and each question has the correct answer (shown by the number at the end of the answer).

Now, edit the question group and uncheck question 2 and 3 from the group, then save:
https://monosnap.com/file/2AoLuagNEINDpPAfHg8q3Qe19KcmfP

What you see in master, is the questions change order slightly but all of those questions still show up:
https://monosnap.com/file/EoNN6STJkn6PIuRCSao8sB2KZegaRA

That's a symptom of this bug.... if the questions are no longer assigned to the event then why are they there? The original query for the export does NOT know about those questions, its working in this example but because of the setup, but its not valid.

Switch to this branch and re-run the same export, you should only see 'Some Question 1,2,3' in the export, like this:
https://monosnap.com/file/z2kqJ7n6NMZ4KvtvSNaxaBhMoCUNp4

That's the correct export for that event.

----

Now we have seen  basic example, lets see the fix for a 'questions/answers mismtaching'.

Switch back to **MASTER** branch.

On the site site above, create 2 more custom questions.  I used 'New Question 1' and 'New Questions 2':
https://monosnap.com/file/q8adHrfNBb8BaB49tMKSZ7K7DzKI5W

Assign them to another new question group and set that question group onto the event you test earlier.

(This highlights what happens when and admin has swapped and changed the questions in use on an event)

Update the event and add another registration, answering all current questions on the event:

https://monosnap.com/file/N5i9S2TFi1X0eAz04uJDXs4Vq0HMa2

It's a little harder to see here, but look here:
https://monosnap.com/file/pIss42yzN1wCvLkGsHwc8kHT8p9bAk

That 0 has 'shifted' over 2 spaces because the rows mismatched... why? Because 'some-question-2' and 'some-question-3' no longer belong in the export and they are working only because of the current setup, as you add more questions this becomes more problematic.

Re-run the export in this branch, this is how it should look: https://monosnap.com/file/HJLPX8Fhfwrw2EKawg7Pj7b8QVIsWw

Pebblo:
Thanks @garthkoyle, merged

Issue #3888, opened 2022-06-15
Fix Critical Pages dropdowns not saving changes


Pebblo:
**Describe the bug**
Changes to EE4's critical page dropdowns are no longer saving.

Go to Event Espresso -> General Settings -> Critical Pages.

Set a different page on any of the dropdowns and hit save, the page will show the original value. Refresh and its still the same, so its not an issue with it just loading the last value etc.

Tested v4.10.29.p and the changes save in that version, so new from .30+

garthkoyle:
This looks to be working fine now (video): https://www.screencast.com/t/gXTMHQU89EN

PR #3889, opened 2022-06-15
Fix Save Issue of Critical Pages Form

hoseinrafiei:
Fixes https://github.com/eventespresso/event-espresso-core/issues/3888 issue.

Before Change

After Change

### DIFF
--- a/admin_pages/general_settings/General_Settings_Admin_Page.core.php
+++ b/admin_pages/general_settings/General_Settings_Admin_Page.core.php
@@ -384,6 +384,7 @@ class General_Settings_Admin_Page extends EE_Admin_Page
      */
     protected function _update_espresso_page_settings()
     {
+        $this->core_config = EE_Registry::instance()->CFG->core;
         // capture incoming request data && set page IDs
         $this->core_config->reg_page_id       = $this->request->getRequestParam(
             'reg_page_id',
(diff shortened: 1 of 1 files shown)


hoseinrafiei:
@tn3rb

`$this->core_config` variable was null in `_update_espresso_page_settings` function. I fixed the issue by defining the variable but I'm curious to find how it was working before and what change created this problem.

I checked the file history and didn't find anything suspicious in the `admin_pages/general_settings/General_Settings_Admin_Page.core.php` file. So I guess maybe recently we changed something in higher level and a little worry about the affects of this change on some other parts of the EE.

Do you think that we should investigate and find the root cause of the problem?

tn3rb:
re:
> Do you think that we should investigate and find the root cause of the problem?

i think it's because originally, the code was referencing `EE_Registry::instance()->CFG->core` everywhere directly and I replaced it with a class property so that it could eventually be injected at run time instead of being hard coupled.  But i missed that the `admin_init()` method isn't called on that update request (still surprised at that)

hoseinrafiei:
> this is fine, another approach would be to remove this line, as well as the duplicate in `admin_init()` and then place it in an earlier method call like `_init_page_props()`

Good point. Done 👍

tn3rb (approved):
yup 👍🏻  plz just verify it still works

hoseinrafiei:
> yup 👍🏻 plz just verify it still works

Yes, already checked.

hoseinrafiei:
@Saam01

Please test this one. It is related to the registration as well. Please check the backend menu and set new pages (with shortcodes) and test whole registration process and post registration process (invoicing, edit, etc) to make sure that it works and new pages are applying.

garthkoyle:
This looks to be working fine now (video): https://www.screencast.com/t/gXTMHQU89EN

Saam01:
Feature seems to be working fine for me also.

PR #3895, opened 2022-06-16
Fix Blank Max Registrations per Transaction on New Event

### LINKED ISSUE #3893: "Max Registrations per Transaction" Does not save on Publish event
When we create an event and press Publish button some data does not save. For example, if we do not change "Max Registrations per Transaction" and leave the default value there it does not save.

Also, it would be good if we check other fields' values and ensure they are saved after Publish.

screen-capture

tn3rb:
This PR:

this was a really tricky issue to figure out and was beginning to drive me bonkers (but why drive when you're in walking distance)
I assumed that the problem was either caused by some unwanted code running, or some wanted code NOT running.
Turns out the issue was the database schema allowing null values if nothing was set, and since the legacy code for the reg options doesn't run if using the Barista plugin with the new EDTR, the value would get blanked out right within the db.

- fixes #3893
- changes the 4.12 DMS so that the event meta table schema no longer allows null values for the additional limit field
- does a BUNCH of formatting and code cleanup (it's what I do when I'm poking around in code)

### DIFF
--- a/core/db_models/EEM_Base.model.php
+++ b/core/db_models/EEM_Base.model.php
@@ -7,11 +7,9 @@ use EventEspresso\core\exceptions\ModelConfigurationException;
 use EventEspresso\core\exceptions\UnexpectedEntityException;
 use EventEspresso\core\interfaces\ResettableInterface;
 use EventEspresso\core\services\container\Mirror;
-use EventEspresso\core\services\dependencies\DependencyResolver;
 use EventEspresso\core\services\loaders\LoaderFactory;
 use EventEspresso\core\services\loaders\LoaderInterface;
 use EventEspresso\core\services\orm\ModelFieldFactory;
-use EventEspresso\core\services\request\RequestInterface;
 
 /**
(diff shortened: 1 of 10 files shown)


core/db_models/EEM_Base.model.php
```diff
@@ -5464,7 +5477,8 @@ protected function _deduce_fields_n_values_from_cols_n_values($cols_n_values)
                 foreach ($this->_get_fields_for_table($table_alias) as $field_name => $field_obj) {
                     if (! $field_obj->is_db_only_field()) {
                         // prepare field as if its coming from db
-                        $prepared_value = $field_obj->prepare_for_set($field_obj->get_default_value());
+                        $prepared_value                            =
+                            $field_obj->prepare_for_set($field_obj->get_default_value());
```
knazart, line 5481:
```suggestion
                        $prepared_value = $field_obj->prepare_for_set($field_obj->get_default_value());
```

knazart (dismissed):

Issue #3234, opened 2020-12-22
Set default recipient as the primary registrant or registrant when clicking to edit or create a custom notification


lorenzocaum:
**Is your feature request related to a problem? Please describe.**

The create custom and edit buttons in the notifications area at the end of the event editor take you to the admin recipient. The create custom button on the default message templates page also takes you to the admin recipient.

![Image 2020-12-22 at 8 56 04 AM](https://user-images.githubusercontent.com/1079138/102896134-e58a5080-4433-11eb-9ceb-b6264957ecbe.jpg)
![Image 2020-12-22 at 8 57 56 AM](https://user-images.githubusercontent.com/1079138/102896145-e7ecaa80-4433-11eb-884d-d8af384c83cb.jpg)

The pattern that I've seen is the customers usually need to edit the notification to the attendees (registrant). Specifically, they'll create a custom template, make changes, and then reach out for help because their changes aren't coming through.

**Describe the solution you'd like**
Can we adjust the default behavior so someone is taken to the default active notification to an attendee? That could be primary registrant or registrant depending on the notification.

Alternatively, can we scale this improvement back and start with these notifications?

- Registration approved
- Registration pending payment

**Describe alternatives you've considered**
The documentation for the message contexts has a couple of nudges to make sure you are viewing the correct recipient:

https://eventespresso.com/wiki/messages-system-working-with-message-contexts/

**Additional context**

I mentioned seeing this confusion here:
https://github.com/eventespresso/event-espresso-core/issues/725

PR #3897, opened 2022-06-20
Missing Message Type & Messenger Parameters

hoseinrafiei:
While checking https://github.com/eventespresso/event-espresso-core/issues/3234 issue, I found that the **Create New Custom** button in the **Notifications** meta-box of event editor throws following error.

An EE_Error exception was thrown!   code: Messages_Admin_Page - add_message_template - 1308
"Sorry, but we can't create new templates because we're missing the messenger or message type"
click to view backtrace and class/method details
/Users/hossein/Projects/ee/wp-content/plugins/event-espresso-core/admin_pages/messages/Messages_Admin_Page.core.php   ( line no: 1308 )

You can see this error only if you try to open the link in a new tab / window.

Screen Shot 2022-06-20 at 17 21 23

This PR fixed this issue.

### DIFF
--- a/core/libraries/messages/EE_messenger.lib.php
+++ b/core/libraries/messages/EE_messenger.lib.php
@@ -567,7 +567,9 @@ abstract class EE_messenger extends EE_Messages_Base
                 $create_url_query_args = [
                     'page' => 'espresso_messages',
                     'action' => 'add_new_message_template',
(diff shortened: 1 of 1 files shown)


tn3rb (approved):
Good find!  Great fix!

Issue #3873, opened 2022-06-07
2022 theme: "Display Venue Details" option for the Event listing page is not reflecting in the frontend


Saam01:
When we activate the option to show "Display Venue Details" and "Display Ticket Selector" option for the Event listing page under EE > Events > Templates, it is not showing the venue and other options.

https://user-images.githubusercontent.com/99984614/172168837-91f0c674-b64d-4d5d-9909-64b643799a09.mp4

**Environment:**
 - Version of Event Espresso: 4.10.31.rc.003 
 - Version of WordPress: 6.0
 - Browser: Chrome 
 - PHP Version: 7.4.1

tn3rb:
@hoseinrafiei && @majidabbasi96 

Can you guys take a look at this issue and see if you can figure out an easy fix for it?
I've messed around with it but can not figure out an easy solution.

You'll want to look at `modules/events_archive/EED_Events_Archive.module.php` as that is the logic that hooks into non-block enabled themes to inject our logic. If it look overly complex, it's because the events list shortcode shares the same logic.

PR #3900, opened 2022-06-21
Fix Issue of Loading Ticket Selector, Event Datetime, and Event Details

hoseinrafiei:
Related to https://github.com/eventespresso/event-espresso-core/issues/3873 issue.

### DIFF
--- a/modules/events_archive/EED_Events_Archive.module.php
+++ b/modules/events_archive/EED_Events_Archive.module.php
@@ -316,7 +316,7 @@ class EED_Events_Archive extends EED_Module
                 EEH_Template::load_espresso_theme_functions();
                 // because we don't know if the theme is using the_excerpt()
                 add_filter(
-                    'the_excerpt',
+                    'get_the_excerpt',
                     array('EED_Events_Archive', 'event_details'),
                     EED_Events_Archive::EVENT_DETAILS_PRIORITY
                 );
@@ -351,7 +351,7 @@ class EED_Events_Archive extends EED_Module
         }
         if (apply_filters('FHEE__EED_Events_Archive__get_the_excerpt__theme_uses_get_the_excerpt', false)) {
             remove_filter(
-                'the_excerpt',
+                'get_the_excerpt',
                 array('EED_Events_Archive', 'event_details'),
                 EED_Events_Archive::EVENT_DETAILS_PRIORITY
             );
@@ -454,7 +454,7 @@ class EED_Events_Archive extends EED_Module
         // we need to first remove this callback from being applied to the_content() or the_excerpt()
         // (otherwise it will recurse and blow up the interweb)
         remove_filter(
-            'the_excerpt',
+            'get_the_excerpt',
             array('EED_Events_Archive', 'event_details'),
             EED_Events_Archive::EVENT_DETAILS_PRIORITY
         );
@@ -470,7 +470,7 @@ class EED_Events_Archive extends EED_Module
         $content = EED_Events_Archive::instance()->template_parts->apply_template_part_filters($content);
         // re-add our main filters (or else the next event won't have them)
         add_filter(
-            'the_excerpt',
+            'get_the_excerpt',
             array('EED_Events_Archive', 'event_details'),
             EED_Events_Archive::EVENT_DETAILS_PRIORITY
         );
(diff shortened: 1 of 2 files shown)


hoseinrafiei:
@tn3rb

It's a quick fix for the https://github.com/eventespresso/event-espresso-core/issues/3873 issue.
I changed all the instances of `the_excerpt` to `get_the_excerpt` hook and now it's loading on both block and classic themes.

What's your idea?

tn3rb:
@hoseinrafiei 
i made a few changes because i'm pretty sure that switching exclusively to using `get_the_excerpt` won't work with some themes, so i used a conditional to flip flop between that and `the_excerpt`.
I also formatted and cleaned up the code in that file. 
Plz review my changes (i'll get Majid to look at it too)

modules/events_archive/EED_Events_Archive.module.php
```diff
@@ -227,21 +218,19 @@ public function run($WP)
...
         // set params that will get used by the filters
         EEH_Event_Query::set_query_params(
-            '',    // month
-            '',    // category
-            $config->display_expired_events,    // show_expired
-            'start_date',    // orderby
-            'ASC'    // sort
+            '',                             // month
+            '',                             // category
+            $config->display_expired_events // show_expired
```
tn3rb, line 223: those last two params were removed because they were the same as the default values for that function

modules/events_archive/EED_Events_Archive.module.php
```diff
@@ -287,11 +273,10 @@ public static function link_target()
 
 
     /**
-     *    template_include
-     *
-     * @access    public
```
tn3rb, line 292: this code is old! 
we used to have to write the function name and the scope (public, private, etc) in the phpdocs because earlier IDEs and code editors could not parse that info from the code itself. Totally not necessary anymore

modules/events_archive/EED_Events_Archive.module.php
```diff
@@ -598,146 +508,203 @@ public static function event_venues($content)
...
-     *
-     * @access    private
-     * @return        void
+     * @return bool
+     * @since $VID:$
+     */
+    public static function isBlockTheme()
+    {
+        return function_exists('wp_is_block_theme') && wp_is_block_theme();
+    }
```
tn3rb, line 517: @hoseinrafiei 
extracted this fix of your from a previous PR

modules/events_archive/EED_Events_Archive.module.php
```diff
@@ -841,45 +796,44 @@ public static function update_template_settings($CFG, $REQ)
...
+                ) : 1;
+            $CFG->EED_Events_Archive->display_venue           =
+                isset($REQ['EED_Events_Archive_display_venue']) ? absint(
+                    $REQ['EED_Events_Archive_display_venue']
+                ) : 0;
+            $CFG->EED_Events_Archive->display_expired_events  = isset($REQ['EED_Events_Archive_display_expired_events'])
                 ? absint($REQ['EED_Events_Archive_display_expired_events']) : 0;
         }
         return $CFG;
     }
```
majidabbasi96, line 819: I see different coding styles for this function for the same logic, I think it is better to have one for all.

modules/events_archive/EED_Events_Archive.module.php
hoseinrafiei: Great.
Actually, I was going to do this myself. My plan was to add this code in the `EEH_Template` class.

modules/events_archive/EED_Events_Archive.module.php
```diff
@@ -598,146 +508,203 @@ public static function event_venues($content)
...
+            $excerpt_filter = EED_Events_Archive::isBlockTheme() ? 'get_the_excerpt' : 'the_excerpt';
+        }
+        return $excerpt_filter;
+    }
+
+
+    /**
+     * @since $VID:$
+     */
+    protected static function addEventDetailsFilters()
```
hoseinrafiei, line 537: Great extraction. 👍

hoseinrafiei:
@Saam01 This is a fix to the issue that you reported. Please check and report.

modules/events_archive/EED_Events_Archive.module.php
tn3rb: oh ya, that would be even better, good thinking

modules/events_archive/EED_Events_Archive.module.php
tn3rb: ahhh ya... autoformatting doesn't always work great and i missed that. 

thnx will fix

Saam01:
Function seems to be working fine now.

PR #3907, opened 2022-06-23
Fix Filter Views for CSV Reg Report

### LINKED ISSUE #3899: Fix searching within the registration list removing filters
Reported here: https://eventespresso.com/topic/csv-exports-ee4/

If you apply a filter within the registration list table and then use search to narrow down your selection... when you export the results none of the filters (including the search) apply to the export.

**To Reproduce**
I use an event with 2 different ticket types to test this.

Have registrations on both ticket types and have some that are pending payment (or just something other than Approved).

Before: https://monosnap.com/file/ujmChfmG47ehfzNPLGZ7X8Lx6lCaNR

Export works fine.

Set the Reg status to Approved, filter: https://monosnap.com/file/nhBQKubaib0RLRjbcRRzRfxYW8cBSv

My export now contains 7 registrations, as expected.

Now add something to the search field, in my example I'm filtering to another specific ticket name: https://monosnap.com/file/9z4TfkaFOEa6u4CPBzPoMHzoTHJ3B5

Export now contains ALL registrations, including the pending payment reg that was filtered previously, so in this example I'm getting 8 regs exported, should be 2 as above.

(Priority high due to previous issues with exports)

### LINKED ISSUE #3901: Fix views no longer being applied to registration reports
@joaovictorangeline mentioned that the 'Views' on the registrations table did not apply to the registration report during testing of #3875 , specifically [HERE](https://github.com/eventespresso/event-espresso-core/pull/3875#issuecomment-1149378953)

Afaik those views did not apply to the report, however, I've been testing the difference between v4.10.29 and the current version with regards to the CSV reports and those views do indeed apply in v29 when using the 'Filtered CSV Report' button (my apologies to @joaovictorangeline )

----

To reproduce simply go to Event Espresso -> Registrations.

Use say the 'This Month' view: https://monosnap.com/file/MHeGrOL1UCpZe969xlR2470qWl7jj4

You can see I get 8 registrations returned and that's what is exported when I click the 'Filtered CSV Report'

In master, I get the same registration list table view but on export, it exports all of my sites registrations

tn3rb:
This PR fixes #3901 && fixes #3899

### DIFF
--- a/core/admin/EE_Admin_List_Table.core.php
+++ b/core/admin/EE_Admin_List_Table.core.php
@@ -486,6 +486,10 @@ abstract class EE_Admin_List_Table extends WP_List_Table
         if (empty($filters)) {
             return;
         }
+        $use_filters = isset($this->_req_data['use_filters'])
+                       && filter_var($this->_req_data['use_filters'], FILTER_VALIDATE_BOOLEAN)
+            ? 'yes'
+            : 'no';
 
         echo '<div class="ee-list-table-filters actions alignleft">';
         foreach ($filters as $filter) {
@@ -495,7 +499,7 @@ abstract class EE_Admin_List_Table extends WP_List_Table
         echo '<input type="submit" class="ee-list-table-filter-submit button button--secondary" value="'
              . esc_html__('Filter', 'event_espresso')
              . '" id="post-query-submit" />';
-        echo '<input type="hidden" id="ee-list-table-use-filters" name="use_filters" value="no"/>';
+        echo '<input type="hidden" id="ee-list-table-use-filters" name="use_filters" value="' . $use_filters .'"/>';
 
         // add reset filters button at end
         echo '<a class="button button--secondary"  href="'
@@ -735,7 +739,7 @@ abstract class EE_Admin_List_Table extends WP_List_Table
      */
     public function single_row_columns($item)
     {
-        [$columns, $hidden, $sortable, $primary] = $this->get_column_info();
+        list($columns, $hidden, $sortable, $primary) = $this->get_column_info();
 
         foreach ($columns as $column_name => $column_display_name) {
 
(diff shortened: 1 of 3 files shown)


core/admin/EE_Admin_List_Table.core.php
```diff
@@ -735,7 +739,7 @@ public function get_hidden_columns()
      */
     public function single_row_columns($item)
     {
-        [$columns, $hidden, $sortable, $primary] = $this->get_column_info();
+        list($columns, $hidden, $sortable, $primary) = $this->get_column_info();
```
hoseinrafiei, line 742: Is something wrong with the [] syntax? Curios to know why do you reverted to the classic style? 🤔

core/admin/EE_Admin_List_Table.core.php
```diff
@@ -806,8 +810,8 @@ public function extra_tablenav($which)
         } else {
             echo '<div class="list-table-bottom-buttons alignleft actions">';
             foreach ($this->_bottom_buttons as $type => $action) {
-                $route         = $action['route'] ?? '';
-                $extra_request = $action['extra_request'] ?? '';
+                $route         = isset($action['route']) ? $action['route'] : '';
+                $extra_request = isset($action['extra_request']) ? $action['extra_request'] : '';
```
hoseinrafiei, line 814: They are equivalent and null coalescing operator is cleaner and more readable. Any reason to switch to classic syntax?

hoseinrafiei (approved):

core/admin/EE_Admin_List_Table.core.php
tn3rb: my IDE converts some of these things automatically, but this branch will be getting merged into master which still allows PHP 5.6, but PHP 5.6 does not allow array destructuring nor null coalescing operators

core/admin/EE_Admin_List_Table.core.php
tn3rb: i agree but master branch still allows PHP 5.6

Pebblo:
@tn3rb I'm still getting the same issues #3899 & #3901 on this branch.

A quick test is to use the 'Today' view, I have 1 registration for today but I'm getting a full export

Using search on a filtered report still returns all of the registrations for the current event.

knazart (approved):

tn3rb:
sry about that Tony

I had tested views while ALSO using the filters, but had not tested using views on their own. 
The way i had the code was only applying the views if you were also using other filters, but that should be fixed now

garthkoyle:
What I found exporting registration data:

* [x] FILTERS ONLY works
* [x] SEARCH + FILTERS outputs all data 💥 when it should be less data (usually a subset of the filters only)
* [ ] SEARCH ONLY exports all data.

garthkoyle:
okay, so now this generates the right csv: 

- no filters
- filters
- search
- filters + search

Issue #3905, opened 2022-06-23
Fix CSV upload not working due to core changes


Pebblo:
The Attendee importer add-on no longer works due to change made in #3806 

To reproduce add the eea-importer add-on to a site: https://github.com/eventespresso/eea-importer

Then.... for later use save a registration CSV export, just go to Event Espresso -> Registrations and run a CSV report.

Then go to Event Espresso -> Importer

Select an Event, then a ticket and it will ask for a CSV file....

The file you use here doesn't matter, select the CSV report you generated earlier and click to upload.

The page will refresh on that same step.

If I checkout right before #3806 was merged, this works as expected.

PR #3911, opened 2022-06-28
Kses - Part 15

hoseinrafiei:
Fixes https://github.com/eventespresso/event-espresso-core/issues/3905 issue.

It was due to lack of one `space` character between the attributes. We had no space between **name** and **enctype** attributes so both get removed by `wp_kses` function.


Pebblo (approved):
Gotta love space issues!

Issue #3871, opened 2022-06-06
Add Session Lifespan Admin Option


tn3rb:
The Session Lifespan controls how long the EE user sessions remain active on the frontend of a website during the registration process. Currently this value defaults to 1 hour and can only be controlled via a filter. This has worked fine for EE plugin users but does not work for Event Smart users as they have no ability to add filters. 

for additional background, plz see: https://github.com/eventespresso/eventsmart.com-website/issues/984 

- create new class `EventEspresso\core\domain\values\session\SessionLifespanOption`
	- have `SessionLifespanOption` extend `EventEspresso\core\services\database\WordPressOption` (search for other classes that do the same for examples)
	- set an appropriately named constant for the option name 
	- add an appropriately named setter and getter (with validation) for value
	- set default value to `(int) HOUR_IN_SECONDS`

- in constructor for `EventEspresso\core\domain\values\session\SessionLifespan`
	- replace ternary checking `$lifespan` and setting default of 1 hour with if conditional
	- if `$lifespan` is not set, instantiate `SessionLifespanOption` 
	- get default value for `$lifespan`  from `SessionLifespanOption` 

- open `caffeinated/admin/extend/registration_form/Extend_Registration_Form_Admin_Page.core.php` 
	- look at `_reg_form_settings()` and `_update_reg_form_settings()`
	- using the code for `email_validation_settings_form()` as an example
		- add methods `setSessionLifespan` and `updateSessionLifespan`
		- add hook in `_reg_form_settings()` using `setSessionLifespan()` as callback
		- position form after `copy_attendee_info_settings_form` (the hook priority likely controls the form position)
		- call `updateSessionLifespan()` in `_update_reg_form_settings()` but do NOT pass or set `EE_Registry::instance()->CFG->registration` as we will not be storing the session lifespan value in that config class.
		- instead of adding method like `_email_validation_settings_form()` create a new folder : `caffeinated/admin/extend/registration_form/forms`
		- create new class in that folder named `SessionLifespanForm` that extends `EE_Form_Section_Proper` and namespace that form accordingly
		- use the form structure in `_email_validation_settings_form()` as an example
		- use the `EE_Admin_Two_Column_Layout` layout strategy
		- add a heading titled "Session Settings"
		- add a select input with following 
			- answer options: 5, 15, 30, 60, & 120 validated as integers
			- input settings: 
				- label text: "Session Lifespan"
[...]

PR #3926, opened 2022-07-02
Session Lifespan Feature

hoseinrafiei:
Added a new feature to manage session lifespan in **Registration Settings**.
Fixes https://github.com/eventespresso/event-espresso-core/issues/3871 issue.

Added a test class to test the functionality of Session Lifespan Option.
Updated the WP GraphQL package.

### DIFF
--- a/caffeinated/admin/extend/registration_form/Extend_Registration_Form_Admin_Page.core.php
+++ b/caffeinated/admin/extend/registration_form/Extend_Registration_Form_Admin_Page.core.php
@@ -1463,15 +1463,14 @@ class Extend_Registration_Form_Admin_Page extends Registration_Form_Admin_Page
 
     public function setSessionLifespan()
     {
-        $session_lifespan_form = new SessionLifespanForm(new SessionLifespanOption());
+        $session_lifespan_form = $this->loader->getNew(SessionLifespanForm::class);
         echo wp_kses($session_lifespan_form->get_html(), AllowedTags::getWithFormTags());
     }
 
 
     public function updateSessionLifespan()
     {
-        $session_lifespan_option = new SessionLifespanOption();
-        $handler = new SessionLifespanFormHandler($session_lifespan_option);
-        $handler->process(new SessionLifespanForm($session_lifespan_option));
+        $handler = $this->loader->getNew(SessionLifespanFormHandler::class);
+        $handler->process($this->loader->getNew(SessionLifespanForm::class));
     }
 }
--- a/caffeinated/admin/extend/registration_form/Extend_Registration_Form_Admin_Page.core.php
+++ b/caffeinated/admin/extend/registration_form/Extend_Registration_Form_Admin_Page.core.php
@@ -1,5 +1,7 @@
 <?php
 
+use EventEspresso\caffeinated\admin\extend\registration_form\forms\SessionLifespanForm;
+use EventEspresso\caffeinated\admin\extend\registration_form\forms\SessionLifespanFormHandler;
 use EventEspresso\core\exceptions\InvalidDataTypeException;
 use EventEspresso\core\exceptions\InvalidInterfaceException;
 use EventEspresso\core\services\request\sanitizers\AllowedTags;
@@ -1108,6 +1110,11 @@ class Extend_Registration_Form_Admin_Page extends Registration_Form_Admin_Page
             array($this, 'copy_attendee_info_settings_form'),
             4
         );
+        add_action(
+            'AHEE__Extend_Registration_Form_Admin_Page___reg_form_settings_template',
+            array($this, 'setSessionLifespan'),
+            4.9
+        );
         $this->_template_args = (array) apply_filters(
             'FHEE__Extend_Registration_Form_Admin_Page___reg_form_settings___template_args',
             $this->_template_args
@@ -1139,6 +1146,7 @@ class Extend_Registration_Form_Admin_Page extends Registration_Form_Admin_Page
         EE_Registry::instance()->CFG->registration = $this->update_copy_attendee_info_settings_form(
             EE_Registry::instance()->CFG->registration
         );
+        $this->updateSessionLifespan();
         EE_Registry::instance()->CFG->registration = apply_filters(
             'FHEE__Extend_Registration_Form_Admin_Page___update_reg_form_settings__CFG_registration',
             EE_Registry::instance()->CFG->registration
@@ -1247,15 +1255,13 @@ class Extend_Registration_Form_Admin_Page extends Registration_Form_Admin_Page
                             __LINE__
                         );
                     }
-                } else {
-                    if ($copy_attendee_info_settings_form->submission_error_message() !== '') {
-                        EE_Error::add_error(
-                            $copy_attendee_info_settings_form->submission_error_message(),
-                            __FILE__,
-                            __FUNCTION__,
-                            __LINE__
-                        );
-                    }
+                } elseif ($copy_attendee_info_settings_form->submission_error_message() !== '') {
+                    EE_Error::add_error(
+                        $copy_attendee_info_settings_form->submission_error_message(),
+                        __FILE__,
+                        __FUNCTION__,
+                        __LINE__
+                    );
                 }
             }
         } catch (EE_Error $e) {
@@ -1378,15 +1384,13 @@ class Extend_Registration_Form_Admin_Page extends Registration_Form_Admin_Page
                             __LINE__
                         );
                     }
-                } else {
-                    if ($email_validation_settings_form->submission_error_message() !== '') {
-                        EE_Error::add_error(
-                            $email_validation_settings_form->submission_error_message(),
-                            __FILE__,
-                            __FUNCTION__,
-                            __LINE__
-                        );
-                    }
+                } elseif ($email_validation_settings_form->submission_error_message() !== '') {
+                    EE_Error::add_error(
+                        $email_validation_settings_form->submission_error_message(),
+                        __FILE__,
+                        __FUNCTION__,
+                        __LINE__
+                    );
                 }
             }
         } catch (EE_Error $e) {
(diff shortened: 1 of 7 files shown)


hoseinrafiei:
@tn3rb

Please check this. All the items that you described here are done: https://github.com/eventespresso/event-espresso-core/issues/3871
I also updated the `wp-graphql` package to the latest production version and the tests are running fine now but we may check the graphql API to make sure there is no problem with the latest package.

Also the test class is added and works fine. Your feedback on the test class is highly appreciated.

caffeinated/admin/extend/registration_form/Extend_Registration_Form_Admin_Page.core.php
```diff
@@ -1247,15 +1255,13 @@ public function update_copy_attendee_info_settings_form(EE_Registration_Config $
...
-                } else {
-                    if ($copy_attendee_info_settings_form->submission_error_message() !== '') {
-                        EE_Error::add_error(
-                            $copy_attendee_info_settings_form->submission_error_message(),
-                            __FILE__,
-                            __FUNCTION__,
-                            __LINE__
-                        );
-                    }
+                } elseif ($copy_attendee_info_settings_form->submission_error_message() !== '') {
```
tn3rb, line 1258: 👍🏻 good stuff

tn3rb (changes requested):
This all looks really really good...

I did have a couple of requests for a few things because I changed my mind about how to approach things (not because of you, you're code was great)

tn3rb:
just looked, and you can ignore this failed unit test:

```
1) CachingLoaderTest::testResetCache
Failed asserting that actual size 3 matches expected size 0.
```

it does not happen locally and I am currently trying to figure it out

caffeinated/admin/extend/registration_form/forms/SessionLifespanForm.php
```diff
@@ -0,0 +1,62 @@
...
+
+
+    /**
+     * SessionLifespanForm constructor.
+     *
+     * @throws EE_Error
+     */
+    public function __construct()
+    {
+        $this->sessionLifespanOption = new SessionLifespanOption();
```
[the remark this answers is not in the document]
hoseinrafiei: Done.

caffeinated/admin/extend/registration_form/forms/SessionLifespanFormHandler.php
```diff
@@ -0,0 +1,70 @@
...
+use EE_Select_Input;
+use EEH_HTML;
+use EventEspresso\core\domain\values\session\SessionLifespanOption;
+
+class SessionLifespanFormHandler
+{
+    /**
+     * @var SessionLifespanOption
+     */
+    private $sessionLifespanOption;
```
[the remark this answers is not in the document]
hoseinrafiei: 👍

caffeinated/admin/extend/registration_form/forms/SessionLifespanFormHandler.php
```diff
@@ -0,0 +1,70 @@
...
+
+    /**
+     * SessionLifespanForm constructor.
+     *
+     * @throws EE_Error
+     */
+    public function __construct()
+    {
+        $this->sessionLifespanOption = new SessionLifespanOption();
+    }
```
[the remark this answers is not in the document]
hoseinrafiei: 👍

core/domain/values/session/SessionLifespan.php
```diff
@@ -37,7 +37,10 @@ class SessionLifespan
     public function __construct($lifespan = 0)
     {
         $lifespan = absint($lifespan);
-        $lifespan = $lifespan > 0 ? $lifespan : (int) HOUR_IN_SECONDS;
+        if ($lifespan <= 0) {
+            $lifespanOption = new SessionLifespanOption();
+            $lifespan = $lifespanOption->getSessionLifespan();
+        }
         $this->setLifespan($lifespan);
     }
```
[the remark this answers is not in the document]
hoseinrafiei: Yes, It was good idea. Thanks for sharing. I did it.

hoseinrafiei:
@tn3rb Please check it again. I implemented your requested changes.

core/EE_Dependency_Map.core.php
```diff
@@ -578,6 +578,9 @@ protected function _register_core_dependencies()
             'EventEspresso\core\domain\services\commands\attendee\CreateAttendeeCommandHandler'                                  => [
                 'EEM_Attendee' => EE_Dependency_Map::load_from_cache,
             ],
+            'EventEspresso\core\domain\values\session\SessionLifespan'                                                    => [
+                'EventEspresso\core\domain\values\session\SessionLifespanOption' => EE_Dependency_Map::load_from_cache,
+            ],
```
tn3rb, line 583: ### AWESOME !!! 😀 

wanted to see if you could figure that out yourself!

core/domain/values/session/SessionLifespan.php
```diff
@@ -31,17 +31,12 @@ class SessionLifespan
...
-    public function __construct($lifespan = 0)
+    public function __construct(SessionLifespanOption $session_lifespan_option)
     {
-        $lifespan = absint($lifespan);
-        if ($lifespan <= 0) {
-            $lifespanOption = new SessionLifespanOption();
-            $lifespan = $lifespanOption->getSessionLifespan();
-        }
-        $this->setLifespan($lifespan);
+        $this->setLifespan($session_lifespan_option->getSessionLifespan());
```
tn3rb, line 39: 👍🏻

caffeinated/admin/extend/registration_form/Extend_Registration_Form_Admin_Page.core.php
```diff
@@ -1462,14 +1463,15 @@ private function _verify_pcre_support(EE_Registration_Config $EE_Registration_Co
 
     public function setSessionLifespan()
     {
-        $session_lifespan_form = new SessionLifespanForm();
+        $session_lifespan_form = new SessionLifespanForm(new SessionLifespanOption());
```
tn3rb, line 1466: if we use the loader, then any dependencies can be automatically injected

```php
$session_lifespan_form = $this->loader->getNew(SessionLifespanForm::class);
```

`$this->loader` should be available within all admin pages

hoseinrafiei: Done.

hoseinrafiei:
@tn3rb Please check it again.

tn3rb (approved):
### Great Job Hossein 
# 💯 ✖️ 🌮 ➡️ 🦖

tn3rb:
@hoseinrafiei 

Once final request... I'm concerned by the 900,000 lines of code being removed from the codebase as part of this PR.
Was there a specific reason why WP GQL needed to be updated as part of this PR?
I'm thinking that that update should be split out into it's own PR because if it causes problems then we may need to revert it, which we won't be able to do if it is embedded within this PR.

hoseinrafiei:
> @hoseinrafiei
> 
> Once final request... I'm concerned by the 900,000 lines of code being removed from the codebase as part of this PR. Was there a specific reason why WP GQL needed to be updated as part of this PR? I'm thinking that that update should be split out into it's own PR because if it causes problems then we may need to revert it, which we won't be able to do if it is embedded within this PR.

@tn3rb

It was to fix the issue of phpunit tests and writing proper test for the new class.
We discussed in Slack channel about it.

I'm fine to remove it and create a separate PR for it. Just let me know how to do it. Just point me to the right direction and then I will search and find.

hoseinrafiei:
@tn3rb I moved the WP GraphQL update to https://github.com/eventespresso/event-espresso-core/pull/3993 branch and removed it from this branch. Please check it and let me know if anything else is needed.

tn3rb:
ok this is looking great, just have a few things that need to be cleaned up before we can merge this:

- there's some linting errors
- needs to be rebased
- the `composer.json` file needs to be updated 


#### `composer.json` update

- add top level item named `allow-plugins`
- add keys for each package listed in `require` or `require-dev`
- set values to `true`
- look at EE core for an example

tn3rb (changes requested):
see above note

hoseinrafiei:
@tn3rb Please check it again. The composer.json file seems to be fixed automatically after rebase.

tn3rb (approved):
# Great Job 💯

garthkoyle:
@hoseinrafiei I don't see this working. I have registrations that are "incomplete" status (no registration form info submitted), and their status is still Incomplete way after the time period selected in the Registration Settings.

Currently, my session length is set to 5 minutes: 
image

The lifespan session length of time has passed, and at least the Incomplete registrations are still `Incomplete`. Based on Brent's description, I expected the `Incomplete` registrations to be set to `Canceled` and thus do not count against the capacity limits set for the datetime and/or ticket.

image


Does the session lifespan apply to each registration individually or does the lifespan time period apply to all registrations no matter when received? For example, if my registration lifespan period was set to 1 hour when I registered 20 minutes ago, and now I set my lifespan to 5 minutes and then register again, what lifespan time period will be applied to each of those two registrations?

Does this code only apply to "Incomplete" registrations or all registration types, e.g. Pending Payment, etc?

garthkoyle:
With more testing I discovered about the lifespan feature. When the lifespan feature triggers according to the selected time period:

- Free or paid registrations without any registration information (Incomplete Registrations), the registration status is never changed in the admin. Only the event tickets become available after the hold time period expires.
- Paid registrations WITH registration form information (Pending Payment) are NOT ever changed in the admin, but the tickets become available after the hold time expires.

So, this probably functions correctly, but maybe the only change I might recommend is to set the registration status to `Canceled` for `Incomplete` registrations only. Would that make sense?

hoseinrafiei:
@garthkoyle 

Thanks for your feedback. This task was to add an ability to change the **session lifespan time** using a configuration form and some code refactor. Previously the time could change only through the code using a WordPress filter.

Changing the status was not included in the task description, but it looks great to add this ability in my idea.
I'm just not sure if `Canceled` is a suitable term for `Incomplete` registrations since we have to differentiate between those registrations that really canceled by user (after complete booking) and those registrations that are `Abandoned` during the booking process.

What do you think?

hoseinrafiei:
@tn3rb I think it will be good if you get involved and provide your feedback and also let me know what is the best way to implement the change status function?

PR #3932, opened 2022-07-04
Redirect SPCO if Payment has Redirect URL

tn3rb:
Old work branch of Michael's that is still valid and should be merged.

comment from Michael on commit:

> instead of relying on the gateways type to determine if SPCO should redirect, use whether or not the payment has a redirect URL

### DIFF
--- a/modules/single_page_checkout/reg_steps/payment_options/EE_SPCO_Reg_Step_Payment_Options.class.php
+++ b/modules/single_page_checkout/reg_steps/payment_options/EE_SPCO_Reg_Step_Payment_Options.class.php
@@ -1745,14 +1745,14 @@ class EE_SPCO_Reg_Step_Payment_Options extends EE_SPCO_Reg_Step
                     $this->_redirect_because_event_sold_out();
                     return false;
                 }
-                $payment_successful = $this->_process_payment();
-                if ($payment_successful) {
+                $payment = $this->_process_payment();
+                if ($payment instanceof EE_Payment) {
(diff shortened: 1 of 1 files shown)


knazart (approved):

tn3rb:
testing:

looks like the following gateways set a redirect url:

- Authorize.net
- PayPal Standard
- PayPal Express

there might be others I don't currently have installed that do so as well, but for 100% thoroughness, ALL offsite gateways should be tested

garthkoyle:
@Pebblo and @knazart is the redirect URL something that's used post-transaction completion? Where is that set in EE or is that in the account settings?

tn3rb:
redirect URL is hardcoded in the payment method code
it's where the browser sends you when you go to an offsite gateway

joaovictorangeline:
@tn3rb How can I test this?

tn3rb:
@joaovictorangeline 
you'll need to setup each of the following payment methods:
- Authorize.net
- PayPal Standard
- PayPal Express

then test a registration using each payment method.

As long as everything redirects correctly and you end up at the Thank You page (you may have to manually click something at the payment gateway site in order to return) then it should be good to go

joaovictorangeline:
Tested and it seems to be working fine.
Tested with Authorize.net, PayPal Standard, and PayPal express. 💯

Pebblo:
Also confirmed this worked fine with the above payment methods.

I also tested Mollie as iirc it does something weird with the redirect. Works fine.

PR #3943, opened 2022-07-06
Bump All GraphQL Max Query Limits Up To 250

tn3rb:
An EventSmart customer has 100+ dates and 100+ tickets but they are not all showing in the EDTR because by default, GraphQL limits the max returned records to 100.

This PR does exactly what the title says 😃 and bumps the limit up to 250

The following is from an event where I used REM to create over 200 dates and tickets
Screenshot 2022-07-06 142443


majidabbasi96 (approved):

garthkoyle:
This looks good. 

on current `dev`
image


on this branch the results of the pagination are more than doubled: 
image

PR #3951, opened 2022-07-07
Fix Command Bus Middleware

tn3rb:
Somehow, when updating FQCNs for commands, my IDE failed to return a result used in SPCO when searching for `EventEspresso\core\services\commands`

This PR:
- updates the import used in the first reg step 
- updates a couple of phpdocs
- used the request class to verify the request type in SPCO when creating registrations

EDIT:
also found a bug in the command bus middleware stack that was preventing middleware from running in composite commands more than once.


hoseinrafiei (approved):

Pebblo:
This works! Merging in.

Issue #3963, opened 2022-07-20
Fix Datetime & Ticket descriptions displaying escaped HTML


Pebblo:
The contents of the ticket and datetime descriptions are being escaped on output.

To reproduce edit an event and click the cog icon for either a datetime or ticket:

![Events ‹ Tony's EE4 Beta Testing — WordPress - Goo](https://user-images.githubusercontent.com/4345417/180002649-0ee5dcd8-c537-4506-9395-216887a5262b.png)

```
Some price test

If I add something here & here what happens?
<br>
And here?
```

Save the event and it will convert the output to be similar to what you see in the image.

Also, the output on the front end does not include any of the line breaks included, can we include those?

![Something something - Tony's EE4 Beta Testing - Go](https://user-images.githubusercontent.com/4345417/180003183-281df2f7-403e-4608-99da-e49329e81df6.png)

----

Note - some users already have `<br>` tags within the content above, I assume to include line breaks on the front end.

Pebblo:
Related issue: https://eventespresso.com/topic/ticket-selector-problem-with-german-characters-after-updating-the-scropt/#post-334436

I think this is also related to the post above, German special characters are not being output correctly:

```
[...]

PR #3964, opened 2022-07-23
Fix Issue of Escaping Content of Ticket Description

hoseinrafiei:
Fixes https://github.com/eventespresso/event-espresso-core/issues/3963 issue.

## How has this been tested

Screen Shot 2022-07-22 at 17 14 41

Screen Shot 2022-07-22 at 17 15 26

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,4 @@ vendor/
 *.DS_Store
 *.idea
 *.vscode
+tests/.phpunit.result.cache
--- a/caffeinated/admin/new/pricing/espresso_events_Pricing_Hooks.class.php
+++ b/caffeinated/admin/new/pricing/espresso_events_Pricing_Hooks.class.php
@@ -1539,7 +1539,7 @@ class espresso_events_Pricing_Hooks extends EE_Admin_Hooks
                     )
                 ),
             'TKT_ID'                        => $default ? 0 : $ticket->ID(),
-            'TKT_description'               => $default ? '' : $ticket->get_f('TKT_description'),
+            'TKT_description'               => $default ? '' : $ticket->get_raw('TKT_description'),
             'TKT_is_default'                => $default ? 0 : $ticket->is_default(),
             'TKT_required'                  => $default ? 0 : $ticket->required(),
             'TKT_is_default_selector'       => '',


.gitignore
```diff
@@ -16,3 +16,4 @@ vendor/
 *.DS_Store
 *.idea
 *.vscode
+tests/.phpunit.result.cache
```
tn3rb, line 19: ahhh beat me to it... :)

caffeinated/admin/new/pricing/espresso_events_Pricing_Hooks.class.php
```diff
@@ -1539,7 +1539,7 @@ protected function _get_ticket_row(
                     )
                 ),
             'TKT_ID'                        => $default ? 0 : $ticket->ID(),
-            'TKT_description'               => $default ? '' : $ticket->get_f('TKT_description'),
+            'TKT_description'               => $default ? '' : $ticket->get_raw('TKT_description'),
```
tn3rb, line 1542: was thinking the issue is escaping on one of the admin page templates that's too strict, 
but if this works, then that's ok with me

tn3rb (approved):
sure 👍🏻

hoseinrafiei:
@Saam01 

Please test this PR and let us know the results. You should test the issue described here: https://github.com/eventespresso/event-espresso-core/issues/3963

Saam01:
The function seems to be working fine.
image
Output:
image

Pebblo:
@tn3rb & @hoseinrafiei 

Whilst I'm fine with releasing this as is if we have a rich text editor for the description fields should we not be outputting that output in the way the user has it?

Meaning rather than users having to manually add br tags into the content, if they have a new line should we not just output the new line?

Also, should this not be escaped on save? Currently, the values save directly as is in the DB: 
https://monosnap.com/file/vekKpDf9R7dQU9D4SOFybm4Tah31hp

tn3rb:
hey @Pebblo 

> if we have a rich text editor

that's just a plain text area, not an RTE. The new EDTR utilizes an RTE for ticket and datetime descriptions, but not the legacy editor.

> should we not be outputting that output in the way the user has it?

because this is just a plain text editor, it will just display any text as is and can not perform any kind of "rendering"

> should this not be escaped on save

no, escaping is only for display. The ticket and datetime description fields for those models are using the `EE_Post_Content_Field` class which will properly sanitize all content prior to saving

Pebblo:
This changes it back to how it used to work so I think it's fine, I just thought we might be able to include the line breaks but I get it now :)

PR #3991, opened 2022-08-05
Fix CNT_ISO3 field not being pulled from the request

Pebblo:
If you edit a countries setting, or make no changes and just hit save, the CNT_ISO3 value is lost.

To reproduce go to Event Espresso -> General Settings -> Countries.

Your sites currently selected country (set in EE -> General Settings -> Your organization) should load its settings there (or US if nothing has been set in your organization yet).

Note the CNT ISO3 field should be populated: https://monosnap.com/file/6BBKOoh9DTrEDH7Dbv8CoCz69W1olJ

Hit the 'Save Country Details' button on that page, on reload the CNT ISO3 field will be blank (its not an empty field in the DB, which you can confirm in esp_country if you'd like see [HERE](https://monosnap.com/file/BloXKNtZSHwZut2QOAKlx6FLiFMqWv))

Manually reading the value, doesn't work.

Now switch to this branch and re-add your country CNT_ISO3 value (you noted it above, right?)

Save. It should save correctly this time.

Select another country from the list, wait for the ajax request to finish so it loads the settings, now make no changes but not the CNT_ISO3 value and hit save.... the CNT_ISO3 value that was there should remain.

### DIFF
--- a/admin_pages/general_settings/General_Settings_Admin_Page.core.php
+++ b/admin_pages/general_settings/General_Settings_Admin_Page.core.php
@@ -1121,7 +1121,7 @@ class General_Settings_Admin_Page extends EE_Admin_Page
         $cols_n_values                    = [];
         $cols_n_values['CNT_ISO3']        = strtoupper(
             $this->request->getRequestParam(
-                "cntry[$CNT_ISO][CNT_ISO3]", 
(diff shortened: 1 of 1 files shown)


tn3rb (approved):
👍🏻

hoseinrafiei (approved):
Great 👍

PR #3995, opened 2022-08-18
Fix EE REST API Debug keys

Pebblo:
Found here: https://eventespresso.com/topic/attendee-mover-stop-working-http520-error/

If you enable `EE_REST_API_DEBUG_MODE`, by adding this to your wp-config.php file:

`define( 'EE_REST_API_DEBUG_MODE', true );`

Then go to `/wp-json/ee/v4.8.36/events/`

You should get a server error.

Turns out this is because with the above enabled EE is adding some debug headers to the request but the keys have spaces.

This just replaces any spaces in the headers with a dash (`-`)

Retest the above endpoint and all it well.


tn3rb (approved):
# 👍🏻

hoseinrafiei (approved):
👍

PR #4002, opened 2022-08-22
Don't try to check if an image exists if the input for the img URL is empty

Pebblo:
Reported here: https://eventespresso.com/topic/paypal-express-cant-be-deactivated/

The change here just checks for a value in the img fields before trying to see if the image actually exists.

This doesn't just affect PayPal Express, it's any of the payment methods which display logos.

To reproduce use Debug bar and activate PayPal Express.

Check the WP_Http queries section and you'll see 2 empty header requests, they fail and EE works as it should but if anything is hooking into [http_request_args](https://developer.wordpress.org/reference/hooks/http_request_args/) expecting a request to actually have a location to send the request to it can cause fatal's

On this branch those empty HEAD requests are skipped, we can't have an image if we don't have a URL either way :)

The thread above is an example, LearnDash is simply doing this:

```
	public function maybe_append_auth_headers( array $parsed_args, string $url ):array {
		$needle = $this->get_api()->base . '/repo/plugin/';
		if ( strpos( $url, $needle ) !== 0 ) {
			return $parsed_args;
		}
		if ( ! is_array( $parsed_args['headers'] ) ) {
			$parsed_args['headers'] = array();
		}

		$parsed_args['headers'] = array_merge( $parsed_args['headers'], $this->get_auth_headers() );

		return $parsed_args;
	}
```

This results in:

```
Fatal error: Uncaught TypeError: Argument 2 passed to LearnDash\Hub\Controller\Projects_Controller::maybe_append_auth_headers() must be of the type string, null given, called in /home/gipenny/public_html/wp-includes/class-wp-hook.php on line 307 and defined in /home/gipenny/public_html/wp-content/plugins/learndash-hub/src/controller/class-projects-controller.php:106
```


core/libraries/form_sections/strategies/display/EE_Admin_File_Uploader_Display_Strategy.strategy.php
```diff
@@ -64,7 +64,11 @@ public function display()
             'ee_media_upload button button--secondary'
         );
         // only attempt to show the image if it at least exists
-        $image = $this->src_exists($this->_input->raw_value()) ? EEH_HTML::br(2) . EEH_HTML::img($this->_input->raw_value(), esc_html__('logo', 'event_espresso'), '', 'ee_media_image') : '';
+        if(!empty($this->_input->raw_value()) && $this->src_exists($this->_input->raw_value())) {
+            $image = EEH_HTML::br(2) . EEH_HTML::img($this->_input->raw_value(), esc_html__('logo', 'event_espresso'), '', 'ee_media_image');
+        } else {
+            $image = '';
+        }
```
hoseinrafiei, line 71:
```suggestion
        $image = '';
        if (! empty($this->_input->raw_value()) && $this->src_exists($this->_input->raw_value())) {
            $image = EEH_HTML::br(2) . EEH_HTML::img($this->_input->raw_value(), esc_html__('logo', 'event_espresso'), '', 'ee_media_image');
        }
```

hoseinrafiei (approved):
The logic is fine. Just posted one comment about code style to make it cleaner and also to fix the linter issue.

tn3rb (approved):
sure 👍🏻

Issue #3973, opened 2022-08-01
Refactor Usage of Third Party Libs for Decaf


tn3rb:
The third party libraries in the `core/third_party_libs/` directory need to have one of the following actions applied to them:

##### ACTION OPTIONS
- `A` moved to global assets folder `core/templates/global_assets` if required by multiple use cases
- `B` co-located to same folder as code that utilizes it if required for a single use case
- `C` move implementation behind feature flag / file check(s) (ie: ` if ( lib exists ) { run lib code }` so that lib can be removed in decaf
- `D` if lib is WP plugin, add logic for recommending download and installation, so that lib can be removed in decaf
- `E` removed altogether because not needed / no longer used
- `F` other possible solutions ???

Here's the libs and the action that I feel may be required (this is subject to change if need be)

- accounting - `A`
- cssinliner - `B`
- dompdf - `C`
- Jaybizzle- `C`
- joyride - `E`
- moment - `A`
- PUE - `C`
- qtip - `A` or `E`
- wp-api-basic-auth - `D`
- wp-graphql - `D`
- jquery.countdown.min.js - `A`
- jquery.plugin.min.js - `A`

If the above action appears to be the best for each lib in question, then plz proceed, otherwise plz provide feedback regarding complications, alternate suggestions, etc

Please perform all work in separate PRs (one for each lib, but some of those smaller assets can be grouped into a single PR). 

Please use `master` branch as the base for all work branches.

Please tag PRs with the following labels: 
- ` category:decaf `
- ` category:3rd-party-integrations `
- ` type:task `

For action option `D` I will likely try to add some kind of helper class for performing plugin recommendations upon activation. So please work on those libs last.

As always, please ask questions if there is *anything* you are unsure about, and I am always open for alternate approaches since there is always more than one way to solve a problem.

hoseinrafiei:
@tn3rb

All the dependencies are done except those that are marked as `D`.
Please let me know when we're ready to work on those items.

tn3rb:
ok great thnx @hoseinrafiei 
`D` was covered by #4043

PR #4005, opened 2022-08-24
Add Feature Flag for Jaybizzle Package

hoseinrafiei:

Moved the `EventEspressoVendor\CrawlerDetect\CrawlerDetect` usage behind the class check to run it safely only whenever the package exists.

### DIFF
--- a/core/services/request/middleware/BotDetector.php
+++ b/core/services/request/middleware/BotDetector.php
@@ -4,7 +4,6 @@ namespace EventEspresso\core\services\request\middleware;
 
 use EventEspresso\core\services\request\RequestInterface;
 use EventEspresso\core\services\request\ResponseInterface;
-use EventEspressoVendor\CrawlerDetect\CrawlerDetect;
 
 /**
  * Class BotDetector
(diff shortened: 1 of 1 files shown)


core/services/request/middleware/BotDetector.php
```diff
@@ -27,12 +26,17 @@ public function handleRequest(RequestInterface $request, ResponseInterface $resp
...
+            $CrawlerDetect = $this->loader->getShared('EventEspressoVendor\CrawlerDetect\CrawlerDetect');
+            if ($CrawlerDetect instanceof \EventEspressoVendor\CrawlerDetect\CrawlerDetect) {
+                // Check and record the user agent of the current 'visitor'
+                $this->request->setIsBot($CrawlerDetect->isCrawler());
+                $this->request->setUserAgent($CrawlerDetect->userAgent());
+            }
+        } else {
+            $this->request->setIsBot(false);
+            $this->request->setUserAgent();
         }
```
hoseinrafiei, line 40: I'm not certain if we need this else statement. Since we don't have an else statement for the if statement on line 32 then I think we can remove this too. Implemented just to discuss it.

tn3rb: it doesn't hurt... 
it at least sets those values to the proper data types, because they are potentially `null` prior to this point

tn3rb (approved):
### Looks Good 🍰

PR #4006, opened 2022-08-24
Add Feature Flag for PluginUpdateEngineChecker Class

hoseinrafiei:

Moved the `PluginUpdateEngineChecker` usage behind the file existence and readable check to run it safely only whenever the file exists so we can safely remove it from `decaf` version.


tn3rb (approved):
# 🍄 🧝🏻 🐲 🍄

PR #4007, opened 2022-08-25
Move CSS Inliner Library to Messages Directory

hoseinrafiei:
Related to https://github.com/eventespresso/event-espresso-core/issues/3973.

Moved the cssinliner library to `core/libraries/messages/` directory.


tn3rb (commented):
oh yuk... how did i not notice this was all a bunch of composer dependencies?!?!?!

hmmmm... having second thoughts about moving this...

can you do me a favour Hossein and research whether there is a better alternative out there these days that we could use instead? 
This basically takes some css that's in separate `.css` files and injects it into the HTML that the CSS was referencing, so that the styles are added inline like ` style="display: none;"`
 
I _think_ it is used for the email message templates, but could also be for the PDF generation for receipts and invoices. You'll need to dig through the messages code to find out exactly how and where it is being used. Be forewarned that the messages code can be very confusing to follow.

hoseinrafiei:
@tn3rb 

Despite some online tools which are not useful for us there are following PHP libraries:

- https://github.com/tijsverkoyen/CssToInlineStyles
- https://github.com/MyIntervals/emogrifier

The first one which we're using already seems to be the best one.

Also there is a free API (without documentation and support) in http://premailer.dialect.ca/api/0.1/documents path that uses a Python library to create HTML with inline CSS codes which we shouldn't use since it will add a dependency to our code.

I think the current library is the best one. We may just need to update it to latest version.
Another idea would be to add one of the above packages to EE production dependencies and include the /vendor directory in the EE decaf and caff packages.

What's your thought?

tn3rb:
ok thnx, for looking into this.

So... because there is actually a lot more to this little lib than i realized, i'm thinking we should maybe leave it where it is within the third party libs folder, and NOT move it into the messages where it is used.

Do you have any other ideas or potential solutions?

hoseinrafiei:
@tn3rb

Thanks for the feedback. I searched more but the only solution that I can suggest at the moment is to use general `/vendor` directory to include production assets as well. We can require this package and include it in the production package of caff and decaf versions.

hoseinrafiei:
If you decided to leave the library in the current path then let me know to remove this branch and ignore it.

tn3rb:
ya that makes sense.

since this is nothing more than some composer dependencies, they can just be added via composer

can you undo what you've done so far in this PR and then add the lib via our main `composer.json`?

or create a new PR if that's easier for you.

hoseinrafiei:
@tn3rb

Would you please check it again?
Removed the old library completely and added the new version as requirement in composer.

core/libraries/messages/messenger/EE_Email_messenger.class.php
```diff
@@ -609,8 +609,7 @@ protected function _body($preview = false)
          */
         if (apply_filters('FHEE__EE_Email_messenger__apply_CSSInliner ', true, $preview)) {
             // require CssToInlineStyles library and its dependencies via composer autoloader
-            require_once EE_THIRD_PARTY . 'cssinliner/vendor/autoload.php';
-
+            require_once EE_VENDOR . 'autoload.php';
```
hoseinrafiei, line 612: I realized that the `autoload.php` file is not loaded by default so added this line to prevent issues. We can require this file in a more general place and then remove it from this part.

core/espresso_definitions.php
```diff
@@ -32,6 +32,7 @@
 define('EE_WIDGETS', EE_PLUGIN_DIR_PATH . 'widgets/');
 define('EE_PAYMENT_METHODS', EE_PLUGIN_DIR_PATH . 'payment_methods/');
 define('EE_CAFF_PATH', EE_PLUGIN_DIR_PATH . 'caffeinated/');
+define('EE_VENDOR', EE_PLUGIN_DIR_PATH . 'vendor/');
```
tn3rb, line 35: # 🌟🌟🌟🌟🌟

tn3rb (approved):
### 💥 BOOM 💥 

much better solution, great job !

PR #4013, opened 2022-08-30
Add Feature Flag for DomPDF Package

hoseinrafiei:

Moved the `Dompdf\Dompdf` usage behind the file existence check to run it safely only whenever the package exists.
Also found an instance in the `EE_Pdf_messenger.class.php` file located in `core/libraries/messages/messenger/` path but did not updated the class since the whole class seems to be deprecated because the class did not used at all.


modules/gateways/Invoice/lib/Invoice.class.php
```diff
@@ -287,7 +287,7 @@ public function send_invoice($download = false)
         $invoice_name .= esc_html__(' for ', 'event_espresso') . $template_args['name'];
         $invoice_name = str_replace(' ', '_', $invoice_name);
         // Create the PDF
-        if ($request->requestParamIsSet('html')) {
+        if ($request->requestParamIsSet('html') || ! is_readable(EE_THIRD_PARTY . 'dompdf/src/Autoloader.php')) {
```
tn3rb, line 290: i get what you are doing, but I don't think this is the best way to handle this because it's essentially changing the intended result. If the `html` request param is set, then the user is not expecting a PDF document (not saying that is happening here), and conversely, the user is not requesting or expecting an HTML rendering if they have requested a PDF. So, although this change _does_ protect against the `Dompdf` library being missing, it will produce unexpected and unwanted results IMHO. 

To fix this, the `is_readable()` check needs to be moved into the the `else` block of code below this, and should likely generate an error if DOMPDF is not readable, then execute the existing code (which is not great and could be improved to be more defensive as it currently breaks the golden rule of programming (“If it can be null, it will be null”).

On a side note, the `else` block below is not necessary if we return from the previous condition after echoing; ie: 

```php
if ($request->requestParamIsSet('html')) {
    echo wp_kses($content, AllowedTags::getWithFormTags());
    return;
}
// no else block necessary here
```

tn3rb (changes requested):
plz refactor as described in my code comment

core/libraries/messages/messenger/EE_Pdf_messenger.class.php
```diff
@@ -1,5 +1,7 @@
 <?php
 
+use Dompdf\Options;
```
tn3rb, line 3: 👍🏻

core/libraries/messages/messenger/EE_Pdf_messenger.class.php
```diff
@@ -318,13 +319,56 @@ protected function _get_main_template($preview = false)
      */
     protected function _do_pdf($content = '')
     {
-        $invoice_name = $this->_subject;
+        // dompdf options
+        $options = $this->get_dompdf_options();
+        if ($options instanceof Options) {
```
tn3rb, line 324: 👍🏻

core/libraries/shortcodes/EE_Messenger_Shortcodes.lib.php
```diff
@@ -114,16 +114,20 @@ protected function _parser($shortcode)
...
+     * @since 4.5.0
+     *
      */
     private function _get_button(EE_Messages_Addressee $recipient, $sending_messenger)
     {
-        $download_text = $sending_messenger == 'pdf'
+        if ($sending_messenger === 'pdf' && ! is_readable(EE_THIRD_PARTY . 'dompdf/src/Autoloader.php')) {
+            return '';
+        }
+        $download_text = $sending_messenger === 'pdf'
```
tn3rb, line 130: `==` > `===`  💯 x 👍🏻

modules/gateways/Invoice/lib/Invoice.class.php
```diff
@@ -289,24 +289,22 @@ public function send_invoice($download = false)
...
+        $pdf_messanger = new EE_Pdf_messenger();
+        // dompdf options
+        $options = $pdf_messanger->get_dompdf_options();
+        if ($options instanceof Dompdf\Options) {
             $dompdf = new Dompdf\Dompdf($options);
             $dompdf->loadHtml($content);
             $dompdf->render();
             $dompdf->stream($invoice_name . ".pdf", ['Attachment' => $download]);
+            exit(0);
         }
```
tn3rb, line 303: this logic is relatively the same as some of the lines in `EE_Pdf_messenger::do_pdf()` and could be extracted into its own method as well. Maybe with parameters for the following:

- `Dompdf\Options` 
- file name
- boolean $attachment

a potentially _better_ idea is to extract ALL of the logic that deals directly with `Dompdf` into some kind of PDF adapter class, maybe located in a new `core/services/helpers/pdf` folder (open to alternate suggestions), so that ALL of the `Dompdf` specific logic is contained in its own class and decoupled from everything else. This would make it easier in the future to update and/or replace it with another library.

This adapter could then be imported into any class that needs to generate a PDF, which i _think_ would only be:

- `core/libraries/messages/messenger/EE_Pdf_messenger.class.php`
- `modules/gateways/Invoice/lib/Invoice.class.php`

but plz search thoroughly as i may be wrong about that.

As a general rule for class creation (keep in mind there's always exceptions) i think it's best to only use class constructors for setting up dependencies and then use another method for initializing with use case and/or request specific details. Specifically, avoid side effects in class constructors and only use them for constructing the absolute minimum required for the class to function so that a class can be constructed when and where it is most useful, but external side effects like db requests don't occur until needed. Basically, classes are more flexible if logic for specific use cases is decoupled from class construction. One obvious exception to the above is DTOs (data transfer objects) where the use case / request specific details **ARE** the essential core of the class. In those cases, passing the data to the constructor makes most sense.

And my apologies for sending you off in another direction. Sometimes things don't appear to me on first review because i get distracted by some other issues. And plz keep in mind that the above is in no way a criticism of your work, as i realize you are working with code that someone else created. My interest now is to improve the existing code and fix any outstanding issues it may have.

tn3rb (changes requested):
your changes so far are looking great but I have some additional requests for improving the existing code

hoseinrafiei:
@tn3rb

Thanks for your comment. Please find the following changes and let me know your feedback.

1. Created `core/services/adapters/PdfAdapter.php` class.
2. Removed the `get_dompdf_options` function from `EE_Pdf_messenger` class.
3. Added the `getOptions` to the new `PdfAdapter` class.
4. Added setters for options and content in the `PdfAdapter` class.
5. Refactored the codes to call the new adapter and removed duplicate codes.
6. Instantiate the PdfAdapter in constructor of `EE_Pdf_messenger` class. I did not use the dependency injection since this class was loaded in `core/libraries/messages/EE_Messenger_Collection_Loader.lib.php` using a different method and modifying the loader code to pass the dependency manually only creates the code more messy.
7. Instantiate the PdfAdapter in `send_invoice` function of the Invoice class since this class was already deprecated.

There might be some more rooms for improvements so feel free to let me know if you have more feedback.

core/services/adapters/PdfAdapter.php
```diff
@@ -0,0 +1,101 @@
...
+
+    /**
+     * @param Options $options
+     * @return $this
+     */
+    public function setOptions(Options $options): self
+    {
+        $this->options = $options;
+        return $this;
+    }
```
tn3rb, line 73: In PHP, **ALL** objects are passed by reference. 

If some client code retrieves the Options object from this adapter, sets some values on it, then passes it back to this method, `$this->options = $options;` will simply be overwriting the property with the _exact same_ object. It is completely unnecessary to call this setter after setting options.

What _would_ be more useful (but completely unnecessary given our current use) would be to combine the logic you have in the `getOptions()` method with this `setOptions()` method, but instead of passing an `Options` instance to it, you pass an array of key value pairs that would get looped over and added to the  `Options` object. 

Maybe something like:

```php
    /**
     * @param array $extra_options
     * @return PdfAdapter
     */
    public function initializeOptions(array $extra_options): PdfAdapter
    {
        $this->options = new Options();
        $this->options->set('isRemoteEnabled', true);
        $this->options->set('isJavascriptEnabled', false);
        if (defined('DOMPDF_FONT_DIR')) {
            $this->options->setFontDir(DOMPDF_FONT_DIR);
            $this->options->setFontCache(DOMPDF_FONT_DIR);
        }
        // Allow changing the paper size.
        if (defined('DOMPDF_DEFAULT_PAPER_SIZE')) {
            $this->options->set('defaultPaperSize', DOMPDF_DEFAULT_PAPER_SIZE);
        }
        // now apply user provided options
        foreach( $extra_options as $key => $value ) {
                $this->options->set( $key, $value );
        }
        return $this;
    }
```

then your client code might be something like:

```php
        $pdf_adapter = new PdfAdapter();
        $options = [
                'key' => 'value'
        ];
        $pdf_adapter
                ->initializeOptions($options)
                ->setContent($content)
                ->generate($invoice_name . ".pdf", (bool) $download);
```

or since we are not actually setting any additional options:

```php
        $pdf_adapter = new PdfAdapter();
        $pdf_adapter
                ->initializeOptions()
                ->setContent($content)
                ->generate($invoice_name . ".pdf", (bool) $download);
```

great decision to use fluent setters btw 👍🏻

core/services/adapters/PdfAdapter.php
```diff
@@ -0,0 +1,101 @@
...
+
+    /**
+     * @param string $content
+     * @return $this
+     */
+    public function setContent(string $content): self
+    {
+        $this->content = $content;
+        return $this;
+    }
```
tn3rb, line 84: we could even just pass the content directly to the `generate()` method since it is only ever used there and then get rid of this setter since there doesn't appear to be any reason to save the content

tn3rb (changes requested):
It's looking good, but i did have some more suggestions and requests

>  I did not use the dependency injection since this class was loaded in core/libraries/messages/EE_Messenger_Collection_Loader.lib.php using a different method and modifying the loader code to pass the dependency manually only creates the code more messy.

that's totally fine and was kind of expected

core/services/adapters/PdfAdapter.php
```diff
@@ -0,0 +1,101 @@
...
+        }
+        return $options;
+    }
+
+
+    /**
+     * @param Options $options
+     * @return $this
+     */
+    public function setOptions(Options $options): self
```
tn3rb, line 69: can you plz verify that `self` is an acceptable return value in PHP 7.1?
I prefer we use the actual classname (in phpdocs too), because sometimes you may be looking at a code snippet posted in chat or elsewhere, and it's impossible to know what class `self` represents since you don't have your IDE to help you out, whereas seeing:

```php
     * @return PdfAdapter
     */
    public function setOptions(Options $options): PdfAdapter
```

is immediately understandable

hoseinrafiei:
@tn3rb

I added the DomPDF as dependency in EE composer and removed it from the `third_party_libs` directory. Also, assuming that DomPDF package will be always available, I removed the feature flag and simplified the codes.

Also implemented your feedback.
Please check it again.

core/services/adapters/PdfAdapter.php
```diff
@@ -18,84 +18,63 @@ class PdfAdapter
...
-        if ($this->options instanceof Options && ! empty($this->content)) {
+        if ($this->options instanceof Options && ! empty($content)) {
             $dompdf = new Dompdf($this->options);
-            $dompdf->loadHtml($this->content);
+            $dompdf->loadHtml($content);
             $dompdf->render();
             $dompdf->stream($filename, ['Attachment' => $download]);
+            return true;
         }
+        return false;
```
tn3rb: 👍🏻

tn3rb (approved):
this all looks great now, much happier with this approach.
let's get this tested 😃

hoseinrafiei:
@Saam01

Please test this PR. You should test all the places that we use PDF Generator to generate a PDF file like invoice and receipt.

Saam01:
@hoseinrafiei  I'm seeing this error everywhere: 

Events-‹-New-Testing-—-WordPress

hoseinrafiei:
@tn3rb 

What do you think about Sam report?
Should we add the vendor directory to git or we should keep it as is and ask the staff to run `composer install` command?

Also, we should include the directory in the production packages. I don't know if we already have this in our package creation process but I just wanted to make sure to mention it.

tn3rb:
@hoseinrafiei 

we'll need to remove the vendor folder from the `.gitignore` AND run `composer install` on this branch (and anytime dependencies are changed/updated/etc) as well as running `composer install --no-dev` somewhere in our deployment process.

hoseinrafiei:
@tn3rb Done.

hoseinrafiei:
@Saam01 Please run `git pull` command on this branch to update your local branch and then test it. The error should be fixed after pulling latest changes.

tn3rb (approved):
👍🏻

Saam01:
Finished this PR and everything is working fine now.

PR #4059, opened 2022-09-29
Fix PHP 7.2 compatibility

Pebblo:
Removing property type hinting for PHP7.2 compatibility and setting min required version to 7.2.0


tn3rb (approved):
# 👍🏻

PR #4072, opened 2022-10-06
no calls to assert() in code

Pebblo:
For more info see: https://github.com/eventespresso/cafe/issues/103


tn3rb:
this was fixed in https://github.com/eventespresso/cafe/pull/105
i can copy over changes from the cafe repo to here if you need to release that fix

PR #4073, opened 2022-10-06
Add a filter to getTickets used for the TicketSelector

Pebblo:
For context, see: https://eventespresso.com/topic/ticket-count-based-on-custom-user-field/?view=all#post-335745

The user wants to limit the TKT qty based on a custom value, I could have filtered just the Qty field but this allows more flexibility overall.

To test this I used this function:

```
add_filter('FHEE__EE_Ticket_Selector__getTickets_tickets', 'tw_filter_ts_tickets', 10, 2);
function tw_filter_ts_tickets($tickets, $query_args) {
	foreach($tickets as $ticket) {
		$ticket->set_max(2);
	}
	return $tickets;
}
```

Which then uses all of the same methods within the class to output as usual, only now the ticket qty is set to 2 in the dropdown. The idea being this could be done on specific tickets based on whatever value you need to pull in within the filter.

Because save() isn't called anywhere within the ticket selector this change only applies when loading,

### DIFF
--- a/modules/ticket_selector/DisplayTicketSelector.php
+++ b/modules/ticket_selector/DisplayTicketSelector.php
@@ -415,7 +415,13 @@ class DisplayTicketSelector
                 : current_time('timestamp');
             $ticket_query_args[0]['TKT_end_date'] = ['>', $current_time];
         }
-        return EEM_Ticket::instance()->get_all($ticket_query_args);
+
+        return apply_filters(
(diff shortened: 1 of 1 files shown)


tn3rb:
changes look fine but plz redo this PR on the cafe repo
then when you want to release i can  update this repo with all of the changes from cafe

tn3rb:
closing as this has been added in https://github.com/eventespresso/cafe/pull/126

PR #4088, opened 2022-11-29
Add translation 'for' text in name_and_info()

man-andy:
There is no translation for the 'for' label in the ticketname

## How has this been tested

Create a waitlist and open the dropdown.
image

## Checklist

* [X] I have read the documentation relating to systems affected by this pull request, see https://github.com/eventespresso/event-espresso-core/tree/master/docs
* [X] User input is adequately validated and sanitized
* [X] all publicly displayed strings are internationalized (usually using `esc_html__()`, see https://codex.wordpress.org/I18n_for_WordPress_Developers)
* [X] My code is tested.
* [X] My code follows the Event Espresso code style.
* [X] My code has proper inline documentation.
* [X] My code accounts for when the site is in Maintenance Mode (MM2 especially disallows usage of models)

### DIFF
--- a/core/db_classes/EE_Ticket.class.php
+++ b/core/db_classes/EE_Ticket.class.php
@@ -1563,7 +1563,7 @@ class EE_Ticket extends EE_Soft_Delete_Base_Class implements EEI_Line_Item_Objec
         foreach ($this->datetimes() as $datetime) {
             $times[] = $datetime->start_date_and_time();
         }
-        return $this->name() . ' @ ' . implode(', ', $times) . ' for ' . $this->pretty_price();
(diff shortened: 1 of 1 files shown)


tn3rb:
OOF! well that was an oversight!

Thanks for the submission @man-andy

tn3rb:
@man-andy 

I'm thinking we should return a single translated string and use placeholders for the dynamic content like:

```php
sprintf(
    esc_html__('%1$s @ $2$s for $3$s', 'event_espresso'),
    $this->name(),
    implode(', ', $times),
    $this->pretty_price()
);
```

this will allow for languages that may want/need to rearrange the order of the words.

I can either go ahead and just make that change, but if you want to have the actual contribution credited to you and appear in your github history, then you can simply ammend your PR. This will also need a translators comment too.

PR #4101, opened 2022-12-12
$this->datetimes_for_event should not be filled with array

Arjen-Smit:
When filtering registrations on an event an error is thrown cause the $datetimes_for_event is filled with an array instead of a DatetimesForEventCheckIn.

## How has this been tested

Steps to reproduce error:
- Go to Admin -> Event Espresso -> Registrations.
- Select Event Check-in
- Select an Event in the Filters and press Filter.

## Checklist

* [ ] My code is tested.


tn3rb:
thanks @Arjen-Smit

this has already been fixed in the upcoming verion of EE

tn3rb:
> This got me a bit confused, I would think dev is the upcoming version of EE?

HA! Yes that would make sense wouldn't it?

Because of the difficulties developing new features and/or refactoring code when changes span across multiple plugins, we decided to merge all of our 50-60 active repos into a single monorepo. So now all PHP development for all of our active plugins happens on a single private repo.

I'm currently wrestling with getting all of the CI/CD stuff configured for automagically deploying code to our various servers, services, and public repos like this one, but that's a crazy jigsaw puzzle to work with. 
Since that's not ready yet, I can try to get dev branch here updated manually within the next week.

